PR #4, opened 2015-03-19
implement MessageHeader

maqi:


src/types.rs
```diff
@@ -1,24 +1,29 @@
...
-    bound by the terms of the MaidSafe Contributor Agreement, version 1.0, found in the root
-    directory of this project at LICENSE, COPYING and CONTRIBUTOR respectively and also
-    available at: http://www.maidsafe.net/licenses
-    Unless required by applicable law or agreed to in writing, the MaidSafe Software distributed
-    under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS
-    OF ANY KIND, either express or implied.
-    See the Licences for the specific language governing permissions and limitations relating to
-    use of the MaidSafe
-    Software.                                                                 */
+// Copyright 2014 MaidSafe.net limited
```
dirvine: 2015

src/types.rs
```diff
@@ -26,4 +31,133 @@ pub enum Authority {
...
+    match *self {
+      Authority::ClientManager => authority = "ClientManager",
+      Authority::NaeManager => authority = "NaeManager",
+      Authority::NodeManager => authority = "NodeManager",
+      Authority::ManagedNode => authority = "ManagedNode",
+      Authority::ManagedClient => authority = "ManagedClient",
+      Authority::Client => authority = "Client",
+      Authority::Unknown => authority = "Unknown",
+    }
+    CborTagEncode { tag : 5483_000 , data : &authority }.encode(e)
```
dirvine: I think this tag needs to be unique across all libs. We need to sort out a scheme for our own libs, but this number is used in types. Perhaps routing should use 5483_1XXX or similar

src/types.rs
```diff
@@ -26,4 +31,133 @@ pub enum Authority {
...
+
+  pub fn generate_address() -> Vec<u8> {
+    let mut address: Vec<u8> = vec![];
+    for _ in (0..64) {
+      address.push(rand::random::<u8>());
+    }
+    address
+  }
+
+  fn test_object<T>(obj_before : T) where T: for<'a> Encodable + Decodable + Eq {
```
dirvine: Nice one Qi

src/types.rs
maqi: agreed. we shall come up with a pattern for this tag.
say : 5483_ for madsafe, then 3 digits, the first for library (0 for types, 1 for routing, 2 for vault, etc.), last two for the number of parameters capsulated (or shall be for each datatype ?)

PR #7, opened 2015-03-23
implement Sentinel::Add

maqi:


src/types.rs
```diff
@@ -14,12 +14,31 @@
...
 
 #![allow(unused_assignments)]
+extern crate sodiumoxide;
+
+use sodiumoxide::crypto;
 
 use cbor::CborTagEncode;
 use rustc_serialize::{Decodable, Decoder, Encodable, Encoder};
 
+pub fn vector_as_u8_64_array(vector: Vec<u8>) -> [u8;64] {
```
dirvine: I think here Qi you can do Array.iter().collect() to go from array to vector.

maqi: I see this kind of function widely used in maidsafe_types,
so thought it is a policy to use so.
does it mean in maidsafe_types will change to use Array.iter().collect() ?

dirvine: Yes I think it should as it's simpler and the compiler does not need to bounds check, so its also more efficient.

PR #16, opened 2015-03-26
Re-enabled inital TCP connection test

Fraser999:
WIP - test passes, but doesn't actually send/recv data.


src/tcp_connections.rs
```diff
@@ -159,37 +158,37 @@ where T: Send + Decodable + 'static {
...
+
+        let (i, mut o) = connect_tcp(SocketAddr::from_str("127.0.0.1:5483").unwrap()).unwrap();
+        for x in 0u64 .. 10u64 {
+            o.send(&x).ok();
+        }
+        o.close();
+        // Print everything that we get back.
+        for a in i.into_blocking_iter() {
+            let (x, fx): (u64, u64) = a;
+            println!("{} -> {}", x, fx);
```
dirvine: if you want to see this output then ```cargo test -nocapture``` should do it.

Fraser999: I had to do `cargo test -- --nocapture`.

PR #18, opened 2015-03-27
Fixed test and bug.

Fraser999:


src/tcp_connections.rs
```diff
@@ -133,19 +138,21 @@ where T: Send + Decodable + 'static {
...
     spawn(move || {
         let mut buffer = BufReader::new(stream);
-        loop {
-            let mut dec = Decoder::from_reader(&mut buffer);
-            match dec.decode().next().unwrap() {
-                Ok(a) => {
-                    // Try to send, and if we can't, then the channel is closed.
-                    if in_snd.send(a).is_err() {
+        {
+            let mut decoder = Decoder::from_reader(&mut buffer);
```
dirvine: Nice one, good catch

Fraser999: Took me long enough!

PR #20, opened 2015-03-28
define and expose some routing types for Vault::DataManager

maqi:


src/bootstrap.rs
```diff
@@ -34,7 +34,7 @@ type BootStrapContacts = Vec<Contact>;
 static MAX_LIST_SIZE: usize = 1500;
 
 fn array_to_vec(arr: &[u8]) -> Vec<u8> {
-  let mut vector: Vec<u8> = Vec::new();
+  let vector: Vec<u8> = Vec::new();
```
dirvine: I think 
 let vector = Vec::new();
Is OK here as the type is inferred

PR #27, opened 2015-03-30
Bootstrap crud test integrated. Serialization test added for Contact.

krishnaIndia:
CRUD test integrated for Bootstrap. Serialization test added for Contact.

### DIFF
--- a/src/bootstrap.rs
+++ b/src/bootstrap.rs
@@ -21,7 +21,6 @@ extern crate sodiumoxide;
 extern crate time;
 extern crate sqlite3;
 extern crate cbor;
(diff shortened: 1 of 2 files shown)


src/bootstrap.rs
```diff
@@ -183,3 +185,58 @@ impl BootStrapHandler {
     ;
   }
 }
+
```
dirvine: Perhaps best to make a ```mod test``` and then if you need a crate for test like rand above it can go in the test module.

krishnaIndia: Sure. I ll make the change and update the commit

PR #31, opened 2015-03-31
Interim commit of UDP broadcast methods.

Fraser999:


src/broadcast.rs
```diff
@@ -0,0 +1,181 @@
...
+                    ((buffer[(2 * i) + 1] as u16) << 8) + (buffer[(2 * i) + 2] as u16);
+            }
+            let port: u16 = ((buffer[17] as u16) << 8) + (buffer[18] as u16);
+            let flowinfo: u32 =
+                ((buffer[19] as u32) << 24) + ((buffer[20] as u32) << 16) +
+                ((buffer[21] as u32) << 8) + (buffer[22] as u32);
+            let scope_id: u32 =
+                ((buffer[23] as u32) << 24) + ((buffer[24] as u32) << 16) +
+                ((buffer[25] as u32) << 8) + (buffer[26] as u32);
+            let peer_socket = SocketAddrV6::new(Ipv6Addr::new(
```
dirvine: Is it possible to receive ip6 broadcasts. I think it uses multicast with a broadcast mullticast group definition or similar?

Fraser999: The broadcast is all done on IPv4 sockets (that's the plan anyway).  This is just parsing the sender's provided address - the one he claims to be listening on.  It could be IPv4 or v6, or it could be a total lie - he may not be listening on the provided address at all.

dirvine: Ah right does that mean a node may return an address as a message as opposed to from the udp packet itself? Could prove to be better as it will allow further flexibility

Fraser999: We're not given access to the UDP packets via Rust's `UdpSocket` - the nodes are just sending data from `u8` arrays.  I couldn't get the cbor serialisation/parsing to work quickly with `SocketAddr`s, so I just went with what made sense to me and extracted the raw bytes from the addresses.

I haven't looked into the endian-safety of this.  If Rust doesn't already handle endianness for us when sending/receiving UDP messages, then the serialise/parse functions aren't safe.

PR #36, opened 2015-03-31
test_multiple_client_small_stream

chandraprakash:


src/tcp_connections.rs
```diff
@@ -206,6 +206,71 @@ mod test {
...
+        }
+
+        // listener
+        thread::spawn(move || {
+            for (connection, u32) in listener.into_blocking_iter() {
+                // Spawn a new thread for each connection that we get.
+                thread::spawn(move || {
+                    let (i, mut o) = upgrade_tcp(connection).unwrap();
+                    for x in i.into_blocking_iter() {
+                        if o.send(&(x, x + 1)).is_err() { break; }
```
dirvine: This will not compile for now, may be a compiler bug, unsure, it needs to be 
```   let i = i.blocking_iter(); 
       let vec : Vec<u64> = i.collect();
        for &x in vec.iter() {
```

PR #37, opened 2015-04-01
destination address needs to be part of the signature of handle_put trait

maqi:


dirvine:
HI @maqi This is failing build with latest nightly. Can you please take a look and perhaps switch to rand crate. ```
src/accumulator.rs:114:20: 114:38 warning: use of deprecated item: use the crates.io `rand` library instead, #[warn(deprecated)] on by default
src/accumulator.rs:114       address.push(rand::random::<u8>());```

PR #39, opened 2015-04-01
Resolve all but one warning

benjaminbollen:
Left one warning untouched, as it would be less trivial change - outwidth my current scope.  Just quick PR to get routing set up and scan through the project


dirvine:
Nice !

PR #46, opened 2015-04-02
Sentinel tests

benjaminbollen:


benjaminbollen:
ASSERT FAILS: Sentinel sends out for GetGroupKey for every copy of PutData received; this assert passed in C++ impl

benjaminbollen:
Expansion of test mechanism paused on Rust language technicalities.  Sentinel test already exposes two failing asserts:  Sentinel does not return a response; and Sentinel calls back to GetGroupKey on every put_data message received

PR #47, opened 2015-04-03
Force Travis Rust on 2015-04-01

benjaminbollen:


benjaminbollen:
Doesn't correctly install rustc, can't find rustc afterwards

PR #50, opened 2015-04-03
Fixed test and bugs.

Fraser999:


dirvine:
Nice now get on holiday and rest :+1:

PR #54, opened 2015-04-05
Pull beta

rossmuir:


dirvine:
I am not sure we should merge this right now @rossmuir as I think routing still needs to work on nightly. self encryption, types and client are currently in Beta channel with crust heading there probably tomorrow.

rossmuir:
OK no worries - I will close this and create PR's for self encryption, types and client appveyor.yml files.

PR #61, opened 2015-04-07
Added tests to assure that quorum is reached.

benjaminbollen:
Additional tests for Accumulator written during debugging of Sentinel, all tests pass


benjaminbollen:
Will reopen to see passes CI (should) and whether it increases coverage.  At least it tests a different case with non-1 quorum size

PR #65, opened 2015-04-08
Routing Client - can be merged

ustulation:
Code for Get for routing client, 
code for Put seems to have different interface , what interface is required here ?


src/routing_client.rs
```diff
@@ -0,0 +1,210 @@
...
+use messages;
+use maidsafe_types::traits::RoutingTrait;
+
+/// DHT node
+pub struct RoutingClient<'a> {
+    facade: &'a (Facade + 'a),
+    sign_public_key: crypto::sign::PublicKey,
+    sign_secret_key: crypto::sign::SecretKey,
+    encrypt_public_key: crypto::asymmetricbox::PublicKey,
+    encrypt_secret_key: crypto::asymmetricbox::SecretKey,
```
chandraprakash: Need to keep a copy of maid id here. (https://github.com/dirvine/maidsafe_types/blob/master/src/id/maid.rs) 
This should be passed in as a parameter.
No need to have other keys.

src/routing_client.rs
```diff
@@ -0,0 +1,210 @@
...
+    sender: Sender<TcpStream>,
+    receiver: Receiver<TcpStream>,
```
chandraprakash: sender & receiver are stray code. 
Need connection manager  https://github.com/inetic/routing/blob/master/src/routing_node.rs#L50-L51

src/routing_client.rs
```diff
@@ -0,0 +1,210 @@
...
+
+        // Serialise PutData message
+        let mut encoder = cbor::Encoder::from_memory();
+        encoder.encode(&[&put_data]).unwrap();
+
+        // Make MessageHeader
+        let header = message_header::MessageHeader::new(
+            self.message_id,
+            types::DestinationAddress {
+                dest: self.own_address.clone(), // Is this Ok? (as put is supposed to go to own MaidManagers, no?)
```
chandraprakash: this is correct for put message

src/routing_client.rs
```diff
@@ -0,0 +1,210 @@
...
+                reply_to: vec![0; 64], // should be Option None
+            },
+            types::SourceAddress {
+                from_node: self.own_address.clone(), // should this be bootstrap address ?
```
chandraprakash: yes

src/routing_client.rs
```diff
@@ -0,0 +1,210 @@
...
+                from_group: vec![0, 64], // Option::None
+                reply_to: self.own_address.clone(),
+            },
+            types::Authority::Client,
+            types::Signature::generate_random(), // should be removed
```
chandraprakash: need signature for put

src/routing_client.rs
```diff
@@ -0,0 +1,210 @@
...
+        // Make MessageHeader
+        let header = message_header::MessageHeader::new(
+            self.message_id,
+            types::DestinationAddress {
+                dest: name.0,
+                reply_to: vec![0; 64], // None // - Option::None
+            },
+            get_data.requester.clone(),
+            types::Authority::Client,
+            types::Signature::generate_random(), // What to do here?  // Shouldn't be present in MessageHeader
```
chandraprakash: signature is not required for get Messages. MessageHeader should have signature as optional field 
https://github.com/chandraprakash/MaidSafe-Routing/blob/inetic-MAID-843-integration-with-crux/src/maidsafe/routing/message_header.h#L173

PR #67, opened 2015-04-08
Sentinel find group

benjaminbollen:


benjaminbollen:
At this point reached: Implemented 'Embedded Signature Group' and PutData with a response from a 'EmbeddedSignatureGroup' passes; suffers from merge conflict and conflict with DhtId

benjaminbollen:
Test for FindGroupResponse in place, and fails as expected

PR #73, opened 2015-04-09
Pmid 

benjaminbollen:


benjaminbollen:
Extends Pmid implementation, PublicPmid, PublicKey, PublicSignKey and Signature.  Shields Vec<u8> inside latter three types, not yet ideal, but good middle group -> allows for compiler type check on using the correct key

PR #80, opened 2015-04-10
more style fixes

steveklabnik:
I didn't go through every file this time, because I got bored :wink: 

This code overall looks good. There's some small things that could be better, but I'd say that in general, you're on the right track.


dirvine:
Thanks Steve much appreciated, you have done a great job with this, very much appreciated

PR #82, opened 2015-04-11
0 state

chandraprakash:
zero state works with a workaround. need to add third node and some generic message handling logic in handle message method. Also need to shut-down the hanging test(commented out currently).
This pull request includes major work from Peter as well.
PB: This code needs clean-up but should be merged for others to progress.


benjaminbollen:
the routing_node test hangs, best to take it out for now, until corrected (or comment it)

PR #83, opened 2015-04-13
Node info

benjaminbollen:
Made the `NodeInfo::endpoint` optional, as to smooth towards routing_node work by Prakash and Peter

### DIFF
--- a/src/routing_table.rs
+++ b/src/routing_table.rs
@@ -69,7 +69,7 @@ impl NodeInfo {
     self.endpoint
   }
 
-  pub fn set_endpoint(&self, new_endpoint: SocketAddr) {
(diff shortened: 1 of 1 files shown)


benjaminbollen:
Closing, as endpoint information should not have been in NodeInfo in the first place?

src/routing_table.rs
```diff
@@ -42,20 +42,36 @@ pub struct NodeInfo {
                    // should be immutable and can be read from
                    // public_pmid.get_name();
     pub fob: PublicPmid,
-    endpoint: SocketAddr,
-    connected: bool,
+    endpoint: Option<SocketAddr>,
+    connected: bool,  // TODO: can we remove this state-flag ? instead base it on Option<endpoint>
```
dirvine: Yes good TODO and we should do this.

PR #84, opened 2015-04-13
Filter

benjaminbollen:
First implementation of filter;  should still be revised with message-filter


src/routing_node.rs
```diff
@@ -14,9 +14,11 @@
 // Software.
 
 extern crate rand;
+extern crate chrono;
```
dirvine: I think the only place needing extern crate is lib.rs here. so this can be replaced with `use rand` and `use chrono` in this case I think anyway.

src/routing_node.rs
```diff
@@ -64,6 +65,7 @@ pub struct RoutingNode<F: Facade> {
     accepting_on: Option<u16>,
     next_message_id: MessageId,
     bootstrap_node_id: Option<DhtId>,
+    filter: LruCache<types::FilterType, ()>,
```
dirvine: I think this should be a message_filter type as opposed to an lru_cache ? (i.e no values)

benjaminbollen: yes, agreed, is add_key already implemented in message_filter?  I figured first put in lru_time_cache and then replace it with message filter.  Guess i lost too much time doing nothing over the weekend :)  I'll update it

PR #87, opened 2015-04-14
Adding DhtId::from_data

vtnerd:
This function will be used [here](https://github.com/dirvine/maidsafe_vault/blob/master/src/vault.rs#L55). Hopefully I didn't misunderstand how this value is calculated.

EDIT:
It will be used in the lines below inside of the function, not at that line obviously.


dirvine:
Looks good Lee, nice one

PR #100, opened 2015-04-20
re-expose crust::Endpoint as routing_client::Endpoint

maqi:


dirvine:
Published to crates.io as 0.1.1 Thx Qi

PR #112, opened 2015-04-23
DhtId to NameType

krishnaIndia:


src/name_type.rs
```diff
@@ -92,11 +100,53 @@ impl fmt::Debug for NameType {
...
+  #[inline]
+  fn ge(&self, other : &NameType) -> bool {
+    PartialOrd::ge(&&self.0[..], &&other.0[..])
+  }
 }
 
+// FIXME(Ben): please fix me
+impl hash::Hash for NameType {
+  fn hash<H: hash::Hasher>(&self, state: &mut H) {
+    state.write(&self.0[..])
```
ustulation: The example shows something like this:
```
impl Hash for Person {
    fn hash<H: Hasher>(&self, state: &mut H) {
        self.id.hash(state);
        self.phone.hash(state);
    }
}
```
Is this the same ?

krishnaIndia: Am not sure either. But I think, Hash trait is not implemented for [T;64], from the documentation (https://doc.rust-lang.org/std/hash/trait.Hash.html). So not sure, whether hash() can be called directly as in the example.

benjaminbollen:
Thanks Krishna !

krishnaIndia:
Cheers Ben :)

PR #113, opened 2015-04-23
Refactoring the GetGroupKeyResponse::merge function

inetic:


src/histogram.rs
```diff
@@ -0,0 +1,28 @@
+
+use std::collections::BTreeMap;
+
+pub struct Histogram<K: Ord + Clone> {
```
benjaminbollen: Nice one Peter, strictly speaking a histogram collects items in defined buckets; so is not a histogram in the strict sense of the word.  Hence I called it 'frequency count' which arguably is not a good name either.  Nice call regardless of semantics to abstract these functions !

inetic: Cheers Ben :)

dirvine:
@chandraprakash I am happy if you are @rossmuir to

chandraprakash:
May be worth having a test along with this PR.

benjaminbollen:
Starting to implement tests on this PR

chandraprakash:
@dirvine  we agreed on merging it. Travis shows failure because of nightly I think.

PR #116, opened 2015-04-23
OPTIONAL fix_build

benjaminbollen:


benjaminbollen:
worked with rustc 1.0.0-nightly (16e1fcead 2015-04-15) (built 2015-04-15); updating nightly

PR #117, opened 2015-04-23
OPTIONAL fix build part 2: cast in accumulator

benjaminbollen:


benjaminbollen:
Documentation will still fail

benjaminbollen:
Fixed documentation macro;  macro however takes ownership of collection.  Tried, but not found how to have it simply borrow the collection.  Adapted therefore additionally the example to prevent double-ownership of data

PR #118, opened 2015-04-23
Frequency test (previously called Histogram)

benjaminbollen:
Renamed Histogram -> Frequency (because a histogram can be constructed from a frequency count, but is not the same object)
Implemented a test to feed a random number of repeated keys in random order; which after processing by Frequency have to be returned as pairs of keys in increasing order; counts in decreasing order.

[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/118)


dirvine:
Reviewed files:
- src/frequency.rs @ r1
- src/histogram.rs @ r1
- src/lib.rs @ r1
- src/messages/find_group_response.rs @ r1
- src/messages/get_group_key_response.rs @ r1

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/routing/118)

PR #122, opened 2015-04-24
Update merge logic of FindGroupResponse GetGroupKeyResponse

benjaminbollen:
includes previous pull request and extends the same logic of merging target_id also for GetGroupKeyResponse; after handling FindGroupResponse in previous PR

[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/122)

### DIFF
--- a/src/messages/find_group_response.rs
+++ b/src/messages/find_group_response.rs
@@ -40,7 +40,7 @@ impl FindGroupResponse {
         for response in responses {
             freq_target_id.update(response.target_id.clone());
         }
-        // first identify the target_ids; Select all that are mentioned more than QuorumSize.
+        // first identify the target_ids;
         let target_ids : Vec<NameType> = freq_target_id.sort_by_highest()
                                        .iter()
                                        .map(|&(ref id, _ )| id.clone())
--- a/src/messages/find_group_response.rs
+++ b/src/messages/find_group_response.rs
@@ -18,7 +18,7 @@
 use cbor::CborTagEncode;
 use rustc_serialize::{Decodable, Decoder, Encodable, Encoder};
 use frequency::Frequency;
-use types::{PublicPmid, GROUP_SIZE};
+use types::{PublicPmid, GROUP_SIZE, QUORUM_SIZE};
 use NameType;
 
 #[derive(PartialEq, Eq, Clone, Debug)]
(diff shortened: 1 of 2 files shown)


dirvine:
Nice work, but we need to test this new code   please. Otherwise looks good. 

---

Reviewed files:
- src/messages/find_group_response.rs @ r1
- src/messages/get_group_key_response.rs @ r1

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/routing/122)

inetic:
@dirvine @benjaminbollen: what is the relationship between message.header.from_group.unwrap() value and the GetGroupKeyResponse::target_id, aren't they the same?

dirvine:
Good point, they are indeed the same AFAIK 

---









Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/routing/122)

inetic:
That was quick :), I asked because if they are the same, then all the messages passed to the merge function will have the same target_id (because they are indexed by the message.header.from_group key in the sentinel). And thus we don't need merge logic for it (the target_id).

benjaminbollen:
Good point.  My argument for merging on target_id was that if one single wrong message would be included, all messages would be unmergeable/dropped.  If Sentinel already performs this selection in the Accumulator, I agree that this merge is redundant.  There is a danger that relying on this selection outside of the merge function might logically break if at some point the external user does not enforce all target_id's to be identical.

So we can reduce the merge logic back down to first forcing them to be identical? and dropping merge if there are is one non-identitcal ?  Or keep the redundant logic?

inetic:
Makes sense Ben, though, I don't think the merge function should suffer in clarity due to possible mistakes done somewhere else. I would be happiest if we would remove the target_id member from the struct as it seems it is redundant anyway. It would save some bandwidth too.

benjaminbollen:
if we can arguably remove the target_id from the message struct then for sure I agree (obviously) that there is no need for merging on it (almost a mute statement).  I agree that the message header has the same information, but the content, i.e. the list of PublicPmids or PublicSignKeys is meaningless if it does not specify what the targeted id was in the first place.  I still favour the redundant storing (in header and message) and even the redundant verifying (in merging explicitly and by construction in Sentinel Accumulator)

inetic:
The way I see it is that having different target_ids in the merge function would be a bug in our code somewhere else than in the `merge` function. Fixing it in the `merge` function would be (IMO) fixing the effect of the bug rather than fixing the bug itself. Removing the target_id seems to only remove the degree of freedom that allows us to introduce such bugs. Then there is code clarity, if a new developer looks in the `merge` function (such as was my case last week) he may have an impression that having different target_ids there is something normal that has to be dealt with, while it is the case that such situation should never happen. All in all, **if** removing the target_id was an option, it would give us: less code (=> cleaner code), less possibility to make errors and would reduce bandwidth. If it wasn't an option there are still other ways to go about it (store only Vec<PublicPmid> inside Sentinel is one of them). Also, there is the question mark about whether this implementation of merge (the one in this PR), is the correct one (e.g. what if the target_id filter leaves you with fewer than GROUP_SIZE of messages) which is a problem we preferable wouldn't need to have.

benjaminbollen:
My argument is the following: the list of keys only has meaning with regards to the specified target_id.  Merging keys only has meaning when merging for the same target_id; otherwise we're comparing apples and pears.  If there are not enough messages with an identical target_id to merge keys then we should either return None or just return the best merge attempt possible - ie the merge function maybe should not be concerned about a QuorumSize; in which case Sentinel has to be doubly careful it is not treating the merge result as a false positive

PR #125, opened 2015-04-27
generic-sendable-type

ustulation:
Vaults will dump message onto this type

[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/125)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -612,44 +612,44 @@ mod test {
 
     struct NullInterface;
 
-    impl Interface for NullInterface {
-        fn handle_get(&mut self, type_id: u64, our_authority: Authority, from_authority: Authority,
-                    from_address: NameType, data: Vec<u8>) -> Result<Action, RoutingError> {
-            Err(RoutingError::Success)
-        }
-        fn handle_put(&mut self, our_authority: Authority, from_authority: Authority,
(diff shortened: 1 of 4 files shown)


src/routing_node.rs
```diff
@@ -612,44 +612,44 @@ mod test {
...
-        fn handle_cache_get(&mut self, type_id: u64, from_authority: Authority,
-                            from_address: NameType, data: Vec<u8>) -> Result<Action, RoutingError> {
-            Err(RoutingError::Success)
-        }
-        fn handle_cache_put(&mut self, from_authority: Authority, from_address: NameType,
-                            data: Vec<u8>) -> Result<Action, RoutingError> {
-            Err(RoutingError::Success)
-        }
-    }
+    // impl Interface for NullInterface {
```
dirvine: I think we would be better removing the commented block here please.

ustulation: Actually this file wasn't coded by me -  it's not used right now so i commented it out only because of new churn signature - do you want me to remove this anyway ?

dirvine:
Reviewed files:
- src/generic_sendable_type.rs @ r1
- src/lib.rs @ r1
- src/node_interface.rs @ r1
- src/routing_node.rs @ r2

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/routing/125)

PR #129, opened 2015-04-27
clean up of functions; better handling of Option<NameType>

benjaminbollen:
Continues our_authority in routing_node

Sorry; I agree it wasn't a priority, but I couldn't leave the code in that state

[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/129)


dirvine:
Reviewed files:
- src/routing_node.rs @ r1
- src/routing_table.rs @ r1

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/routing/129)

PR #134, opened 2015-04-27
Application of special merge logic for GetGroupKeyResponse

inetic:
Similar to the one that is used for FindGroupResponse. Also adds default merge logic for all other messages (AccountTransfer inclusive).

[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/134)


dirvine:
Reviewed files:
- src/sentinel.rs @ r1
- src/types.rs @ r1

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/routing/134)

PR #142, opened 2015-04-28
Our authority test

benjaminbollen:
copied test into fresh branch to get rid of Cargo.toml mess up

[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/142)


benjaminbollen:
this test also sets up a routing node and calls on routing table; necessarilly

PR #155, opened 2015-04-29
Fixes FindGroupResponse & GetGroupKeyResponse merge tests.

BSMaidSafe:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/155)


dirvine:
Are you still on the `embedded_find_group_response` test @BSMaidSafe ? It seems to still fail. Or is that another task ? Thanks for this one though

dirvine:
I wonder if we went with a GROUP_SIZE of 32 and QUORUM of 19 would it perhaps help the issue we were discussing earlier ?

---









Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/routing/155)

BSMaidSafe:
I had a word with Peter and he suggested looking into these. The Sentinel tests are pending, although, if the tests are correct, which still has to be verified, and the merge function is correct, which likewise needs verifying, then it's likely the failure won't be fixable without further work on the Sentinel code. I can take a look at varying GROUP/QUORUM size though. I'd prefer adding the code from the C++ version I was working on, it's not obvious what's going on in those tests.

dirvine:
I think more tests the better in sentinel so yes absolutely if you want to cut some over then I think its a great move. Also checking the GROUP of 32 would be nice to know.

BSMaidSafe:
Cool. I'll get on with that next.

dirvine:
Do you want this merged to start fresh or will you work on this PR

BSMaidSafe:
If this gets merged then there's failing tests to contend with, without a clear explanation of why they're failing. So closing this PR might be best and keep working from it. What do you think?

BSMaidSafe:
Ah, sorry, forget that last comment, there's already failing tests. Yes, merge this since it fixes two of the failures.

dirvine:
Ok Cheers chap.

PR #157, opened 2015-04-30
Routing actions - Done

ustulation:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/157)


src/node_interface.rs
```diff
@@ -20,6 +20,15 @@ use name_type::NameType;
 use types::{Authority, DestinationAddress};
 use super::{Action, RoutingError};
 
+pub enum RoutingNodeAction {
```
benjaminbollen: in lib.rs we have an `Action` object: 
```pub enum Action {
  Reply(Vec<u8>),
  SendOn(Vec<NameType>),
}```

Should this be taken into consideration?

dirvine:
Reviewed files:
- src/node_interface.rs @ r2
- src/routing_node.rs @ r2

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/routing/157)

PR #162, opened 2015-04-30
Changed GROUP_SIZE to 32.

BSMaidSafe:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/162)

### DIFF
--- a/src/messages/find_group_response.rs
+++ b/src/messages/find_group_response.rs
@@ -88,18 +88,15 @@ mod test {
 
     #[test]
     fn merge() {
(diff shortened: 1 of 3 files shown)


BSMaidSafe:
There was assertions in FindGroupResponse and GetGroupKeyResponse merge tests as follows,
assert!(GROUP_SIZE == 23) assert!(QUORUM_SIZE == 19). By removing those checks coverage went down 0.02%.

dirvine:
Yes seen that, I think coverage is more accurate than we think but how the source gets covered is weird.

PR #164, opened 2015-04-30
Message header sendon reply

benjaminbollen:
Create a single place to mutate headers on Action::SendOn or Action::Reply; this gets called by a managed node after the interface has returned an Action to routing

[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/164)

### DIFF
--- a/src/message_header.rs
+++ b/src/message_header.rs
@@ -118,6 +118,24 @@ impl MessageHeader {
     pub fn from_authority(&self) -> types::Authority {
         self.authority.clone()
     }
+
+    /// Only a node in a manager will send_on a routing message;
+    /// after this message is processed by the interface.
+    /// Note: this is not for XOR-forwarding; then the header is preserved!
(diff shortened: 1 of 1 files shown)


src/message_header.rs
```diff
@@ -118,6 +118,50 @@ impl MessageHeader {
...
+        // implicitly preserve all non-mutated fields.
+        let reply_header = self.clone();
+        reply_header.source = types::SourceAddress {
+            from_node : our_name.clone(),
+            from_group : self.destination.dest.clone(),
+            reply_to : self.source.reply_to.clone()
+        };
+        reply_header.destination = types::DestinationAddress {
+            dest : self.source.from().clone(),
+            reply_to : self.destination.reply_to.clone()
```
dirvine: Should this not be form the `self.source.reply_to()` so we swap it around form a source to destination when we reply?

PR #173, opened 2015-05-01
handle_get_data replies or sends on data received from interface.handle_...

mmoadeli:
Please note RoutingError needs further work

[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/173)


dirvine:
Again coverage decrease ;-) Apart from that nice work.

mmoadeli:
it is right though as untested code added

PR #175, opened 2015-05-01
test handle put_public_pmid

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/175)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -683,7 +683,7 @@ impl<F> RoutingNode<F> where F: Interface {
         // if data type is public pmid and our authority is nae then add to public_pmid_cache
         // don't call upper layer if public pmid type
         let put_public_pmid = try!(self.decode::<PutPublicPmid>(&body).ok_or(()));
-        let our_authority = match self.our_authority(&put_public_pmid.public_pmid.name, &header) {
+        match self.our_authority(&put_public_pmid.public_pmid.name, &header) {
             Authority::NaeManager => {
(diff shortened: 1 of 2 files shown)


src/routing_table.rs
```diff
@@ -266,7 +266,7 @@ impl RoutingTable {
         if self.routing_table.len() < GROUP_SIZE {
             return true;
         }
-        let furthest_close_node = self.routing_table[GROUP_SIZE].clone();
+        let furthest_close_node = self.routing_table[GROUP_SIZE - 1].clone();
```
benjaminbollen: Found a bug I introduced in RoutingTable before; will write accompanying test for `address_in_our_close_group`

benjaminbollen:
werid about the coverage result here:  the PR only adds a test; and removes a commented code block.  Probably the latter made coverage drop.

src/routing_node.rs
```diff
@@ -677,18 +677,24 @@ impl<F> RoutingNode<F> where F: Interface {
         Ok(())
     }
 
+    /// Every node needs to republish its public key directly to PublicPmid.name every 10 minutes.
+    /// Sentinel will query this pool.  No handle_get_public_pmid is needed.
```
chandraprakash: Can you remove above comment. there is no api for PutPublicPmid...
This only happens while bootstrapping.

src/routing_table.rs
chandraprakash: good catch !

PR #180, opened 2015-05-01
expand test our_close_group and _in_range

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/180)


dirvine:
can you please look at failures

---









Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/routing/180)

benjaminbollen:
Test `our_close_group_and_in_range` passes.  IMPORTANT change: `address_in_our_close_group_range` also returns true on the border: the furthest close_node in our close group is considered in our close_group_range

src/sentinel.rs
```diff
@@ -1269,6 +1269,7 @@ mod test {
   }
 
   #[test]
+  #[ignore]
```
dirvine: OK But needs added to readme.md  to fix

src/name_type.rs
```diff
@@ -111,6 +111,21 @@ pub fn closer_to_target(lhs: &NameType, rhs: &NameType, target: &NameType) -> bo
     false
 }
 
+/// Returns true if `lhs` is closer to `target` than `rhs`, or when `lhs == rhs`.
+/// "Closer" here is as per the Kademlia notion of XOR distance,
+/// i.e. the distance between two `NameType`s is the bitwise XOR of their values.
+pub fn closer_to_target_or_equal(lhs: &NameType, rhs: &NameType, target: &NameType) -> bool {
```
dirvine: How can it ever be equal?

benjaminbollen: address_in_our_close_group_range passes as right-hand side the furthest close group node; as such without the equal part, the last close node is not considered in range.  I can either make the tests exclude the last close node from being considered 'in range' or we change the implementation of 'address_in_our_close_group_range' to include up to the last close node.  This is done as such in this PR

PR #182, opened 2015-05-01
Sentinel test additions and type fixes.

BSMaidSafe:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/182)


dirvine:
The fake messages test fails Brian, have we a plan for that? thx

BSMaidSafe:
In effect it's a Sybil attack, allowing a node to store data freely, in this case. I think we need to revisit Sentinel to come up with a solution.

PR #183, opened 2015-05-02
Fix for message_index condition in tests.

BSMaidSafe:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/183)


dirvine:
Reviewed files:
- src/sentinel.rs @ r1

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/routing/183)

PR #184, opened 2015-05-04
Example routing node

maqi:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/184)

### DIFF
--- a/examples/routing.rs
+++ b/examples/routing.rs
@@ -133,7 +133,7 @@ impl Interface for TestNode {
                   from_authority: types::Authority, from_address: NameType)
                    -> Result<Action, RoutingError> {
         let stats = self.stats.clone();
-        let mut stats_value = stats.lock().unwrap();
+        let stats_value = stats.lock().unwrap();
         for data in stats_value.stats.iter().filter(|data| data.1.name() == name) {
             return Ok(Action::Reply(data.1.serialised_contents().clone()));
         }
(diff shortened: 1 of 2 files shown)


dirvine:
Reviewed files:
- examples/routing.rs @ r1
- src/routing_node.rs @ r1

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/routing/184)

examples/routing.rs
```diff
@@ -180,43 +210,92 @@ fn main() {
...
+            let v: Vec<&str> = command.split(' ').collect();
+            match v[0].trim() {
+                "stop" => break,
+                "bootstrap" => {
+                    let endpoint_address = match SocketAddr::from_str(v[1].trim()) {
+                        Ok(addr) => addr,
+                        Err(_) => continue
+                    };
+                    println!("bootstrapping to {} ", endpoint_address);
+                    let _ = mutate_node.lock().unwrap().bootstrap(Some(vec![Endpoint::Tcp(endpoint_address)]), None);
```
chandraprakash: @maqi  We always need to pass None for the bootstrap list option. Then only crust will try default methods like beacon.
If user provides an explicit bootstrap endpoint, then only we need to provide Some(Vec<Endpoint>) to the bootstrap method.

Refer crust documentation http://dirvine.github.io/crust/crust/connection_manager/struct.ConnectionManager.html#method.bootstrap
I will try to add more documentation to routing headers tomorrow to make this clear.

PR #206, opened 2015-05-05
adds unauthorised_put in routing_client

mmoadeli:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/206)


mmoadeli:
there is no test for the code, as it is identical to the one in routing_node. also, adding test for this requires adding complete test code available in routing_node

chandraprakash:
Thanks mahmoud !  I will have another task to write routing_client tests.
@dirvine  good to go

PR #214, opened 2015-05-06
WIP:blocking bootstrap

mmoadeli:
the entries in bootstrap_list are tried one by one. on connection establishment, if the bootstrap_node_id is not received in 3 seconds, another endpoint is tried.
If none of the bootstrap_list entries succeed, beacon if tried. On failure, error is returned,

[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/214)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -129,11 +129,11 @@ impl<F> RoutingNode<F> where F: Interface {
     pub fn get(&mut self, type_id: u64, name: NameType) {
         let destination = types::DestinationAddress{ dest: NameType::new(name.get_id()),
                                                      reply_to: None };
-        let header = MessageHeader::new(self.get_next_message_id(), 
-                                        destination, self.our_source_address(), 
+        let header = MessageHeader::new(self.get_next_message_id(),
+                                        destination, self.our_source_address(),
                                         types::Authority::Client);
-        let request = GetData{ requester: self.our_source_address(),  
-                               name_and_type_id: NameAndTypeId{name: NameType::new(name.get_id()), 
+        let request = GetData{ requester: self.our_source_address(),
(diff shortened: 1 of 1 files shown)


mmoadeli:
This work requires that in each iteration, on calling connection_manager::bootstrap, that any existing connection to be dropped.

mmoadeli:
After talking to David, It was mentioned that due to security issues, multiple bootstrap nodes are to be tried in parallel. This is not conforming to the initial specification of the task and needs to be re-written.

mmoadeli:
A node connects to the network through  bootstrap node/nodes. As no single node can be trusted, routing tries to connects to a number of nodes to assist it join the network. The RPCs required for a node to join the network should go through these multiple bootstrap nodes. Achieving this requires works on routing and crust libraries.

In crust:
    bootstrap function which currently returns the endpoint to one connected node should be changed to return connect to multiple nodes and return list of all successfully connected  nodes.

In routing: 
-bootstrap node related variables should become a vector 
-bootstrap should be changed to send bootstrap info request in parallel. 
-changing bootstrap(...) to join(...) as Prakash suggested should be considered
-send_to_bootstrap_node should potentially offer sending to multiple bootstrap nodes in parallel
- ....

dirvine:
Is this still valid? 

---

Review status: 0 of 1 files reviewed, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/routing/214)

maidsafe-highfive:

maidsafe-highfive:

maidsafe-highfive:

benjaminbollen:
@mmoadeli do we close this PR?

PR #221, opened 2015-05-07
Issue 204 routing_client::run() needs to be internal

### LINKED ISSUE #204: routing_client::run() should be internal and not public
routing_client::run() needs to be internal to routing and upper layer should not have any control on this. This should be called inside bootstrap method (non public method). This method allows routing to accept messages from crust.

This PR actually made run public method. https://github.com/dirvine/routing/pull/200
Not reverting the PR as its merged and other parts of the PR are valid.

similar to routing node task https://github.com/dirvine/routing/issues/202

maqi:
resolve Issue #204

[Review on Reviewable](https://reviewable.io/reviews/dirvine/routing/221)


dirvine:
<sup>**[src/routing_client.rs, line 218 \[r1\]](https://reviewable.io:443/reviews/dirvine/routing/221#-JohXatoX1kQCZszRUis)** ([raw file](https://github.com/dirvine/routing/blob/4c113cad340212a4441ceb6f3377e52fe9ef7ab2/src/routing_client.rs#L218)):</sup>
Could we perhaps rename this `join` Bootstrap will really be handled by crust and we can define join more accurately so for client it's when you can login and node when you have your close nodes in your routing_table. 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/routing/221)

src/routing_client.rs
```diff
@@ -156,36 +158,82 @@ impl Decodable for ClientIdPacket {
     }
 }
 
-pub struct RoutingClient<F: Interface> {
+pub struct RoutingClientState<F: Interface> {
```
chandraprakash: Is this not an internal state ? so , don't need pub

benjaminbollen:
@maqi; can you in a spare moment resolve these merge conflicts, then I'm happy to merge, with the remark that we need to reconsider the pattern.  Mutable state should only every be owned in a single thread; instead ownership (of messages) should be passed over through channels.

But we will address this later in this sprint.

Cheers
Ben

maqi:
although the conflict is resolved, due to the change of bootstrap id request/response flow, the previous workable routing client example becomes invalid.
It will be better to have someone familiar with the new bootstrap id request/response flow to debug it.

As the current pattern seems not ideal, currently closing this PR.

benjaminbollen:
Ok thanks Qi.  Noted

PR #237, opened 2015-05-11
test connection lost

mmoadeli:
tests connection lost. There seem to be a bug in crust which causes connection to be disconnected immediately after it is established.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/237)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -214,6 +214,7 @@ impl<F> RoutingNode<F> where F: Interface {
                 self.handle_connect(endpoint);
             },
             crust::Event::LostConnection(endpoint) => {
+                println!("lost connection");
                 self.handle_lost_connection(endpoint);
             }
         }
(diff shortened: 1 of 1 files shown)


dirvine:
It will be good to let @chandraprakash  and @Fraser999  take a look here Mahmoud as the CRUST tests do not show this behaviour (of disconnecting). Thx though nice work

dirvine:
Review status: all files reviewed, all discussions resolved, some commit checks pending.
Reviewed files:
- src/routing_node.rs @ r1

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/routing/237)

maidsafe-highfive:

maidsafe-highfive:

maidsafe-highfive:

dirvine:
@mmoadeli  Can you please confirm you still want this PR and if so please look at conflicts.

mmoadeli:
@dirvine This is the PR which you asked to be stopped as it had some issues, mainly the fact that it does not conform to the real use case, where each routing node is run as one process. Therefore, I close it.

PR #238, opened 2015-05-12
Further refactor of error handling

inetic:
Some errors were being ignored at low levels (such as sending), so I tried to move the 'ignoring' upwards. Also trying to make the ignoring explicit using the `ignore` free function.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/238)

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -173,11 +173,6 @@ impl Decodable for RoutingMessage {
 }
 
 impl RoutingMessage {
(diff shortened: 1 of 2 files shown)


dirvine:
Reviewed files:
- src/messages.rs @ r1
- src/routing_node.rs @ r1

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/routing/238)

PR #251, opened 2015-05-13
PAUSED Initial commit for using on_churn

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/251)

### DIFF
--- a/docs/messages.md
+++ b/docs/messages.md
@@ -1,0 +1,5 @@
+# Messages
+
+|Message | test|
+ |--------|------|
+ |fsdl|dfkjdsfhkj|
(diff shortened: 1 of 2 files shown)


dirvine:
In this can we also add in a 'fix' in swarm the swarm target is your group, this is wrong. The swarm target should be only to those in your group close to the target. So the swarm overlaps the group but only just. At the moment is swarms to your group and will do this across the network, so every node will see every other node, which is a mistake. 
So swarm should only send messages to nodes in your group that are in the group close to the target and then only sent onwards to any node closer than you. So is stops transmitting across the whole network. So you sort only your close group nodes and the target and send to only nodes closer.
The first time a message arrives that is in your close group (closer than furthest close nodes) then we transmit to whole close group, on retransmit then only to nodes closer.  We can tell second retransmit as target is not closer than your furthest close node.

PR #253, opened 2015-05-16
replace owner with owner_sign_public_key

mmoadeli:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/253)

### DIFF
--- a/src/sendable.rs
+++ b/src/sendable.rs
@@ -16,6 +16,7 @@
 // relating to use of the SAFE Network Software.
 
 use name_type;
+use sodiumoxide::crypto;
 
 /// This trait is required for any type of message to be
 /// passed to routing, refresh / account transfer is optional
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @benjaminbollen

dirvine:
Review status: 0 of 1 files reviewed, 1 unresolved discussion, some commit checks pending.

---

<sup>**[src/sendable.rs, line 28 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/253#-JpTmphhmftDz-znpqgw)** ([raw file](https://github.com/maidsafe/routing/blob/8f3cc7023071136dd05d2ff45eeb360ec2c0dc86/src/sendable.rs#L28)):</sup>
Is this valid as sentinel for instance needs to know the actual owner to validate the data/message. Is the idea we will provide the key and if so how do we get it?  Usually for instance sentinel will get the owner id then ask for the keys to be also downloaded and match them.  If this is just for certain ID types where we can get teh actual public key from them I would see the point, but this trait is to cover all data types so this change would break that. I.e. if we get a bit of structured data we need to know the owner and then get their keys from the network. 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/routing/253)

maidsafe-highfive:

maidsafe-highfive:

maidsafe-highfive:

mmoadeli:
According to @dirvine and @Fraser999 said. I revert Sendable to what it was and apply changes asked by Fraser.

PR #256, opened 2015-05-20
Add Network Visualiser

benjaminbollen:
Add Network Visualiser at end of sprint


maidsafe-highfive:
r? @inetic

PR #258, opened 2015-05-21
Include signature in derivation of name.

s-brian:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/258)

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -23,7 +23,7 @@ use cbor::CborTagEncode;
 use rustc_serialize::{Decodable, Decoder, Encodable, Encoder};
 use rand::random;
 use sodiumoxide;
-use sodiumoxide::crypto::sign::SIGNATUREBYTES;
+use sodiumoxide::crypto::sign::{PUBLICKEYBYTES, SIGNATUREBYTES};
 use NameType;
 use std::fmt;
 use error::ResponseError;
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @inetic

benjaminbollen:
seems good to me. I'll leave it to you Peter

inetic:
Hi Brian, looks good. But I have a remark about the original code, maybe we could address it in this PR? Gonna write another comment.

s-brian:
Is this what you were thinking?

src/types.rs
```diff
@@ -271,7 +272,7 @@ impl Id {
     let sign_arr = &pub_sign_key.0;
     let asym_arr = &pub_asym_key.0;
 
-    let mut arr_combined = [0u8; 64 * 2];
+    let mut arr_combined = [0u8; 64 * 2 + SIGNATUREBYTES];
```
inetic: I'm wondering why we use the number 64 here. If I understand it correctly `arr_combined` consists of:

[PublicKey|PublicKey|Signature]

The first PublicKey I think is from [sodiumdioxide/crypto/sign/ed25519.rs file](https://github.com/dnaq/sodiumoxide/blob/master/src/crypto/sign/ed25519.rs#L11) and the second one is from [sodiumdioxide/crypto/asymmetricbox/curve25519xsalsa20poly1305.rs file](https://github.com/dnaq/sodiumoxide/blob/4429e27ae6359434e01d088f7dc6351a90e5e070/src/crypto/asymmetricbox/curve25519xsalsa20poly1305.rs#L12). Both I believe have size 32 bytes which would make the `64*2` calculation incorrect.

Ideally it would be written something like:
[0u8; sign::PUBLICKEYBYTES + asymmetricbox::PUBLICKEYBYTES + SIGNATUREBYTES]

inetic:
yap

s-brian:
I'll change it to your suggestion on the outside chance that these two values ever differ.

benjaminbollen:
so hold the horses, Brian can you make the correction as suggested: that asymmetric box key and signing key are better mentioned explicitly, rather than *2 ?

PR #259, opened 2015-05-21
Use values specific to key type.

s-brian:


maidsafe-highfive:
r? @benjaminbollen

PR #260, opened 2015-05-22
ConnectResponse-ConnectRequest

s-brian:


maidsafe-highfive:
r? @inetic

src/routing_node.rs
```diff
@@ -593,6 +592,14 @@ impl<F> RoutingNode<F> where F: Interface {
...
         println!("{:?} received ConnectResponse", self.own_name);
         let connect_response = try!(decode::<ConnectResponse>(&body));
+
+        // Verify a connect request came from us.
+        if !verify_detached(&connect_response.connect_request_signature.get_crypto_signature(),
+                            &connect_response.serialised_connect_request[..],
+                            &self.id.get_crypto_public_sign_key()) {
+            return Err(RoutingError::Response(ResponseError::InvalidRequest));
+        }
+
```
benjaminbollen: Can we parse the serialised_connect_request again and check that the connection request corresponds to the connect response? Otherwise we can have a replay attack where someone picks up an arbitrary connect_request we've signed and uses that to send us an unwanted connect_response.

PR #261, opened 2015-05-22
MAID-1032 update.

s-brian:


maidsafe-highfive:
r? @inetic

src/types.rs
```diff
@@ -271,26 +271,33 @@ impl Id {
...
+
+    let mut combined = [0u8; KEYS_SIZE + sign::SIGNATUREBYTES];
+
+    for i in 0..KEYS_SIZE {
+        combined[i] = keys[i];
+    }
 
     for i in 0..sign::SIGNATUREBYTES {
-        arr_combined[sign::PUBLICKEYBYTES + asymmetricbox::PUBLICKEYBYTES + i] = validation_token.signature[i];
+        combined[KEYS_SIZE + i] = validation_token.signature[i];
```
inetic: Looks correct, but perhaps a bit inefficient. You could have avoided some copying with code such as this one:
```
let i: usize = 0;
for value in sign_key.iter().chain(asym_key.iter()).chain(validation_token.signature.iter()) {
    combined[i] = value;
    i += 1;
}
```
Though, who knows what's happening behind the scenes with iterators, so this may be subjective. Let me know what you think.

benjaminbollen: I find it a bit harder to read and probably takes the same amount of 'copying work'?  Good to go for now?

inetic: Fine with me, Brian's call though.

s-brian: The amount of copying will be the same, I do like the use of iterators though. I guess it depends on how the compiler optimizes the code and since Id's aren't created continually I'd leave it as it is. You can always sneak it in as part of another task.

PR #264, opened 2015-05-22
Tidied function address_in_close_group_range, and miraculously fixes …

s-brian:
…'not for us' output.

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -33,7 +33,7 @@ use crust;
 use lru_time_cache::LruCache;
 use message_filter::MessageFilter;
 use NameType;
-use name_type::{closer_to_target, NAME_TYPE_LEN};
+use name_type::{closer_to_target_or_equal, NAME_TYPE_LEN};
 use node_interface;
 use node_interface::Interface;
 use routing_table::{RoutingTable, NodeInfo};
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/routing_node.rs
```diff
@@ -1039,8 +1039,12 @@ impl<F> RoutingNode<F> where F: Interface {
             return true;
         }
 
-        let close_group = self.routing_table.our_close_group();
-        closer_to_target(&address, &self.routing_table.our_close_group().pop().unwrap().id(), &self.own_name)
+        match self.routing_table.our_close_group().pop() {
+            Some(furthest_close_node) => {
+                closer_to_target_or_equal(&address, &furthest_close_node.id(), &self.own_name)
```
dirvine: think the or_equal part may be where this is happening so multiple calls from same node is not closer but may be equal ? May be worth checking here with routing_table.check_node() call that we do not already have the node perhaps?

s-brian: Where what is happening? Turns out the' not for us' issue is a non issue just now anyway. The 'or_equal' is including the furthest close node as part of the group which both Ben and I think is correct.

dirvine: No worries, I was concerned we were seeing a node that matched here as OK to add to routing_table, if not that is totally correct. It was a quick scan so I am happy if you guys are.

s-brian: Got it, no, it occurs after that check. Thanks for clarifying.

dirvine: :+1:

benjaminbollen: Happy with it too.  Will Merge

PR #269, opened 2015-05-25
Interface change for async operations.

ustulation:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/269)

### DIFF
--- a/examples/routing.rs
+++ b/examples/routing.rs
@@ -345,7 +345,7 @@ fn main() {
         let sign_keypair = crypto::sign::gen_keypair();
         let encrypt_keypair = crypto::asymmetricbox::gen_keypair();
         let client_id_packet = ClientIdPacket::new((sign_keypair.0, encrypt_keypair.0), (sign_keypair.1, encrypt_keypair.1));
-        let test_client = RoutingClient::new(TestClient { stats: Arc::new(Mutex::new(Stats {stats: Vec::<(u32, TestData)>::new()})) }, client_id_packet);
+        let test_client = RoutingClient::new(Arc::new(Mutex::new(TestClient { stats: Arc::new(Mutex::new(Stats {stats: Vec::<(u32, TestData)>::new()})) })), client_id_packet);
         let mutate_client = Arc::new(Mutex::new(test_client));
         let copied_client = mutate_client.clone();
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @benjaminbollen

examples/routing.rs
```diff
@@ -345,7 +345,7 @@ fn main() {
         let sign_keypair = crypto::sign::gen_keypair();
         let encrypt_keypair = crypto::asymmetricbox::gen_keypair();
         let client_id_packet = ClientIdPacket::new((sign_keypair.0, encrypt_keypair.0), (sign_keypair.1, encrypt_keypair.1));
-        let test_client = RoutingClient::new(TestClient { stats: Arc::new(Mutex::new(Stats {stats: Vec::<(u32, TestData)>::new()})) }, client_id_packet);
+        let test_client = RoutingClient::new(Arc::new(Mutex::new(TestClient { stats: Arc::new(Mutex::new(Stats {stats: Vec::<(u32, TestData)>::new()})) })), client_id_packet);
```
benjaminbollen: Can we try to avoid introducing more mutexes.  I do not like it.

ustulation: Ok i went through the Routing code. Here's what i see: `RoutingClient` object wants a cross-thread shareable instance of interface anyway. That is why it has `Arc<Mutex<F>>` as its member variable. So the constructor `new(..)` must also accept `Arc<Mutex<F>>` instead of just `F` otherwise it doesn't make much sense. The calls to `RoutingClient` are non-blocking so there will eventually be a need to synchronize somewhere and this is possible only if I pass the **same** object (which implements `Interface`) which I create in `MaidSafe-Client` crate and for this I would need to package it in `Arc<Mutex<F>>`. 

It is not very apparent why this is so in your examples because the only thing that happens there is the printing of the obtained result which is trivial. In actual scenarios (like in `Clients` crate once the data is obtained from routing there needs to be a cross threaded signalling to allow obtaining and dealing with that data in appropriate way which will require either:
<1> an exposure of the passed interface object through `RoutingClient`
OR
<2> the way I've done it above which will help retain the interface object at the callee site.

So API change is in order anyway i think.

PR #276, opened 2015-05-27
MAID-1040 Added assign_relocated_name and is_relocated methods to PublicId and Id.

chandraprakash:
Almost done need to get commented out test compiling.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/276)

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -220,7 +220,7 @@ pub struct PublicId {
   pub public_sign_key: PublicSignKey,
   pub validation_token: Signature,
   name: NameType,
-  original_name : NameType,
+  relocated: bool,
 }
 
 impl PublicId {
@@ -230,7 +230,7 @@ impl PublicId {
         public_sign_key : id.get_public_sign_key(),
         validation_token : id.get_validation_token(),
         name : id.get_name().clone(),
-        original_name : id.get_original_name(),
+        relocated : id.relocated,
       }
     }
 
@@ -245,12 +245,14 @@ impl PublicId {
     }
 
     pub fn is_relocated(&self) -> bool {
-        self.name != self.original_name
+        self.relocated
     }
 
     pub fn assign_relocated_name(&mut self, relocated_name: NameType) {
-        debug_assert!(!self.is_relocated(), "already assigned relocated name !");
-        self.name = relocated_name
+        debug_assert!(!self.relocated, "already assigned relocated name !");
+        debug_assert!(self.name != relocated_name, "using original name as relocated name !");
+        self.name = relocated_name;
+        self.relocated = true;
     }
 }
 
@@ -260,7 +262,7 @@ impl Encodable for PublicId {
                                    &self.public_sign_key,
                                    &self.validation_token,
                                    &self.name,
-                                   &self.original_name)).encode(e)
+                                   &self.relocated)).encode(e)
   }
 }
 
@@ -268,11 +270,11 @@ impl Decodable for PublicId {
   fn decode<D: Decoder>(d: &mut D)->Result<PublicId, D::Error> {
     try!(d.read_u64());
     let (public_key, public_sign_key,
-         validation_token, name, original_name) = try!(Decodable::decode(d));
+         validation_token, name, relocated) = try!(Decodable::decode(d));
     Ok(PublicId { public_key: public_key,
                     public_sign_key : public_sign_key,
                     validation_token: validation_token, name : name,
-                    original_name : original_name })
+                    relocated : relocated })
   }
 }
 
@@ -285,7 +287,7 @@ pub struct Id {
   secret_keys: (crypto::sign::SecretKey, crypto::asymmetricbox::SecretKey),
   validation_token: Signature,
   name: NameType,
-  original_name : NameType,
+  relocated : bool,
 }
 
 impl Id {
@@ -326,7 +328,7 @@ impl Id {
       secret_keys : (sec_sign_key, sec_asym_key),
       validation_token : validation_token,
       name : NameType::new(digest.0),
-      original_name : NameType::new(digest.0),
+      relocated : false,
     }
   }
 
@@ -334,9 +336,6 @@ impl Id {
       &self.name
   }
 
-  pub fn get_original_name(&self) -> NameType {
-      self.original_name.clone()
-  }
   pub fn get_public_key(&self) -> PublicKey {
       PublicKey::new(self.public_keys.1.clone())
   }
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @inetic

src/types.rs
```diff
@@ -226,22 +229,38 @@ impl PublicId {
...
         validation_token : id.get_validation_token(),
-        name : id.get_name()
+        name : id.get_name(),
+        original_name : id.get_original_name(),
       }
     }
 
+    pub fn name(&self) -> NameType {
+      self.name.clone()
+    }
```
inetic: Can we do it without the implicit cloning? In many places we just need the reference.

src/types.rs
```diff
@@ -303,34 +325,44 @@ impl Id {
       public_keys : (pub_sign_key, pub_asym_key),
       secret_keys : (sec_sign_key, sec_asym_key),
       validation_token : validation_token,
-      name : NameType::new(digest.0)
+      name : NameType::new(digest.0),
+      original_name : NameType::new(digest.0),
```
inetic: Are we ever going to reuse the `original_name`? If not, we can just store the `is_relocated` information in a `bool`. Ids are going to be copied over the network, so this would be a wasted 64Bytes.

inetic: Whops, Id isn't serialisable, but the PublicId is. Still would be 64B wasted in memory if we don't need it.

benjaminbollen: I'm open for either; but if we don't store original_name, then I would also not store a bool, because at that point it's better to just recalculate the name from the provided fields and compare

benjaminbollen: So I did advise Prakash to keep the original name field for now, and we can take it out later?  Would that be okay for you Peter?

chandraprakash: Yeah .. bool make more sense to me. I will update it to bool.

inetic:
The changes look correct to me. I am a little unsure about the task itself though. The Id and PublicId structures looked to me like a self sufficient structure that contained exactly what they needed to go about cryptography, signature verification and authentication. Now they contain a field that to me seems to have different semantics. Maybe I'm wrong, I would need more time to know better. An alternative that I would like to propose is something like this (replace with better names):

```
enum PossiblyRelocatedId {
    NonRelocated(Id),
    Relocated(Id)
}
```

That is, we're now not implanting the `boolean` into the structures itself. That way, for a function that previously took Id (or PublicId) we would provide it with the same amount of information - no more, no less. Canonical example is sending, while sending the boolean is much better than sending another NameType, if the other end doesn't need to know about it, then I think we shouldn't send it.

inetic:
As per discussion with Ben: r?@benjaminbollen

src/types.rs
benjaminbollen: +1

src/types.rs
```diff
@@ -212,12 +212,15 @@ impl Decodable for PublicKey {
...
+// Note: name field is initially same as original_name, this should be replaced by relocated name
+// calculated by the nodes close to original_name by using assign_relocated_name method
 #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Debug)]
 pub struct PublicId {
   pub public_key: PublicKey,
   pub public_sign_key: PublicSignKey,
   pub validation_token: Signature,
-  pub name: NameType
+  name: NameType,
+  relocated: bool,
```
benjaminbollen: I would favour to not have a bool at all, simply the name field.  either it corresponds to the calculated result on the other fields, or it doesn't.  If it doesn't, it is only valid if it is already stored in a routing table of a close group.  This only works because vaults are non-persistent.

src/types.rs
```diff
@@ -226,22 +229,40 @@ impl PublicId {
...
+      self.name.clone()
+    }
+
     pub fn serialised_contents(&self)->Vec<u8> {
         let mut e = cbor::Encoder::from_memory();
         e.encode(&[&self]).unwrap();
         e.into_bytes()
     }
+
+    pub fn is_relocated(&self) -> bool {
```
benjaminbollen: I asked for a calculated function here, not the implicit bool.  The preceding alternative was to have both name fields, but that has been discarded as an idea

src/types.rs
```diff
@@ -226,22 +229,40 @@ impl PublicId {
...
+        self.relocated
+    }
+
+    pub fn assign_relocated_name(&mut self, relocated_name: NameType) {
+        debug_assert!(!self.relocated, "already assigned relocated name !");
+        debug_assert!(self.name != relocated_name, "using original name as relocated name !");
```
benjaminbollen: this is a hard condition, where name should be recalculated, and if the stored name is already different from the calculated name, refuse re-assigning it.  The function is best to return the resulting name

src/types.rs
```diff
@@ -303,34 +338,48 @@ impl Id {
...
       secret_keys : (sec_sign_key, sec_asym_key),
       validation_token : validation_token,
-      name : NameType::new(digest.0)
+      name : NameType::new(digest.0),
     }
   }
 
-  pub fn get_name(&self) -> NameType {
-    self.name.clone()
+  pub fn get_name<'a>(&'a self) -> &'a NameType {
```
benjaminbollen: Can you return this to returning the actual NameType with implicit clone inside of the function?

chandraprakash: It was initially doing implicit copy and then you and Peter both agreed on changing it (see +1 on your previous comment). 
Its not that clean either way actually. Can you guys give me a final comment saying what is the final call on this? I am happy to revert it back to implicit copy. @inetic @benjaminbollen

benjaminbollen: I'm happy to make a final call on this; sorry about the 'switching back' but I'd prefer for now to see it back as an original pass back a clone, not a borrow

src/types.rs
benjaminbollen: -1 ;-) reverting my support for this

inetic: Can you explain why?

src/types.rs
```diff
@@ -480,4 +530,56 @@ mod test {
         let obj_after: PublicId = d.decode().next().unwrap().unwrap();
         assert_eq!(obj_before, obj_after);
     }
+
+
+#[test]
+    fn assign_relocated_name_public_id() {
+        let before = PublicId::generate_random();
+        let original_name = before.name();
+        assert!(!before.is_relocated());
```
benjaminbollen: Nice, this verifies that the double implementation generates the same name!

src/types.rs
benjaminbollen: because that is the dominant way it is currently being done.

PR #280, opened 2015-05-29
Initial commit to allow collaboration with MAID-1039

benjaminbollen:
Put bare signature of handlers for handle_put_public_id_response

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/280)


maidsafe-highfive:
r? @inetic

PR #284, opened 2015-06-01
Maid-1112 new put flow

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/284)

### DIFF
--- a/src/node_interface.rs
+++ b/src/node_interface.rs
@@ -29,7 +29,6 @@ pub enum MethodCall {
     Post,
     Refresh { content: Box<Sendable>, },
     PutResponse { destination: NameType, payload: Vec<u8> },
-    Error
 }
 
 pub trait Interface : Sync + Send {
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/node_interface.rs
```diff
@@ -28,6 +28,7 @@ pub enum MethodCall {
     Get { type_id: u64, name: NameType, },
     Post,
     Refresh { content: Box<Sendable>, },
+    PutResponse { destination: NameType, payload: Vec<u8> },
```
benjaminbollen: Looks good to me; routing can cater for this

benjaminbollen: Can we rename this though to 'SendOn { destination : NameType }`;  A PutResponse is only generated after conversion from a Put on Error or Reply;  So you can add an error with with payload Vec<u8> to have this payload included.

maqi: OK, if routing  guarantees the payload will be attached to the PutResponse msg.

benjaminbollen:
I'll do it myself

PR #285, opened 2015-06-01
Revert "Maid-1112 new put flow"

benjaminbollen:
Reverts maidsafe/routing#284

The decoding and encoding of the error would lose the payload of the error message

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/285)


maidsafe-highfive:
r? @inetic

benjaminbollen:
Better to only revert the last commit by Qi for now

PR #298, opened 2015-06-05
Maid 1126: Reflect changes in Source / Destination

mmoadeli:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/298)

### DIFF
--- a/src/message_header.rs
+++ b/src/message_header.rs
@@ -161,7 +161,13 @@ impl MessageHeader {
             relayed_for: None
         };
         reply_header.destination = types::DestinationAddress {
-            dest : self.from().clone(),
+            dest : match self.source.reply_to.clone() {
+                       Some(reply_to) => reply_to,
+                       None => match self.source.from_group.clone() {
+                           Some(group_name) => group_name,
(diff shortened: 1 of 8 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/message_header.rs
```diff
@@ -156,11 +157,12 @@ impl MessageHeader {
         reply_header.source = types::SourceAddress {
             from_node : our_name.clone(),
             from_group : Some(self.destination.dest.clone()),
-            reply_to : None
+            reply_to : None,
+            relayed_for: None
         };
         reply_header.destination = types::DestinationAddress {
             dest : self.from().clone(),
```
benjaminbollen: This was wrong in the old code: can we make
dest : match self.source.reply_to {
   Some(reply_to) => reply_to,
   None => match self.from_group {
      Some(group_name) => group_name,
      None => self.from_node

benjaminbollen:
Looks good to me !  Just the one comment that was not part of the task, but if you can correct that as well, then thats perfect.

PR #301, opened 2015-06-08
WIP: 1066 - Routing node example

mmoadeli:
the task is suspended as it requires some work done on routing_node and routing_membrane

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/301)


maidsafe-highfive:
r? @benjaminbollen

PR #303, opened 2015-06-08
simplifies routing node creation

mmoadeli:
suggestion aiming to simplify routing node creatiion


maidsafe-highfive:
r? @inetic

PR #304, opened 2015-06-08
Maid 1071 routing node

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/304)


maidsafe-highfive:
r? @inetic

src/routing_membrane.rs
```diff
@@ -340,6 +352,36 @@ impl<F> RoutingMembrane<F> where F: Interface {
         // TODO: trigger churn on boolean
     }
 
+    /// Parse and update the header with us as a relay node;
```
inetic: I think we shouldn't wrap the `name` in the `Relay` enum at line 206 and then check whether it really is wrapped in `Relay` here. Instead, the `received_from` should have directly the type `NameType`.

benjaminbollen: I agree it is not necessary; I introduced it as a logical-debug-tool;  I wanted to make sure that I could not by accident pass a relay message into the SAFE network without the compiler complaining.  in L206 we tag the name with its source (ie a message that was already on the SAFE network, or a to-be relayed message/to-be introduced onto the SAFE network by the relay node

src/relay.rs
```diff
@@ -170,6 +172,12 @@ impl RelayMap {
     pub fn pop_accepted_connect_request(&mut self, endpoint: &Endpoint) -> Option<PublicId> {
         self.accepted_connect_requests.remove(endpoint)
     }
+
+    /// Returns true if the relay map was instantiated with a self_relocated id.
+    /// A self_relocated id should only be used by the first node to start a network.
+    pub fn zero_node(&self) -> bool {
```
inetic: Just a suggestion, but I would prefer if this was called `is_self_relocated` instead of introducing new terminology.

PR #309, opened 2015-06-09
Maid 1070 tie up loose ends

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/309)

### DIFF
--- a/src/routing_membrane.rs
+++ b/src/routing_membrane.rs
@@ -596,30 +596,33 @@ impl<F> RoutingMembrane<F> where F: Interface {
 
     fn our_source_address(&mut self, from_group: Option<NameType>) -> types::SourceAddress {
         // first check whether we are safe to drop our bootstrap connection
+        let mut relayed_for : Option<NameType> = None;
         match self.bootstrap_endpoint.clone() {
             Some(endpoint) => {
                 // this threshold is set arbitrarily
                 if self.routing_table.size() > 5 {
                     self.connection_manager.drop_node(endpoint);
                     self.bootstrap_endpoint = None;
+                } else {
+                    relayed_for = Some(self.own_name.clone());
                 }
             },
             None => {}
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @inetic

src/routing_membrane.rs
```diff
@@ -590,19 +629,35 @@ impl<F> RoutingMembrane<F> where F: Interface {
...
+                              // if a message is sent over a relay connection,
+                              // the relay node will fill-in reply_to field with its name
+                              // if there is no reply_to field, relayed_for is ignored;
+                              // so a message can be sent out over both routing table and
+                              // a bootstrap connection and both should find their way back
+                              // (provided the bootstrap connection is not destroyed since;
+                              // but then then routing table should amply have it covered.)
+                              reply_to: None,
+                              relayed_for: relayed_for
+        }
```
inetic: Ben, I might be mistaken, but I think this is a significant change from how the header was previously meant to work. I believe that previously, if we had a message sent from node A to a relay node R, the relay node would set the `from_node` to R and `reply_to` to A. While with this code, the comment suggests that the relay node would leave `from_node` intact and set `reply_to` to R. Subjectively I liked the previous approach better, objectively, the header gets bigger and we'll need to be extra careful when reusing the code from the old RoutingNode. I'll merge if you confirm you know what you're doing, but would like to revisit the necessity of this change once there is more time.

benjaminbollen: Thanks for catching this !  You are right to point this out; I'll make a change and that should have this covered.

benjaminbollen:
Thanks Peter!  That was a good catch; I made it explicit with a separate function; the only use case for now is that we need to send out the initial FindGroup message and the initial connect requests over this bootstrap endpoint

Issue #318, opened 2015-06-10
Unable to create connection with simple_key_value_store example - failure due to Connection refusedbootstrap


rossmuir:
Testing environment : `Ubuntu 15.04 / rustc 1.2.0-nightly (2015-06-09)`

Unable to create connection between two nodes using simple_key_value_store example (in either debug or release).

Starting the first node (output from debug example) >
```
$ ./simple_key_value_store --type=first
ZERO listening on Some(Tcp(V4(127.0.0.1:50665)))
Input command (stop)
Refused unknown connection from Tcp(V4(127.0.0.1:36002))
Refused unknown connection from Tcp(V4(192.168.1.93:50229))
Received new message from Tcp(V4(192.168.1.93:50229))
New message came from unknown endpoint Tcp(V4(192.168.1.93:50229))
5fc6b4..02b3a2 construct_connect_response_msg 
Setup relay for node 44bcf1..df76b2 on Tcp(V4(127.0.0.1:38504))
Received new message from Tcp(V4(192.168.1.93:50229))
New message came from unknown endpoint Tcp(V4(192.168.1.93:50229))
```
Starting second node >
```
$ ./simple_key_value_store --type=node
NOTE: Transport connect 127.0.0.1:41873 failure due to Connection refused (os error 111)
NOTE: Transport connect 192.168.1.93:41873 failure due to Connection refused (os error 111)
NOTE: Transport connect [::1]:41873 failure due to Connection refused (os error 111)
NOTE: Transport connect 127.0.0.1:55228 failure due to Connection refusedbootstrap Tcp(V4(192.168.1.93:50665))
 (os error 111)
NOTE: Transport connect [::1]:50665 failure due to Connection refused (os error 111)
NOTE: Transport connect 127.0.0.1:54153 failure due to Connection refused (os error 111)
NOTE: Transport connect [::1]:54153 failure due to Connection refused (os error 111)
NOTE: Transport connect 127.0.0.1:42124 failure due to Connection refused (os error 111)
NOTE: Transport connect [::1]:42124 failure due to Connection refused (os error 111)
NOTE: Transport connect 127.0.0.1:46859 failure due to Connection refused (os error 111)
Waiting for responses from network
Received connect response
NOTE: Transport connect 192.168.1.105:46859 failure due to No route to host (os error 113)
NOTE: Transport connect 192.168.1.105:54153 failure due to No route to host (os error 113)
[...]

Issue #317, opened 2015-06-10
NewConnection precedes NewMessage


benjaminbollen:
CRUST NewConnection precedes NewMessage, it was assumed that a new connection was only established after both sides connect

routing_node_new::bootstrap adds listeners.0 to the connect_request;  this should not happen

zero node:
> ZERO listening on Some(Tcp(V4(127.0.0.1:38036)))
> Refused unknown connection from Tcp(V4(127.0.0.1:42208))
> Refused unknown connection from Tcp(V4(192.168.1.205:60393))
> Refused unknown connection from Tcp(V4(192.168.1.205:43550))
> Received new message from Tcp(V4(127.0.0.1:42208))
> New message came from unknown endpoint Tcp(V4(127.0.0.1:42208))
> ec895f..354829 construct_connect_response_msg 
> Setup relay for node c530cc..531206 on Tcp(V4(127.0.0.1:48344))
> Received new message from Tcp(V4(127.0.0.1:42208))
> New message came from unknown endpoint Tcp(V4(127.0.0.1:42208))
1st Node:
> BOOTSTRAP Tcp(V4(127.0.0.1:38036))
> NODE listening on Some(Tcp(V4(127.0.0.1:48344)))

benjaminbollen:
This assumption is OK, we just don't need to add "our_endpoints" to a connect_request; will be disabled in a coming PR

benjaminbollen:
Also CRUST gives a NewConnection from a one-sided attempt; Routing needs to clean these out if no connect request is received on them.  CRUST might address this by not allowing a single sided connection as a new connection

PR #319, opened 2015-06-10
Directly add to relay_map on connect_request

benjaminbollen:
Patch: on connect_request, directly add to relay_map - bypass the "remember public_id"
The assumption was that first a connect_request was received on our listening port, and when both sides 'connection_manager.connect' a 'CRUST::NewConnection' would be triggered.

Turns out that CRUST gives a new CRUST::NewConnection for every single sided connect, and the NewMessage comes after that, for a given endpoint.  This patch should do, but post-sprint this logic needs to be reworked.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/319)

### DIFF
--- a/src/routing_membrane.rs
+++ b/src/routing_membrane.rs
@@ -269,16 +269,23 @@ impl<F> RoutingMembrane<F> where F: Interface {
         // Try to connect to the peer.
         // when CRUST succeeds at establishing a connection,
         // we use this register to retrieve the PublicId
-        self.relay_map.register_accepted_connect_request(&connect_request.external_endpoints,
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @inetic

benjaminbollen:
#318 #317 

Now bumps on debug assert when FirstNode enters the membrane; it does not have a relocated (by zero node) address; I will raise an issue and progress

Fraser999:
I'll raise the issue - you code :-)

benjaminbollen:
resolves #318 

resolves #317

Issue #321, opened 2015-06-10
Comment block of simple_key_value_store out of date


Fraser999:
[This block](https://github.com/Fraser999/routing/blob/db57b7f6de6997b7166216b303413b83a1185095/examples/simple_key_value_store.rs#L17-L25) is outdated.

Issue #315, opened 2015-06-10
Issues with simple_key_value_store help output


Fraser999:
Running `-h` yields:

>     Usage:
      routing_new [--type=<node_type>] [--peer=<end_point>]
      routing_new (-h | --help)

>       If no arguments are passed, this will try to connect to an existing network
      using Crust's discovery protocol.  If this is unsuccessful, you can provide
      a list of known endpoints (other running instances of this example) and the node
      will try to connect to one of these in order to connect to the network.

>     Options:
      -h --help           Show this screen.
      --type=<node_type>  node_type = client | node | first.
      --peer=<end_point>  end_point

* The binary's name is no longer `routing_new`
* The term "client" was removed entirely from the Crust example to avoid giving the impression of a client/server network - this may be a requirement here too.
* There's no need to have `--peer` since any endpoints passed can only be peers (c.f. Crust example)
* `end_point` should be `endpoint`
* Except for `-h`, the descriptions for the options aren't descriptive (c.f. Crust example)
* Validation of `node_type` isn't being set, e.g. you can successfully call `--type=a`.  I'd suggest doing away with the `--type` arg as well and change the Usage part to:

        simple_key_value_store [--client | --node | --first] [<peer>...]
        simple_key_value_store (-h | --help)

* The description doesn't say which value for `type` will be chosen if no args are passed.

mmoadeli:
I started working on this

PR #322, opened 2015-06-10
Issues 315 & 321

mmoadeli:
addresses issues #315 && #321 

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/322)

### DIFF
--- a/examples/simple_key_value_store.rs
+++ b/examples/simple_key_value_store.rs
@@ -61,18 +61,22 @@ use routing::error::{ResponseError, InterfaceError};
 // ==========================   Program Options   =================================
 static USAGE: &'static str = "
 Usage:
-  routing_new [--type=<node_type>] [--peer=<end_point>]
-  routing_new (-h | --help)
+  simple_key_value_store [--client | --node ] [<peer>...]
+  simple_key_value_store --first
+  simple_key_value_store (-h | --help)
 
+  Running with '--node' or '--client' requires an existing network to connect to.
+  'first' should be passed as parameter to start first node in the network.
   If no arguments are passed, this will try to connect to an existing network
   using Crust's discovery protocol.  If this is unsuccessful, you can provide
-  a list of known endpoints (other running instances of this example) and the node
-  will try to connect to one of these in order to connect to the network.
+  an endpoint (other running instances of this example) and the node
+  will try to connect to the node in order to connect to the network.
 
 Options:
   -h --help           Show this screen.
-  --type=<node_type>  node_type = client | node | first.
-  --peer=<end_point>  end_point
+  --client            node runs as a client
+  --first             node runs as the first non-client node in the network
+  --node              node runs as the non-first non-client node in the network
 ";
 
 // ==========================   Helper Function   =================================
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @inetic

Fraser999:
This has stopped the ability to run the example with no arguments.  Also, the second and last bullet points of #315 aren't addressed by this commit.

examples/simple_key_value_store.rs
```diff
@@ -71,8 +74,9 @@ Usage:
 
 Options:
   -h --help           Show this screen.
-  --type=<node_type>  node_type = client | node | first.
-  --peer=<end_point>  end_point
+  --client            node runs as a client
+  --first             node runs as the first non-client node in the network
+  --node              node runs as the non-first non-client node in the network
```
Fraser999: Please use proper sentences for the descriptions (capital initial letter, full stop at end).  We probably need to explain somewhere the difference between client and node modes.

examples/simple_key_value_store.rs
```diff
@@ -61,9 +61,12 @@ use routing::error::{ResponseError, InterfaceError};
...
 static USAGE: &'static str = "
 Usage:
-  routing_new [--type=<node_type>] [--peer=<end_point>]
-  routing_new (-h | --help)
+  simple_key_value_store [--client | --node ] [<peer>...]
+  simple_key_value_store --first
+  simple_key_value_store (-h | --help)
 
+  Running with '--node' or '--client' requires an existing network to connect to.
+  'first' should be passed as parameter to start first node in the network.
```
Fraser999: `'first'` should be `'--first'`.

mmoadeli:
1- It was easier to implement it the way you mention but it seems to be more reasonable to explicitly mention what type of node is created.
2- how do you want to say a node is a client then?
3- It is the first one. Do we really want to have a default?

examples/simple_key_value_store.rs
```diff
@@ -316,98 +320,94 @@ fn main() {
...
-                command.clear();
-                println!("Input command (stop)");
-                let _ = io::stdin().read_line(&mut command);
-                let v: Vec<&str> = command.split(' ').collect();
-                match v[0].trim() {
-                    "stop" => break,
-                    _ => println!("Invalid Option")
+    if args.get_bool("--node") {
+        let mut test_node = RoutingNode::<TestNode, TestNodeGenerator>::new(TestNodeGenerator);
+        if !args.get_str("peer").is_empty() {
```
Fraser999: This doesn't work - I think you need `args.get_vec("<peer>")`.

mmoadeli: this is another issue https://github.com/maidsafe/routing/issues/324

Fraser999: #324 doesn't cover it - this existing code doesn't even handle a single endpoint.

`args.get_str("peer")` will always be empty even if the user enters a peer endpoint since docopt sees `peer` as a vector.

mmoadeli:
@Fraser999 please let me know if your concerns are addressed in the PR. Thanks

examples/simple_key_value_store.rs
mmoadeli: I see your point

Issue #320, opened 2015-06-10
simple_key_value_store crashing when second node joins first


Fraser999:
Having run an instance of `simple_key_value_store --type=first` then running `simple_key_value_store --type=node` results in the following crash of the second instance:

>     BOOTSTRAP Tcp(V4(192.168.0.35:61781))
    NODE listening on Some(Tcp(V4(192.168.0.35:62104)))
    Waiting for responses from network
    Received connect response
    Received PutPublicId response
    thread '<main>' panicked at 'assertion failed: relocated_id.is_relocated()', src\routing_membrane.rs:114

PR #323, opened 2015-06-10
Fix was trivial, first node started membrane explicitly unrelocated_id

benjaminbollen:
debug remnant;

Added FIXME: probably we can drop the zero_node special handler for PutPublicId

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/323)


maidsafe-highfive:
r? @inetic

benjaminbollen:
resolved #320

benjaminbollen:
Now the FindGroup message sent out needs to be investigated; the first, second etc nodes stay only connected over the bootstrap_connection - which is preserved!! great +1 for Rust, we move the whole crust manager and the active connections and event_receiver successfully to a new thread

Issue #325, opened 2015-06-10
FindGroupResponse does not find its way back


benjaminbollen:
FindGroupResponse does not find its way back.

benjaminbollen:
resolved

PR #326, opened 2015-06-10
Send FindGroupResponse back directly to bootstrap if we can

benjaminbollen:
#325

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/326)


maidsafe-highfive:
r? @inetic

PR #327, opened 2015-06-10
Fix minor issues with simple_key_value_store output, issue #316

### LINKED ISSUE #316: Minor issues with simple_key_value_store output
Once running, simple_key_value_store outputs:

>     Input command (stop, put <key> <value>, get <key>)

To keep this consistent with Crust example and docopt syntax it should be

    Enter command (stop | put <key> <value> | get <key>)>

<hr/>

Entering `put k v` and `get k` yield respectively:

>     putting data TestData( key: "k" , value: "v" ) to network with name as 8335fa..017c09
    getting data having key k from network using name as 8335fa..017c09

The outputs should have initial capital letters, and they should either both insert quotes round the key or neither should.  I'd prefer the output to be more like:

    Putting "v" to the network under key "k".
    Getting value for key "k" from network.

I don't think we should be showing the hex-encoded hash in the output - it's maybe useful for us to debug stuff, but it's confusing as part of an example.

rossmuir:
Fixed #316

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/327)

### DIFF
--- a/examples/simple_key_value_store.rs
+++ b/examples/simple_key_value_store.rs
@@ -386,7 +386,7 @@ fn main() {
             }
             loop {
                 command.clear();
-                println!("Input command (stop, put <key> <value>, get <key>)");
+                println!("Enter command (stop | put <key> <value> | get <key>)>");
                 let _ = io::stdin().read_line(&mut command);
                 let v: Vec<&str> = command.split(' ').collect();
                 match v[0].trim() {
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @inetic

rossmuir:
@Fraser999

rossmuir:
@Fraser999 Sorry I will sort the conflicts tomorrow

PR #337, opened 2015-06-11
Add with_keys constructor to Id type. Fixes #336.

### LINKED ISSUE #336: Routing 0.1.62 Id type change causes problem
think with the present code `client` will not be able to work with the new `Id` thing in routing.
`maidsafe_client` requires this to be present in the session packet (along with mpid groups) and also to encrypt/decrypt DirectoryListings etc.

It contacts `maidsafe-types` for that

`maidsafe_types` does the random symmetric and asymmetric key-pair generations

`maidsafe-types` has no provision to construct it from given keys - it is who generates it - see:
https://github.com/maidsafe/maidsafe_types/blob/master/src/id/revocation_id_type.rs#L67
for example

And that is how i think it should be - generation of secure keys in `types` library so that there is one place to go to if someone wants to audit/tinker it and also fits with the paradigm of maidsafe-types crate itself.

`routing` should not be the one generating keys - it should accept it in a package (like clientid stuff previously) and then do what ever it wants.

Anyway - unless types crate is changed it will break `client` as of now.

s-brian:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/337)

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -386,6 +386,44 @@ impl Id {
     }
   }
 
+  pub fn with_keys(public_keys: (crypto::sign::PublicKey, crypto::asymmetricbox::PublicKey),
+                   secret_keys: (crypto::sign::SecretKey, crypto::asymmetricbox::SecretKey)) -> Id {
+    let sign_key = &(public_keys.0).0;
+    let asym_key = &(public_keys.1).0;
+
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @benjaminbollen

benjaminbollen:
looks good to me

PR #340, opened 2015-06-11
Not pop from rotuing table when query just peek

dirvine:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/340)


maidsafe-highfive:
r? @benjaminbollen

Issue #347, opened 2015-06-12
Debug Put / Get flow for client on network


benjaminbollen:

benjaminbollen:
Put and Get end up on the network and get processed.  Seems the only remaining issue is for the responses to return back to the relayed client

PR #350, opened 2015-06-12
Debug PUT/GET flow - Issue 347

benjaminbollen:
Turned out it responses were being given from cache; took a long time to remember that after long hours :)

Inserted interception points to relay back to relay node / client;  It is clear we need a more elegant approach; to rush things I have replicated the same code on several locations; at least the should be put in one function-  issue raised to clean this up

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/350)


maidsafe-highfive:
r? @inetic

benjaminbollen:
resolves #347

PR #356, opened 2015-06-12
Removes some unwraps.

s-brian:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/356)

### DIFF
--- a/examples/simple_key_value_store.rs
+++ b/examples/simple_key_value_store.rs
@@ -184,8 +184,10 @@ impl Sendable for TestData {
 
     fn serialised_contents(&self) -> Vec<u8> {
         let mut e = cbor::Encoder::from_memory();
-        e.encode(&[&self]).unwrap();
-        e.into_bytes()
+        match e.encode(&[&self]) {
+            Ok(()) => e.into_bytes(),
+            Err(_) => Vec::<u8>::new()
+        }
     }
 
     fn refresh(&self) -> bool {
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @inetic

examples/simple_key_value_store.rs
```diff
@@ -184,8 +184,10 @@ impl Sendable for TestData {
 
     fn serialised_contents(&self) -> Vec<u8> {
         let mut e = cbor::Encoder::from_memory();
-        e.encode(&[&self]).unwrap();
-        e.into_bytes()
+        match e.encode(&[&self]) {
+            Ok(()) => e.into_bytes(),
+            Err(_) => Vec::<u8>::new()
+        }
```
inetic: If an error happens when encoding, the client will send an empty vector of bytes. I don't think this is the correct behavior (is it?). In an example I would suggest to either `assert`, which would be the simplest solution. I think the ideal solution would be for this function to return Result<Vec<u8>, some-cbor-error-type>, that way, the code that uses this function can handle the situation and just not send anything. The ideal solution might be an overkill here though as encoding will hardly ever fail in this example.

benjaminbollen:
I agree this is worse than the unwrap

examples/simple_key_value_store.rs
s-brian: As written your correct. I think it's another task though to update the example to return a Result. I was only tasked to remove unwrap's, maybe this one could have been left as it was, it probably is overkill, and since it should never really fail an empty vector will rarely get returned.

dirvine:
Review status: 0 of 1 files reviewed, 2 unresolved discussions, some commit checks pending.

---

<sup>**[examples/simple_key_value_store.rs, line 251 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/356#-Jrc9HozgPUfDnyK1_RW)** ([raw file](https://github.com/maidsafe/routing/blob/ba45aee1be1d6714e8ac373aa514948312a58611/examples/simple_key_value_store.rs#L251)):</sup>
Perhaps tidy up message a little here. 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/routing/356)

benjaminbollen:
looks good to me

Issue #360, opened 2015-06-12
test relay::test::add ... FAILED (debug)


rossmuir:
`Ubuntu 15.04`
`test relay::test::add ... FAILED` Failing intermittently 

```
failures:

---- relay::test::add stdout ----
	thread 'relay::test::add' panicked at 'assertion failed: `(left == right)` (left: `true`, right: `false`)', src/relay.rs:217



failures:
    relay::test::add

test result: FAILED
```

benjaminbollen:
similar intermittent problem occurs on relay::test::add_multiple_endpoints

PR #384, opened 2015-06-18
check that mock-endpoint is not already in the relay_map

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/384)


maidsafe-highfive:
r? @inetic

benjaminbollen:
Failed intermittently because of same port in random generation

benjaminbollen:
resolves #360

Issue #395, opened 2015-06-19
Correct preservation of message_id


benjaminbollen:
Preserve message_id on MethodCall::SendOn; (currently msg_id -> msg_id+1)
MaidSafe_client relies on the message_id to relate a response to an original request, so routing needs to return the origial message id;

an alternative fix to adding to the message_id on SendOn, is for the message_filter in the routing_nodes to also take the destination into account; as such in a small network different groups are likely to overlap physically and we can still SendOn between without being falsely blocked by the filter.

PR #396, opened 2015-06-19
preserve message_id on SendOn

benjaminbollen:
Needs correction to message_filter now such that filter does not falslely block a sent-on message

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/396)


maidsafe-highfive:
r? @inetic

benjaminbollen:
resolves #395

PR #416, opened 2015-06-24
Activate Churn Handle;

benjaminbollen:
We don't act on the resulting MethodCall

Signed-off-by: Benjamin Bollen <benjamin.bollen@maidsafe.net>

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/416)


maidsafe-highfive:
r? @inetic

Issue #423, opened 2015-06-25
patch FailureToStoreData


benjaminbollen:
Routing will interpret here that a FailureToStoreData with different data then received from a ManagedNode, indicates a Success on storing the original data. It will send the error separately to preceding group as normal, but with different (deterministic) message_id. 

This is a patch to allow Vaults to work with the limited interface they are currently offered.  An update is planned in routing and at that point, this logic needs to be moved out of routing again and should purely be a choice on Vaults.

PR #424, opened 2015-06-25
Patch for Vault Handle Put behaviour;

benjaminbollen:
send back put response to Client;
on FailedToStoreData from ManagedNode with different data, apply patch

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/424)

### DIFF
--- a/src/routing_membrane.rs
+++ b/src/routing_membrane.rs
@@ -932,7 +932,7 @@ impl<F> RoutingMembrane<F> where F: Interface {
             },
             Err(InterfaceError::Response(error)) => {
                 try!(self.send_put_reply(&header.from(), our_authority, &header,
-                &put_data, Err(error)));
+                    &put_data, Err(error)));
             }
         }
         Ok(())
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @inetic

benjaminbollen:
successfully confirmed consistent behaviour for routing example with this patch; should fix PutResponse on REply to actual client even in big networks -- not tested on big enough network to confirm

benjaminbollen:
Also successfully passes the self_authentication example;

conclusion: this PR patches routing to ensure that Vaults Safecoin measurement is not explicitly broken;

we can only explicitly test the safecoin farming rate after the payload-issue is resolved in Vaults

resolves #423

dirvine:
Reviewed 1 of 1 files at r2.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/routing/424)

PR #426, opened 2015-06-25
Act on churn

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/426)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,24 +1,16 @@
 # Routing - Change Log
-## [0.2.5]
+## [0.2.6] - Temporary patch for Vault behaviour
 
-- [#421](https://github.com/maidsafe/routing/pull/421) Set Authority unauthorised put to ManagedNode to accommodate Vaults for now
+- [#424](https://github.com/maidsafe/routing/pull/424) Patch for Vaults handle put behaviour
 
-## [0.2.4]
+## [0.2.1 - 0.2.5] - debug with upper layers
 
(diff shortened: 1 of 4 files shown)


maidsafe-highfive:
r? @inetic

benjaminbollen:
running local network tests

benjaminbollen:
also addresses #427

benjaminbollen:
r? @dirvine 

works really well; stopping work because I started to pile up several clean up issues in one commit;

the clean up work was provoked by the need for efficient traffic if we activate churn; the network should be much more lean with these cleanups already

dirvine:
Reviewed 1 of 3 files at r1, 1 of 3 files at r2.
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/routing_membrane.rs, line 624 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/426#-JslDMoFmGjs6ELtY6C2)** ([raw file](https://github.com/maidsafe/routing/blob/c5c1548281e95f2ec9df3ad167d30fbdbdb755c6/src/routing_membrane.rs#L624)):</sup>
Should we should use lru cache for this rather than arbitrary timer

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/routing/426)

dirvine:
Reviewed 2 of 3 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/routing/426)

Issue #252, opened 2015-05-14
Send Findgroup if we lose a close group node


chandraprakash:
Send FindGroup if we lose connection to a close group node

benjaminbollen:
Can we not extract updates on our close group from swarmed messages?

chandraprakash:
I guess, on any routing table change, just check if the node which went down was close group member and fire a findgroup and forget.

benjaminbollen:
Messages are filtered on new interesting nodes;

nodes should primarily be dropped as "online" when connected node is lost; when the whole group agrees to drop the node, we should trigger a churn event

PR #431, opened 2015-07-01
Send Findgroup if we lose a close group node

mmoadeli:
addresses issue #252

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/431)

### DIFF
--- a/src/routing_membrane.rs
+++ b/src/routing_membrane.rs
@@ -388,10 +388,10 @@ impl<F> RoutingMembrane<F> where F: Interface {
         // The relay map will automatically drop the Name if the last endpoint to it is dropped
         self.relay_map.remove_unknown_connection(&endpoint);
         self.relay_map.drop_endpoint(&endpoint);
-        let mut trigger_handle_churn = false;
+        let mut close_node_lost = false;
         match self.routing_table.lookup_endpoint(&endpoint) {
             Some(name) => {
-                trigger_handle_churn = self.routing_table
+                close_node_lost = self.routing_table
                     .address_in_our_close_group_range(&name);
                 self.routing_table.drop_node(&name);
                 println!("RT (size : {:?}) connection {:?} disconnected for {:?}.",
@@ -412,7 +412,8 @@ impl<F> RoutingMembrane<F> where F: Interface {
         };
         if drop_bootstrap { self.bootstrap_endpoint = None; }
 
-        if trigger_handle_churn {
+        if close_node_lost {
+            self.send_find_group_request();
             println!("Handle CHURN lost node");
             let mut close_group : Vec<NameType> = self.routing_table
                 .our_close_group().iter()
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @benjaminbollen

benjaminbollen:
Is this work in progress or ready for review?

src/routing_membrane.rs
```diff
@@ -388,10 +388,10 @@ impl<F> RoutingMembrane<F> where F: Interface {
         // The relay map will automatically drop the Name if the last endpoint to it is dropped
         self.relay_map.remove_unknown_connection(&endpoint);
         self.relay_map.drop_endpoint(&endpoint);
-        let mut trigger_handle_churn = false;
```
benjaminbollen: why do you rename this?

src/routing_membrane.rs
```diff
@@ -434,6 +435,34 @@ impl<F> RoutingMembrane<F> where F: Interface {
...
+        let serialised_msg = match encode(&find_group_msg) {
+            Ok(msg) => msg,
+            Err(_) => panic!("Failure to serialise find group message.")
+        };
+
+        if self.routing_table.size() != 0 {
+            return self.send_swarm_or_parallel(&our_name, &serialised_msg);
+        }
+
+        loop {
```
benjaminbollen: This is not correct within address relocation; so please limit the function send_find_group to only do what the function says. nothing more.

also, if you are making a send_find_group function, update all cases where we send find_group_request consistently

src/routing_membrane.rs
```diff
@@ -434,6 +435,34 @@ impl<F> RoutingMembrane<F> where F: Interface {
...
+    /// If routing table is not empty, the find group request is sent through available node(s),
+    /// else if bootstrap_endpoint is non-empty, request sent through bootstrap_endpoint.
+    /// If routing table is empty & bootstrap_endpoint is empty, attempt is made to
+    /// send the request through the endpoint acheived from re-bootstrapping.
+    fn send_find_group_request(&mut self) {
+        let our_name = self.own_name.clone();
+        let find_group_msg = self.construct_find_group_msg(&our_name);
+        let serialised_msg = match encode(&find_group_msg) {
+            Ok(msg) => msg,
+            Err(_) => panic!("Failure to serialise find group message.")
```
benjaminbollen: just as below, have the function return a result that we can ignore( ) for now, but throw an error, don't panic

src/routing_membrane.rs
```diff
@@ -434,6 +435,34 @@ impl<F> RoutingMembrane<F> where F: Interface {
...
+        };
+
+        if self.routing_table.size() != 0 {
```
benjaminbollen: you don't need to check this; the function is simply a no-op; and no bootstrapping connections should be used by default

src/routing_membrane.rs
```diff
@@ -434,6 +435,34 @@ impl<F> RoutingMembrane<F> where F: Interface {
         };
     }
 
+    /// If routing table is not empty, the find group request is sent through available node(s),
+    /// else if bootstrap_endpoint is non-empty, request sent through bootstrap_endpoint.
+    /// If routing table is empty & bootstrap_endpoint is empty, attempt is made to
+    /// send the request through the endpoint acheived from re-bootstrapping.
+    fn send_find_group_request(&mut self) {
```
benjaminbollen: additionally, send_ functions are collected below, please group them where they are

src/routing_membrane.rs
```diff
@@ -434,6 +435,34 @@ impl<F> RoutingMembrane<F> where F: Interface {
...
+        let our_name = self.own_name.clone();
```
benjaminbollen: you don't need to clone this

PR #434, opened 2015-07-07
Rfc0

dirvine:
Implement types for rfc0
Update Signature as real type
Update Authority enum to take params which identify location of claim (i.e. group_name for NaeManager, PublicKey for client etc.)
Several tidy up of code (lots to go)

Important tests not refactored but require to be. There may be some slight changes such as ManagedNode not needing a parameter etc. quick chat makes it an easy fix though. Also important this needs my master github repo for sodiumoxide for now as there are several extra traits there we are using and not in main repo yet

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/434)


maidsafe-highfive:
r? @benjaminbollen

src/structured_data.rs
```diff
@@ -0,0 +1,235 @@
...
+        // seems overkill to use serialisation here, but done
+        // to ensure cross platform signature handling is OK
+        let mut enc = cbor::Encoder::from_memory();
+        try!(enc.encode(self.type_tag.to_string().as_bytes()));
+        try!(enc.encode(&self.identifier[..]));
+        try!(enc.encode(&self.data));
+        try!(enc.encode(&self.previous_owner_keys));
+        try!(enc.encode(&self.current_owner_keys));
+        try!(enc.encode(self.version.to_string().as_bytes()));
+        Ok(enc.as_bytes().into_iter().map(|&x| x).collect())
```
ustulation: `Ok(enc.into_bytes())`

dirvine: Ah that seems easier :-)

inetic:
Looks good

src/structured_data.rs
```diff
@@ -0,0 +1,235 @@
...
+use error::RoutingError;
+use NameType;
+use sodiumoxide::crypto;
+use std::str;
+
+/// StructuredData
+/// These types may be stored unsigned with previous and current owner keys
+/// set to the same keys. Updates though require a signature to validate
+#[derive(Clone, RustcDecodable, RustcEncodable)]
+pub struct StructuredData {
```
ustulation: I've coded the following mock.
Point of differences:
<1> no signature parameter in `new()`. User must not be trusted to sign properly, so the function does it itself.
<2> identifier may not follow utf8 for everyone (i could give in anything and it should be possible). So changes in `name()`
<3> `data_to_sign` function should be public so that order of concatenation is centralized just in-case someone needs it to sign externally or vaults need to verify signatures with customized rule (and not `verify_previous_owner_signatures`)
<4> Previous_owner_keys is Option<...> to make it more consistent with the notion that it may not be filled.
<5> This is minor: can previous_owner_signatures be made just `signatures` ? It is sometimes confusing to have that field populated and still have previous_owner_keys contain nothing.

Other than that, the deviations in naming and return types, I'll fix at my end.

```
#[derive(Clone)]
pub struct StructuredData {
    tag_type: u64,
    identifier: ::routing::NameType,//::sodiumoxide::crypto::hash::sha512::Digest,
    version: u64,
    data: Vec<u8>,
    current_owner_keys: Vec<::sodiumoxide::crypto::sign::PublicKey>,
    previous_owner_keys: Option<Vec<::sodiumoxide::crypto::sign::PublicKey>>,
    signatures: Vec<::sodiumoxide::crypto::sign::Signature>,
}

impl StructuredData {
    pub fn new(tag_type: u64,
               identifier: ::routing::NameType,//::sodiumoxide::crypto::hash::sha512::Digest,
               version: u64,
               data: Vec<u8>,
               current_owner_keys: Vec<::sodiumoxide::crypto::sign::PublicKey>,
               previous_owner_keys: Option<Vec<::sodiumoxide::crypto::sign::PublicKey>>,
               sign_key: &::sodiumoxide::crypto::sign::SecretKey) -> StructuredData {
        let mut structured_data = StructuredData {
            tag_type: tag_type,
            identifier: identifier,
            version: version,
            data: data,
            current_owner_keys: current_owner_keys,
            previous_owner_keys: previous_owner_keys,
            signatures: vec![],
        };

        structured_data.add_signature(sign_key);
        structured_data
    }

    pub fn add_signature(&mut self, sign_key: &::sodiumoxide::crypto::sign::SecretKey) {
        let signable = self.get_signable_data();
        self.signatures.push(::sodiumoxide::crypto::sign::sign_detached(&signable[..], sign_key));
    }

    pub fn get_signable_data(&self) -> Vec<u8> {
        let mut vec = self.data.iter().chain( self.version.to_string().as_bytes().iter().chain(
                self.current_owner_keys.iter().fold(
                    Vec::<u8>::new(), |mut key_vec, key| { key_vec.extend(key.0.iter().map(|a| *a).collect::<Vec<u8>>()); key_vec }
                    ).iter())).map(|a| *a).collect::<Vec<u8>>();

        if let Some(ref previous_owner_keys) = self.previous_owner_keys {
            vec.extend(previous_owner_keys.iter().fold(Vec::<u8>::new(), |mut key_vec, key| { key_vec.extend(key.0.iter().map(|a| *a).collect::<Vec<u8>>()); key_vec }));
        }

        vec
    }

    pub fn name(&self) -> ::routing::NameType {
        use ::sodiumoxide::crypto::hash::sha512::hash;
        ::routing::NameType(
            hash(&hash(&self.identifier.0).0.iter().chain(self.tag_type.to_string().as_bytes().iter()).map(|a| *a).collect::<Vec<u8>>()[..]).0)
    }
}
```

dirvine: Cheers @ustulation 

Thoughts 

1. Agreed 

2. Agreed (although on the fence as to whether NameType should exist at all and instead not be actually a hash, which I think it always is) Should be note sodium seems to not like copying hashes (timing attacks). So this needs security review 

3. Agreed 

4. Not sure, @inetic made the point a vector by default provides the ability to be zero length, which does make sense as it's implicit in the type. (it used to be an Option)

5. Again not sure, if your prev owners field is empty the signature check would fail would it not ? So an invariant is prev owners must have signed data over to new owner. 

And lastly PR will be welcomed by @inetic I am sure :-)

ustulation: <2> Ah no, NameType is just temporarily there (as i am currently not using the sodiumoxide fork). It'll be Digest eventually. Even then i guess we shouldn't assume utf8 for it can come from any source, not necessarily hash of utf8 encoded string.

<4> Ok

<5> Ok too. This was just about naming. Most of the time we will not be transferring ownership and simply be having `previous_owner_keys` field empty and dealing with signatures of `current_owner_keys` only to update the StructuredData. So it seemed strange to call it `previous_owner_signatures`. But that's just naming, so it's Ok.

inetic: 4. Yes, I don't like it in the rest of the code neither but especially here.

The main reason is that we'll now have two instances of the field, namely `None` and `Some(empty_vec)` to mean exactly the same thing. So instead of asking `if !owners.is_empty() ...` you'll be asking `if owners.is_some() && !owners.is_empty() ...`. And the reason I don't like it 'especially' in this case is that it adds more data to the message.

inetic: IMO the only scenario where `Option<Vec<>>` makes sense is when `None` and `Some(empty_vector)` can have different meaning, but I don't believe it is the case here.

src/test_utils/messages_util.rs
```diff
@@ -78,7 +68,7 @@ impl Random for messages::connect_response::ConnectResponse {
             receiver_id: Random::generate_random(),
             receiver_fob: Random::generate_random(),
             serialised_connect_request: generate_random_vec_u8(64),
-            connect_request_signature: Random::generate_random()
+            connect_request_signature: crypto::sign::Signature([0; 64]),
```
dirvine: I see this part as WIP and we should have no invalid sigs, this should be foxed in tech debt sprint though. Testing could be better.

PR #442, opened 2015-07-08
Dirvine rfc0

mmoadeli:
make tests compile

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/442)


maidsafe-highfive:
r? @inetic

PR #444, opened 2015-07-09
New message infrastructure. (not passing tests or fully implemented yet)

dirvine:
messages taking shape, implemented client interface (put get post delete) to allow clients to at least have an API to match. 

Will require 4 of us later today go right through and get client / node and membrane updated to match new messages and types. Followed by ensuring all tests up to spec (again, sorry) and very likely removing more functions that are no longer required. 

Largest obvious changes 

1. We will sign complete message (Except Get) 
2. Relay node will set node address it is relayed around and create the complete message (signed if necessary). 
3. Message wrapper looks like this 

Message -> (RoutingMessage Or SignedRoutingMessage) [SignedRoutingMessagse is only a wrapper fro RoutingMessage and used when we require messages signed]

RoutingMessage contains all header info and relay addresses as well as a Data or DataRequest enum. These enums contain the routing data types Or an indication of the type and type_tag (enum variant) respectively. This seems slightly complex but saves a lot of unnecessary and complex code.

DestinationAddress and SourceAddress are also refactored into enums to ensure complete matching when relay etc. is involved. 

Authority is also a ADT now to ensure we know the address for the persona claim being made and can calculate the correctness of such a claim.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/444)


maidsafe-highfive:
r? @inetic

inetic:
The PR is marked WIP, is that still accurate?

src/messages.rs
```diff
@@ -15,178 +15,186 @@
...
+pub struct SignedRoutingMessage {
+    pub encoded_routing_message : Vec<u8>,
+    pub signature               : Signature
+}
+
+#[derive(PartialEq, Eq, Clone, Debug, RustcEncodable, RustcDecodable)]
+enum Message {
+Signed(SignedRoutingMessage),
+Unsigned(RoutingMessage)    
+}
```
inetic: Wouldn't it be cleaner to just add `Option<Signature>` to the `RoutingMessage`?

dirvine: I did look at that, but the issue came down to multiple serialisation, I am not 100% of the efficiency, but we have to serialise the RoutingMessage anyway to sign, so I figured wrapping that was easiest. Ideally though I feel we need to get the DestinationAddress and SourceAddress easily parsable on their own so intermediate nodes do not need to parse everything all of the time.

PR #449, opened 2015-07-09
RelayedForClient from_address set back to bootstrap node

inetic:
As it is the correct id of the relay node. (I would prefer the approach discussed where there would be just `header` instead of `destination` and `source`, but did it this way to avoid confusion for now)

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/449)


maidsafe-highfive:
r? @benjaminbollen

PR #450, opened 2015-07-09
Sign RoutingMessage not Unsigned(RoutingMessage)

inetic:
When creating Message::SignedRoutingMessage

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/450)

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -29,6 +29,8 @@ use types::{DestinationAddress, SourceAddress, FromAddress, ToAddress, NodeAddre
 use error::{RoutingError, ResponseError};
 use NameType;
 use cbor;
+use utils::{encode};
+use cbor::{CborError};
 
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/messages.rs
```diff
@@ -187,10 +189,28 @@ impl RoutingMessage {
 
 #[derive(PartialEq, Eq, Clone, Debug, RustcEncodable, RustcDecodable)]
 pub struct SignedRoutingMessage {
+    // FIXME: The field `encoded_routing_message` should be private to
+    // avoid setting it with data that represent something other than
+    // serialised RoutingMessage. (`signature` should probably be
+    // private as well for similar reason).
```
dirvine: Yes, there is a load of `Pub`statements that need removed, secret keys as well We should make these private with accessors to allow borrowing

PR #456, opened 2015-07-10
PUT messages should go to ClientManager

inetic:
And few RelayedForClient takes a PublicKey
not name of the client (this would have been caught
by the compiler later though).

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/456)


maidsafe-highfive:
r? @benjaminbollen

PR #461, opened 2015-07-11
Fix handling caching get

inetic:
Few commits back I made an invalid assumptions that get
messages only arrive from non-group sources.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/461)


maidsafe-highfive:
r? @benjaminbollen

PR #463, opened 2015-07-13
Further refactoring

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/463)

### DIFF
--- a/src/authority.rs
+++ b/src/authority.rs
@@ -63,12 +63,35 @@ pub enum Authority {
 pub fn our_authority(message       : &RoutingMessage,
                      routing_table : &RoutingTable) -> Authority {
 
+    // Purposely listing all the cases and not using wild cards so
+    // that if a new message is added to the MessageType enum, compiler
+    // will warn us that we need to add it here.
     let element = match message.message_type {
-        MessageType::GetData(data_request)  => Some(message.non_relayed_destination()),
-        MessageType::PutData(data)          => Some(data.name()),
-        MessageType::PutPublicId(public_id) => public_id.name(),
-        // MessageType::Refresh(_, _)          => Some(message.from_group()),
-        _                                   => None,
+        M::BootstrapIdRequest     => None,
+        M::BootstrapIdResponse    => None,
+        M::ConnectRequest(_)      => None,
+        M::ConnectResponse(_)     => None,
+        M::FindGroup(_)           => None,
+        M::FindGroupResponse(_)   => None,
+        M::GetData(data_request)  => Some(message.non_relayed_destination()),
+        M::GetDataResponse(_)     => None,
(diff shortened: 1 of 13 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/authority.rs
```diff
@@ -62,33 +62,36 @@ pub enum Authority {
...
+        MessageType::PostResponse(_)        => None,
+        MessageType::PutData(data)          => Some(data.name()),
+        MessageType::PutDataResponse(_)     => None,
+        MessageType::PutKey                 => None,
+        MessageType::AccountTransfer(_)     => None,
+        MessageType::PutPublicId(public_id) => Some(public_id.name()),
+        MessageType::PutPublicIdResponse(_) => None,
+        //MessageType::Refresh(_, _)          => Some(message.from_group()),
+        MessageType::Refresh(_,_)           => None,
+        MessageType::Unknown                => None,
```
dirvine: Yes I agree no wildcards here is better.

src/messages.rs
```diff
@@ -55,6 +55,13 @@ pub struct ConnectResponse {
     pub connect_request_signature: Signature
 }
 
+/// Respond wiht Data or Error
+#[derive(PartialEq, Eq, Clone, Debug, RustcEncodable, RustcDecodable)]
+pub enum DataOrError {
+Data(Data),
+Err(ResponseError),
+}
```
dirvine: Yea I am thinking this is a mistake here as it really needs `Err(RoutingError, SignedMessage)` where the message is the original message and error states clearly why it failed. I do not think we should ever send success messages (waste of b/w). Only default return is Get where if we have data we return it, Error cases for Get are something to further consider, but not for now.

src/messages.rs
```diff
@@ -179,30 +194,38 @@ impl RoutingMessage {
...
+        //send_on_message.destination = match self.destination {
+        //    DestinationAddress::RelayToClient(_, b) => DestinationAddress::RelayToClient(destination, b),
+        //    DestinationAddress::RelayToNode(_, b)   => DestinationAddress::RelayToNode(destination, b),
+        //    DestinationAddress::Direct(_)           => DestinationAddress::Direct(destination),
+        //};
+        //send_on_message.authority = our_authority.clone();              send_on_message.authority = our_authority.clone();
+        //send_on_message
+
+        send_on_message.source      = SourceAddress::Direct(our_name);
+        send_on_message.destination = DestinationAddress::Direct(destination);
```
dirvine: I think this should be 
```send_on_message.destination = reply_destination();```

inetic: No, this is the "send_on" function, not reply :)

PR #464, opened 2015-07-13
Gets rid of the Message type and introduces SignedMessage

inetic:
As per discussion where the consensus was that all messages should be signed

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/464)


maidsafe-highfive:
r? @benjaminbollen

PR #476, opened 2015-07-17
Simplifying and proper name when msg arrives from bootstrap node

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/476)


maidsafe-highfive:
r? @benjaminbollen

dirvine:
Yes this is much better. Will also help when we move to parallel bootstrap and multiple bootstrap connections.

PR #478, opened 2015-07-17
Our authority

benjaminbollen:
tests not checked, as last remaining compilation errors still to be resolved

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/478)

### DIFF
--- a/src/authority.rs
+++ b/src/authority.rs
@@ -23,11 +23,8 @@ use messages::{RoutingMessage, MessageType};
 
 #[derive(RustcEncodable, RustcDecodable, PartialEq, PartialOrd, Eq, Ord, Debug, Clone)]
 pub enum Authority {
-  ClientManager(NameType),  // from a node in our range but not routing table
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @inetic

benjaminbollen:
Important points:
1. ClientManager(NameType) always points to the ClientName, not the element of the message
2. removed OurCloseGroup as Refresh messages will be handled without Authority (only valid when received from our RT)
3. updated the tests to refactor
4. excluded false positive from NaeManager (for ManagedNode)
5. updated documentation

benjaminbollen:
6. also for NodeManager(NameType) points to the Node name, not the element

dirvine:
I think that's good as these all have the address that shows why they are that authority, so it == element or nodendame etc. but whichever the authority address is quite explicit, so we do not need to work out outside this method if the name == element or node or client etc.

PR #483, opened 2015-07-17
Correct Interface for example; disabled example on test

benjaminbollen:
TestData in the example still has to be converted to PlainData

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/483)


maidsafe-highfive:
r? @inetic

PR #485, opened 2015-07-17
MAID-1262.

s-brian:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/485)

### DIFF
--- a/src/routing_membrane.rs
+++ b/src/routing_membrane.rs
@@ -955,7 +955,6 @@ impl<F> RoutingMembrane<F> where F: Interface {
                                            message: RoutingMessage,
                                            response: ErrorReturn) -> RoutingResult {
         info!("Handle PUT data response.");
-        let our_authority = our_authority(&message, &self.routing_table);
         let from_authority = message.from_authority();
         let from = message.source.clone();
 
@@ -970,8 +969,7 @@ impl<F> RoutingMembrane<F> where F: Interface {
                 MethodCall::Forward { destination } =>
                     ignore(self.forward(&signed_message, &message, destination)),
                 MethodCall::Reply { data: _data } =>
-                    ignore(self.send_reply(&message, our_authority.clone(),
-                        MessageType::PutDataResponse(ErrorReturn { error: response.error.clone(), orig_request: signed_message.clone() }))),
+                    info!("IGNORED: on handle_put_data_response MethodCall:Reply is not a Valid action")
             }
         }
         Ok(())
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/routing_membrane.rs
```diff
@@ -361,6 +361,7 @@ impl<F> RoutingMembrane<F> where F: Interface {
 
     /// When CRUST reports a lost connection, ensure we remove the endpoint anywhere
     /// TODO: A churn event might be triggered
+    #[allow(unused_variables)]
```
benjaminbollen: starting a variable with an _ allows for that variable to be unused

s-brian: Just added that to the PR.

src/routing_membrane.rs
```diff
@@ -720,6 +715,7 @@ impl<F> RoutingMembrane<F> where F: Interface {
         }
     }
 
+    #[allow(unused_variables)]
```
benjaminbollen: again, better to not name or _name unused variables; thanks

src/routing_membrane.rs
```diff
@@ -950,25 +952,29 @@ impl<F> RoutingMembrane<F> where F: Interface {
...
+            match method_call {
+                MethodCall::Put { destination: x, content: y, } => self.put(x, y),
+                MethodCall::Get { name: x, data_request: y, } => self.get(x, y),
+                MethodCall::Refresh { type_tag, from_group, payload } => self.refresh(type_tag, from_group, payload),
+                MethodCall::Post { destination: x, content: y, } => self.post(x, y),
+                MethodCall::Delete { name: x, data : y } => self.delete(x, y),
+                MethodCall::None => (),
+                MethodCall::Forward { destination } =>
+                    ignore(self.forward(&signed_message, &message, destination)),
+                MethodCall::Reply { data } =>
```
benjaminbollen: this should probably not be supported, "put_response reply" as data is discarded

src/routing_membrane.rs
```diff
@@ -1228,23 +1235,29 @@ impl<F> RoutingMembrane<F> where F: Interface {
...
+            match method_call {
+                MethodCall::Put { destination: x, content: y, } => self.put(x, y),
+                MethodCall::Get { name: x, data_request: y, } => self.get(x, y),
+                MethodCall::Refresh { type_tag, from_group, payload } => self.refresh(type_tag, from_group, payload),
+                MethodCall::Post { destination: x, content: y, } => self.post(x, y),
+                MethodCall::Delete { name: x, data : y } => self.delete(x, y),
+                MethodCall::None => (),
+                MethodCall::Forward { destination } =>
+                    ignore(self.forward(&orig_message, &message, destination)),
+                MethodCall::Reply { data } =>
```
benjaminbollen: again, on get_data_response, reply should not be supported

benjaminbollen:
please look at the other comments that are now on "an outdated diff"; they are still relevant

PR #486, opened 2015-07-18
Fix for routing_table tests, Implement public_id name

dirvine:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/486)

### DIFF
--- a/src/id.rs
+++ b/src/id.rs
@@ -68,10 +68,10 @@ impl Id {
       self.name
     }
 
-    pub fn get_name(&self) -> NameType {
+    pub fn set_name(&mut self, name: NameType) {
         // This function should not exist, it is here only temporarily
         // to fix compilation.
-        self.name
+        self.name = name;
     }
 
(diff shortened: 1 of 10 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/id.rs
```diff
@@ -53,10 +56,11 @@ impl Id {
 
     pub fn with_keys(sign_keys: (crypto::sign::PublicKey, crypto::sign::SecretKey),
                      encrypt_keys: (crypto::box_::PublicKey, crypto::box_::SecretKey))-> Id {
+        let name = NameType::new(crypto::hash::sha512::hash(&sign_keys.0[..]).0);
```
inetic: Could we use this function?
https://github.com/maidsafe/routing/blob/master/src/utils.rs#L42
Or does that one have different semantic?

dirvine: Yes I think we will end up being more explicit here and using only the publicKey for non xor nodes.

PR #491, opened 2015-07-20
Implementes RoutingMessage::create_reply + few test fixes

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/491)

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -224,21 +224,27 @@ impl RoutingMessage {
     /// Authority is changed at this point as this method is called after
     /// the interface has processed the message.
     /// Note: this is not for XOR-forwarding; then the header is preserved!
-    pub fn create_reply(&self, _our_name : &NameType, _our_authority : &Authority)
-        -> RoutingMessage {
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/messages.rs
```diff
@@ -224,21 +224,27 @@ impl RoutingMessage {
...
+        reply_message.destination = match self.orig_message {
+            Some(ref orig_message) => {
+                try!(orig_message.get_routing_message()).reply_destination()
+            },
+            None => {
+                self.reply_destination()
+            }
+        };
+
+        reply_message.orig_message = None;
```
benjaminbollen: Should the original message not be attached again when it is present?

inetic: It should, but not in `reply_message.orig_message`, instead it should be in `reply_message.message_type.<response-type>.orig_request`. So, currently it is the responsibility of the one who calls `create_reply` to set it explicitly.

benjaminbollen: you are right, I warned I found this highly confusing :)

benjaminbollen:
looks good to me

PR #495, opened 2015-07-21
interface addition and correction

ustulation:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/495)

### DIFF
--- a/src/structured_data.rs
+++ b/src/structured_data.rs
@@ -57,6 +57,19 @@ impl StructuredData {
                    previous_owner_signatures: previous_owner_signatures
                  }
     }
+
+    /// This is a static function required for computing a name give the tag-type and identifier to
+    /// be use by GETs
+    pub fn compute_name(type_tag: u64, identifier: &NameType) -> NameType {
+        let type_tag_as_string = type_tag.to_string();
+
+        let chain = identifier.0.iter()
+                    .chain(type_tag_as_string.as_bytes().iter())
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/structured_data.rs
```diff
@@ -57,6 +57,19 @@ impl StructuredData {
                    previous_owner_signatures: previous_owner_signatures
                  }
     }
+
+    /// This is a static function required for computing a name give the tag-type and identifier to
+    /// be use by GETs
+    pub fn compute_name(type_tag: u64, identifier: &NameType) -> NameType {
```
benjaminbollen: What is the use-case for having compute_name as a public function, when we already have pub fn name?

benjaminbollen: Ok I see your use-case

benjaminbollen: Never mind my comments

ustulation: Explanation's in the documentaion :) .. Here it is in detail: In GETs I only have the type_tag and identifier, and i have to get the location by doing magic on it. That magic is the way structureddata name is calculated which also maintains an invariant. It needs to be available else i need to replicate the logic locally eveytime and it's not safe (if you ever decide to change the way name is calculated, none of my GETs will work)

benjaminbollen:
looks good to me; passing on to a second reviewer

PR #499, opened 2015-07-21
Refactor RoutingMembrane::handle_put_data as preparation for adding PureSentinel: MAID-1164

mmoadeli:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/499)

### DIFF
--- a/src/routing_membrane.rs
+++ b/src/routing_membrane.rs
@@ -915,11 +915,16 @@ impl<F> RoutingMembrane<F> where F: Interface {
         let from = message.source_address();
         //let to = message.send_to();
         let to = message.destination_address();
+        let quorum = types::QUORUM_SIZE;
+
+        if self.routing_table.size() < types::QUORUM_SIZE {
+            quorum = self.routing_table.size();
+        }
 
         let resolved = match self.put_sentinel.add_claim(
                         SentinelPutRequest::new(message, data, our_authority),
                         source, signed_message.signature.clone(),
-                        signed_message.encoded_body.clone(), QUORUM_SIZE) {
+                        signed_message.encoded_body.clone(), quorum) {
                             Some(result) =>  match  result {
                                 AddResult::RequestKeys(name) => {
                                     // Get Key Request
@@ -943,7 +948,7 @@ impl<F> RoutingMembrane<F> where F: Interface {
                         MethodCall::Forward { destination } => {
                             let message_id = self.get_next_message_id();
                             let msg = RoutingMessage {
-                                destination : DestinationAddress::Direct(destination),
+                                destination : DestinationAddress::Direct(resolved.0.destination_group),
                                 source      : SourceAddress::Direct(self.id.name()),
                                 orig_message: None,
                                 message_type: MessageType::PutData(resolved.0.data.clone()),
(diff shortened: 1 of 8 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/sentinel_request.rs
```diff
@@ -0,0 +1,49 @@
...
+// relating to use of the SAFE Network Software.
+
+use authority::Authority;
+use messages::{RoutingMessage};
+use name_type::NameType;
+use sentinel::pure_sentinel::Source;
+use data::Data;
+
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq)]
+pub struct SentinelPutRequest {
```
benjaminbollen: Can the structure be a "Request" without "Put" being an internal parameter? also it should not refer to "Sentinel", as we can want to use this struct to pass it to vaults.

src/routing_membrane.rs
```diff
@@ -55,6 +55,8 @@ use node_interface::MethodCall;
 use refresh_accumulator::RefreshAccumulator;
 use id::Id;
 use public_id::PublicId;
+use sentinel::pure_sentinel::*;
+use sentinel_request::SentinelPutRequest;
```
inetic: I would prefer not to use the wildcards, it makes it hard to see where types come from.

src/routing_membrane.rs
```diff
@@ -557,7 +561,12 @@ impl<F> RoutingMembrane<F> where F: Interface {
                         message.clone(), request.clone()),
                     MessageType::GetDataResponse(ref response) =>
                         self.handle_get_data_response(message_wrap, message.clone(), response.clone()),
-                    MessageType::PutData(ref data) => self.handle_put_data(message_wrap, message.clone(), data.clone()),
+                    MessageType::PutData(ref data) => {
+                        match message.source.actual_source() {
+                            Address::Node(name) => self.handle_group_put_data(message_wrap, message.clone(), data.clone(), name),
+                            Address::Client(name) => self.handle_client_put_data(message_wrap, message.clone(), data.clone(), name),
+                        }
+                    },
```
inetic: I think it is not correct to use `match message.source.actual_source()` here to determine whether the message came from a group. It is because source being a node does not imply it came from a group. I think better would be to use `message.from_authority().is_group()` (it has been recently added, so you'll need to pull from master).

mmoadeli: can you think of any scenario where a message from one single non-client node be processed?
My impression is that If it comes from a non-client node, then it should belong to a group.

inetic: I think e.g. a node that has not yet joined the network, or a PmidNode.

mmoadeli: a node which is not fully joined the network, depending on the size of its routing table expects a minimum number of messages.
Regarding to PmidNode, sorry, I am not sure I am aware of any operation in PmidNode that can be done by only one request. Both Put & Delete requires quorum size messages, I think.

inetic: > a node which is not fully joined the network, depending on the size of its routing table expects a minimum number of messages.

Sorry, I don't understand this argument. What I meant was when a normal node receives a message from a not-yet-connected node.

Though the real problem I see here is that we're assuming something from the upper layers (that a node never does non-group put request), if such assumption was valid, we should explicitly check it somewhere to make it clear that we require it. In here, the question is "does it come from a group or not" and `message.from_authority().is_group()` answers this question without any implicit assumptions (which may turn out true or false later).

benjaminbollen: mahmoud, please use message.from_group() as this is the way the sender can and will indicate how it thinks we should validate its message.

src/sentinel_request.rs
```diff
@@ -0,0 +1,52 @@
...
+use authority::Authority;
+use messages::{RoutingMessage};
+use name_type::NameType;
+use sentinel::pure_sentinel::Source;
+use types::MessageId;
+use data::Data;
+
+#[derive(Clone, Ord, PartialOrd, Eq, PartialEq)]
+pub struct SentinelPutRequest {
+    pub data: Data,
```
benjaminbollen: if instead of data we put a MessageType, then this structure is generic and we can have a group_sentinel instead of a put_group_sentinel

PR #500, opened 2015-07-22
Adds information for sentinel (name, pub-key)s to response messages

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/500)


maidsafe-highfive:
r? @benjaminbollen

inetic:
For reference: https://maidsafe.atlassian.net/browse/MAID-1180?filter=12102

src/messages.rs
```diff
@@ -92,20 +98,18 @@ impl ErrorReturn {
...
     DeleteData(DataRequest),
     DeleteDataResponse(ErrorReturn),
-    GetKey,
-    GetKeyResponse(NameType, sign::PublicKey),
     GetGroupKey,
-    GetGroupKeyResponse(Vec<(NameType, sign::PublicKey)>),
+    GetGroupKeyResponse(BTreeMap<NameType, sign::PublicKey>),
     Post(Data),
-    PostResponse(ErrorReturn),
+    PostResponse(ErrorReturn, BTreeMap<NameType, sign::PublicKey>),
```
benjaminbollen: What's your game here?

inetic: The idea is to not ask the network to pick members of the group twice. With non-response messages, any set of nodes {N1,...,Nk} can claim that they are a close group, so we can't trust that they really are. To fix that, we send the GetGroupKey message and the _network_ decides who is in the group by only sending the message to them (intermediate nodes will also catch this, but that is why we have the quorum). With response messages, we don't need to do this step, because we've already done it when we sent the request.

src/routing_membrane.rs
```diff
@@ -1326,6 +1363,16 @@ impl<F> RoutingMembrane<F> where F: Interface {
         Ok(())
     }
 
+    fn group_pub_keys(&self) -> BTreeMap<NameType, sign::PublicKey> {
+        let name_and_key_from_info = |node_info : NodeInfo| {
+            (node_info.fob.name(), node_info.fob.signing_public_key())
+        };
+
+        self.routing_table.target_nodes(&self.id.name())
```
benjaminbollen: we need to use explicitly RT::our_close_group and add our own name and key to it

inetic: Thanks, I missed that function when I was looking. I'll use that one then.

inetic:
You're right, I'll include that.

benjaminbollen:
divided on whether it "looks good to me" :-), but looks good to me, merging

PR #504, opened 2015-07-22
Separate RoutingMembrane::handle_{get,put}_data_response in preparati…

s-brian:
…on for Sentinel integration.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/504)

### DIFF
--- a/src/routing_membrane.rs
+++ b/src/routing_membrane.rs
@@ -45,7 +45,7 @@ use relay::{RelayMap, IdType};
 use sendable::Sendable;
 use data::{Data, DataRequest};
 use types;
-use types::{MessageId, Bytes, DestinationAddress, SourceAddress};
+use types::{MessageId, Bytes, DestinationAddress, SourceAddress, Address};
 use authority::{Authority, our_authority};
 use who_are_you::{WhoAreYou, IAm};
 use messages::{RoutingMessage, SignedMessage, MessageType,
@@ -552,11 +552,19 @@ impl<F> RoutingMembrane<F> where F: Interface {
                     //MessageType::FindGroupResponse(find_group_response) => self.handle_find_group_response(find_group_response),
                     MessageType::GetData(ref request) => self.handle_get_data(message_wrap,
                         message.clone(), request.clone()),
-                    MessageType::GetDataResponse(ref response) =>
-                        self.handle_get_data_response(message_wrap, message.clone(), response.clone()),
+                    MessageType::GetDataResponse(ref response) => {
+                        match message.from_group() {
+                            None => self.handle_client_get_data_response(message_wrap, message.clone(), response.clone()),
+                            Some(_) => self.handle_group_get_data_response(message_wrap, message.clone(), response.clone()),
+                        }
+                    },
                     MessageType::PutData(ref data) => self.handle_put_data(message_wrap, message.clone(), data.clone()),
-                    MessageType::PutDataResponse(ref response) => self.handle_put_data_response(message_wrap, message.clone(),
-                        response.clone()),
+                    MessageType::PutDataResponse(ref response) => {
+                        match message.from_group() {
+                            None => self.handle_client_put_data_response(message_wrap, message.clone(), response.clone()),
+                            Some(_) => self.handle_group_put_data_response(message_wrap, message.clone(), response.clone()),
+                        }
+                    },
                     MessageType::PutPublicId(ref id) => self.handle_put_public_id(message_wrap, message.clone(), id.clone()),
                     MessageType::Refresh(ref tag, ref data) => { self.handle_refresh(message.clone(), tag.clone(), data.clone()) },
                     MessageType::Post(ref data) => self.handle_post(message_wrap, message.clone(), data.clone()),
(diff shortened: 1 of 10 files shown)


maidsafe-highfive:
r? @benjaminbollen

inetic:
@s-brian @benjaminbollen, sorry, this is not yet finished. The main objective is to remove any mention of  any invariant message information from `handle_{get,put}_data_response` function right after they are passed to the sentinel. The confusing part is that there is currently no sentinel for these messages because its implementation is blocked by this task. But it will be right before `self.mut_interface().handle_XY(...)` function is called (see @mmoadeli's work on it [here](https://github.com/maidsafe/routing/pull/499/files#diff-647ebc1293551c577c303914f4c91ce6R937)).

The reason is, that after the messages pass the sentinel, every information that was unique per message is lost.

s-brian:
The task description states it's in preparation for use with Sentinel not that Sentinel should be used in the task. I'm failing to see how this task blocks writing the Sentinel that will handle the messages, if anything I would say it's better to have the Sentinel written first. The work that Mahmoud has done creates a struct specific to the message for a dedicated Sentinel, can't that be made generic? Also, after the Sentinel check it appears that the MethodCall's are going to be repeated, can't we have the forward and reply functions do the work, specifically removing unique information while retaining the invariants? In fact, I was discussing this earlier with Ben and I think we agree that once it can be seen definitively how the MethodCall's are used in all instances it would probably be better to factor them out to a separate function. Apologies if I'm missing something here but essentially I disagree about the status of the stated task.

inetic:
> The task description states it's in preparation for use with Sentinel not that Sentinel should be used in the task. I'm failing to see how this task blocks writing the Sentinel that will handle the messages, if anything I would say it's better to have the Sentinel written first.

This is a chicken and an egg thing. Doing this task first allows for finishing the task and not breaking compilation or tests. Imagine if we implemented sentinel first in e.g. handle_get_data_response function, after the message passes the sentinel it goes to the node_interface, that in turn returns the MethodCall action. Handling of the MethodCall currently uses e.g. the `message` which is just not available after sentinel, so you'll get compilation issues.

> The work that Mahmoud has done creates a struct specific to the message for a dedicated Sentinel, can't that be made generic?

Yes, the intention is to make it generic, if you can do it as part of this task that would be great :)

> Also, after the Sentinel check it appears that the MethodCall's are going to be repeated, can't we have the forward and reply functions do the work, specifically removing unique information while retaining the invariants?

Also, yes, I just did not think of it as part of this task. But if it's done then I'd be supper happy about it :)

> In fact, I was discussing this earlier with Ben and I think we agree that once it can be seen definitively how the MethodCall's are used in all instances it would probably be better to factor them out to a separate function.

Yap, that is the direction I would like to go too.

s-brian:
Sounds like the scope of the task has changed I guess it can be put in as an additional task.

inetic:
I'm not sure Brian, the task descriptions sais:

* Split the function into two: one for messages that came from a group and one for non-group messages
* The one handling group messages must not use 'variant' data after it is handled by the Interface.

You've done the first bullet, but the second one is more important (the first bullet is really just precondition for the second one). So I'm not sure why we should create a new task for it. That said, I only care that second bullet is implemented, I don't really mind creating new tasks, so feel free to suggest task name and description.

Note also that I do agree that the task description is quite vague, I really though it would be implemented by the same person who finishes MAID-1164. That is why I'll try to be available when I can to answer questions if something isn't clear.

dirvine:
Yes in a sprint if a task takes longer then we can see a planning deficiency on the task, so good. Creating new tasks means teh whole sprint planning is broken, but not as clear where or why. It's a balance and sometime swe may need new tasks where large things are missing, but at that stage we consider abandoning the sprint altogether I would think. In this case looks good to add more comment and revert to in progress etc.

src/routing_membrane.rs
```diff
@@ -1032,8 +1070,62 @@ impl<F> RoutingMembrane<F> where F: Interface {
...
+                    let message = RoutingMessage {
+                        destination  : DestinationAddress::Direct(destination),
+                        source       : SourceAddress::Direct(self.id.name()),
+                        orig_message : None,
+                        message_type : MessageType::PutDataResponse(response.clone(), BTreeMap::<NameType, sign::PublicKey>::new()),
+                        message_id   : message_id,
+                        authority    : our_authority.clone(),
+                    };
+                    let signed_message = SignedMessage::new(&message, self.id.signing_private_key());
+                    ignore(self.forward(&signed_message.unwrap(), &message, destination));
```
inetic: Can we do `try!(SignedMessage::new(...))` instead of unwraping?

src/routing_membrane.rs
```diff
@@ -1716,7 +1869,7 @@ fn populate_routing_node() -> RoutingMembrane<TestInterface> {
         assert_eq!(Tester::new().call_operation(put_data_response,
             SourceAddress::Direct(Random::generate_random()),
             DestinationAddress::Direct(Random::generate_random()),
-            Authority::NaeManager(Random::generate_random())).call_count, 1usize);
+            Authority::NaeManager(Random::generate_random())).call_count, 0usize);
```
inetic: This gives us false hope that everything works as expected. I think we should have it set to `1usize` and ignore the test for now. Then we can re-enable it as part of [MAID-1265](https://maidsafe.atlassian.net/browse/MAID-1265).

s-brian: If sentinel is doing what it's supposed to we would expect the return to be zero given the current setup of the test. Presumably additional or updated tests can still be addressed with MAID-1265.

inetic: fair enough

PR #512, opened 2015-07-24
Implements KeySentinel for FindGroupResponse.

s-brian:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/512)

### DIFF
--- a/src/routing_membrane.rs
+++ b/src/routing_membrane.rs
@@ -528,31 +528,9 @@ impl<F> RoutingMembrane<F> where F: Interface {
         // Handle FindGroupResponse
         match  message.message_type {
             MessageType::FindGroupResponse(ref vec_of_public_ids) => {
-                let mut quorum = types::QUORUM_SIZE;
-                if self.routing_table.size() < types::QUORUM_SIZE {
-                    quorum = self.routing_table.size();
-                }
-
-                let source = match message.source.actual_source() {
-                    Address::Node(source) => source,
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @inetic

src/routing_membrane.rs
```diff
@@ -524,10 +527,33 @@ impl<F> RoutingMembrane<F> where F: Interface {
...
+                    Some((request, vec_of_public_ids)) => (request, vec_of_public_ids),
+                    None => return Ok(())
+                };
+
                 ignore(self.handle_find_group_response(
-                            vec_of_public_ids.clone(),
-                            address_in_close_group_range.clone())),
+                            resolved.1.clone(),
+                            address_in_close_group_range.clone()));
+            },
```
inetic: Could this be a dedicated handle_XY function?

s-brian: Not sure exactly what you mean here Peter, is handle_find_group_response not a dedicated function?

inetic: Yes, sorry, missed the fact we have the function already. Then, what I was trying to ask is whether you could move the added lines to the `handle_find_group_response` function's body. It would be consistent with the other `handle_` functions and we would have less code inside this (already quite big) `message_received` function.

benjaminbollen:
implementation of IdTrait for PublicId is useful, but more easily repeated when putting in the sentinel after refactor

PR #522, opened 2015-07-27
identify all connections with I Am

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/522)


maidsafe-highfive:
r? @inetic

src/routing_membrane.rs
```diff
@@ -795,19 +795,13 @@ impl<F> RoutingMembrane<F> where F: Interface {
 
     fn handle_who_are_you(&mut self, endpoint: &Endpoint, serialised_message: Bytes)
```
inetic: Minor, should be renamed to `handle_i_am` I think

benjaminbollen:
failing tests are false negative from OSX Travis

PR #529, opened 2015-07-28
Review find group

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/529)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -215,11 +215,38 @@ impl<F, G> RoutingNode<F, G> where F : Interface + 'static,
             }
         }
 
+        let our_bootstrap = match possible_first {
+            // we bootstrapped to a node
+            false => {
+                // verify bootstrap connection
+                let our_bootstrap = match self.bootstrap {
+                    Some((ref endpoint, ref opt_name)) => {
+                        match *opt_name {
+                            Some(name) => {
+                                (endpoint.clone(), name.clone())
+                            },
+                            None => return Err(RoutingError::FailedToBootstrap)
+                        }
+                    },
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @inetic

benjaminbollen:
the ConnectRequests are still not compatible

src/routing_node.rs
```diff
@@ -215,11 +215,38 @@ impl<F, G> RoutingNode<F, G> where F : Interface + 'static,
...
+                // send FindGroup request before moving to Membrane
+                let find_group_msg =
+                    try!(self.construct_find_group_msg_as_client(&our_bootstrap.1));
+                ignore(cm.send(our_bootstrap.0.clone(), try!(encode(&find_group_msg))));
+
+                Some(our_bootstrap)
+            },
+            // someone tried to bootstrap to us
+            true => None
+        };
```
inetic: `if {} else {}` would do as well here.

src/routing_node.rs
```diff
@@ -215,11 +215,38 @@ impl<F, G> RoutingNode<F, G> where F : Interface + 'static,
             }
         }
 
+        let our_bootstrap = match possible_first {
+            // we bootstrapped to a node
+            false => {
+                // verify bootstrap connection
+                let our_bootstrap = match self.bootstrap {
+                    Some((ref endpoint, ref opt_name)) => {
```
inetic: Pattern matching can go several layers deep:
```
match self.bootstrap {
    Some((ref endpoint, Some(ref name)) => {
    ....
    }
    _ => return Err(RoutingError::FailedToBootstrap)
}
```

PR #533, opened 2015-07-29
Patch signature

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/533)

### DIFF
--- a/src/routing_membrane.rs
+++ b/src/routing_membrane.rs
@@ -503,7 +503,8 @@ impl<F> RoutingMembrane<F> where F: Interface {
         }
 
         // Forward
-        ignore(self.send_swarm_or_parallel_or_relay(&message));
+        ignore(self.send_swarm_or_parallel_or_relay_with_signature(
+            &message, &message_wrap.signature));
 
         let address_in_close_group_range =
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @inetic

benjaminbollen:
patch ready for review

src/routing_membrane.rs
```diff
@@ -502,7 +496,8 @@ impl<F> RoutingMembrane<F> where F: Interface {
         }
 
         // Forward
-        ignore(self.send_swarm_or_parallel_or_relay(&message));
+        ignore(self.send_swarm_or_parallel_or_relay_with_signature(
+            &message, message_wrap.signature().clone()));
```
inetic: Could this new `send_swarm_or_parallel_or_relay_with_signature` function just take `SignedMessage` as argument? In this case the `message_wrap` variable.

benjaminbollen: I went for this approach - fully in agreement it is not the most elegant - as it is clear, correct and consistent with the approach where no destination is passed into the send_* function (from handlers side) as to make the destination always read from the RoutingMessage;

equally taking into account that no additional unwrapping of the RoutingMessage needs to occur.

PR #534, opened 2015-07-29
Block bootstrap fn until IAm msg arrives

inetic:
So that client can start sending/receiving messages right after
bootstrap fn returns. Also renames 'run_one' fn to 'poll_one' to
be use consistent terminology with boost asio.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/534)


maidsafe-highfive:
r? @benjaminbollen

benjaminbollen:
looks good to me

PR #548, opened 2015-07-30
Remove bootstrap connections which were not used

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/548)


maidsafe-highfive:
r? @benjaminbollen

benjaminbollen:
looks good to me;

peter is addressing a final concern to ensure we use the endpoint that gave us the PutPublicIdResponse, as an additional cleanup;

PR #550, opened 2015-07-30
Store NameType for each bootstrap node

inetic:
until fully initialized (because we don't know which
bootstrap node initializes first).

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/550)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -36,7 +36,7 @@ use authority::{Authority};
 use messages::{RoutingMessage, SignedMessage, MessageType};
 use error::{RoutingError};
 use std::thread::spawn;
-use std::collections::BTreeSet;
+use std::collections::BTreeMap;
 
 static MAX_BOOTSTRAP_CONNECTIONS : usize = 1;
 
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/routing_node.rs
```diff
@@ -95,50 +95,59 @@ impl<F, G> RoutingNode<F, G> where F : Interface + 'static,
...
-                                    match message.message_type {
-                                        MessageType::PutPublicIdResponse(
-                                            ref new_public_id, ref _orig_request) => {
-                                              relocated_name = Some(new_public_id.name());
-                                              info!("Received name {:?} from network; original name was {:?}",
-                                                  relocated_name, self.id.name());
-                                              break;
+                    let opt_bootstrap_name = self.bootstraps.get(&endpoint).cloned();
+
+                    if let Some(opt_name) = opt_bootstrap_name {
```
benjaminbollen: I understand this correct, that this ensures that we only listen to NewMessage's from registered bootstrap connections; correct?

benjaminbollen:
looks good to me

PR #551, opened 2015-07-30
Don't drop all bootstrap nodes

inetic:
Fixes stupid mistake from my previous commit

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/551)


maidsafe-highfive:
r? @benjaminbollen

benjaminbollen:
I was observing the same :)
looks good to me !

PR #554, opened 2015-07-31
Test 531

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/554)


maidsafe-highfive:
r? @inetic

benjaminbollen:
CI might still be down; hence the local test, only on linux though

PR #555, opened 2015-07-31
POC to channels based simple routing

benjaminbollen:
Initial start to RoutingNode and RoutingHandler

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/555)


maidsafe-highfive:
r? @inetic

src/event.rs
```diff
@@ -16,138 +16,33 @@
...
-
-impl Source<NameType> for Event {
-    fn get_source(&self) -> NameType {
-        match self {
-            &Event::PutDataRequest(_, _, source_group, _, _, _, _) => source_group,
-            &Event::PutDataResponse(_, _, source_group, _, _, _, _) => source_group,
-            &Event::GetDataResponse(_, _, source_group, _, _, _, _) => source_group,
-        }
-    }
+    MessageSecured(RoutingMessage, Authority),
```
inetic: This event should provide enough information to create a reply message. Now I'm not sure whether this has changed or not, but my understanding is that responses need to contain the original signed message together with the signature. Now, since we decided to only support one hop in routing, this only applies for messages that came from clients and managed nodes (because messages originating in groups would all have a different signature due to the claimant being different per each message).

src/messages.rs
```diff
@@ -267,6 +263,9 @@ impl RoutingMessage {
 #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Debug, RustcEncodable, RustcDecodable)]
 pub struct SignedMessage {
     encoded_body : Vec<u8>,
+    claimant     : Address,
```
inetic: Just realized that this field will be redundant for cases when `from_authority` is `ManagedNode` and `undefined` when it is `Client`.

inetic: Oh, sorry, missed that the type is `Address`. It will be redundant in both cases.

src/event.rs
dirvine: I think possibly messages also `To` Clients / ManagedNodes (I am thinking of PmidNodes) need to be returned where they cannot take an action the network requests. I agree usually a group message is consensus checked and does not perhaps need to original signed message though. Interesting observation! 

Possibly if a group gives us back a error we can consider we asked for it, so security wise that is fine, the other side of the coin is getting the actual request back as the network cannot fix it, meaning we need the original message to be able to resolve the error.

benjaminbollen: So first remark, I removed the impl, to clear compilation errors, but obviously part of this needs to be restored.  Although i think reply would now not be on the `Event`.

It's a good remark.  Would create_reply on RoutingMessage be a good place for the implementation (also providing the original signature) returning a Result<RoutingMessage, _>

src/messages.rs
benjaminbollen: if it is ManagedNode, then yes it is redundant;
when it is client, it is not undefined, because Address is an enum {Node(nametype), Client(PublicKey)}

the only comment I have is that Address seems to be a misnomer, I'd prefer "Identity", but (re)naming isnt a priority just now

inetic: That is why I corrected myself, that is, it will not be undefined for `Client` case, but will be redundant the same way as it is redundant for `ManagedNode`.

benjaminbollen: true, but that hardly seems a problem.  It just states that in these two cases the SignedMessage is invalidly signed if they differ

src/event.rs
benjaminbollen: To be clear, the message_type can contain an original_message, and it is up to Vaults on what to do with such original message

src/messages.rs
inetic: Apart from the obvious problems that it increases bandwidth and that it will cause additional runtime checks - more lines of code and more CPU hungry (even if only little) - my biggest concern is that to me redundancy indicates bad design. I'll let you two be judges of it though, just wanted you to know about it.

src/event.rs
inetic: Yes, but the message_type doesn't contain the signature, you can't just make it up, so it needs to be part of the `Event`. (btw, I have no problem with you removing the `impl`).

PR #556, opened 2015-07-31
WIP: Integrate channel in routing_client

mmoadeli:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/556)

### DIFF
--- a/src/routing_membrane.rs
+++ b/src/routing_membrane.rs
@@ -948,10 +948,7 @@ impl<F> RoutingMembrane<F> where F: Interface {
                     Authority::ClientManager(name) => name,
                     Authority::NaeManager(name)    => name,
                     Authority::NodeManager(name)   => name,
-                    Authority::ManagedNode      => return Err(RoutingError::BadAuthority),
-                    Authority::ManagedClient(_) => return Err(RoutingError::BadAuthority),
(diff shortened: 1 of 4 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/routing_membrane.rs
```diff
@@ -948,10 +948,7 @@ impl<F> RoutingMembrane<F> where F: Interface {
                     Authority::ClientManager(name) => name,
                     Authority::NaeManager(name)    => name,
                     Authority::NodeManager(name)   => name,
-                    Authority::ManagedNode      => return Err(RoutingError::BadAuthority),
-                    Authority::ManagedClient(_) => return Err(RoutingError::BadAuthority),
-                    Authority::Client(_)        => return Err(RoutingError::BadAuthority),
-                    Authority::Unknown          => return Err(RoutingError::BadAuthority),
```
inetic: This was purposely exhaustive so that if a new Authority is added or some is changed/removed, compiler would warn us, that we need to check the code here.

mmoadeli: I think I wrote it myself :). I can revert it. I believe, this is better, though.

benjaminbollen:
will be outdated by refactor; routing_client is deprecated

PR #557, opened 2015-08-04
Routing Refactor to channels and core

benjaminbollen:
Full refactor of routing:
1. replace interface with channel architecture
2. invert hierarchy: vault/client now owns Routing; where before routing wrapped around vault and client was left dangling
3. simplify handlers, anything Routing does not act on is passed to the event channel
4. unify routing node
5. provide a clonable interface to routing_node

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/557)


maidsafe-highfive:
r? @inetic

src/authority.rs
```diff
@@ -82,7 +89,9 @@ pub fn our_authority(message       : &RoutingMessage,
         MessageType::ConnectResponse(_)     => None,
         MessageType::FindGroup              => None,
         MessageType::FindGroupResponse(_)   => None,
-        MessageType::GetData(_)             => Some(message.non_relayed_destination()),
+        MessageType::GetData(_)             => {
+            Some(message.destination().get_location().clone())
```
dirvine: Should this not be `Some(message.data.name())` or similar?

src/routing.rs
```diff
@@ -0,0 +1,102 @@
...
+//use error::{RoutingError};
+//use std::thread::spawn;
+//use std::collections::BTreeMap;
+
+/// Routing provides an actionable interface to RoutingNode.
+/// On constructing a new Routing object a RoutingNode will also be started.
+/// Routing objects are clonable for multithreading, or a Routing object can be
+/// cloned with a new set of keys while preserving a single RoutingNode.
+#[derive(Clone)]
+pub struct Routing {
```
dirvine: This is maybe only useful for trying connect requests/responses I imagine when we connect on one endpoint we then keep only that one endpoint for this connection. So we would send these as favoured endpoints when trying to connect (tcp/udt then ..)

src/authority.rs
inetic: I think so. IIRC in pre-rust-2 sprint code we had `GetData.data.name()` same as the destination the same, and it was carried here with it.

benjaminbollen: yes, I agree

src/routing.rs
benjaminbollen: This relates to the introduction of `peer` I assume. It might be worth if we can discuss this later, because I might not have understood fully.

benjaminbollen:
Can't be merged until LINT is re-enabled !

PR #561, opened 2015-08-07
remove feature gate core

benjaminbollen:
still leaves feature(convert)

needs to address the unit tests of NameType; convert is used to generate the full debug id (which we never use) but it is used in the unit tests to test the debug id output

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/561)


maidsafe-highfive:
r? @inetic

benjaminbollen:
I was too quick there, simply removing the #[feature(core)] is not good enough :)

PR #564, opened 2015-08-09
Pruning, also, cleaning up NameType traits a bit

dirvine:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/564)


maidsafe-highfive:
r? @inetic

src/id.rs
```diff
@@ -22,7 +22,6 @@ use NameType;
 // Note: name field is initially same as original_name, this should be later overwritten by
 // relocated name provided by the network using assign_relocated_name method
 // TODO (ben 2015-04-01) : implement order based on name
-#[derive(Clone)]
```
benjaminbollen: I was planning on cloning the Id from the initial Routing object into the Routing core, as vaults acting on Routing need to sign messages as the RoutingNode; alternatively we need to have a separate Action:: that takes the elements of a message, but not a fully SignedMessage, and then finish construction in RoutingNode

benjaminbollen:
looks good to me
raises few discussion points for Monday

inetic:
@dirvine CI check are failing, is that expected?

PR #570, opened 2015-08-10
Run

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/570)

PR #571, opened 2015-08-11
Example refactoring

inetic:
To the new Routing API

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/571)


maidsafe-highfive:
r? @benjaminbollen

examples/simple_key_value_store.rs
```diff
@@ -43,11 +43,12 @@ extern crate maidsafe_sodiumoxide as sodiumoxide;
 extern crate crust;
 extern crate routing;
 
-use core::iter::FromIterator;
+//use core::iter::FromIterator;
 use std::io;
 use std::net::SocketAddr;
 use std::str::FromStr;
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, mpsc};
```
benjaminbollen: Can we now do without an Arc / Mutex anymore?

inetic: I think we can.

examples/simple_key_value_store.rs
```diff
@@ -139,294 +144,156 @@ struct CliArgs {
...
+}
 
-    fn handle_post_response(&mut self, _from_authority: Authority,
-        _from_address: types::SourceAddress, _response: ResponseError)
-        -> Vec<MethodCall> {
-        unimplemented!();
-    }
+impl Client {
+    fn new(_bootstrap_peers: Vec<Endpoint>) -> Result<Client, RoutingError> {
+        let (routing_sender, routing_receiver) = mpsc::channel::<Event>();
```
benjaminbollen: just a style issue to consider: naming a channel WHAT_sender, and WHAT_receiver is more clear than WHO_sender/_receiver, as it feels to me less clear the direction of the channel

src/who_are_you.rs
```diff
@@ -20,9 +20,9 @@ use public_id::PublicId;
 use types::Address;
```
dirvine: We should rename the module / file here I think

benjaminbollen: I'll rename it in #570

benjaminbollen:
will already merge (in concert with @inetic s review of #570 - as both contain the others commits) 
the libsodium issue needs resolving independent from this PR

PR #572, opened 2015-08-12
RequestNetworkName

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/572)


maidsafe-highfive:
r? @inetic

src/routing_core.rs
```diff
@@ -365,4 +367,52 @@ impl RoutingCore {
...
+                let mut close_group : Vec<NameType> = routing_table
+                        .our_close_group().iter()
+                        .map(|node_info| node_info.fob.name())
+                        .collect::<Vec<NameType>>();
+                close_group.insert(0, self.id.name());
+                Some(close_group)
+            },
+            None => None,
+        }
+    }
```
inetic: Can we not just return `Vec<NameType>`, i.e. return an empty vector if `routing_table == None`?

inetic: Also, I think there are cases when we don't want us inside the group, e.g. sentinel will explicitly reject "self assigned" public keys.

benjaminbollen: That is what I did first, but corrected it afterwards based on the argument that if we are a client and not connected "our close group" is not defined; so seemed more correct to wrap it in an Option to allow for its non-existance

benjaminbollen: On the conversation topic: when writing the beginning of the article on formalising the logic, it became clear that the close group always includes ourselves; I think we should cater for it in Sentinel to detect ourselves, should such a problem arise

PR #573, opened 2015-08-12
use SignedMessage::Signature as FilterType for filtering messages

benjaminbollen:
The Signature in the SignedMessage is already the hash of all fields, including claimant and RoutingMessage, making it a unique identifier for detecting duplicate versions of the same message without additional overhead

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/573)

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -135,14 +135,13 @@ pub struct RoutingMessage {
     pub from_authority : Authority,
     pub to_authority   : Authority,
     pub content        : Content,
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @inetic

benjaminbollen:
added a TODO note to discuss at a later point the options for a user to explicitly send out multiple "identical" requests for eg increasing/decreasing a network based reference counter

benjaminbollen:
this removes message_id

PR #576, opened 2015-08-12
Accumulator

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/576)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -42,6 +42,7 @@ use utils::{encode, decode};
 use utils;
 use data::{Data, DataRequest};
 use authority::{Authority, our_authority};
+use std::cmp::min;
 
 use messages::{RoutingMessage,
                SignedMessage, SignedToken,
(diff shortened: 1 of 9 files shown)


maidsafe-highfive:
r? @benjaminbollen

benjaminbollen:
looks good ! great that we can indeed consider a general RoutingMessage accumulator !

dirvine:
:+1: , nice one @inetic

src/routing_node.rs
```diff
@@ -348,6 +352,11 @@ impl RoutingNode {
             return Err(RoutingError::BadAuthority);
         }
 
+        let (message, opt_token) = match self.accumulate(message_wrap) {
```
benjaminbollen: Can we exclude `InternalResponse::CacheNetworkName` from this accumulator; we are currently not using `FindGroup`/`Response::FindGroup` so no need to mind that; for now just don't accumulate CacheNetworkName Responses

benjaminbollen: Sorry you deal with this in `fn accumulate`; ignore my comment !

benjaminbollen: Sorry no :), this function does not exclude this one message type; can we patch this here?

benjaminbollen:
looks good but can you apply the patch to exclude InternalResponse::CacheNetworkName from the accumulator (it will never resolve as every message is seen as different, as it includes the close group)

benjaminbollen:
looks good to me

PR #577, opened 2015-08-13
Event sender

benjaminbollen:
includes #572

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/577)


maidsafe-highfive:
r? @inetic

PR #579, opened 2015-08-13
Simpler example

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/579)


maidsafe-highfive:
r? @benjaminbollen

benjaminbollen:
maybe we should rename this example "simpler_key_value_store" ;) - just joking

benjaminbollen:
looks good to me

PR #583, opened 2015-08-13
handle_cache_get & handle_cache_put

mmoadeli:
handle_get_cache returns Err(NoData) if it fails to provide data. This helps routing later to decide to  send request further. Returning Ok() indicates the request is handled.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/583)


maidsafe-highfive:
r? @benjaminbollen

src/routing_node.rs
```diff
@@ -30,8 +30,10 @@ use lru_time_cache::LruCache;
 
 use action::Action;
 use event::Event;
+use immutable_data::{ImmutableData, ImmutableDataType};
 use NameType;
 use name_type::{closer_to_target_or_equal};
+use plain_data::PlainData;
```
dirvine: What we want here is a cache per data type. The routing API will provide setters for each cache that will take the size (number of elements) and or time (number of minutes). This will allow fine grained control for users of routing and also allow us to test with no caching. The value should be the data type for each cache.

src/routing_node.rs
```diff
@@ -836,6 +840,86 @@ impl RoutingNode {
                                   refresh_our_own_group: bool) -> RoutingResult {
         unimplemented!()
     }
+
+    fn handle_cache_get(&mut self, request : ExternalRequest,
```
dirvine: Here we try to get the data direct from the lru_time cache, no need for authority etc.

src/routing_node.rs
```diff
@@ -836,6 +840,86 @@ impl RoutingNode {
...
+
+        match SignedMessage::new(Address::Node(self.core.id().name()),
+            routing_message, self.core.id().signing_private_key()) {
+                Ok(signed_message) => ignore(self.send(signed_message)),
+                Err(e) => return Err(RoutingError::Cbor(e)),
+            };
+        Ok(())
+    }
+
+    fn handle_cache_put(&mut self, request : ExternalRequest,
```
dirvine: Here just try and insert to lru_time_cache. No need to return errors etc. it's a best effort.

src/routing_node.rs
```diff
@@ -78,6 +80,7 @@ pub struct RoutingNode {
     public_id_cache     : LruCache<NameType, PublicId>,
     connection_cache    : BTreeMap<NameType, SteadyTime>,
     accumulator         : MessageAccumulator,
+    data_cache          : LruCache<NameType, Vec<u8>>,
```
benjaminbollen: we can store the actual types in the cache, without a need to serialise them into Vec<u8>

src/routing_node.rs
mmoadeli: Unused parameters are removed.

src/routing_node.rs
mmoadeli: immutable and plain data caches are added

src/routing_node.rs
```diff
@@ -65,19 +67,21 @@ static MAX_BOOTSTRAP_CONNECTIONS : usize = 1;
...
+    action_sender        : mpsc::Sender<Action>,
+    action_receiver      : mpsc::Receiver<Action>,
+    event_sender         : mpsc::Sender<Event>,
+    filter               : MessageFilter<types::FilterType>,
+    core                 : RoutingCore,
+    public_id_cache      : LruCache<NameType, PublicId>,
+    connection_cache     : BTreeMap<NameType, SteadyTime>,
+    accumulator          : MessageAccumulator,
+    immutable_data_cache : LruCache<NameType, Data>,
+    plain_data_cache     : LruCache<NameType, Data>,
```
dirvine: Are we missing a structuredData cache ?

src/routing_node.rs
```diff
@@ -65,19 +67,21 @@ static MAX_BOOTSTRAP_CONNECTIONS : usize = 1;
...
-    accumulator         : MessageAccumulator,
+    action_sender        : mpsc::Sender<Action>,
+    action_receiver      : mpsc::Receiver<Action>,
+    event_sender         : mpsc::Sender<Event>,
+    filter               : MessageFilter<types::FilterType>,
+    core                 : RoutingCore,
+    public_id_cache      : LruCache<NameType, PublicId>,
+    connection_cache     : BTreeMap<NameType, SteadyTime>,
+    accumulator          : MessageAccumulator,
+    immutable_data_cache : LruCache<NameType, Data>,
```
dirvine: Why is the Value Data and Not ImmutableData? etc.

src/routing_node.rs
mmoadeli: @benjaminbollen mentioned we do not cache structureddata

src/routing_node.rs
```diff
@@ -836,6 +842,64 @@ impl RoutingNode {
...
+                        _ => return Err(From::from(ResponseError::NoData)),
+                    }
+                }
+            },
+            _ => return Err(From::from(ResponseError::NoData)),
+        };
+
+        let routing_message = RoutingMessage {
+            from_authority : Authority::ManagedNode(self.core.id().name()),
+            to_authority   : Authority::ManagedNode(from_authority.get_location().clone()),
```
dirvine: Should to_authority not be the from_authority of the request?

src/routing_node.rs
```diff
@@ -836,6 +842,64 @@ impl RoutingNode {
                                   refresh_our_own_group: bool) -> RoutingResult {
         unimplemented!()
     }
+
+    fn handle_cache_get(&mut self, request : ExternalRequest,
+                                   from_authority : Authority, response_token : SignedToken)
+         -> RoutingResult {
```
dirvine: I wonder if this is not more complex than needed. It could return an Option<Data> so if the result is `Some(data)` we could call the reply method with the data and appropriate info. Then this is just a fn called when there is a get_data_request message.

src/routing_node.rs
mmoadeli: I could use ImmutableData. The only drawback would be converting it to the Data to send it in handle_cache_get.
I can change if you think ImmutableData is better

src/routing_node.rs
```diff
@@ -836,6 +842,64 @@ impl RoutingNode {
...
+
+        match SignedMessage::new(Address::Node(self.core.id().name()),
+            routing_message, self.core.id().signing_private_key()) {
+                Ok(signed_message) => ignore(self.send(signed_message)),
+                Err(e) => return Err(RoutingError::Cbor(e)),
+            };
+        Ok(())
+    }
+
+    fn handle_cache_put(&mut self, request : ExternalRequest) -> RoutingResult {
```
dirvine: This should only try and put to cache when the messagetype is a get_response, so if a cache is there it will add it, otherwise ignore it. I think this function requires no return.

src/routing_node.rs
mmoadeli: The function does not return Data. If data is found it is sent.

src/routing_node.rs
benjaminbollen: I did not relay our discussion to Mahmoud yesterday, David; so that's my bad

benjaminbollen:
I renamed your PR to WIP, as it needs rebasing after all the crazy merges just now; if not for other comments

src/routing_node.rs
mmoadeli: I change it

mmoadeli:
handle_cache_get & put is done. The test needs more improvement. Only immutableData is tested.

src/routing_node.rs
```diff
@@ -65,19 +68,22 @@ static MAX_BOOTSTRAP_CONNECTIONS : usize = 1;
...
-    accumulator         : MessageAccumulator,
+    action_sender         : mpsc::Sender<Action>,
+    action_receiver       : mpsc::Receiver<Action>,
+    event_sender          : mpsc::Sender<Event>,
+    filter                : MessageFilter<types::FilterType>,
+    core                  : RoutingCore,
+    public_id_cache       : LruCache<NameType, PublicId>,
+    connection_cache      : BTreeMap<NameType, SteadyTime>,
+    accumulator           : MessageAccumulator,
+    immutable_data_cache  : LruCache<NameType, ImmutableData>,
```
dirvine: These should all be `Option` We will provide setters to any lib user who wishes to set size or time. It's up to users to set these. As this code stands we are setting them regardless. I don't think this is how we wanted it.

src/routing_node.rs
```diff
@@ -92,21 +98,42 @@ impl RoutingNode {
...
+            bootstraps            : BTreeMap::new(),
+            action_sender         : action_sender,
+            action_receiver       : action_receiver,
+            event_sender          : event_sender.clone(),
+            filter                : MessageFilter::with_expiry_duration(Duration::minutes(20)),
+            core                  : RoutingCore::new(event_sender),
+            public_id_cache       : LruCache::with_expiry_duration(Duration::minutes(10)),
+            connection_cache      : BTreeMap::new(),
+            accumulator           : MessageAccumulator::new(),
+            immutable_data_cache  : LruCache::with_expiry_duration(Duration::minutes(10)),
```
dirvine: These should all be `None`

src/routing_node.rs
```diff
@@ -836,6 +863,114 @@ impl RoutingNode {
...
+                LruCache::<K, V>::with_expiry_duration_and_capacity(lru_time_to_live,
+                                                                    lru_capacity);
+         },
+         (Some(lru_time_to_live), None) => {
+             *lru_cache = LruCache::<K, V>::with_expiry_duration(lru_time_to_live);
+         },
+         (None, Some(lru_capacity)) => {
+             *lru_cache = LruCache::<K, V>::with_capacity(lru_capacity);
+         },
+         (None, None) => {}
```
benjaminbollen: This is not the expected behaviour; as setting none, none, now leaves it unchanged, where it would be expected to be changed to not having the cache; this function leaves the existing cache.  Rather than having this templated construction, an option on the actual cache seems simpler and more correct, would it not?

If any templating should occur, why not template it on the actual data types that are there?

benjaminbollen:
Can you resolve the merge conflicts for this work; from a quick glance you'll have to revisit the error types as eg ResponseError::NoData does no longer exist; a default paradigm is to return an Option<Data>, returning None if there is no Data;  Errors are reserved for when the message itself hits an insurmountable conflict

src/routing_node.rs
```diff
@@ -836,6 +863,139 @@ impl RoutingNode {
                                   refresh_our_own_group: bool) -> RoutingResult {
         unimplemented!()
     }
+
+    fn handle_cache_get(&mut self, request : ExternalRequest) -> Option<Data> {
+        match request {
+            ExternalRequest::Get(data_request) => {
+                match data_request {
+                    DataRequest::StructuredData(data_name, value) => {
+                        match self.structured_data_cache.as_mut() {
```
benjaminbollen: why is it borrowed .as_mut() here?

benjaminbollen:
Sorry, it seems I was going on an old memory when referring to the ResponseError::NoData, so you can ignore the above comment.  Looks good.  Can you just resolve the merge conflict though?

src/routing_node.rs
benjaminbollen: I just looked it up, yes it is, as LruCache updates itself on get; so looks good !

benjaminbollen:
Can you just address the merge conflict here; the new branch you opened pulls the old routing code again.

benjaminbollen:
the history is here https://github.com/maidsafe/routing/pull/583

PR #595, opened 2015-08-15
Fix appveyor sodium dependency

Viv-Rajkumar:
Fix the `libsodium` dependency by setting the env var `SODIUM_LIB_DIR` used by the custom build script for sodiumoxide.

While this PR fixes the dependency issue it also highlights the current compile failure in sodiumoxide module for x86 windows.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/595)


maidsafe-highfive:
r? @inetic

PR #596, opened 2015-08-16
handle_lost_connection

dirvine:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/596)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -168,18 +168,18 @@ impl RoutingNode {
         debug!("New connection on {:?}", endpoint);
         // only accept new connections if we are a full node
         // FIXME(dirvine) I am not sure we should not accept connections here :16/08/2015
-        // let has_bootstrap_endpoints = self.core.has_bootstrap_endpoints();
-        // if !self.core.is_node() {
-        //     if has_bootstrap_endpoints {
-        //         // we are bootstrapping, refuse all normal connections
-        //         self.connection_manager.drop_node(endpoint);
-        //         return;
-        //     } else {
-        //         let assigned_name = NameType::new(crypto::hash::sha512::hash(
-        //             &self.core.id().name().0).0);
-        //         let _ = self.core.assign_name(&assigned_name);
-        //     }
-        // }
+        let has_bootstrap_endpoints = self.core.has_bootstrap_endpoints();
+        if !self.core.is_node() {
+            if has_bootstrap_endpoints {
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/routing_node.rs
```diff
@@ -167,18 +167,19 @@ impl RoutingNode {
...
-            }
-        }
+        // FIXME(dirvine) I am not sure we should not accept connections here :16/08/2015
+        // let has_bootstrap_endpoints = self.core.has_bootstrap_endpoints();
+        // if !self.core.is_node() {
+        //     if has_bootstrap_endpoints {
+        //         // we are bootstrapping, refuse all normal connections
+        //         self.connection_manager.drop_node(endpoint);
+        //         return;
+        //     } else {
```
benjaminbollen: We can debate whether "normal connections" should not be accepted, but by disabling the self_assignment of a name the network will for sure not start, as everyone will stay a Client, and no-one will become a Node.

The condition upon which to refuse connections here is very subtle, and tests out to be correct - at least in the current implementation:
 if ! core.is_node()  ===>   if we do not have a Node(name)/ a routing table
    if has_bootstrap_endpoints ===> but we do have connections that are labelled as bootstrap;
===> we are connected to relay nodes (either trying to request a network name, or as pure clients), so we should not attempt to help other nodes bootstrap into the network, as we are ourselves still clients

dirvine: Updated now

src/routing_node.rs
```diff
@@ -190,8 +191,9 @@ impl RoutingNode {
 
     /// When CRUST reports a lost connection, ensure we remove the endpoint anywhere
     fn handle_lost_connection(&mut self, endpoint : Endpoint) {
-        error!("Lost connection on {:?}, but CORE IS NOT UPDATED!", endpoint);
-        //unimplemented!()
+        error!("Lost connection on {:?}", endpoint);
+        let connection_name = self.core.lookup_endpoint(&endpoint);
```
benjaminbollen: It is possibly more desirable to have an explicit function core::drop_endpoint (currently only have drop_peer), but that's up for discussion

dirvine: Yes agreed, it will do for now I suppose. Not to bad.

benjaminbollen:
looks good to me !

PR #597, opened 2015-08-16
Synchronise core

benjaminbollen:
This is a second patch to node::handle_hello to ensure that mutliple connections between two nodes only have a single resulting connection, correctly labelled on both sides.

The first patch ensured that a single connection was left standing after duplicate connections were detecting, by introducing rules to decide which node should take action (named alpha); however the non-alpha node would already update the core, potentially leaving it to refuse the the correct identification from the core, as decided upon by the alpha.

This is resolved here by having the alpha send a confirmation hello on the connection that he has added to his core, initiating the non-alpha to also add that connection as the connection to use to its core.

There are possibly security loop holes / opportunities for hacking in this setup; but the current focus is to establish a single connection between nodes (trimming out multiple connections), and having correct core information on both sides afterwards.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/597)

### DIFF
--- a/src/hello.rs
+++ b/src/hello.rs
@@ -21,8 +21,9 @@ use types::Address;
 
 #[derive(Debug, Eq, PartialEq, RustcEncodable, RustcDecodable)]
 pub struct Hello {
-    pub address   : Address,
-    pub public_id : PublicId,
+    pub address       : Address,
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @inetic

PR #602, opened 2015-08-17
DISCUSS: Cause for external request

benjaminbollen:
Solution proposed here; but to be discussed tomorrow morning

items that are debatable;
1. we only need a limited number of bits, as the risk of collision is rather small - this only needs to differentiate multiple identical ExternalRequests in a short timeframe (where filter remembers);
Hash is a good candidate as we need deterministic differentiation (a single client can use random numbers, but vaults starting a Request need to do this uniformly over a group.  As such not providing a "Cause" gives, identical Requests, but would not be able to differentiate multiple (legitimate ?) requests;

Equally we can reduce the filter time out;

in either case std::hash  ::SipHash -> u64 would suffice, but the function hash was marked unstable.  As such in this POC I used crypto::hash ::sha256 which is far more than needed - again, we are looking for collision avoidance in a small space.

2. I made it external in the Routing API, as a vault close group on churn should provide a uniform cause. if we would exclude vaults (ie don't give them a cause-differentiator) then multiple identical requests would not register - again it is unlikely that identical requests in vault in a short time frame would even occur; but it is not excluded either

### DIFF
--- a/examples/simple_key_value_store.rs
+++ b/examples/simple_key_value_store.rs
@@ -65,7 +65,7 @@ use routing::event::Event;
 use routing::data::{Data, DataRequest};
 use routing::plain_data::PlainData;
 use routing::utils::{encode};
-use routing::{ExternalRequest, SignedToken};
+use routing::{ExternalRequest, SignedToken, Cause};
 
 // ==========================   Program Options   =================================
 static USAGE: &'static str = "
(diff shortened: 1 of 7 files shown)


maidsafe-highfive:
r? @inetic

benjaminbollen:
We will NOT apply a measure to avoid the network from collapsing identical requests; even identical get requests (this is compensated with cache)

dirvine:
I know not a security issue, but siphash has been a cause of alarm a few time in industry so far, collisions. anyway SHA256 is very fast now on modern CPU's so the issue here will be we are transferring 256 bits instead of 32 or 64 bits. For me at the moment sha256 is probably OK until we know we have header size issues. I suspect we can reduce this header a lot anyway.

Agree on reduction of filter timeout, as this makes memory use better for the applications on top of routing.

dirvine:
>We will NOT apply a measure to avoid the network from collapsing identical requests; even identical get requests (this is compensated with cache)

:+1:

PR #604, opened 2015-08-17
disable connect on ConnectResponse (only for TCP)

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/604)


maidsafe-highfive:
r? @inetic

benjaminbollen:
we don't want to do this.

PR #605, opened 2015-08-17
Correcting Structured Data (conceptually and functionally)

ustulation:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/605)

### DIFF
--- a/src/structured_data.rs
+++ b/src/structured_data.rs
@@ -41,21 +41,24 @@ impl StructuredData {
     /// Constructor
     pub fn new(type_tag: u64,
                identifier: NameType,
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @inetic

src/structured_data.rs
```diff
@@ -75,23 +78,31 @@ impl StructuredData {
     /// To transfer ownership the current owner signs over the data, the previous owners field
     /// must have the previous owners of version - 1 as the current owners of that last version.
```
benjaminbollen: This documentation also needs update; but not crucial

PR #606, opened 2015-08-17
API update for Event::Connected

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/606)

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -16,7 +16,7 @@
 // relating to use of the SAFE Network Software.
 
 use authority::Authority;
-use messages::{RoutingMessage, ExternalRequest, ExternalResponse, Content, SignedToken};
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @inetic

src/event.rs
```diff
@@ -55,5 +56,10 @@ pub enum Event {
     //    ~~|~~~~~~~~~~
     //      | our close group sorted from our name; always including our name
     //      | if size > 1, we are connected to the network
+    Connected,
+    Disconnected,
+    FailedRequest(Authority, ExternalRequest, InterfaceError),
+    FailedResponse(Authority, ExternalResponse, InterfaceError),
+
```
dirvine: Looks good to me, I think this is minimal info but required to allow client apps to know the current state. :+1:

PR #613, opened 2015-08-18
Finalise api 0.3

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/613)

### DIFF
--- a/examples/simple_key_value_store.rs
+++ b/examples/simple_key_value_store.rs
@@ -134,7 +134,7 @@ struct Node {
 impl Node {
     fn new(_bootstrap_peers: Vec<Endpoint>) -> Result<Node, RoutingError> {
         let (sender, receiver) = mpsc::channel::<Event>();
-        let routing = try!(Routing::new(sender));
+        let routing = Routing::new(sender);
 
         Ok(Node {
             routing  : routing,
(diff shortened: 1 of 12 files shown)


maidsafe-highfive:
r? @inetic

examples/simple_key_value_store.rs
```diff
@@ -280,7 +278,8 @@ struct Client {
 impl Client {
     fn new(_bootstrap_peers: Vec<Endpoint>) -> Result<Client, RoutingError> {
```
dirvine: Are these bootstrap things not completely take care of in crust ? i.e. via the config file.

src/error.rs
```diff
@@ -30,20 +30,25 @@ use data::Data;
...
 impl error::Error for ResponseError {
     fn description(&self) -> &str {
         match *self {
-            ResponseError::NoData => "No Data",
-            ResponseError::InvalidRequest => "Invalid request",
-            ResponseError::FailedToStoreData(_) => "Failed to store data",
+            ResponseError::Abort => "Abort",
+            ResponseError::InvalidRequest(_) => "Invalid request",
+            ResponseError::FailedRequestForData(_) => "Failed request for data",
+            ResponseError::HadToClearSacrificial(size) => "Had to clear {:?} bytes of Sacrificial
```
dirvine: I am not sure this should be even known to routing.

examples/simple_key_value_store.rs
benjaminbollen: yes they are, and I have another WIP (not yet opened PR, where Im taking that out of the example)  good spotting !

src/error.rs
benjaminbollen: I agree!  I would want to minimise what Routing knows/defines about ExternalRequest ExternalResponse and associated errors; it's stupid that the user can only pass these error types - as one example

benjaminbollen: Is there a quick way to fix it now, or do we mark that for the technical death sprint?

PR #623, opened 2015-08-19
putting maximum allowed size of structured data

ustulation:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/623)


maidsafe-highfive:
r? @benjaminbollen

benjaminbollen:
not 100% sure this should be defined by routing, but I'll merge it for now to make sure safe_vault and safe_client use the same value!

PR #624, opened 2015-08-19
Example032

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/624)


maidsafe-highfive:
r? @inetic

benjaminbollen:
raised issue for intermittent failure in authority test (which has option to show intermittent failures for bad parameters in test) https://github.com/maidsafe/routing/issues/625

dirvine:
I believe those tests will (like routing table tests) need to run X times allowing Y failures really. Even then we could get Y failures and still within bounds (random bah!).

PR #637, opened 2015-08-23
Routing interface

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/637)


src/lib.rs
```diff
@@ -32,7 +32,7 @@
        html_favicon_url = "http://maidsafe.net/img/favicon.ico",
        html_root_url = "http://maidsafe.github.io/routing")]
 // #![warn(missing_docs)]
-// #![deny(dead_code, unused_variables, unused_features, unused_attributes)]
```
dirvine: nice

src/action.rs
```diff
@@ -30,6 +30,7 @@ pub enum Action {
     //          ~~|~~~~~~~~~~
     //            | a fully signed message with a given claimant
     SendContent(Authority, Content),
+    ClientSendContent(Authority, Content),
```
dirvine: needs discussion with you me and Brian

src/refresh_accumulator.rs
```diff
@@ -21,19 +21,17 @@ use NameType;
...
 //                     |
 pub type Request = (NameType, u64);
-
 pub struct RefreshAccumulator {
     //                                 +-> Who sent it
     //                                 |
     requests: LruCache<Request, Map<NameType, Bytes>>,
 }
 
+const MAX_REQUEST_COUNT: usize = 1000;
```
dirvine: Again works but should be calculated I think

src/routing_client.rs
```diff
@@ -0,0 +1,106 @@
...
+use messages::{ExternalRequest, ExternalResponse, InternalRequest, Content};
+
+type RoutingResult = Result<(), RoutingError>;
+
+/// Routing provides an actionable interface to RoutingNode.
+/// On constructing a new Routing object a RoutingNode will also be started.
+/// Routing objects are clonable for multithreading, or a Routing object can be
+/// cloned with a new set of keys while preserving a single RoutingNode.
+#[derive(Clone)]
+pub struct RoutingClient {
```
dirvine: I see why, but we are so close. Needs discussion as well. for now it will work though

src/routing_node.rs
```diff
@@ -497,10 +520,21 @@ impl RoutingNode {
                             None => return Err(RoutingError::UnknownMessageType),
                         }
                     }
-                    InternalRequest::Refresh(_, _) => {
-                        // TODO (ben 17/08/2015) implement handle Refresh identical to preceding
-                        // implementation
-                        Ok(())
+                    InternalRequest::Refresh(type_tag, bytes) => {
```
dirvine: :+1:

src/routing_client.rs
benjaminbollen: It's only necessitated because the user libraries still split in two different paths;  That is the sole reason routing gets split

PR #640, opened 2015-08-24
Adds nonce to ExternalRequest. Fixes #629.

### LINKED ISSUE #629: Nonce for client GetData Requests
filter blocks identical messages (same content from same claimant/client), including get data requests.  There are use-cases where the client might within filter-release time want to get the same data; in particular for Structured Data get the latest version.

This issue was recently explored with "cause for external request", but we decided to not go for this approach, as there are strong reservations against it.  see #602 

To avoid problems both for routing and safe_client at this point in time though, we will add a nonce (internally), specifically to `ExternalRequest::Get(DataRequest, nonce)`, where nonce is deliberately (comments?) kept small as a random `u8` to avoid a big space for frequent identical requests from the same client.

For full nodes (safe_vault) we need to ensure that groups act in unison, so no random values can be used and the nonce will be set as a fixed 0u8 if our authority is a group.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/640)

### DIFF
--- a/src/routing_client.rs
+++ b/src/routing_client.rs
@@ -72,10 +72,11 @@ impl RoutingClient {
 
     /// Send a Get message with a DataRequest to an Authority, signed with given keys.
     pub fn get_request(&mut self, location: Authority, data_request: DataRequest) {
+        self.get_counter = self.get_counter.wrapping_add(1);
         let _ = self.action_sender.send(Action::ClientSendContent(
                 location,
                 Content::ExternalRequest(
-                    ExternalRequest::Get(data_request, self.get_counter.wrapping_add(1)))));
+                    ExternalRequest::Get(data_request, self.get_counter))));
     }
 
     /// Add something to the network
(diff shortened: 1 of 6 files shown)


src/routing_client.rs
```diff
@@ -67,14 +67,15 @@ impl RoutingClient {
...
+        RoutingClient { action_sender: action_sender, get_counter: 0u8 }
     }
 
     /// Send a Get message with a DataRequest to an Authority, signed with given keys.
-    pub fn get_request(&self, location: Authority, data_request: DataRequest) {
+    pub fn get_request(&mut self, location: Authority, data_request: DataRequest) {
         let _ = self.action_sender.send(Action::ClientSendContent(
                 location,
-                Content::ExternalRequest(ExternalRequest::Get(data_request))));
+                Content::ExternalRequest(
```
benjaminbollen: this will always give 1u8; you need to set get_counter

benjaminbollen:
looks good to me

PR #650, opened 2015-08-26
Cache data in ttl lru cache in line with current routing requirements.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/650)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -79,9 +79,7 @@ pub struct RoutingNode {
     connection_cache: BTreeMap<NameType, SteadyTime>,
     accumulator: MessageAccumulator,
     refresh_accumulator: RefreshAccumulator,
-    immutable_data_cache: Option<LruCache<NameType, ImmutableData>>,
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @benjaminbollen

dirvine:
This is much cleaner, my only concern is that we should be able to set a cache per type, so may need setters anyway to let routing know if it should even try a data type (i.e. we won't use structuredData types in cache). So routing needs to know what to cache, apart from that looks much nicer.

benjaminbollen:
Can you consider following points
1. maybe expand the functionality in a cache.rs object
2. make an elegant way of passing in the constructor of routing -> routing_node which types to cache 
3. actually use the functions in RoutingNode::message_received post filter; if the cache.rs has a nice interface that takes &RoutingMessage or a $Content, then the message_received function can remain concise with single (or a few cache lines)

benjaminbollen:
already merging in; but should be continues

PR #654, opened 2015-08-26
enable Scanning

benjaminbollen:
and temporarily disable connection_filter (as unwraps in that crate are hit)

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/654)


maidsafe-highfive:
r? @brian-js

PR #655, opened 2015-08-26
WIP enable crude data churn in the simple key value store example

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/655)


maidsafe-highfive:
r? @brian-js

benjaminbollen:
I could establish crude data churn on the example; there is a problem with the accumulator threshold that now needs to be calculated as planned

dirvine:
@benjaminbollen Is this good to merge now?

PR #660, opened 2015-08-30
Refactor main event loop in routing 

dirvine:
To accommodate both efficiency and also reduction in CPU usage. Can possibly be improved by only having a nanosleep when loop has not actually matched any receive channel. For now this appears to be successful in ensuring nodes are stable. There was an issue waiting on crust actions whilst routing was attempting to handle an event.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/660)

### DIFF
--- a/examples/simple_key_value_store.rs
+++ b/examples/simple_key_value_store.rs
@@ -132,6 +132,9 @@ impl Node {
                                         from_authority,
                                         response_token);
                 },
+                Event::Churn(our_close_group) => {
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @brian-js

benjaminbollen:
Also merges my own work, but i have tested that so I would not need to hold that back.
The change to the event loop now handles actions and churn on a 1-to-1 basis, which possibly is justified as actions are becoming increasingly important too

PR #662, opened 2015-08-31
Enable caching in RoutingNode.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/662)


maidsafe-highfive:
r? @benjaminbollen

src/routing_node.rs
```diff
@@ -1138,35 +1154,139 @@ impl RoutingNode {
...
+                                                            data_request,
+                                                            None);
+                                                    return Some(Content::ExternalResponse(response));
+                                                },
+                                                None => return None
+                                            }
+                                        }
+                                        return None;
+                                    }
+                                    DataRequest::StructuredData(data_name, version) => {
```
benjaminbollen: It is only a name, but for code clarity it is worthwhile fixing, the `u64` in the `DataRequest::StructuredData(identifier_name, type_tag)` - it is not the version number of the StructuredData (the SAFE network does not maintain the version history of an SD, the internal version number functions only to linearise the history in clients control)

src/routing_node.rs
```diff
@@ -1138,35 +1154,139 @@ impl RoutingNode {
...
+                                        if self.cache_options.structured_data_caching_enabled() {
+                                            match data_cache.get(&data_name) {
+                                                Some(data) => {
+                                                    match *data {
+                                                        Data::StructuredData(ref structured_data) => {
+                                                            if structured_data.get_version() == version {
```
benjaminbollen: So here and on line 1239 the name of the structured data needs to be calculated and used - no version number checking required.  The data_cache::get (and corresponding put) needs to follow

DataRequest::StructuredData(ref name, tag) => StructuredData::compute_name(tag, name)

benjaminbollen: It is worth noting that this is the important reason why safe_vault will not enable caching for structured data.  Routing is purely catering for generic uses here, and leaves it to the discretion of the user to assert that caching of structured data is sensible.

src/routing_node.rs
```diff
@@ -402,6 +407,19 @@ impl RoutingNode {
...
         }
 
+        let message = signed_message.get_routing_message().clone();
+
+        // Cache a response if from a GetRequest and caching is enabled for the Data type.
+        self.handle_cache_put(&message);
+        // Get from cache if it's there.
+        match self.handle_cache_get(&message) {
+            Some(content) => {
+                return self.send_content(
```
benjaminbollen: current practice is not good either, but it would currently be written
let _ = self.send_content(...);
return Ok(())

but this is just a note, and not needed to be changed.

src/routing_node.rs
brian-js: Thanks for the clarification Ben.

benjaminbollen:
looks good to me

PR #663, opened 2015-09-01
Connected vaults

benjaminbollen:
Limit churn signaling to user until a routing node has at least QUORUM_SIZE routing table connections

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/663)


maidsafe-highfive:
r? @brian-js

PR #667, opened 2015-09-02
PAUSED introduce RoutingError::Crust(::crust::error::Error)

benjaminbollen:
waiting on changes from Crust, and resolution of https://github.com/maidsafe/crust/issues/303

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/667)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -165,6 +165,8 @@ pub enum RoutingError {
     Cbor(CborError),
     /// invalid response
     Response(ResponseError),
+    /// crust error
+    Crust(::crust::error::Error),
 }
 
 impl From<str::Utf8Error> for RoutingError {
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @brian-js

src/error.rs
```diff
@@ -165,6 +165,8 @@ pub enum RoutingError {
     Cbor(CborError),
     /// invalid response
     Response(ResponseError),
+    /// crust error
+    Crust(::crust::error::Error),
```
Fraser999: I think you should be able to use `::crust::Error` here (http://maidsafe.net/crust/master/crust/index.html#reexports).

benjaminbollen:
Has been taken over by MAID-1366 and #704

PR #672, opened 2015-09-07
Maid 1349 refresh authority

benjaminbollen:
The code needed for the library is limited.  The key_value_store however, has no refresh message to test this behaviour at the routing layer.  I will further expand the example in the 8 points, to include an account, and corresponding refresh message.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/672)


maidsafe-highfive:
r? @brian-js

Issue #683, opened 2015-09-11
Refine Churn and Connected events for vaults


benjaminbollen:
`Event::Churn`​ from the very first new connection with the ​`length -1`​ indicating the number of connected nodes (maxed at ​`group_size`​) and ​`Event::Connected`​ only at reaching ​`GroupSize`​ number of connections (+1 ourselves)
and ​`Event::Disconnected`​ will be (for now) asymmetrical; kept at the original threshold: triggered on `1->0` routing table connections
and then ​`Terminate`​ WILL NOT automatically be called.  The user is then expected to know that on `Disconnected` routing has become useless.

PR #684, opened 2015-09-11
address Connected and Churn behaviour as discussed in #683

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/684)


maidsafe-highfive:
r? @brian-js

benjaminbollen:
fixes #683

PR #696, opened 2015-09-16
expose const NAME_TYPE_LEN and cfg(test) random traits

### LINKED ISSUE #674: Expose `NameType`'s length and/or expose `NameType::generate_random()`
It would be useful if Routing either exposed [`NAME_TYPE_LEN`](https://github.com/maidsafe/routing/blob/master/src/name_type.rs#L24) or provided a static getter in the `NameType` struct, e.g.

    pub fn len() -> usize { NAME_TYPE_LEN }

The only alternatives I think we currently have now are to use the `len()` of an actual instance, or just use a hard-coded value of `64`, neither of which is ideal.

The need arises in my case because I'm creating a random `NameType`for testing Vault accounts.  I have to go through the horribly-convoluted process of:

    ::routing::NameType(::routing::types::vector_as_u8_64_array(
        ::routing::types::generate_random_vec_u8(64)))

(well - obviously I don't *have* to - I could write my own functions for this, but in the interests of DRY...)

I would much rather just use:

    ::routing::NameType::generate_random()

but this hasn't been exposed and I seem to remember that was by design.  So failing this becoming available, I'd rather not hard-code `64` - i.e. I'd rather write:

    ::routing::NameType(::routing::types::vector_as_u8_64_array(
        ::routing::types::generate_random_vec_u8(::routing::NameType::len())))
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^

Even this doesn't shield us from Routing changing from 64, since I imagine the public function `vector_as_u8_64_array` would get renamed.

benjaminbollen:
fixes #674

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/696)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,6 +84,6 @@ pub mod plain_data;
 pub mod data;
 
 /// NameType is a 512bit name to address elements on the DHT network.
-pub use name_type::{NameType, closer_to_target};
+pub use name_type::{NameType, closer_to_target, NAME_TYPE_LEN};
 pub use messages::{SignedToken, ExternalRequest, ExternalResponse};
 pub use authority::Authority;
--- a/src/name_type.rs
+++ b/src/name_type.rs
@@ -38,12 +38,7 @@ impl NameType {
     pub fn new(id: [u8; NAME_TYPE_LEN]) -> NameType {
         NameType(id)
     }
-
-    // TODO(Ben): Resolve from_data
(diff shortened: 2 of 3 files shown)


maidsafe-highfive:
r? @brian-js

benjaminbollen:
@Fraser999 would you want to review and merge this so that I can put it into the version change for routing 0.3.11 ?

PR #704, opened 2015-09-18
MAID-1366 BREAKS ROUTING BOOTSTRAPPING crust 3 refactor

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/704)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,8 @@ sodiumoxide = "*"
 #crust = "0.2.*"
 
 [dependencies.crust]
-git = "https://github.com/maidsafe/crust"
+branch = "accept-returns-endpoint"
+git = "https://github.com/inetic/crust"
 
 [dev-dependencies]
 bit-vec = "*"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,8 +25,7 @@ sodiumoxide = "*"
 #crust = "0.2.*"
 
 [dependencies.crust]
-branch = "pub-proto"
-git = "https://github.com/inetic/crust"
+git = "https://github.com/maidsafe/crust"
 
 [dev-dependencies]
 bit-vec = "*"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,10 +22,7 @@ itertools = "*"
 log = "*"
 env_logger = "*"
 sodiumoxide = "*"
-#crust = "0.2.*"
-
-[dependencies.crust]
-git = "https://github.com/maidsafe/crust"
+crust = "0.3.*"
 
 [dev-dependencies]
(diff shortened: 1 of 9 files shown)


maidsafe-highfive:
r? @brian-js

Cargo.toml
```diff
@@ -22,7 +22,10 @@ itertools = "*"
 log = "*"
 env_logger = "*"
 sodiumoxide = "*"
-crust = "0.2.*"
+#crust = "0.2.*"
+
+[dependencies.crust]
+path = "/home/peter/work/rust/crust"
```
benjaminbollen: yeah, that's not going to work well on the CI machines ;)

Cargo.toml
```diff
@@ -22,7 +22,11 @@ itertools = "*"
 log = "*"
 env_logger = "*"
 sodiumoxide = "*"
-crust = "0.2.*"
+#crust = "0.2.*"
+
+[dependencies.crust]
+branch = "pub-proto"
```
benjaminbollen: at what point will this be merged into master and published to crates.io?

inetic: Changed to point to maidsafe's git repo, going to discuss with Viv adding it to crates.

src/routing_node.rs
```diff
@@ -74,15 +73,27 @@ impl RoutingNode {
...
 
         let (crust_sender, crust_receiver) = ::std::sync::mpsc::channel::<::crust::Event>();
-        let cm = ::crust::ConnectionManager::new(crust_sender);
-        let accepting_on = cm.get_own_endpoints();
+        let mut crust_service = match ::crust::Service::new(crust_sender) {
+            Ok(service) => service,
+            Err(what) => panic!(format!("Unable to start crust::Service {}", what)),
+        };
+
+        let accepting_on = crust_service.start_default_acceptors().into_iter()
```
benjaminbollen: is this now a non-blocking call, but with the iterator it is made blocking, correct?

inetic: No, the `Service::start_default_acceptors()` returns simply a `Vec<io::Result<Endpoint>>` which are the (local) endpoints crust is now accepting on.

src/routing_node.rs
```diff
@@ -939,6 +963,13 @@ impl RoutingNode {
         }
     }
 
+    fn connect(&mut self, endpoints: &Vec<::crust::Endpoint>) {
+        for e in endpoints {
+            self.connect_requests.insert(e.clone());
```
benjaminbollen: this is not okay :).  We do not want to keep any state on ::connect; but I think I can easily enough fix it, so I'll raise an issue and take it out afterwards

PR #715, opened 2015-09-23
Test on droplets - restoring state to compare for routing issue observed on droplets

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/715)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -61,6 +61,7 @@ pub struct RoutingNode {
     refresh_accumulator: ::refresh_accumulator::RefreshAccumulator,
     cache_options: CacheOptions,
     data_cache: Option<LruCache<NameType, Data>>,
+    connect_requests: ::std::collections::BTreeSet<::crust::Endpoint>,
 }
 
 impl RoutingNode {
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @brian-js

benjaminbollen:
I have tested this locally on my MacBook, but the problem predominantly shows on the droplet network, and I have not recently run performance tests from my MacBook, so hard to compare.  The network seems to connect. 

This is a guess on whether this can abate the problem reported this morning.  In either case, with upcoming connection management review in routing this needs to be further examined

benjaminbollen:
did it help?

brian-js:
Peter resolved some crust issues and it appears to be ok now. Still to test vaults though.

benjaminbollen:
ok :) so there was not really a reason to merge this then, as it introduces a pile-up of state :).  No worries, we'll address it Monday.

brian-js:
No and yes.

Viv-Rajkumar:
well if we don't need this and peter's fixes worked without this, why don't we just have this reverted before the version bump in routing? @brian-js @benjaminbollen

PR #717, opened 2015-09-25
Revert-715, Interface addresses and crust dependency

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/717)


maidsafe-highfive:
r? @brian-js

PR #730, opened 2015-10-11
Optional min/max dates in StructuredData for proof of storage date

Thierry61:
This is an implementation of my proposal in following post:

https://forum.safenetwork.io/t/any-time-consensus-within-safenet/2891/14?u=tfa

The aim is to allow an optional proof of storage date of StructuredData that doesn’t need a complex time synchronization mechanism in the safe network.

The following elements have been added:
  - Two fields in StructuredData (min_date/max_date)
  - Two default values written to these fields in the existing constructor (StructuredData::new) so that the behavior of classic StructuredData remains unchanged (no date validation)
  - A new constructor (StructuredData::with_dates) with 2 supplementary arguments specifying a range of dates for the proof of storage.
  - A validation function (validate_date) that checks that current date is within the specified range.
  - inclusion of the new fields in the signed part of StructuredData
  - Two public getter functions (get_min_date/get_max_date)
  - Display of [min_date..max_date] range in the format debug trait of StructuredData
  - Three tests functions (default_dates, valid_dates, invalid_dates) that check the validation function in the three possible cases.
  - A new error code (RoutingError::OutOfRangeDate) which can be returned by the validation function

Usage notes:
  - The new constructor has to be called from the clients to activate this functionality
  - The validation function has to be called in each put and post request of SD. I searched for site calls for existing functions like replace_with_other or validate_self_against_successor but without success (including safe_vault crate). I have the feeling that these kind of controls are not yet activated (except in test functions). When they are, the validate_date function should also be called.

Potential problem:

This solution works only if validation is done when the user put or post the SD. It doesn’t work if the validation must be done again at later times (for example during churn events).

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/730)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -145,6 +145,8 @@ pub enum RoutingError {
     NotEnoughSignatures,
     /// Duplicate signatures
     DuplicateSignatures,
+    /// Current date is ouside specified range [min_date .. max_date] of structured data
+    OutOfRangeDate,
     /// duplicate request received
     FilterCheckFailed,
     /// failure to bootstrap off the provided endpoints
@@ -212,6 +214,7 @@ impl ::std::error::Error for RoutingError {
             RoutingError::FailedSignature => "Signature check failure",
             RoutingError::NotEnoughSignatures => "Not enough signatures",
             RoutingError::DuplicateSignatures => "Duplicated signatures",
+            RoutingError::OutOfRangeDate => "Current date is outside specified range",
             RoutingError::FailedToBootstrap => "Could not bootstrap",
             RoutingError::RoutingTableEmpty => "Routing table empty",
             RoutingError::RejectedPublicId => "Rejected Public Id",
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
Thanks for the pull request, and welcome! The MaidSafe team is excited to review your changes, and you should hear from @brian-js (or someone else) soon.

dirvine:
Thanks you for this, but a massive caveat. There is no time used between nodes anywhere in the network and we have been adamant about this since the inception. It causes cases where logic is confused and will for sure require some synchronisation which gets back to ntp servers etc. 

It would be a very deep rooted change to include time between nodes without at minimum an RFC which will be very distracting right now, but if in the pipeline would be discussed. 

It is not the code that would be the issue but the factor of time, new folk to the code talk of time like this in house and are very quickly astonished at the response of the Engineers regarding time. So I cannot really emphasise the enormity of such a change. 

With this in mind I would suggest taking this to an RFC instead of a pull request and follow that process which will help guide us through the pros cons and drawabcks as well as forcing a loot at alternatives. 

In terms of proof of storage, it seems strongly tied into the functioning logic at the moment so it would be an interesting RFC if that were to be changed at this stage. It is the right way to go though. 

So thanks again, but for such changes we must follow the RFC process.

Thierry61:
Okay, I will write a RFC.

But I think you didn't understand my proposal. It is really a simple thing that don't need NTP servers or even synchronization between nodes. 

To workaround the date variability among the nodes, the client adds a date range in the SD. For the validation consensus a node has just to check that its own date is inside this range.

To maximize the chances that its request is valid the client can give a very large range like current date ± one hour or even ± one day. Most of the times people don't need a precise dating for the proof of storage. 

Another unrelated thing: my latest commit is a correction of a regression brought by the latest nightly rust version. You should take it otherwise people will soon complain that routing crate doesn't compile anymore.

brian-js:
@Thierry61 Closing PR in favour of pending RFC. The types changes have been committed.

PR #732, opened 2015-10-12
ExpirationMap.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/732)

### DIFF
--- a/src/utilities/expiration_map.rs
+++ b/src/utilities/expiration_map.rs
@@ -15,19 +15,17 @@
 // Please review the Licences for the specific language governing permissions and limitations
 // relating to use of the SAFE Network Software.
 
-extern crate time;
-
 /// A time limited map of key-value pairs.
 #[allow(unused)]
 pub struct ExpirationMap<K, V> {
-    map: ::std::collections::BTreeMap<K, (V, time::SteadyTime)>,
-    time_to_live: time::Duration,
+    map: ::std::collections::BTreeMap<K, (V, ::time::SteadyTime)>,
+    time_to_live: ::time::Duration,
 }
 
 #[allow(unused)] // remove when an ExpirationMap has been included elsewhere in the code
 impl<K, V> ExpirationMap<K, V> where K: PartialOrd + Ord + Clone, V: Clone {
     /// Constructor
-    pub fn with_expiry_duration(time_to_live: time::Duration) -> ExpirationMap<K, V> {
+    pub fn with_expiry_duration(time_to_live: ::time::Duration) -> ExpirationMap<K, V> {
         ExpirationMap {
             map: ::std::collections::BTreeMap::new(),
             time_to_live: time_to_live,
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/utilities/expiration_map.rs
```diff
@@ -0,0 +1,203 @@
...
+// Licenses can be found in the root directory of this project at LICENSE, COPYING and CONTRIBUTOR.
+//
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+extern crate time;
```
benjaminbollen: is this not already included in lib.rs?

src/utilities/expiration_map.rs
```diff
@@ -0,0 +1,201 @@
...
+        match self.map.remove(key) {
+            Some((value, _)) => Some(value),
+            None => None,
+        }
+    }
+
+    /// Retrieves a value for the given key if present in the map.
+    pub fn get(&mut self, key: &K) -> Option<&V> {
+        match self.map.get(key) {
+            Some(&(ref value, _)) => Some(value),
```
benjaminbollen: should the expiration not be validated on get and ignored if expired?  similar for contains_key.  On remove we can silently remove and has_expired is then not needed.  Sounds good? :)

src/utilities/expiration_map.rs
```diff
@@ -0,0 +1,201 @@
...
+            assert!(expiration_map.get(&i).is_none());
+        }
+
+        assert!(expiration_map.contains_key(&11));
+        assert_eq!(expiration_map.get(&11).unwrap(), &11);
+    }
+
+    #[test]
+    fn insert_same_key_with_different_value() {
+        let duration = ::time::Duration::milliseconds(500);
```
benjaminbollen: can we reduce this to 10 ms?

src/utilities/expiration_map.rs
brian-js: Already done.

src/utilities/expiration_map.rs
```diff
@@ -0,0 +1,210 @@
...
+            None => false
+        }
+    }
+
+    /// Recover expired key-value pairs removing any such from the map.
+    pub fn remove_expired(&mut self) -> Option<Vec<(K,V)>> {
+        let mut expired = Vec::new();
+
+        for (key, &(ref value, time)) in self.map.iter() {
+            if time + self.time_to_live < ::time::SteadyTime::now() {
```
benjaminbollen: probably better to only call ::time::SteadyTime::now() only once, but good for now

benjaminbollen:
looks good to me

PR #737, opened 2015-10-15
Maid 1380 use connection map

benjaminbollen:
Continues #734 for MAID-1380

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/737)


maidsafe-highfive:
r? @brian-js

PR #738, opened 2015-10-16
OnConnect.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/738)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -185,15 +185,7 @@ impl RoutingNode {
                     };
                 }
                 Ok(::crust::Event::OnConnect(connection)) => {
-                    // FIXME (ben 21/09/2015) new logic needs to be considered to properly remove
-                    // the concept of a first node, as it is just hidden, not logically removed
-                    // refactoring to crust 0.3 has made this logic even worse than it was.
-                    if self.core.is_node() {
-                        // self.handle_new_connection(connection);
-                        self.handle_on_connect(connection);
-                    } else {
-                        self.handle_new_bootstrap_connection(connection);
-                    }
+                    self.handle_on_connect(connection);
                 }
                 Ok(::crust::Event::OnAccept(connection)) => {
                     self.handle_new_connection(connection);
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/routing_node.rs
```diff
@@ -189,7 +189,8 @@ impl RoutingNode {
                     // the concept of a first node, as it is just hidden, not logically removed
                     // refactoring to crust 0.3 has made this logic even worse than it was.
                     if self.core.is_node() {
```
benjaminbollen: can we get rid of this self.core.is_node() altogether; directly go to handle_on_connect() and inside that function we can send a (client_)hello (if we are a client)

src/routing_node.rs
```diff
@@ -232,6 +233,33 @@ impl RoutingNode {
...
 
+    fn handle_on_connect(&mut self, connection: ::crust::Connection) {
+        match self.core.state() {
+            &::routing_core::State::Disconnected => {
+                self.core.add_bootstrap_connection(connection.clone());
+                ignore(self.send_hello(connection, None));
+                return;
+            },
+            &::routing_core::State::Bootstrapped => {
+                self.crust_service.drop_node(connection);
```
benjaminbollen: combined with above comment, here we would also send a hello (which will indicate that we are a client connecting)

src/routing_node.rs
```diff
@@ -232,6 +233,33 @@ impl RoutingNode {
           self.set_cache_options(preserve_cache_options);
     }
 
+    fn handle_on_connect(&mut self, connection: ::crust::Connection) {
+        match self.core.state() {
+            &::routing_core::State::Disconnected => {
```
benjaminbollen: can you add small inline comments on the motivation behind these steps?  that would be very valuable here I think.  Thanks!

benjaminbollen:
looks really good thx

benjaminbollen:
src/routing_node.rs:287:5: 306:6 error: method is never used: `handle_new_bootstrap_connection` ;)

PR #743, opened 2015-10-19
OnConnectRequest.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/743)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -195,9 +195,9 @@ impl RoutingNode {
                 }
                 Ok(::crust::Event::BootstrapFinished) => {
                 }
-                Ok(::crust::Event::ExternalEndpoints(ext_endpoints)) => {
-                    for ext_ep in ext_endpoints {
-                        self.accepting_on.push(ext_ep);
+                Ok(::crust::Event::ExternalEndpoints(external_endpoints)) => {
+                    for external_endpoint in external_endpoints {
+                        self.accepting_on.push(external_endpoint);
                     }
                 }
             };
@@ -623,18 +623,17 @@ impl RoutingNode {
         let threshold = self.group_threshold();
         debug!("Accumulator threshold is at {:?}", threshold);
 
-        let claimant : NameType = match *signed_message.claimant() {
+        let claimant: NameType = match *signed_message.claimant() {
             Address::Node(ref claimant) => claimant.clone(),
             Address::Client(_) => {
-                error!("Claimant is a Client, but passed into accumulator for a group. dropped.");
-                debug_assert!(false);
+                error!("Claimant is a Client, but passed into accumulator for a group, dropping.");
+                // debug_assert!(false);
                 return None;
             }
         };
 
         debug!("Adding message from {:?} to accumulator", claimant);
-        self.accumulator.add_message(threshold, claimant, message)
-                        .map(|msg| (msg, None))
+        self.accumulator.add_message(threshold, claimant, message).map(|msg| (msg, None))
     }
 
     // ---- Direct Messages -----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/routing_node.rs
```diff
@@ -623,18 +623,17 @@ impl RoutingNode {
...
         debug!("Accumulator threshold is at {:?}", threshold);
 
-        let claimant : NameType = match *signed_message.claimant() {
+        let claimant: NameType = match *signed_message.claimant() {
             Address::Node(ref claimant) => claimant.clone(),
             Address::Client(_) => {
-                error!("Claimant is a Client, but passed into accumulator for a group. dropped.");
-                debug_assert!(false);
+                error!("Claimant is a Client, but passed into accumulator for a group, dropping.");
+                // debug_assert!(false);
```
benjaminbollen: why uncomment the debug_assert!(false) ? is it then not better to remove it? or do you want it restored?

src/routing_node.rs
```diff
@@ -864,55 +863,45 @@ impl RoutingNode {
...
+                    // (TODO Brian 19.10.15) Shouldn't happen since we should have at least one
+                    // bootstrap connection, but should be acted on explicitly if it we get here.
+                    None => return Err(RoutingError::Interface(InterfaceError::NotConnected))
+                };
+                let signing_key = self.core.id().signing_public_key();
+                (Authority::Client(name, signing_key), Address::Client(signing_key))
+            },
+            &::routing_core::State::Terminated => {
+                // (TODO Brian 19.10.15) A new error code may be more appropriate here.
+                return Err(RoutingError::Interface(InterfaceError::NotConnected))
```
benjaminbollen: I'd prefer to stick to the RoutingError::NotBootstrapped for now (over using InterfaceError)

src/routing_node.rs
```diff
@@ -864,55 +863,45 @@ impl RoutingNode {
...
-                    match SignedMessage::new(Address::Client(self.core.id().signing_public_key()),
-                                             routing_message,
-                                             self.core.id().signing_private_key()) {
-                        Ok(signed_message) => ignore(self.send(signed_message)),
-                        Err(e) => return Err(RoutingError::Cbor(e)),
-                    };
-                    Ok(())
-                }
-                None => return Err(RoutingError::NotBootstrapped),
+        let (from_authority, address) = match self.core.state() {
```
benjaminbollen: perhaps more clear to call this `address` `claimant` it had me confused

src/routing_node.rs
brian-js: It was commented to draw attention to the fact that it's there. Either it's ok to return None in debug or release mode or since there's an error logged should we not return an error?

src/routing_node.rs
```diff
@@ -921,33 +910,37 @@ impl RoutingNode {
...
                 };
-                // first verify that the message is correctly self-signed
-                if !response_token.verify_signature(&connect_request.requester_fob
-                    .signing_public_key()) {
+                // First verify that the message is correctly self-signed.
+                if !response_token.verify_signature(
+                        &connect_request.requester_fob.signing_public_key()) {
                     return Err(RoutingError::FailedSignature);
                 };
                 if !self.core.check_node(&ConnectionName::Routing(
```
benjaminbollen: can you add a TODO that the address relocation cache needs to be checked at this point?

src/routing_node.rs
brian-js: Either way the error is wrong. Should we not create a new error for the terminated state?

benjaminbollen:
looks good to me

PR #747, opened 2015-10-21
Maid 1437.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/747)

### DIFF
--- a/src/routing_core.rs
+++ b/src/routing_core.rs
@@ -776,15 +776,24 @@ impl RoutingCore {
         match unconnected_key {
             Some(key) => {
                 let _ = self.expected_connections.insert(key.clone(), Some(connection.clone()));
+                // self.action_sender.send(::action::Action::MatchConnection(
+                //     Some(key.clone(), Some(connection.clone())), None));
                 true
             },
             None => false,
         }
     }
 
-    /// Check whether the connection has been sent in a ConnectRequest/ConnectResponse.
-    pub fn match_unknown_connection(&self, _hello: ::direct_messages::Hello) -> bool {
-        unimplemented!();
+    /// Check whether the connection has been made.
+    pub fn match_unknown_connection(&mut self, connection: &::crust::Connection,
+            hello: &::direct_messages::Hello) -> bool {
+        if self.unknown_connections.contains_key(connection) {
+            let _ = self.unknown_connections.insert(connection.clone(), Some(hello.clone()));
+            true
+        }
+        else {
+            false
+        }
     }
 
     /// Add a bootstrap connection.
--- a/src/routing_core.rs
+++ b/src/routing_core.rs
@@ -84,13 +84,13 @@ pub enum State {
 }
 
 /// ExpectedConnection.
-#[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Clone)]
+#[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Clone, RustcEncodable, RustcDecodable)]
 #[allow(unused)]
 pub enum ExpectedConnection {
     /// ConnectRequest sent by peer.
     Request(::messages::ConnectRequest),
     /// ConnectResponse in response to a ConnectRequest sent by peer.
-    Response(::messages::ConnectResponse),
+    Response(::messages::ConnectResponse, ::messages::SignedToken),
 }
 
 /// RoutingCore provides the fundamental routing of messages, exposing both the routing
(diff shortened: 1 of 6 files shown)


maidsafe-highfive:
r? @benjaminbollen

src/routing_node.rs
```diff
@@ -690,36 +699,36 @@ impl RoutingNode {
...
-        };
-        match SignedMessage::new(Address::Client(core_id.signing_public_key()),
-                                 routing_message,
-                                 core_id.signing_private_key()) {
-            Ok(signed_message) => ignore(self.send(signed_message)),
-            Err(e) => return Err(RoutingError::Cbor(e)),
-        };
-        Ok(())
-    }
+    // fn request_network_name(&mut self,
```
benjaminbollen: request_network_name should be `used`

src/routing_core.rs
```diff
@@ -743,9 +743,313 @@ impl RoutingCore {
...
+                            None => {},
+                        }
+                    }
+                }
+            },
+            (None, Some((unknown_connection, Some(hello)))) => {
+                // At matching from unknown_connection against expected connection
+                let mut opt_connection = None;
+                match hello.expected_connection {
+                    Some(ref hello_expected_connection) => {
```
benjaminbollen: ref and * can possibly be collapsed into single destructuring

src/routing_core.rs
```diff
@@ -743,9 +743,313 @@ impl RoutingCore {
...
+                        match *hello_expected_connection {
+                            ExpectedConnection::Request(ref request) => {
+                                for (key, value) in self.expected_connections.iter() {
+                                    match key {
+                                        &ExpectedConnection::Request(_) => {
+                                            // Don't match on ConnectRequest.
```
benjaminbollen: I do believe we need to act here too.  We have a Hello and found a ConnectResponse

src/routing_core.rs
```diff
@@ -743,9 +743,313 @@ impl RoutingCore {
...
+                                        },
+                                    }
+                                }
+                            },
+                        }
+                    },
+                    None => {},
+                };
+
+                match opt_connection {
```
benjaminbollen: make comment: this is the primary connection, either the expected connect_request connection for node B, or the unknown connection for node A - diagram RFC-0011

src/routing_core.rs
```diff
@@ -743,9 +743,313 @@ impl RoutingCore {
...
+                    Some(connection) => {
```
benjaminbollen: but here we need to determine whether we are a client bootstrapping.

Alternatively, and as suggested by Brian, we need to earlier split the program flow and handle as a client before already

src/routing_core.rs
```diff
@@ -743,9 +743,313 @@ impl RoutingCore {
...
+                        if self.add_peer(ConnectionName::Routing(
+                                hello.public_id.name()), connection, hello.public_id.clone()) {
+                            if connection != unknown_connection {
+                                self.action_sender.send(::action::Action::SendConfirmationHello(
+                                    connection, ::types::Address::Node(hello.public_id.name())));
+                                let _ = self.action_sender.send(::action::Action::DropConnections(
+                                    vec![unknown_connection]));
+                            } else {
```
benjaminbollen: the "confirmation of the hello" is not explicitly checked; which raises some questions, as if we already have a hello stored in unknown connections, then we drop any subsequent action on a hello.  So we need to check within handle_hello whether it is a consistent confirmation and do match.

src/routing_core.rs
```diff
@@ -743,9 +743,313 @@ impl RoutingCore {
...
+
+    /// Check whether the connection has been accepted.
+    pub fn match_unknown_connection(&mut self, connection: &::crust::Connection,
+            hello: &::direct_messages::Hello) {
+        for (key, value) in self.unknown_connections.iter_mut() {
+            if key == connection {
+                match value.0 {
+                    None => {
+                        value.0 = Some(hello.clone());
+                        let _ = self.action_sender.send(::action::Action::MatchConnection(
```
benjaminbollen: Here we need to check whether it is a confirmation hello, consistent with the hello stored and continue to Match is so.

src/routing_node.rs
```diff
@@ -293,142 +295,150 @@ impl RoutingNode {
...
-                // to avoid both sides drop the other connection, possibly leaving none
-                self.crust_service.drop_node(connection.clone());
-                debug!("Core refused {:?} on {:?} and dropped the connection",
-                    hello_address, connection);
-            };
-        } else {
-            debug!("We are not alpha and the hello was not confirmed yet, awaiting alpha.");
-        }
-        Ok(())
+        self.match_unknown_connection(&connection, &hello)
```
benjaminbollen: handle bootstrapping

benjaminbollen:
looks good to me, without bootstrapping finished

PR #755, opened 2015-10-27
Fix types vector functions

canndrew:
This PR removes/changes some of the functions in types.rs

* The `array_as_vector` function is pointless - you can already do this with `Vec::from(slice)` or `slice.to_owned()`.
* The `vector_as_u8_{32,64}_array` functions should takes slices not vectors - there's no reason for them to be taking owned data.

Note that several other repos will have to be updated to bring them in line with these changes.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/755)


maidsafe-highfive:
r? @benjaminbollen

PR #756, opened 2015-10-27
Replace Random with rand::Rand

canndrew:
The `rand` crate (which we already depend on) implements a trait and function for generating random instances of types. It's redundant for us to reimplement our own versions of these.

This PR remove the `Random` and `generate_random` and replaces their usages with `rand::Rand` and `rand::random`.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/756)


maidsafe-highfive:
r? @benjaminbollen

benjaminbollen:
very nice cleanup work!! Thanks a lot!

PR #765, opened 2015-11-06
response tokens for new crust

benjaminbollen:
Update routing to crust 0.5 cargo toml;
introduce dummy response tokens for now;
0u32 is reserved for bootstrapping; any connect request has a wrapping u32 counter

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/765)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -48,6 +48,7 @@ pub struct RoutingNode {
     crust_receiver: ::std::sync::mpsc::Receiver<::crust::Event>,
     crust_service: ::crust::Service,
     accepting_on: Vec<::crust::Endpoint>,
+    connection_counter: u32,
     // for RoutingNode
     client_restriction: bool,
     action_sender: ::std::sync::mpsc::Sender<Action>,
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @brian-js

src/routing_node.rs
```diff
@@ -208,6 +215,7 @@ impl RoutingNode {
                         self.accepting_on.push(external_endpoint);
                     }
                 }
+                _ => {}
```
brian-js: We should probably include the remaining events as unimplemented also.

src/routing_node.rs
```diff
@@ -864,7 +872,11 @@ impl RoutingNode {
     }
 
     fn connect(&mut self, endpoints: &Vec<::crust::Endpoint>) {
-        self.crust_service.connect(endpoints.clone());
+        let new_response_token = self.connection_counter.clone();
+        self.connection_counter = self.connection_counter.wrapping_add(1u32);
```
brian-js: The counter should form part of the key to the ExpirationMap.

PR #770, opened 2015-11-16
Code clean up - intra-thread channel communication removal

ustulation:
<1>
Current flow:

Now the flow will be:

<2>
Integrating Using maidsafe_utilities

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/770)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -447,7 +447,9 @@ impl RoutingNode {
         debug!("Lost connection on {:?}.\n", connection);
         let connection_name = self.lookup_connection(&connection);
         if connection_name.is_some() {
-            self.drop_peer(&connection_name.unwrap());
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @canndrew

src/routing_node.rs
```diff
@@ -1604,123 +1606,125 @@ impl RoutingNode {
...
                 self.routing_table = None;
-                match self.action_sender.send(::action::Action::Rebootstrap) {
-                    Ok(()) => {},
-                    Err(_) => {
-                        error!("Action receiver in RoutingNode disconnected. \
-                            Terminating from core.");
-                        self.state = State::Terminated;
-                    }
-                };
+                self.reset();
```
dirvine: :+1:  much much better

PR #771, opened 2015-11-16
add_peer() purged of action_sender

ustulation:
Current flow:

Now the flow will be:

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/771)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -1770,7 +1770,7 @@ impl RoutingNode {
                                                line!()),
                             }
                         },
-                        None => ()
+                        None => info!("No node removed from RT as a result of node addition"),
                     }
 
                     if add_node_result.0 {
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -1785,7 +1785,9 @@ impl RoutingNode {
                             info!("Routing Node has connected to {:?} nodes.", size);
 
                             self.state = State::GroupConnected;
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @canndrew

src/routing_node.rs
```diff
@@ -1737,61 +1737,92 @@ impl RoutingNode {
...
-                                    None => ()
-                                }
-                            },
-                            None => ()
+                if public_id.name() != routing_name {
+                    return false
+                }
+
+                // TODO(Spandan) - Suspicious code - is there a possiblility of getting add_peer()
+                // even if there is no routing table ? When and why would this happen ?
```
dirvine: The routing table should always exist IMHO, I suspect the creation of a routing table was used to promote to a node instead of being a client. So to be clear we should not do this, we should have a routing table if we are a node & only  a node can get an add_peer(). 
To add a peer we make sure the call comes only from a connect_request (if we are receiving this) or a connect_response (if we are receiving this) and we add it if it passes the `routing_table.add()`. IF we think a peer is worth connecting to we do a `routing_table.check()` & if successful then we initiate the connect_request (or connect_response).

Hope that makes sense :-)

dirvine: If you add info to the TODO then I will merger this one @ustulation

PR #774, opened 2015-11-17
Replacing Intra Thread signalling

ustulation:
Current flow:

Now the flow will be:

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/774)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -167,6 +167,8 @@ pub enum RoutingError {
     Cbor(::cbor::CborError),
     /// invalid response
     Response(ResponseError),
+    /// Current state is invalid for the operation
+    InvalidStateForOperation,
 }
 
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @brian-js

dirvine:
Reviewed 3 of 3 files at r1.

---

<sup>**[src/error.rs, line 171 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/774#-K3JdlZKuYL7w9eSYGOE)** ([raw file](https://github.com/maidsafe/routing/blob/44e445762f82bea924fc2a35bf26f080fca0d2a5/src/error.rs#L171)):</sup>
I think we will get to a stage where we do not / cannot get this, for now though it's good.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/routing/774)

PR #777, opened 2015-11-17
Cleanup

ustulation:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/777)

### DIFF
--- a/src/routing_node.rs
+++ b/src/routing_node.rs
@@ -232,13 +232,24 @@ impl RoutingNode {
     pub fn run(&mut self) {
         self.crust_service.bootstrap(0u32);
         debug!("RoutingNode started running and started bootstrap");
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @canndrew

ustulation:
@dirvine : This is mainly for printing the Routing Table size in fixed intervals so that we don't have to keep scrolling everytime we add/drop a new node in the terminal

dirvine:
As a temp measure then this is OK but I would recommend not keeping it too long. 

---

Reviewed 1 of 1 files at r1.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/routing/777#-:-K3KPTd5RjXjYh1kQVmx:-2072124729)_

PR #779, opened 2015-11-18
Use LruCaches in place of ExpirationMaps.

Fraser999:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/779)


maidsafe-highfive:
r? @canndrew

src/routing_node.rs
```diff
@@ -392,27 +393,11 @@ impl RoutingNode {
...
-                            }
-                            Err(_) => {}
-                        }
-                    }
-                    ExpectedConnection::Response(_, _, _) => {
-                        ignore(self.send_hello(connection, None));
-                    }
-                }
-            }
+            Some(_) => ignore(self.send_hello(connection, None)),
```
dirvine: tomorrow we should match and print debug statements here if we are swallowing the error I think. Maybe a wee TODO?

Fraser999: I'm just starting work on the outputs now :)

PR #783, opened 2015-11-19
Cleanups

ustulation:
<1> Replace custom filter with message_filter crate
<2> Replace custom accumulator with message_accumulator crate

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/783)


maidsafe-highfive:
r? @brian-js

PR #786, opened 2015-11-19
Fix message_received checks

Viv-Rajkumar:
In `message_received` ignore incoming message correctly when message exists in `claimant_message_filter `.

Also add our own public_id to the close_group list when sending the `public_ids` back for `InternalResponse::RelocatedNetworkName`

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/786)


maidsafe-highfive:
r? @brian-js

dirvine:
Nice catch Viv

PR #797, opened 2015-11-25
Cleanup

ustulation:
<1> Common drop_connection
<2> marking unreachable! instead of panic!

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/797)


maidsafe-highfive:
r? @canndrew

PR #801, opened 2015-11-28
maidsafe_utilities integration

Viv-Rajkumar:
Minor addition to PR #800 to update crust dependency to `0.6.1` and also add `State::Bootstrapping`

`Bootstrapping` state is set in `handle_on_connect` when current state is `Disconnected`. We then use this state as an indicator in the `bootstrap_finished` event to know if we should start listening.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/801)


maidsafe-highfive:
r? @canndrew

PR #802, opened 2015-11-29
SignedMessage changes.

brian-js:
Updates SignedMessage/Token and RoutingMessage as agreed. No attempt at optimisation has been made.

SignedMessage contains signed RoutingMessage and public signing key of sender.
RoutingMessage has additional optional vector of public signing keys field.
Acummulator accumulates RoutingMessages w.r.t public signing keys.
MessageFilter filters messages w.r.t public signing keys instead of claimant.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/802)


maidsafe-highfive:
r? @canndrew

PR #809, opened 2015-12-01
Cleanup

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/809)


maidsafe-highfive:
r? @canndrew

src/routing_table.rs
```diff
@@ -46,23 +46,23 @@ impl NodeInfo {
...
-            id: id,
+            name: name,
         }
     }
 
     #[cfg(test)]
-    pub fn id(&self) -> &NameType {
-        &self.id
+    pub fn name(&self) -> &NameType {
+        &self.name
```
dirvine: it may be better to not store the name but make it `public_id.name().clone();` if this is in fact the name? or is this only the initial name? If only for tests as it seems we should consider this as we move on.

brian-js: It's as written specifically for tests. I'll remove the `name` field so that the tests better represent the production code.

PR #825, opened 2015-12-14
Minor fixes to message handling

Viv-Rajkumar:
- Added `want_address_in_routing_table` so we dont ignore nodes already existing in the RT with the `routing_table.want_to_add()`
- Handle case where we are the proxy node for a client not in our close group

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/825)


maidsafe-highfive:
r? @brian-js

PR #853, opened 2015-12-25
Key_value_store example updated

ustulation:
<1> Initial implementation for churn handling and lost close node
<2> updated PlainData flow from Key value store example to match Immutable data flow

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/853)


maidsafe-highfive:
r? @canndrew

PR #854, opened 2015-12-26
Churn handling fixes

Viv-Rajkumar:
- Add NONCE to External Requests

This prevents repeated requests from getting filtered out, thereby allowing clients to request the same chunk twice. Also in the process of churn, NaeManagers might request a ManagedNode for a chunk which it might have previously retrieved for a client within the filter duration.

`Client` API generates a Nonce internally per request made by upper libraries.

`Node` API expects a Nonce in the API function for requests and responses since these will purely be used to respond to requests made by client from which the nonce is expected to be used to continue the message flow. The only action initiated by vaults directly for now is when chunks are relocated for `LostCloseNode`, for this a nonce is derived from the `lost_node` which should thereby yield the same nonce to the entire group.

- Emit `LostCloseNode` before `Churn`

This allows the refresh messages synchronised between persona groups of upper libraries hold accurate info.

__________

With these changes, Churn/LostCloseNode function as expected and is tested with the `key_value_store` example running 15 Nodes and killing 8. Data stored by client continue to be available even after the original nodes storing the chunks have gone offline.

TODO: Updates to make the examples more robust and simulate vault flow more accurately namely respond with failure responses when applicable across the personas. Also stop refreshing group info received in refresh messages once suitable entries are accumulated for all personas (This will prevent slow responding nodes within the accumulator expiry duration from resetting data held by the group if other nodes have updated corresponding info in the meantime)

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/854)

### DIFF
--- a/examples/churn_node.rs
+++ b/examples/churn_node.rs
@@ -38,6 +38,7 @@ extern crate maidsafe_utilities;
 extern crate routing;
 extern crate xor_name;
 extern crate rustc_serialize;
+extern crate sodiumoxide;
 extern crate lru_time_cache;
 extern crate time;
 
--- a/examples/node.rs
+++ b/examples/node.rs
@@ -37,6 +37,7 @@ extern crate log;
 extern crate maidsafe_utilities;
 extern crate routing;
 extern crate xor_name;
+extern crate sodiumoxide;
 extern crate rustc_serialize;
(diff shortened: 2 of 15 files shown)


maidsafe-highfive:
r? @canndrew

PR #856, opened 2015-12-29
Churn handling updates

Viv-Rajkumar:
Addition to PR #855 

Removed refresh_accumulator and corresponding event for refresh. Refresh is treated as a standard routing message

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/856)


maidsafe-highfive:
r? @brian-js

PR #859, opened 2016-01-02
Use a dereference on the right side of = and also combine 2 if statem…

dirvine:
…ents into one.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/859)


maidsafe-highfive:
r? @brian-js

PR #861, opened 2016-01-04
WIP - Examples logging

Viv-Rajkumar:
Addition to PR #860 

- Fixed logging to file from `ci_test` example

Needs to be confirmed as fine from Windows/Linux too


maidsafe-highfive:
r? @brian-js

PR #862, opened 2016-01-05
Relaxed churn threshold

Viv-Rajkumar:
ci_test example is functional with default parameters.

Added ci_test run to ci builds

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/862)


maidsafe-highfive:
r? @brian-js

PR #867, opened 2016-01-06
Only send messages on one connection

Viv-Rajkumar:
This is even if we have multiple connections to a node

also updated the ci scripts to pipe output to file and only print output if successful. This helps ci machines time out the run if test takes a long time.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/867)


maidsafe-highfive:
r? @canndrew

PR #870, opened 2016-01-07
Churn fixes

Viv-Rajkumar:
MessageId generated for churn event raised by routing is unique between a node added / lost.

This prevents churn getting ignored when a Node is Added and soon after lost. In this case previously both `MessageId`'s were the same thereby getting filtered by `signed_msg_filter`

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/870)


maidsafe-highfive:
r? @canndrew

PR #872, opened 2016-01-07
Expand documentation for the main structures.

afck:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/872)

### DIFF
--- a/src/authority.rs
+++ b/src/authority.rs
@@ -21,7 +21,8 @@ use std::fmt::{Debug, Formatter};
 
 /// An entity that can act as a source or destination of a message.
 ///
-/// An `Authority` can be an individual client or node, or a group of nodes.
+/// An `Authority` can be an individual `Client` or `ManagedNode`, or a group of nodes, like a
+/// `NodeManager`, `ClientManager` or `NaeManager`.
 #[derive(RustcEncodable, RustcDecodable, PartialEq, PartialOrd, Eq, Ord, Clone, Hash)]
 pub enum Authority {
     /// Manager of a Client.  XorName is the hash of the Client's `client_key`.
--- a/src/authority.rs
+++ b/src/authority.rs
@@ -19,10 +19,12 @@ use xor_name::XorName;
 use sodiumoxide::crypto::{hash, sign};
 use std::fmt::{Debug, Formatter};
 
-/// Persona types recognised by network.
+/// An entity that can act as a source or destination of a message.
+///
+/// An `Authority` can be an individual client or node, or a group of nodes.
 #[derive(RustcEncodable, RustcDecodable, PartialEq, PartialOrd, Eq, Ord, Clone, Hash)]
 pub enum Authority {
-    /// Manager of Client.  XorName is the hash of the Client's `client_key`.
+    /// Manager of a Client.  XorName is the hash of the Client's `client_key`.
     ClientManager(XorName),
     /// Manager of a network-addressable element.  XorName is the name of the element in question.
     NaeManager(XorName),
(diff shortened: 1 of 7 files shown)


maidsafe-highfive:
Thanks for the pull request, and welcome! The MaidSafe team is excited to review your changes, and you should hear from @brian-js (or someone else) soon.

src/authority.rs
```diff
@@ -19,10 +19,12 @@ use xor_name::XorName;
 use sodiumoxide::crypto::{hash, sign};
 use std::fmt::{Debug, Formatter};
 
-/// Persona types recognised by network.
+/// An entity that can act as a source or destination of a message.
+///
+/// An `Authority` can be an individual client or node, or a group of nodes.
```
dirvine: This maybe should be more explicit. So node, NodeManager, NaeManager, ClientManager

src/client.rs
```diff
@@ -40,10 +43,12 @@ pub struct Client {
...
 
 impl Client {
-    /// Starts a new RoutingIdentity, which will also start a new Core.
-    /// The Core will only bootstrap to the network and not attempt to
-    /// achieve full routing node status.
-    /// If the client is started with a relocated id (ie the name has been reassigned),
+    /// Create a new `Client`.
+    ///
+    /// It will automatically connect to the network, but not attempt to achieve full routing node
+    /// status. The name of the client will be the name of the `PublicId` of the `keys`.
```
dirvine: and cryptographically secure, where groups use group consensus

src/immutable_data.rs
```diff
@@ -22,18 +22,20 @@ use xor_name::XorName;
 use sodiumoxide::crypto;
 
 #[derive(Hash, Clone, Eq, PartialEq, Ord, PartialOrd, RustcEncodable, RustcDecodable, Debug)]
-/// ImmutableDataType.
+/// The type of an individual copy of immutable data.
```
dirvine: Self validating data - i.e. name = Hash(content) or for backup name = Hash(Hash(content) etc.

src/plain_data.rs
```diff
@@ -18,7 +18,7 @@
 use rustc_serialize::{Decoder, Encodable, Encoder};
 use xor_name::XorName;
 
-/// PlainData
+/// Plain data with a name and a value.
```
dirvine: Name is set and not derived in this case. Allowing for contents to be set by user i.e. may contain and be restricted to a DSL

src/structured_data.rs
```diff
@@ -62,8 +65,8 @@ impl StructuredData {
         Ok(structured_data)
     }
 
-    /// This is a static function required for computing a name give the tag-type and identifier to
```
dirvine: The calculation is set by routing,

src/immutable_data.rs
afck: Done.
(Added it to the documentation of the enum variants and of `ImmutableData`.)

src/structured_data.rs
afck: I'm not entirely sure what you mean. I expanded the text to make it more precise; is the new version correct?

dirvine: :+1:

src/plain_data.rs
afck: Do you mean a Domain-Specific Language?
I added that any restrictions are left to the implementation - or should I explicitly mention DSLs?

dirvine: Yes I did, but it is fine Andreas, much better shape now. Thx for these PR's they have helped a lot.

PR #878, opened 2016-01-12
Join limit

afck:
Limit the number of prospective nodes in the bootstrapping process to one per proxy node.

We also disabled the direction check for now, since we found that it makes the ci_test less reliable. This is to be expected since the notion of a "close node" is currently ill-defined, but we should consider re-enabling it once that issue is fixed.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/878)

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -132,10 +132,8 @@ pub struct Core {
     routing_table: RoutingTable<::id::PublicId, ::crust::Connection>,
     // our bootstrap connections
     proxy_map: ::std::collections::HashMap<::crust::Connection, PublicId>,
-    // any clients we have proxying through us
-    client_map: ::std::collections::HashMap<sign::PublicKey, ::crust::Connection>,
-    // all prospective nodes trying to join the network through us
-    joining_nodes_map: ::std::collections::HashMap<sign::PublicKey, ::crust::Connection>,
+    // any clients we have proxying through us, and whether they have `client_restriction`
+    client_map: ::std::collections::HashMap<sign::PublicKey, (::crust::Connection, bool)>,
     data_cache: LruCache<XorName, Data>,
 }
 
@@ -197,7 +195,6 @@ impl Core {
                 routing_table: RoutingTable::new(&our_name),
                 proxy_map: ::std::collections::HashMap::new(),
                 client_map: ::std::collections::HashMap::new(),
-                joining_nodes_map: ::std::collections::HashMap::new(),
                 data_cache: LruCache::with_expiry_duration(::time::Duration::minutes(10)),
             };
 
@@ -357,9 +354,8 @@ impl Core {
             if let Some(&NodeInfo { ref public_id, ..}) = self.routing_table.get(hop_msg.name()) {
                 try!(hop_msg.verify(public_id.signing_public_key()));
             } else if let Some((ref pub_key, _)) = self.client_map
-                                                .iter()
-                                                .chain(self.joining_nodes_map.iter())
-                                                .find(|ref elt| &connection == elt.1) {
+                                                       .iter()
+                                                       .find(|ref elt| connection == (elt.1).0) {
                 try!(hop_msg.verify(pub_key));
             } else {
                 // TODO drop connection ?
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @brian-js

afck:
Also, retrying currently doesn't work because crust doesn't seem to be able to establish connections if several clients try to contact the beacon at once. Brian is currently trying to isolate that issue.

src/core.rs
```diff
@@ -1046,6 +1100,16 @@ impl Core {
...
     }
 
+    fn retry_bootstrap(&mut self) {
+        self.crust_service.stop_bootstrap();
+        self.state = State::Disconnected;
+        for &connection in self.proxy_map.keys() {
+            self.crust_service.drop_node(connection);
+        }
+        self.proxy_map.clear();
+        self.crust_service.bootstrap(0u32, Some(CRUST_DEFAULT_BEACON_PORT));
```
Fraser999: Shouldn't this be `self.crust_service.bootstrap_with_blacklist` so we don't try and immediately bootstrap off the same peer?

Also, I'm not sure if we maybe need to wait for a small time before retrying?  Possibly we need to see how hard on the single beacon listener immediate rebootstrapping is.  If it can handle it, we're fine.  I'm just imagining 20 peers all repeatedly clamouring to bootstrap flooding the beacon listener.  If it *is* a problem, I think Crust should handle it itself (i.e. Crust should add a delay between bootstrap attempts if required), but until then we can get the example working by adding a sleep here temporarily.  Just a thought :)

afck: I agree; I'll change that, run `ci_test` and add it to the PR.
(It won't change anything for `ci_test`, though, as the nodes currently can only try to connect to the beacon, I think.)

@brian-js Are you fine with that?

brian-js: Yes.

src/core.rs
```diff
@@ -1022,6 +1071,14 @@ impl Core {
                             return Ok(());
                         }
 
+                        if self.routing_table.len() >= ::kademlia_routing_table::group_size()
+                                && !self.proxy_map.is_empty() {
+                            trace!("Routing table reached quorum size. Dropping proxy.");
```
dirvine: msg should be group_size here I think not quorum size

afck:
Right, thanks! I changed it.

dirvine:
Was there a reason not to use a tuple in client map for (client, joining_node) rather than have two containers we now have to synchronise (or us let_ in places). It may make sense to combine these into a single container?

afck:
You mean with a boolean `joining_node` (or better `client_restriction`, as we already use elsewhere)? I don't know whether that would make it easier.

Currently `client_map` and `joining_nodes_map` are disjoint: The joining nodes contains those nodes that don't want to remain in `Client` state. However, we still act as a proxy for them so in some places we have to look through both maps.

Alternatively, we could continue adding everything to `client_map` and just have a `HashSet` of the names of those clients that intend to become a node.

Shall I try that? Or the `(client, client_restriction)` tuple? Not sure which one will be nicer.

dirvine:
No I think you are correct they seem disjoint, I was prompted by the code at line 1669 in this PR. I am happy to merge though :-)

afck:
But you're right: We have to mix the two maps in four places. Probably there would be fewer places where we'd have to consider the `joining_nodes` `HashSet`.

Let me try out whether that's nicer. I think it might simplify the code.

dirvine:
Ok no problem, thx

afck:
I used a tuple so we don't have to remember keeping two fields in sync, as you suggested.

dirvine:
:+1

src/core.rs
```diff
@@ -1581,12 +1656,13 @@ impl Core {
...
-                             .find(|&(_, client)| client == connection)
-                             .map(|entry| entry.0.clone());
-        if let Some(public_key) = public_key {
-            let _ = self.client_map.remove(&public_key);
+        if let Some(public_key) = self.client_map
+                                      .iter()
+                                      .find(|entry| (entry.1).0 == *connection)
+                                      .map(|entry| entry.0.clone()) {
+            if let Some((_, false)) = self.client_map.remove(&public_key) {
+                trace!("Joining node dropped. {} remaining.", self.joining_nodes_num());
```
dirvine: Yes it is nicer. Nice code :-)

PR #880, opened 2016-01-13
Fix computation of remaining required signatures for StructuredData.

afck:
Deal with the underflow for required signatures, and change the
requirement to a strict majority.

I added a test for the strict majority (with four owners) and shortened
the tests somewhat. Also, using `unwrap()` instead of `ok()` would print
more information in case of failure.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/880)

### DIFF
--- a/src/structured_data.rs
+++ b/src/structured_data.rs
@@ -109,7 +109,7 @@ impl StructuredData {
     /// Verifies that `other` is a valid update for `self`; returns an error otherwise.
     ///
     /// An update is valid if it doesn't change type tag or identifier (these are immutable),
-    /// increases the version by 1 and is signed by (at least 50% of) the owners.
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @brian-js

src/structured_data.rs
```diff
@@ -196,7 +196,7 @@ impl StructuredData {
         } else {
             &self.previous_owner_keys
         };
-        Ok(((owner_keys.len() + 1) / 2) - self.previous_owner_signatures.len())
+        Ok(((owner_keys.len() / 2) + 1).saturating_sub(self.previous_owner_signatures.len()))
```
dirvine: nice

PR #881, opened 2016-01-13
Replace explicit ASCII codes.

afck:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/881)


maidsafe-highfive:
r? @brian-js

PR #882, opened 2016-01-14
Some improvements to the message.rs documentation.

afck:
I also made a small change to `HopMessage`'s API to make it consistent with `SignedMessage`.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/882)


maidsafe-highfive:
r? @canndrew

PR #886, opened 2016-01-15
Reject clients if the routing table is too small.

afck:
A client should never set a small dynamic quorum size, as that is
currently never updated. Instead of introducing a new message type to
reset the client's quorum size, let's just not accept any clients until
there are at least 8 nodes in our routing table.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/886)

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -950,7 +950,12 @@ impl Core {
                 Ok(())
             }
             DirectMessage::BootstrapDeny => {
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @canndrew

dirvine:
Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/routing/886)*

PR #890, opened 2016-01-20
Maid 1547

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/890)


maidsafe-highfive:
r? @canndrew

afck:
Looks good to me, but I think it should also test that `signed_message.check_integrity().is_err()` if the signature is wrong. You could change the `signature` field at the end of `signed_message_check_integrity` and assert that.

brian-js:
^^^ Done.

afck:
That's weird; do you know why the test suddenly failed in the doc examples on Travis? Doesn't seem to have anything to do with this change ... but maybe we should call that variable `_receiver` then?

brian-js:
Looks like it arbitrarily decides to enforce checks. receiver and node both look like the culprits.

PR #891, opened 2016-01-20
Add test for node connection (MAID-1534)

madadam:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/891)

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -1,0 +1,117 @@
+// Copyright 2015 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under (1) the MaidSafe.net Commercial License,
+// version 1.0 or later, or (2) The General Public License (GPL), version 3, depending on which
+// licence you accepted on initial access to the Software (the "Licences").
+//
+// By contributing code to the SAFE Network Software, or to this project generally, you agree to be
+// bound by the terms of the MaidSafe Contributor Agreement, version 1.0.  This, along with the
+// Licenses can be found in the root directory of this project at LICENSE, COPYING and CONTRIBUTOR.
+//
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
Thanks for the pull request, and welcome! The MaidSafe team is excited to review your changes, and you should hear from @canndrew (or someone else) soon.

src/tests.rs
```diff
@@ -0,0 +1,117 @@
...
+fn connect() {
+  let (event_sender, event_receiver) = mpsc::channel();
+
+  let nodes_count = 4;
+  let timeout = Duration::from_secs(10);
+
+  let mut nodes = Vec::with_capacity(nodes_count);
+  let mut connection_counts = Vec::with_capacity(nodes_count);
+
+  for i in 0..4 {
```
afck: Should probably be `0..nodes_count`.

afck:
The indentation is off in some lines: It should always be 4 spaces.

src/tests.rs
```diff
@@ -0,0 +1,117 @@
...
+use std::time::Duration;
+use maidsafe_utilities::thread::RaiiThreadJoiner;
+
+use event::Event;
+use node::Node;
+
+struct TestNode {
+  // id: usize,
+  _node: Node,
+  _thread_joiner: RaiiThreadJoiner,
```
afck: I don't think these need leading underscores. They are being used, so they shouldn't cause any "unused" errors.

madadam: The linter is complaining about unused struct fields. I guess because they are only assigned, but never referenced again.

afck: Ah, right - and they are needed so that the node isn't dropped and the thread isn't joined.

madadam: That's right.

Issue #721, opened 2015-09-28
Clean up debug option


rossmuir:
With `RUST_LOG=debug` ON you see very these long number lists (example below), these make finding patterns difficult - is it possible to suppress these? 

```
DEBUG:routing::routing_node: Sending InternalRequest(Refresh(3, [162, 100, 110, 97, 109, 101, 152,
64, 24, 186, 24, 95, 24, 234, 24, 218, 24, 102, 24, 122, 24, 193, 9, 24, 162, 24, 132, 9, 24, 45, 24,
186, 24, 204, 24, 220, 24, 253, 24, 93, 24, 197, 24, 31, 24, 96, 24, 151, 24, 128, 24, 62, 8, 24, 115,
24, 249, 24, 54, 24, 49, 24, 189, 24, 75, 24, 199, 24, 253, 24, 73, 24, 186, 24, 199, 24, 106, 24, 184,
24, 55, 24, 161, 24, 208, 12, 24, 211, 24, 237, 24, 216, 24, 157, 24, 104, 24, 171, 24, 131, 24, 158,
24, 164, 24, 127, 24, 211, 24, 247, 24, 56, 24, 100, 24, 140, 24, 139, 24, 108, 24, 50, 24, 36, 24, 136,
24, 240, 24, 47, 24, 36, 110, 114, 101, 115, 111, 117, 114, 99, 101, 95, 105, 110, 100, 101, 120, 0],
c87df5..5c0755)) to 8 target connection(s)
```

canndrew:
Instead of using `Vec<u8>` everywhere for serialised data we could wrap it in something like:

    struct Bytes(pub Vec<u8>);

And then `impl Debug for Bytes` so that it prints `Bytes [0xa2 0x64 ... 0x78 0x00]` or something of the sort. Alternatively we could wrap the `Vec` like this:

    struct Serialised<T> {
        data: Vec<u8>,
        _pd: std::marker::PhantomData<T>,
    }

This has the advantage of being slightly more type-safe (you can't confuse a `Serialised<A>` for a `Serialised<B>`). It would also let us define an `impl<T: Debug> Debug for Serialised<T>` that deserialises the data and prints the result (although that might be overkill).

Fraser999:
I agree here too, and have a slight preference for the latter option, which is really a fleshed-out version of what I was thinking of in [my other comment](https://github.com/maidsafe/routing/issues/723#issuecomment-144752640) about `NameType`.

[...]

PR #892, opened 2016-01-20
Maid 1546

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/892)

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -483,7 +483,7 @@ mod test {
 
         assert!(signed_message_result.is_ok());
 
-        let signed_message = signed_message_result.unwrap();
+        let signed_message = unwrap_result!(signed_message_result);
 
         assert_eq!(routing_message, *signed_message.content());
         assert_eq!(full_id.public_id(), signed_message.public_id());
@@ -515,7 +515,7 @@ mod test {
 
         assert!(hop_message_result.is_ok());
 
-        let hop_message = hop_message_result.unwrap();
+        let hop_message = unwrap_result!(hop_message_result);
 
         assert_eq!(signed_message, *hop_message.content());
         assert_eq!(hop_name, *hop_message.name());
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -484,7 +484,7 @@ mod test {
 
         assert!(signed_message_result.is_ok());
 
-        let signed_message = unwrap_result!(signed_message_result);
+        let mut signed_message = unwrap_result!(signed_message_result);
 
         assert_eq!(routing_message, *signed_message.content());
         assert_eq!(full_id.public_id(), signed_message.public_id());
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -15,6 +15,7 @@
 // Please review the Licences for the specific language governing permissions and limitations
 // relating to use of the SAFE Network Software.
 
+use std::fmt;
 use data::{Data, DataRequest};
 use id::{FullId, PublicId};
 use types::MessageId;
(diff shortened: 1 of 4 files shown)


maidsafe-highfive:
r? @canndrew

src/messages.rs
```diff
@@ -40,7 +41,7 @@ pub enum Message {
 /// Messages sent via a direct connection.
 ///
 /// Allows routing to directly send specific messages between nodes.
-#[derive(Debug, RustcEncodable, RustcDecodable)]
+#[derive(RustcEncodable, RustcDecodable)]
```
afck: Would it make sense to create some wrapper type `BinaryData(Vec<u8>)` instead, manually implement `Debug` for that one, and then keep the `#[derive(Debug)]` on all the message types?

brian-js: Maybe create a task for that if it's desired.

Fraser999: This was previously discussed in https://github.com/maidsafe/routing/issues/721 which was closed without a decent resolution in my opinion (I would have preferred something similar to what you're suggesting).

src/messages.rs
```diff
@@ -363,3 +364,169 @@ pub enum ResponseContent {
...
+                dst: Authority::ClientManager(name),
+                content: RequestContent::Connect,
+            }
+        );
+        let full_id = FullId::new();
+        let signed_message_result = SignedMessage::new(routing_message.clone(), &full_id);
+
+        assert!(signed_message_result.is_ok());
+
+        let signed_message = signed_message_result.unwrap();
```
Fraser999: Can you replace the bare `unwrap` with our `unwrap_result` macro please?  There are a few instances in these tests.

src/messages.rs
```diff
@@ -363,3 +364,169 @@ pub enum ResponseContent {
...
+    extern crate rand;
+
+    use super::{HopMessage, SignedMessage, RoutingMessage, RequestMessage, RequestContent};
+    use id::FullId;
+    use authority::Authority;
+    use xor_name::XorName;
+    use sodiumoxide::crypto::sign;
+
+    #[test]
+    fn signed_message_check_integrity() {
```
Fraser999: Looks like we're not really checking anything but the success cases here.  Could we add checks for failure of integrity if any of the three fields of the signed message is tampered with or corrupted please?

brian-js: The tests are based on the task description, "The non-trivial functions of these structs need unit tests." Andreas asked for the error case for the signature when corrupted in the PR for MAID-1547, that's been done. Do we want all fields to be altered in various combinations then checked in each test?

src/messages.rs
```diff
@@ -364,6 +365,98 @@ pub enum ResponseContent {
...
+        match *self {
+            DirectMessage::BootstrapIdentify { ref public_id, ref current_quorum_size } =>
+                write!(formatter, "DirectMessage {{ BootstrapIdentify {{ {:?}, {:?} }} }}",
+                    public_id, current_quorum_size),
+            DirectMessage::BootstrapDeny =>
+                write!(formatter, "DirectMessage {{ BootstrapDeny }}"),
+            DirectMessage::ClientIdentify { ref client_restriction, .. } =>
+                write!(formatter, "DirectMessage {{ ClientIdentify {{ .., {:?} }} }}", client_restriction),
+            DirectMessage::NodeIdentify { .. } =>
+                write!(formatter, "DirectMessage {{ NodeIdentify {{ .. }} }}"),
```
afck: Maybe the `serialised_public_id` should be included for `ClientIdentify` and `NodeIdentify`, using `format_binary_array`.

afck:
In `messages.rs`, a lot of content is currently completely omitted from the debug strings (`..`). Maybe it should in general be included, using `format_binary_array`.

brian-js:
I agree there's a range of possibilities here, although I mainly went with task description at https://maidsafe.atlassian.net/browse/MAID-1546, replacing <binary data> with .. in order to reduce the useful debug output to a minimal, and using format_binary_array for unserialised, unencrypted data, with the exception of nonce arguments which may or may not prove useful during debugging. If we need this changed further could you push the task back and put updated specs in a task comment please, and I'll close this PR for now.

afck:
OK, makes sense.

PR #897, opened 2016-01-23
Maid 1536

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/897)

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -17,22 +17,22 @@
 
 #![cfg(test)]
 
-use maidsafe_utilities::serialisation::serialise;
-use maidsafe_utilities::thread::RaiiThreadJoiner;
-use rand::{thread_rng, Rng};
 use sodiumoxide::crypto;
 use sodiumoxide::crypto::hash::sha512;
-use std::cmp::Ordering::{Less, Greater};
-// use std::iter;
+use std::iter;
 use std::sync::mpsc::{self, Sender, Receiver, TryRecvError};
 use std::thread;
 use std::time::Duration;
+use std::cmp::Ordering::{Less, Greater};
 use time;
+use rand;
 use itertools::Itertools;
+
 use xor_name;
 use xor_name::XorName;
 use kademlia_routing_table;
-
+use maidsafe_utilities::serialisation::serialise;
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @canndrew

src/tests.rs
```diff
@@ -17,26 +17,32 @@
...
 use data::Data;
 use event::Event;
 use id::FullId;
-use messages::{RequestContent, ResponseContent, ResponseMessage};
+use messages::{RequestContent, ResponseContent, RequestMessage, ResponseMessage};
 use node::Node;
 use plain_data::PlainData;
 
-const GROUP_SIZE: usize = 8;
+const GROUP_SIZE: usize = kademlia_routing_table::GROUP_SIZE as usize;
```
dirvine: `kademlia_routing_table::group_size()` is maybe what you are after here ?

brian-js: I could use that wherever it's required, but defining GROUP_SIZE once using the existing constant seems desirable. Rust won't allow us to do that with the function, yet.

afck: Why do we have those `u8` constants anyway? We seem to be using `usize` almost everywhere, so wouldn't it make more sense to make the constants `usize` and convert them to `u8` where needed?

dirvine: Possibly. On the lazy static Brian there is a crate for that I think.

brian-js: @afck I don't think we want to do those conversions, or in fact use them as `u8`'s, since AFAICT we don't currently. So 'usize`'s look fitting. @dirivine  It appears to be in the pipeline for the language to handle. More problematic is the tests time out in Appveyor.

dirvine: This is the crate Brian https://github.com/rust-lang-nursery/lazy-static.rs if it helps

afck: But what's the advantage of using lazy_static, compared to just leaving it as it is for now?
`const GROUP_SIZE: usize = kademlia_routing_table::GROUP_SIZE as usize;`
And later we'll change `GROUP_SIZE` to be `usize` in kademlia_routing_table anyway.

PR #898, opened 2016-01-25
Ci test binary

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/898)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,9 @@ test = true
 bench = false
 name = "ci_test"
 test = true
+
+[[test]]
+bench = false
+harness = false
+name = "ci"
+test = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,4 +45,3 @@ test = true
 bench = false
 harness = false
 name = "ci"
-test = true
--- a/tests/ci.rs
+++ b/tests/ci.rs
@@ -206,7 +206,6 @@ fn core() {
 
     {
         // request and response
-        println!("request and response");
         let client = TestClient::new(nodes.len(), event_sender.clone());
         let data = gen_plain_data();
 
@@ -246,7 +245,6 @@ fn core() {
 
     {
         // request to group authority
-        println!("request to group authority");
         let node_names = nodes.iter().map(|node| node.name()).collect();
         let client = TestClient::new(nodes.len(), event_sender.clone());
         let data = gen_plain_data();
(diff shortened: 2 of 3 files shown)


maidsafe-highfive:
r? @canndrew

tests/ci.rs
```diff
@@ -24,23 +32,19 @@ use std::sync::mpsc::{self, Sender, Receiver, TryRecvError};
...
-use node::Node;
-use plain_data::PlainData;
+use routing::Authority;
+use routing::Client;
+use routing::Data;
+use routing::Event;
+use routing::FullId;
+use routing::{RequestContent, ResponseContent, RequestMessage, ResponseMessage};
+use routing::Node;
+use routing::PlainData;
```
Fraser999: These `routing` statements could all be a single one.

Cargo.toml
```diff
@@ -40,3 +40,9 @@ test = true
 bench = false
 name = "ci_test"
 test = true
+
+[[test]]
+bench = false
+harness = false
+name = "ci"
+test = true
```
Fraser999: `test` is true by default.

Fraser999:
Please run rustfmt on this file.

Cargo.toml
brian-js: It's explicitly true now though.

Fraser999: I see my mistake - I wasn't explicit enough :)  So... could you remove the unnecessary `test = true` lines please?

madadam:
There are two errors in this PR, which were actually introduced in #897, but they sort of cancel each other out so I only noticed them now:

First one is here: https://github.com/brian-js/routing/blob/ci_test_binary/tests/ci.rs#L112, originally, the line was `let interval = Duration::from_millis(100)`. This change makes the `recv_with_timeout` function less useful, because it now always waits the full `timeout` duration (unless there already was a message in the queue at the time of calling it). Before the change, the function only waited until a message was received. So this change makes the tests take longer to finish. 

The second one is here: https://github.com/brian-js/routing/blob/ci_test_binary/tests/ci.rs#L263 and here: https://github.com/brian-js/routing/blob/ci_test_binary/tests/ci.rs#L312 - the comparison operator is reversed. It should be `start + timeout < time::SteadyTime::now()` - so the loop breaks when it takes more than the given timeout, not less :)

My suggestion is to revert the `recv_with_timeout` function to how it was before and then you can remove the timeout checks in request/response to group authority tests completely, because `recv_with_timeout` takes care of it automatically.

brian-js:
Having the comparison operator inverted certainly doesn't help. Cheers @madadam.

PR #900, opened 2016-01-26
Tests for less/more than quorum responses.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/900)

### DIFF
--- a/tests/ci.rs
+++ b/tests/ci.rs
@@ -431,7 +431,7 @@ fn core() {
     }
 
     {
-        // message from quorum - 1 group members
+        // message from more than quorum group members
         let client = TestClient::new(nodes.len(), event_sender.clone());
         let data = gen_plain_data();
         let mut number_of_events = 0;
--- a/tests/ci.rs
+++ b/tests/ci.rs
@@ -42,6 +42,7 @@ use routing::{Client, Data, Event, FullId, Node, PlainData, RequestContent, Requ
               ResponseMessage};
 
 const GROUP_SIZE: usize = kademlia_routing_table::GROUP_SIZE as usize;
+const QUORUM_SIZE: usize = 5;
 
 #[derive(Debug)]
 struct TestEvent(usize, Event);
@@ -108,13 +109,13 @@ fn spawn_select_thread(index: usize, main_sender: Sender<TestEvent>) -> (Sender<
     (sender, RaiiThreadJoiner::new(thread_handle))
 }
 
-fn recv_with_timeout<T>(receiver: &Receiver<T>, timeout: Duration) -> T {
+fn recv_with_timeout<T>(receiver: &Receiver<T>, timeout: Duration) -> Option<T> {
     let interval = Duration::from_millis(100);
     let mut elapsed = Duration::from_millis(0);
 
     loop {
         match receiver.try_recv() {
-            Ok(value) => return value,
+            Ok(value) => return Some(value),
             Err(TryRecvError::Disconnected) => break,
             _ => (),
         }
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @canndrew

tests/ci.rs
```diff
@@ -42,6 +42,7 @@ use routing::{Client, Data, Event, FullId, Node, PlainData, RequestContent, Requ
               ResponseMessage};
 
 const GROUP_SIZE: usize = kademlia_routing_table::GROUP_SIZE as usize;
+const QUORUM_SIZE: usize = 5;
```
afck: We should make that `pub` in kademlia_routing_table! (Maybe add a TODO here, so we use the constant once that's done.)

tests/ci.rs
```diff
@@ -347,15 +368,104 @@ fn core() {
...
+            if let Some(test_event) = recv_with_timeout(&event_receiver, Duration::from_secs(5)) {
+                match test_event {
+                    TestEvent(index, Event::Connected) if index == client.index => {
+                        unwrap_result!(client.client
+                                             .send_put_request(Authority::ClientManager(*client.name()), data.clone()));
+                    }
+                    TestEvent(index,
+                              Event::Request(RequestMessage{ src: Authority::Client{ .. },
+                                                                    dst: Authority::ClientManager(name),
+                                                                    content: RequestContent::Put(data, id) })) => {
```
afck: The indentation is a bit off here. (Maybe run rustfmt.)

tests/ci.rs
```diff
@@ -347,15 +368,104 @@ fn core() {
...
+                    _ => (),
+                }
+            } else {
+                break;
+            }
+        }
+    }
+
+    {
+        // message from quorum - 1 group members
```
afck: This is actually from `GROUP_SIZE` members, is it?
Maybe we should test it with exactly `QUORUM_SIZE` responses here instead.

tests/ci.rs
brian-js: Ok, but it was rustfmt that done the initial indentation.

afck: Ah. OK, if rustfmt insists, let's leave it like that. :-)

tests/ci.rs
brian-js: No, this one is for more than quorum responses but one event. The comment is wrong, I'll update it.

afck: But exactly `QUORUM_SIZE` responses should already suffice to cause the event, right? It might make sense to test this case, i. e. send 5 responses instead of 8 and assert that the event still occurs.

tests/ci.rs
brian-js: Ok.

tests/ci.rs
brian-js: It doesn't really matter how many responses are sent though what we're testing is that only one event occurs. From the task, "If `quorum_size` or more members send them, they do arrive and exactly one routing Event is sent." But what your suggesting also makes sense to test it works for quorum responses.

afck: Yes, if in doubt, I'd always opt for the "hardest" test: We want to check that 4 (or less) doesn't cause an event, and 5 (or more) does, so we should probably test these two cases.

brian-js: Those two cases are tested. Tests for exactly quorum succeeding might also be useful though.

tests/ci.rs
Fraser999: I think it was deliberately left private so we're forced to use [`dynamic_quorum_size`](https://github.com/maidsafe/kademlia_routing_table/blob/master/src/lib.rs#L331-L337).

However, we could remove the local `GROUP_SIZE` variable once https://github.com/maidsafe/kademlia_routing_table/pull/19 is merged since I think it's only there to cast the existing `u8` from routing table.

afck:
As dicussed, I'll create another issue for the case with exactly five responses.

PR #901, opened 2016-01-27
ulimit fix

Fraser999:
This avoids unconditionally setting the open file limit in the Travis script.

Now only the CI test (which is unrepresentative of normal usage conditions) on OS X will try to adjust the limit to 1024 if it's set below this.  Other tests and other OSs are unaffected.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/901)


maidsafe-highfive:
r? @canndrew

PR #914, opened 2016-02-23
Mock crust

madadam:
Minimal version of mock crust. Currently simulates only idealized network conditions where all messages are always delivered. More functionality coming later, for example simulating various NAT types, or unstable network connections, etc...

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/914)


maidsafe-highfive:
r? @brian-js

src/core.rs
```diff
@@ -800,6 +831,7 @@ impl Core {
     fn dispatch_request_response(&mut self,
                                  routing_msg: RoutingMessage)
                                  -> Result<(), RoutingError> {
+        trace!("{:?}Handling - {:?}", self, routing_msg);
```
afck: I wouldn't reintroduce those trace statements: We log most of the requests and responses in the handling methods.
(Or was that a git rebase gone wrong?)

src/core.rs
```diff
@@ -236,11 +270,7 @@ impl Core {
                                                         category_tx);
 
         // TODO(afck): Add the listening port to the Service constructor.
```
afck: That comment now probably belongs in the closure in non-test `new`.

src/core.rs
madadam: Definitely a rebase hiccup :grin: Will remove it.

madadam:
This PR is superseded by https://github.com/maidsafe/routing/pull/920, closing.

PR #915, opened 2016-02-25
Send NodeIdentify before ClientToNode.

afck:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/915)


maidsafe-highfive:
r? @canndrew

PR #920, opened 2016-03-01
Mock Crust + remove threads from Core

madadam:
- Add mock crust.
- Remove thread from `Core`. Users of `Core` now have to poll for events manually.
- `Client` and `Node` still use threads by default (so nothing changes for the upper layers), but when compiled with `--features="use-mock-crust"`, they use the polling interface too.
- Add some very simple unit tests for `Core`.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/routing/920)


maidsafe-highfive:
r? @canndrew

src/core.rs
```diff
@@ -211,16 +233,18 @@ pub struct Core {
     our_connection_info_map: LruCache<PublicId, OurConnectionInfo>,
     their_connection_info_map: LruCache<PublicId, TheirConnectionInfo>,
     get_request_count: usize,
```
afck: Thanks for cleaning up the mess with the statistics! Shouldn't `get_request_count` also go into `DebugStats`?

src/core.rs
```diff
@@ -250,146 +273,183 @@ impl Core {
...
+    }
 
-                            Action::Name { result_tx, } => {
-                                if result_tx.send(*self.name()).is_err() {
-                                    return;
-                                }
-                            }
+    /// If there is an event in the queue, processes it and returns true.
+    /// otherwise returns false. Never blocks.
+    #[allow(unused)]
```
afck: Why don't you use `#[cfg(feature = "use-mock-crust")]` here? (Similarly in `run`.)

src/core.rs
```diff
@@ -2124,3 +2200,21 @@ fn swap_remove_if<T, F>(vec: &mut Vec<T>, pred: F) -> Option<T>
         None
     }
 }
+
+#[cfg(not(feature = "use-mock-crust"))]
+fn restart_crust_service(service: &mut Service, sender: crust::CrustEventSender) {
+    use std::mem;
+
+    let _ = mem::replace(service,
```
afck: Wouldn't it be easier if this method were a member? Then instead of `replace`, you could probably just `self.crust_service = Service::new(...)`?

src/core_tests.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    while run {
+        run = false;
+
+        for node in nodes.iter_mut() {
+            if node.poll() {
+                run = true;
+            }
+        }
+    }
```
afck: `while nodes.iter_mut().any(TestNode::poll) {}` should be equivalent.

src/core.rs
```diff
@@ -250,146 +273,183 @@ impl Core {
...
 
-                            Action::Name { result_tx, } => {
-                                if result_tx.send(*self.name()).is_err() {
-                                    return;
-                                }
-                            }
+    /// If there is an event in the queue, processes it and returns true.
+    /// otherwise returns false. Never blocks.
+    #[allow(unused)]
+    pub fn poll(&mut self) -> bool {
```
afck: Maybe `handle_next_event` would be a better name for this.

src/lib.rs
```diff
@@ -55,7 +55,7 @@
 //! to the network through any node, and exchange public keys with it. That node becomes a
 //! bootstrap node for the client, and messages to and from the client will be routed over it.
 //!
-//! ```
+//! ```no_run
```
afck: Do these really break with mock Crust? After all, the signature of `Client::new` didn't change ...?

src/mock_crust/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+pub mod crust;
+mod support;
+#[cfg(test)] mod tests;
```
afck: I'm okay with this. But is rustfmt, too? :-D

src/mock_crust/support.rs
```diff
@@ -0,0 +1,449 @@
...
+//
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+// It seems that code used only in tests is considered unused by rust.
+#![allow(unused)]
```
afck: Shouldn't the whole module then be `#![cfg(test)]` instead?

src/mock_crust/support.rs
```diff
@@ -0,0 +1,449 @@
...
+        Network(Rc::new(RefCell::new(NetworkImpl {
+            services: HashMap::new(),
+            next_endpoint: 0,
+            queue: VecDeque::new(),
+        })))
+    }
+
+    /// Create new Device.
+    pub fn new_device(&self, config: Option<Config>, endpoint: Option<Endpoint>) -> Device {
+        let config = config.unwrap_or_else(|| Config::new());
```
afck: `config.unwrap_or_else(Config::new)`

src/mock_crust/support.rs
```diff
@@ -0,0 +1,449 @@
...
+            }
+        }
+    }
+
+    fn find_service(&self, endpoint: Endpoint) -> Option<Rc<RefCell<ServiceImpl>>> {
+        self.0.borrow().services.get(&endpoint).and_then(|s| s.upgrade())
+    }
+}
+
+/// Device represents a mock version of a real machine connected to the network.
```
afck: Isn't it a mock version of a Crust instance, rather than a machine?

src/mock_crust/support.rs
```diff
@@ -0,0 +1,449 @@
...
+        self.send_event(Event::BootstrapConnect(peer_id));
+        self.decrement_pending_bootstraps();
+    }
+
+    fn handle_bootstrap_failure(&mut self, _peer_endpoint: Endpoint) {
+        self.decrement_pending_bootstraps();
+    }
+
+    fn handle_connect_request(&mut self, peer_endpoint: Endpoint, peer_id: PeerId) {
+        // TODO: ignore if already connected?
```
afck: I would rather log a warning in that case, so that our tests let us know if Routing is attempting duplicate connections.

src/mock_crust/tests.rs
```diff
@@ -0,0 +1,304 @@
...
+use std::sync::mpsc::{self, Receiver};
+
+use super::crust::{CrustEventSender, Event, Service};
+use super::support::{Config, Network};
+
+fn get_event_sender()
+    -> (CrustEventSender,
+        Receiver<MaidSafeEventCategory>,
+        Receiver<Event>)
+{
```
afck: Is rustfmt OK with this?

tests/utils.rs
```diff
@@ -0,0 +1,76 @@
...
+use std::time::Duration;
+
+#[derive(Debug)]
+pub enum RecvWithTimeoutError {
+    Disconnected,
+    Timeout,
+}
+
+/// Blocks until something is received on the Receiver, or timeout, whichever
+/// hapens sooner.
```
afck: "happens"

src/core.rs
madadam: That was just a quick and dirty change to get those stats out of the way, but yeah, anything that is just statistics for debugging purposes should probably go there too.

src/core.rs
madadam: Yes, but what I'd like the best is if even the real `crust::Service` has a `restart` method. Then we wouldn't need this hack at all.

src/core_tests.rs
madadam: That is much nicer. And more idiomatic!

src/lib.rs
madadam: Yes, it breaks, because mock crust service needs to be bound to the mock `Device` object, which is done using the `make_current(device, || { /* create Core here */ })` hack. Without it, the mock Service asserts. Not sure it's worth doing anything about it. The examples still need to compile anyway, even with the `no_run` annotation.

src/mock_crust/support.rs
madadam: What is a "Crust instance" ?

src/lib.rs
afck: Ah, right, as long as it's still compiled, that's fine! :+1:

src/mock_crust/support.rs
afck: I mean a `crust::Service` instance, i. e. an instance of a software that runs a client or node ... I was just confused by the words "device" and "machine": It doesn't really correspond to a network card or a computer, but rather to individual programs, does it?

src/core_tests.rs
madadam: Actually, it's not completely equivalent: Your version stops on the first node whose `poll` returns true, mine always goes through all of them. In both versions all events are always processed, but the order this happens might be different.

src/mock_crust/mod.rs
madadam: Almost. It doesn't like the space between `]` and `mod` :)

src/core_tests.rs
afck: That's true. To make it completely equivalent, you'd probably have to do `fold` with `||` instead ... and then it's very debatable whether it's still nicer. :-D

But eventually, we'll want to control the exact order anyway, I think.

src/mock_crust/support.rs
madadam: No, it doesn't work. Without `#![allow(unused)]` it complains about unused stuff. With `#![cfg(test)]` it complains about unresolved imports. It seems that the `test` and `feature = "..."` cfgs interact in some weird and confusing ways. Don't want to spend too much time on this, so I'd keep it as is. We should probably extract mock crust into a separate crate or move it to the crust create, as that would solve this issue (I think).

src/mock_crust/support.rs
madadam: I agree the name can be confusing, and am open to suggestions here. Maybe `Device` => `Application`? Or `Program`?

src/mock_crust/support.rs
afck: :+1:

src/mock_crust/support.rs
afck: That would probably be better - but still not completely accurate, as applications may start several `crust::Service` instances (and safe_launcher does, I think). Maybe just `ServiceHandle` or `ServiceRef`?

src/mock_crust/tests.rs
madadam: It seems so.

src/mock_crust/support.rs
madadam: Ok, let's go with `ServiceHandle`.

afck:
Reviewed 9 of 17 files at r1, 1 of 2 files at r2, 8 of 8 files at r5.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/routing/920)*

afck:

---

*[src/core.rs, line 312 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/920#-KBmfHAyh8YFo_kyKz9I-r3-312:-KBqt4J8Hm7t1tYC1kgx:-275907566) ([raw file](https://github.com/maidsafe/routing/blob/1286c9b7f215eeb17f827cd4a71468c5f1b08163/src/core.rs#L312)):*
As discussed, let's use `poll` for now.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/routing/920)*

PR #942, opened 2016-03-16
Allow ci_tests to run client against droplets.

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/942)


maidsafe-highfive:
r? @canndrew

PR #944, opened 2016-03-18
Simplify bootstrap and disable node_id_cache check.

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/944)


maidsafe-highfive:
r? @brian-js

PR #946, opened 2016-03-24
Updated ImmutableData

Fraser999:
Affects mainly immutable_data.rs, but also minor impacts on other files.

This removes the tag from the ImmutableData struct and instead uses a new Name enum to define the type.

+@dirvine +@qima +@brian-js

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/946)


maidsafe-highfive:
r? @canndrew

PR #948, opened 2016-03-28
Use Crust master, and several other fixes.

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/948)


maidsafe-highfive:
r? @canndrew

PR #949, opened 2016-03-28
Add a message ID to GetNetworkName and ExpectCloseNode

afck:
This prevents them from being filtered out if a client retries relocating.

(Also: Some fixes to log messages in example_node.)

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/949)


maidsafe-highfive:
r? @canndrew

PR #952, opened 2016-03-30
1) Using lastest maidsafe_utilities 2) Resolving a build error - kade…

ustulation:
…mlia_routing_table QUORUM_SIZE is private and the code here used it which gives a compilation error in cargo test. Using a locally defined QUORUM_SIZE to get it to compile

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/952)


maidsafe-highfive:
r? @canndrew

PR #954, opened 2016-03-31
Make cache an option and remove name from HopMessage.

afck:
And disable the cache in the tests.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/954)


maidsafe-highfive:
r? @canndrew

PR #957, opened 2016-04-01
Send periodic keepalive messages to all peers.

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/957)


maidsafe-highfive:
r? @canndrew

PR #961, opened 2016-04-06
Implement responses in the example node.

afck:
The `ClientManager` now waits for a `PutSuccess` from the `NaeManager`
before sending it on to the client, more success and failure responses
are now handled, and cause corresponding updates to the `dm_accounts`.

Also, the output of ci_test is now a bit less redundant.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/961)


maidsafe-highfive:
r? @brian-js

PR #974, opened 2016-04-14
feat/types: Add a zero constructor for MessageId.

afck:
This returns a MessageId with value 0 that can be used if there is no
danger of duplication anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/974)


maidsafe-highfive:
r? @canndrew

PR #976, opened 2016-04-17
Redefine immutable data types

dirvine:
Scope: Data

Create Backup and Sacrifical data types as structs as opposed to enum variant
These types include ImmutableData themselves. Very simple really but should be effective and remove some complexity form their use. They are pub visibility.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/976)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Routing - Change Log
 
+## [0.15.1]
+- Fix a message handling bug.
+- Add `MessageId::zero` constructor.
+- Always send `NodeAdded` for a new peer, even if not in a common group.
+
 ## [0.15.0]
 - Implement Rand for mock PeerId.
 - Add data name to Put, Post and Delete success responses.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ license = "GPL-3.0"
 name = "routing"
 readme = "README.md"
 repository = "https://github.com/maidsafe/routing"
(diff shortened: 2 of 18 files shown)


maidsafe-highfive:
r? @brian-js

afck:
The core_tests still need to be updated.

---

Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/immutable_data.rs, line 58 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/976#-KFXouIzMcW6JiF784f5:-KFXouIzMcW6JiF784f6:-941389928) ([raw file](https://github.com/maidsafe/routing/blob/32f3ab1e74114b8f65e4a797b7268eb3cb16000a/src/immutable_data.rs#L58)):*
I think Clippy will want backticks around `ImmutableData`. (Also below.)

---

*[src/immutable_data.rs, line 86 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/976#-KFXp5bdNBaPVF9TaT9x:-KFXp5bdNBaPVF9TaT9y:677121513) ([raw file](https://github.com/maidsafe/routing/blob/32f3ab1e74114b8f65e4a797b7268eb3cb16000a/src/immutable_data.rs#L86)):*
Typo

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/976#-:-KFXpPR6YHChsQdMKZUx:1700671186)*

PR #990, opened 2016-04-22
Add GetNetworkNameFailed and NetworkStartupFailed events.

afck:
Also dont' send `GetNetworkName` if we just sent it to another node.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/990)


maidsafe-highfive:
r? @canndrew

Issue #1008, opened 2016-05-04
AppVeyor fails with "OS Error 8" on x86 but not on x64.


afck:
See e. g.:
https://ci.appveyor.com/project/MaidSafe-QA/routing/build/1.0.916/job/bbb2k2s6towm0k88#L628

PR #1009, opened 2016-05-05
refactor/timer: change `Timer` to use a single thread and be stoppable

Fraser999:
This changes `Timer` so that it doesn't spawn a detached thread per scheduled task.  It also makes
`Timer` stop running the thread when dropped, whereas previously the spawned threads would sleep
unconditionally until their exipiry times.  A test for `Timer` is also added.

FIX: #1008

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1009)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,7 +21,7 @@ install:
         . ".\Install libsodium.ps1"
 
 platform:
-  #- x86 // TODO: Re-enable once https://github.com/maidsafe/routing/issues/1008 is fixed.
+  - x86
   - x64
 
 configuration:
--- a/src/structured_data.rs
+++ b/src/structured_data.rs
@@ -481,7 +481,7 @@ mod test {
                     Ok(mut new_structured_data) => {
                         assert_eq!(new_structured_data.add_signature(&keys2.1).ok(), Some(0));
                         match orig_structured_data.replace_with_other(new_structured_data) {
-                            Ok(()) => println!("All good"),
+                            Ok(()) => (),
                             Err(e) => panic!("Error {:?}", e),
                         }
                         // transfer ownership back to keys1 only
@@ -494,7 +494,7 @@ mod test {
                                                          Some(&new_owner.1)) {
                             Ok(another_new_structured_data) => {
                                 match orig_structured_data.replace_with_other(another_new_structured_data) {
-                                    Ok(()) => println!("All good"),
+                                    Ok(()) => (),
                                     Err(e) => panic!("Error {:?}", e),
                                 }
                             }
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -15,44 +15,183 @@
 // Please review the Licences for the specific language governing permissions and limitations
 // relating to use of the SAFE Network Software.
 
-use std::thread;
-use std::time::Duration;
+use std::collections::BTreeMap;
+use std::sync::{Arc, Mutex, Condvar};
+use std::time::{Duration, Instant};
 
 use action::Action;
+use itertools::Itertools;
+use maidsafe_utilities::thread::RaiiThreadJoiner;
 use types::RoutingActionSender;
+
+struct Detail {
+    deadlines: BTreeMap<Instant, Vec<u64>>,
+    cancelled: bool,
+}
 
 /// Simple timer.
 pub struct Timer {
-    sender: RoutingActionSender,
     next_token: u64,
+    detail_and_cond_var: Arc<(Mutex<Detail>, Condvar)>,
+    _worker: RaiiThreadJoiner,
 }
 
 impl Timer {
-    /// Create new timer passing a sender used to send Timeout events.
+    /// Creates a new timer, passing a channel sender used to send `Timeout` events.
     pub fn new(sender: RoutingActionSender) -> Self {
+        let detail = Detail {
+            deadlines: BTreeMap::new(),
+            cancelled: false,
+        };
+        let detail_and_cond_var = Arc::new((Mutex::new(detail), Condvar::new()));
+        let detail_and_cond_var_clone = detail_and_cond_var.clone();
+        let worker = RaiiThreadJoiner::new(thread!("Timer",
+                                                   move || Self::run(sender, detail_and_cond_var)));
         Timer {
-            sender: sender,
             next_token: 0,
+            detail_and_cond_var: detail_and_cond_var_clone,
+            _worker: worker,
         }
     }
 
-    /// Schedule a timeout event after `duration`. Returns a token that can be
-    /// used to identify the timeout event.
+    /// Schedules a timeout event after `duration`. Returns a token that can be used to identify
+    /// the timeout event.
     pub fn schedule(&mut self, duration: Duration) -> u64 {
-        let sender = self.sender.clone();
-        let token = self.next_token();
-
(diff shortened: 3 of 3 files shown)


afck:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/timer.rs, line 57 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1009#-KGziA1JvV_eezngHqZG:-KGziA1KemXhc7wR1Zke:585426474) ([raw file](https://github.com/maidsafe/routing/blob/e953e35b8670c334c0c0f7402202a721ccd29f57/src/timer.rs#L57)):*
You must mistakenly have inserted another space by mistake there mistakenly ...
Also, the convention seems to be that docs are third person: `Schedules`, `Creates`, ...

---

*[src/timer.rs, line 65 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1009#-KGzilC5WRo2DLYYTubK:-KGzilC6DnIQy-VxN8DE:-1164087073) ([raw file](https://github.com/maidsafe/routing/blob/e953e35b8670c334c0c0f7402202a721ccd29f57/src/timer.rs#L65)):*
Good point!
We should at least detect if that happens, and log an error!
Actually, I'd even vote for making it a multimap (i. e. a `BTreeMap<Instant, Vec<u64>>`) to be on the safe side, or for adding a nanosecond until we have a key that doesn't exist yet.

---

*[src/timer.rs, line 158 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1009#-KGzw8y06FSpUuze_pNO:-KGzw8y06FSpUuze_pNP:264796862) ([raw file](https://github.com/maidsafe/routing/blob/e953e35b8670c334c0c0f7402202a721ccd29f57/src/timer.rs#L158)):*
Are these exactly identical?
Is that because of some borrow checker issue?

---

*[src/timer.rs, line 178 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1009#-KGzxc4s5MR_eQI4yKmq:-KGzxc4t2GQFnFYTZDXY:-1055549691) ([raw file](https://github.com/maidsafe/routing/blob/e953e35b8670c334c0c0f7402202a721ccd29f57/src/timer.rs#L178)):*
Is that `clone()` necessary?
You could use a catch all `category` instead of `Crust` below, then you don't have to use this variable there.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1009)*

Fraser999:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/timer.rs, line 65 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1009#-KGzilC5WRo2DLYYTubK:-KH-6tp2W0tajt1vkQyO:198448500) ([raw file](https://github.com/maidsafe/routing/blob/e953e35b8670c334c0c0f7402202a721ccd29f57/src/timer.rs#L65)):*
I'm reasonably sure that it won't be possible to get duplicate entries - after all, we're locking a mutex between each call.  Nonetheless, I guess it won't hurt much to add a nanosecond to the ejected entry and re-add it.

---

*[src/timer.rs, line 158 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1009#-KGzw8y06FSpUuze_pNO:-KH-8cDthajYd5GT5uXz:2053083745) ([raw file](https://github.com/maidsafe/routing/blob/e953e35b8670c334c0c0f7402202a721ccd29f57/src/timer.rs#L158)):*
Ah - that's a mistake :)  They are indeed identical!  I added the second one while debugging so that the failing backtrace showed which instance of `check_no_events_received()` was being run.

---

*[src/timer.rs, line 178 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1009#-KGzxc4s5MR_eQI4yKmq:-KH-9y89KZw8YeiHf41o:-407911312) ([raw file](https://github.com/maidsafe/routing/blob/e953e35b8670c334c0c0f7402202a721ccd29f57/src/timer.rs#L178)):*
Not sure I follow.  If I replace `MaidSafeEventCategory::Crust` with `_`, I'd still want to use the `category` variable in the `panic!`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1009)*

afck:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/timer.rs, line 65 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1009#-KGzilC5WRo2DLYYTubK:-KH-CgRe_mfGUllILuXO:-956076067) ([raw file](https://github.com/maidsafe/routing/blob/e953e35b8670c334c0c0f7402202a721ccd29f57/src/timer.rs#L65)):*
I'd feel better with that approach, yes. :+1: 

---

*[src/timer.rs, line 178 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1009#-KGzxc4s5MR_eQI4yKmq:-KH-CIoQ92Tq39AMGr6x:-385558733) ([raw file](https://github.com/maidsafe/routing/blob/e953e35b8670c334c0c0f7402202a721ccd29f57/src/timer.rs#L178)):*
But if you then replace `_` with `category`, you can use _that_ variable in the panic message.
Also, please replace `unwrap_result` with `expect`, as the long-term plan is to remove that macro.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1009)*

Fraser999:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/timer.rs, line 65 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1009#-KGzilC5WRo2DLYYTubK:-KH-Ed9l5QpBF9F4b0J6:-122454121) ([raw file](https://github.com/maidsafe/routing/blob/e953e35b8670c334c0c0f7402202a721ccd29f57/src/timer.rs#L65)):*
On second thoughts, I think I prefer the multimap approach.  Seems cleaner and more accurate.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1009)*

Fraser999:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/timer.rs, line 178 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1009#-KGzxc4s5MR_eQI4yKmq:-KH-GHU5sxvoJ7xbpzjV:975027706) ([raw file](https://github.com/maidsafe/routing/blob/e953e35b8670c334c0c0f7402202a721ccd29f57/src/timer.rs#L178)):*
> But if you then replace _ with category, you can use that variable in the panic message.

Ah - I see now :)

> Also, please replace unwrap_result with expect, as the long-term plan is to remove that macro.

Was this discussed anywhere?  The last discussion I remember is that we're allowed to use these in tests.  In my experience, they're far more useful than `expect`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1009)*

afck:
Reviewed 2 of 3 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1009)*

PR #1012, opened 2016-05-06
fix/core: don't send tick events in clients

afck:
Don't initialise `tick_timer_token` to `0` as that is the first valid
timer token and may trigger `Tick` events in clients.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1012)


maidsafe-highfive:
r? @canndrew

Viv-Rajkumar:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1012)*

PR #1017, opened 2016-05-12
feat/core: use a Crust priority based on message type

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1017)


maidsafe-highfive:
r? @canndrew

PR #1021, opened 2016-05-13
fix/core: small networks only accept new nodes via the first node

afck:
Otherwise the quorum size might be inaccurate.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1021)


maidsafe-highfive:
r? @canndrew

Viv-Rajkumar:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1021)*

PR #1028, opened 2016-05-26
refactor/core: deduplicate and extract some logic from Core::send

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1028)


maidsafe-highfive:
r? @canndrew

PR #1029, opened 2016-05-26
feat/core: reduce XorName size from 512 to 256 bits

afck:
This reduces the size of several messages and the danger of any hash
collisions is still negligible.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1029)


maidsafe-highfive:
r? @canndrew

PR #1034, opened 2016-06-03
feat/account_info: add handling for retrieving Client account information

Fraser999:
This adds new message variants and API functions to allow Clients to request information about their
accounts.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1034)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,8 @@ repository = "https://github.com/maidsafe/routing"
 version = "0.21.0"
 
 [dependencies]
-accumulator = "~0.3.0"
-clippy = {version = "~0.0.69", optional = true}
+accumulator = "~0.4.0"
+clippy = {version = "~0.0.71", optional = true}
 crust = "~0.14.0"
 itertools = "~0.4.15"
 kademlia_routing_table = "~0.6.0"
--- a/examples/utils/example_node.rs
+++ b/examples/utils/example_node.rs
@@ -123,6 +123,10 @@ impl ExampleNode {
                 warn!("{:?} ExampleNode: Delete unimplemented.",
                       self.get_debug_name());
             }
+            Request::GetAccountInfo(..) => {
+                warn!("{:?} ExampleNode: GetAccountInfo unimplemented.",
+                      self.get_debug_name());
+            }
             Request::Refresh(content, id) => {
                 self.handle_refresh(content, id);
             }
--- a/src/client.rs
+++ b/src/client.rs
@@ -31,6 +31,7 @@ use error::{InterfaceError, RoutingError};
 use authority::Authority;
 use messages::Request;
 use types::MessageId;
+use xor_name::XorName;
 
 type RoutingResult = Result<(), RoutingError>;
 
@@ -43,6 +44,7 @@ pub struct Client {
     interface_result_tx: Sender<Result<(), InterfaceError>>,
     interface_result_rx: Receiver<Result<(), InterfaceError>>,
     action_sender: ::types::RoutingActionSender,
+    name: XorName,
 
     #[cfg(feature = "use-mock-crust")]
     core: RefCell<Core>,
@@ -71,6 +73,7 @@ impl Client {
 
         // start the handler for routing with a restriction to become a full node
         let (action_sender, mut core) = Core::new(event_sender, Role::Client, keys, use_data_cache);
+        let name = *core.name();
         let (tx, rx) = channel();
 
         let raii_joiner = RaiiThreadJoiner::new(thread!("Client thread", move || {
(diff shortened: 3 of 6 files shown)


maidsafe-highfive:
r? @afck

Viv-Rajkumar:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/client.rs, line 166 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1034#-KJLlC5MlOpsum-w9d77:-KJLlC5MlOpsum-w9d78:-803944246) ([raw file](https://github.com/maidsafe/routing/blob/a87a7e76849c073096a833d208205ef204b1d08f/src/client.rs#L166)):*
> ```Rust
>                                          message_id: MessageId)
>                                          -> Result<(), InterfaceError> {
>         let dst = Authority::ClientManager(self.name);
> ```

why are we choosing to save `self.name` locally btw? and not just call `self.core.name()` here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1034)*

Viv-Rajkumar:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/client.rs, line 166 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1034#-KJLlC5MlOpsum-w9d77:-KJLnCxtrx5BoNcoaDeN:1206079082) ([raw file](https://github.com/maidsafe/routing/blob/a87a7e76849c073096a833d208205ef204b1d08f/src/client.rs#L166)):*
<details><summary><i>Previously, Viv-Rajkumar (Viv Rajkumar) wrote…</i></summary>
> why are we choosing to save `self.name` locally btw? and not just call `self.core.name()` here?
</details>
Oh guess its to maybe not go cross thread for every call of this? If thats the case for now it could maybe just take the `dst` auth as an argument like what `send_put_request` does to be consistent?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1034)*

Fraser999:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/client.rs, line 166 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1034#-KJLlC5MlOpsum-w9d77:-KJLt5xMNxYMzx5o9u-R:676493922) ([raw file](https://github.com/maidsafe/routing/blob/a87a7e76849c073096a833d208205ef204b1d08f/src/client.rs#L166)):*
<details><summary><i>Previously, Viv-Rajkumar (Viv Rajkumar) wrote…</i></summary>
> Oh guess its to maybe not go cross thread for every call of this? If thats the case for now it could maybe just take the `dst` auth as an argument like what `send_put_request` does to be consistent?
</details>
Yeah - it's to avoid concurrent access to the variable in core.  I know we don't modify Client names just now, but if that changes, we'd possibly get problems there.  The chances are slim, so if you prefer I can change this to:

```rust
let dst = Authority::ClientManager(*self.core.borrow().name());
```
If we want to avoid the concurrent access issue in the future, we'd have to change this to something more like the node's `name()` call where we send a request to core via the action channel, wait for a response, then use that.

I'd really rather avoid having the `dst` in this function's signature, since we'd really have to check they'd passed the correct destination or ignore whatever they passed or something like that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1034)*

Viv-Rajkumar:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/client.rs, line 166 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1034#-KJLlC5MlOpsum-w9d77:-KJLzGGJz4yzcp2tAtK7:-598944484) ([raw file](https://github.com/maidsafe/routing/blob/a87a7e76849c073096a833d208205ef204b1d08f/src/client.rs#L166)):*
> Yeah - it's to avoid concurrent access to the variable in core.  I know we don't modify Client names just now, but if that changes, we'd possibly get problems there.  The chances are slim, so if you prefer I can change this to:
> 
> ```rust
> let dst = Authority::ClientManager(*self.core.borrow().name());
> ```
> If we want to avoid the concurrent access issue in the future, we'd have to change this to something more like the node's `name()` call where we send a request to core via the action channel, wait for a response, then use that.
> 
> I'd really rather avoid having the `dst` in this function's signature, since we'd really have to check they'd passed the correct destination or ignore whatever they passed or something like that.
</details>
yep `let dst = Authority::ClientManager(*self.core.borrow().name());` sounds good to me man. :thumbsup: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1034)*

afck:
Reviewed 3 of 5 files at r1, 1 of 1 files at r2, 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1034)*

PR #1040, opened 2016-06-09
fix/core: fix message accumulation for hash messages

afck:
Since the accumulator values could be either hashes or full messages, in
some cases they were counted twice: If a group resends a message,
another node will send the full content, so two nodes will send
something different than before, and they will be counted twice.

This separates the accumulator from the hash map.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1040)


maidsafe-highfive:

PR #1046, opened 2016-06-13
feat/mock_crust: unidirectional blocking

madadam:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1046)


maidsafe-highfive:
r? @afck

PR #1050, opened 2016-06-14
feat/core: mio-based crust, message splitting, group deduplication

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1050)


maidsafe-highfive:

PR #1055, opened 2016-06-16
refactor/core_tests: Use real Node and Client in tests

madadam:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1055)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -31,6 +31,7 @@ use error::{InterfaceError, RoutingError};
 use authority::Authority;
 use messages::{Request, DEFAULT_PRIORITY, CLIENT_GET_PRIORITY};
 use types::MessageId;
+use xor_name::XorName;
 
 type RoutingResult = Result<(), RoutingError>;
 
@@ -157,6 +158,14 @@ impl Client {
                          CLIENT_GET_PRIORITY)
     }
 
+    /// Returns the name of this node.
+    pub fn name(&self) -> Result<XorName, InterfaceError> {
+        let (result_tx, result_rx) = channel();
+        try!(self.action_sender.send(Action::Name { result_tx: result_tx }));
+
+        self.receive_action_result(&result_rx)
+    }
+
     fn send_action(&self,
                    content: Request,
                    dst: Authority,
@@ -170,18 +179,18 @@ impl Client {
         };
 
         try!(self.action_sender.send(action));
-        self.receive_action_result()
+        try!(self.receive_action_result(&self.interface_result_rx))
     }
 
     #[cfg(not(feature = "use-mock-crust"))]
-    fn receive_action_result(&self) -> Result<(), InterfaceError> {
-        try!(self.interface_result_rx.recv())
+    fn receive_action_result<T>(&self, rx: &Receiver<T>) -> Result<T, InterfaceError> {
+        Ok(try!(rx.recv()))
     }
 
     #[cfg(feature = "use-mock-crust")]
-    fn receive_action_result(&self) -> Result<(), InterfaceError> {
+    fn receive_action_result<T>(&self, rx: &Receiver<T>) -> Result<T, InterfaceError> {
         while self.poll() {}
-        try!(self.interface_result_rx.recv())
+        Ok(try!(rx.recv()))
     }
 }
 
--- a/src/core.rs
+++ b/src/core.rs
@@ -333,17 +333,6 @@ impl Core {
         self.full_id.public_id().name()
     }
 
-    /// Returns the names of all nodes in the close group of this node.
-    #[allow(unused)]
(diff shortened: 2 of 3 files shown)


maidsafe-highfive:
r? @afck

afck:
Reviewed 3 of 3 files at r1.

---

*[src/client.rs, line 192 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1055#-KKO-evderZ1LEGzNKkb:-KKO-everA8VDJoUSe7F:43736998) ([raw file](https://github.com/maidsafe/routing/blob/6a5361a79b7a90a41f98c0ee16a1230de2769e58/src/client.rs#L192)):*
> ```Rust
>     #[cfg(feature = "use-mock-crust")]
>     fn receive_action_result<T>(&self, rx: &Receiver<T>) -> Result<T, InterfaceError> {
>         while self.poll() {}
> ```

Actually, should we even do that here? I think in most cases the test itself wants to control when to poll and when to check for events. And that would even allow us to merge the two implementations.

---

*[src/core_tests.rs, line 81 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1055#-KKO0sKE7e4vZnS4iDjo:-KKO0sKE7e4vZnS4iDjp:-1854013164) ([raw file](https://github.com/maidsafe/routing/blob/6a5361a79b7a90a41f98c0ee16a1230de2769e58/src/core_tests.rs#L81)):*
> ```Rust
>         TestNode {
>             handle: handle,
>             node: node,
> ```

I wonder whether this (and `client` in `TestClient`) should be called `inner` ... or should (some of) the `node` variables below be renamed to `test_node` ... not sure. It's okay as is, too. Your call. :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1055)*

madadam:
*[src/client.rs, line 192 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1055#-KKO-evderZ1LEGzNKkb:-KKOBLPT0mXBrufS8umu:-1362570227) ([raw file](https://github.com/maidsafe/routing/blob/6a5361a79b7a90a41f98c0ee16a1230de2769e58/src/client.rs#L192)):*
> Actually, should we even do that here? I think in most cases the test itself wants to control when to poll and when to check for events. And that would even allow us to merge the two implementations.
</details>
The problem with doing what you suggest is the way methods like `Node::name` work. The method creates a channel, then sends a message to `Core`, then `recv` on the channel to retrieve the result. So it hides the async internals of `Core` behind a sync interface. If we didn't `poll` before `recv`, the test would get stuck.

It all feels quite hacky, so I wouldn't mind at all to refactor it into more sensible design.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1055)*

madadam:
*[src/core_tests.rs, line 81 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1055#-KKO0sKE7e4vZnS4iDjo:-KKODSL-lhJz83nig8h-:1411505674) ([raw file](https://github.com/maidsafe/routing/blob/6a5361a79b7a90a41f98c0ee16a1230de2769e58/src/core_tests.rs#L81)):*
> I wonder whether this (and `client` in `TestClient`) should be called `inner` ... or should (some of) the `node` variables below be renamed to `test_node` ... not sure. It's okay as is, too. Your call. :-)
</details>
Yeah, `inner` is probably better. I'll change it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1055)*

afck:
Review status: 2 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/client.rs, line 192 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1055#-KKO-evderZ1LEGzNKkb:-KKOLB-DdWZCWlxSyFfv:-960082980) ([raw file](https://github.com/maidsafe/routing/blob/6a5361a79b7a90a41f98c0ee16a1230de2769e58/src/client.rs#L192)):*
> The problem with doing what you suggest is the way methods like `Node::name` work. The method creates a channel, then sends a message to `Core`, then `recv` on the channel to retrieve the result. So it hides the async internals of `Core` behind a sync interface. If we didn't `poll` before `recv`, the test would get stuck.
> 
> It all feels quite hacky, so I wouldn't mind at all to refactor it into more sensible design.
</details>
Ugh, of course.
Then let's leave it as is for now: first the tests, then the refactoring.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1055)*

afck:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1055)*

PR #1057, opened 2016-06-16
refactor/tests: use the new libsodium_seeded_prng crate in tests

Fraser999:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1057)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,18 +22,20 @@ maidsafe_utilities = "~0.6.0"
 rand = "~0.3.14"
 rustc-serialize = "~0.3.19"
 sodiumoxide = "~0.0.10"
+libsodium_seeded_prng = { version = "~0.1.1", optional = true }
 term = "~0.4.4"
 
 [dev-dependencies]
 docopt = "~0.6.80"
 libc = "~0.2.12"
+libsodium_seeded_prng = "~0.1.1"
 
 [build-dependencies]
 hyper = {version = "~0.9.7", optional = true}
 
 [features]
 generate-diagrams = ["hyper"]
-use-mock-crust = []
+use-mock-crust = ["libsodium_seeded_prng"]
 
 [[example]]
 bench = false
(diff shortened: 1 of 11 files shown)


maidsafe-highfive:
r? @afck

afck:
Reviewed 11 of 11 files at r1.

---

*[src/core_tests.rs, line 399 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1057#-KKS3AUumMKmFM4JQN6T:-KKS3AUvkmyk03ojAQCV:1410755205) ([raw file](https://github.com/maidsafe/routing/blob/49ae28930e7c9c1ae57d72d52174f8d422bc4163/src/core_tests.rs#L399)):*
> ```Rust
>     // Send a message from the group `src` to the node `dst`. Only the `QUORUM_SIZE`-th sender
>     // should cause accumulation and a `Response` event. The event should only occur once.
>     let message_id = MessageId::new();
> ```

Probably not relevant for this particular test, but eventually we should also make `MessageId` implement `Rand` to make this deterministic here.
But that might turn out to be a bigger change: We should then also make sure that all the message IDs generated by clients and nodes are deterministic (especially for the safe_vault tests), so maybe not in this PR.

---

*[src/core_tests.rs, line 492 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1057#-KKS1czPvKb1aixiPIGz:-KKS1czQBDiCdUougnX5:192759963) ([raw file](https://github.com/maidsafe/routing/blob/49ae28930e7c9c1ae57d72d52174f8d422bc4163/src/core_tests.rs#L492)):*
> ```Rust
>     for i in 0..100 {
>         let len = nodes.len();
>         if len > GROUP_SIZE + 2 && Range::new(0, 3).ind_sample(&mut *rng_ptr.borrow_mut()) == 0 {
> ```

In tests with lots of rng uses, we might also just initialise a new, local `XorShiftRng` with bytes taken from `libsodium_seeded_prng`'s. Maybe that would be more readable. Not sure ...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1057)*

Fraser999:
Closed in favour of https://github.com/maidsafe/routing/pull/1060.

PR #1058, opened 2016-06-17
test/churn: Add tests for requests during churn

madadam:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1058)

### DIFF
--- a/src/core_tests.rs
+++ b/src/core_tests.rs
@@ -19,6 +19,7 @@ use rand::{self, Rng, SeedableRng, XorShiftRng};
 use rand::distributions::{IndependentSample, Range};
 use std::cmp;
 use std::collections::HashSet;
+use std::ops;
 use std::sync::mpsc;
 use std::thread;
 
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @afck

afck:
Reviewed 1 of 1 files at r2.

---

*[a discussion](https://reviewable.io:443/reviews/maidsafe/routing/1058#-KKU0ExWLOppzrfc5utP:-KKU0ExWLOppzrfc5utQ:411352293) (no related file):*
:+1: to all the deduplications!

There are a few new Clippy errors, though.

---

*[src/core_tests.rs, line 104 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1058#-KKU3DCCDpyPHhEsbQyJ:-KKU3DCDo4xvW3e4jY_o:-1064838246) ([raw file](https://github.com/maidsafe/routing/blob/c7648fc127e749c0ca464cdeb5170351f8a928a4/src/core_tests.rs#L104)):*
> ```Rust
>             }
> 
>             return result;
> ```

This could be simplified to `if Some(result) != exclude { return result; }`.

Alternatively, the `loop` could be avoided completely, although that would be more code then:
```
match exclude {
  None => self.gen_range(low, high),
  Some(exclude) => {
    let mut r = self.gen_range(low, high - 1);
    if r >= exclude {
      r += 1
    }
    r
  }
}
```

---

*[src/core_tests.rs, line 552 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1058#-KKU4pPc-Vmu0euyxAWE:-KKU4pPdEXKJQGdoz7Q1:72392515) ([raw file](https://github.com/maidsafe/routing/blob/6f66432f475f20efd3f43e747e885630b3b69069/src/core_tests.rs#L552)):*
> ```Rust
> }
> 
> #[ignore]
> ```

I actually added that back in because the test was flaky and failed occasionally.
Did you run this a few times in a loop? Let's only re-enable it once we are confident that it now always passes. (Maybe it does now - not sure whether the last time I saw if fail was just due to a `Tick`.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1058)*

madadam:

---

*[src/core_tests.rs, line 552 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1058#-KKU4pPc-Vmu0euyxAWE:-KKgK_dl_Fp38LbpPuxt:1447156184) ([raw file](https://github.com/maidsafe/routing/blob/6f66432f475f20efd3f43e747e885630b3b69069/src/core_tests.rs#L552)):*
> I actually added that back in because the test was flaky and failed occasionally.
> Did you run this a few times in a loop? Let's only re-enable it once we are confident that it now always passes. (Maybe it does now - not sure whether the last time I saw if fail was just due to a `Tick`.)
</details>
Oh, sorry about that. I thought it got lost in a rebase or something. Might have been just the `Tick` thing though. I'll try running this test several times  and add the `ignore` back if it fails.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1058)*

afck:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1058)*

afck:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1058#-:-KKhaH4pdvLMl7BpbM9n:1417088721)*

PR #1059, opened 2016-06-20
Response caching

madadam:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1059)

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -1,0 +1,37 @@
+// Copyright 2016 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under (1) the MaidSafe.net Commercial License,
+// version 1.0 or later, or (2) The General Public License (GPL), version 3, depending on which
+// licence you accepted on initial access to the Software (the "Licences").
+//
+// By contributing code to the SAFE Network Software, or to this project generally, you agree to be
+// bound by the terms of the MaidSafe Contributor Agreement, version 1.0.  This, along with the
+// Licenses can be found in the root directory of this project at LICENSE, COPYING and CONTRIBUTOR.
+//
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+use messages::{Request, Response};
+
+/// A cache that stores `Response`s keyed by `Requests`. Should be implemented
+/// by layers above routing.
+pub trait Cache : Send {
+    /// Retrieve cached response for the given request.
+    fn get(&self, request: &Request) -> Option<Response>;
+
+    /// Cache the given response.
+    fn put(&self, response: Response);
+}
+
+/// A no-op implementation of the `Cache` trait. Throws everything away on put
+/// and always returns `None` on get.
+pub struct NullCache;
+
+impl Cache for NullCache {
+    fn get(&self, _: &Request) -> Option<Response> { None }
+    fn put(&self, _: Response) {}
+}
--- a/src/client.rs
+++ b/src/client.rs
@@ -26,6 +26,7 @@ use std::sync::mpsc::{Receiver, Sender, channel};
 use id::FullId;
 use action::Action;
 use event::Event;
+use cache::NullCache;
 use core::{Core, Role};
 use data::{Data, DataIdentifier};
 use error::{InterfaceError, RoutingError};
@@ -69,7 +70,10 @@ impl Client {
         sodiumoxide::init();  // enable shared global (i.e. safe to multithread now)
 
         // start the handler for routing with a restriction to become a full node
-        let (action_sender, mut core) = Core::new(event_sender, Role::Client, keys);
+        let (action_sender, mut core) = Core::new(event_sender,
+                                                  Role::Client,
+                                                  keys,
+                                                  Box::new(NullCache));
         let (tx, rx) = channel();
 
         let raii_joiner = RaiiThreadJoiner::new(thread!("Client thread", move || {
@@ -88,7 +92,10 @@ impl Client {
     #[cfg(feature = "use-mock-crust")]
     pub fn new(event_sender: Sender<Event>, keys: Option<FullId>) -> Result<Client, RoutingError> {
         // start the handler for routing with a restriction to become a full node
-        let (action_sender, core) = Core::new(event_sender, Role::Client, keys);
+        let (action_sender, core) = Core::new(event_sender,
+                                              Role::Client,
+                                              keys,
+                                              Box::new(NullCache));
         let (tx, rx) = channel();
 
         Ok(Client {
@@ -109,6 +116,12 @@ impl Client {
     /// Resend all unacknowledged messages.
     pub fn resend_unacknowledged(&self) -> bool {
         self.core.borrow_mut().resend_unacknowledged()
+    }
+
+    #[cfg(feature = "use-mock-crust")]
+    /// Are there any unacknowledged messages?
+    pub fn has_unacknowledged(&self) -> bool {
+        self.core.borrow().has_unacknowledged()
     }
 
     /// Send a Get message with a `DataIdentifier` to an `Authority`, signed with given keys.
--- a/src/core.rs
+++ b/src/core.rs
@@ -36,7 +36,7 @@ use rand;
 use sodiumoxide::crypto::{box_, sign};
 use sodiumoxide::crypto::hash::sha256;
 use std::{cmp, iter, fmt};
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{HashMap, HashSet};
 use std::fmt::{Debug, Formatter};
 use std::net::SocketAddr;
 use std::sync::mpsc;
@@ -46,6 +46,7 @@ use xor_name::{XorName, XOR_NAME_BITS};
 
 use action::Action;
 use authority::Authority;
+use cache::Cache;
 use error::{RoutingError, InterfaceError};
 use event::Event;
 use id::{FullId, PublicId};
@@ -53,7 +54,7 @@ use stats::Stats;
 use timer::Timer;
 use types::{MessageId, RoutingActionSender};
 use messages::{DirectMessage, HopMessage, Message, MessageContent, RoutingMessage, SignedMessage,
-               UserMessage, DEFAULT_PRIORITY};
+               UserMessage, UserMessageCache, DEFAULT_PRIORITY};
 use utils;
 
 /// The group size for the routing table. This is the maximum that can be used for consensus.
@@ -213,9 +214,9 @@ pub struct Core {
     tunnels: Tunnels,
     stats: Stats,
     send_filter: LruCache<(u64, PeerId, u8), ()>,
-    /// This maps `(hash, part_count)` of an incoming `UserMessage` to the map containing all
-    /// `UserMessagePart`s that have already arrived, by `part_index`.
-    user_msg_cache: LruCache<(u64, u32), BTreeMap<u32, Vec<u8>>>,
+    user_msg_cache: UserMessageCache,
+    cacheable_user_msg_cache: UserMessageCache,
+    response_cache: Box<Cache>,
     peer_mgr: PeerManager,
     // List of addresses of peers who have previously sent `BootstrapDeny` messages.
     bootstrap_blacklist: HashSet<SocketAddr>,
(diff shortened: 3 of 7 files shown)


maidsafe-highfive:
r? @afck

afck:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/core.rs, line 879 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKiMZOUVgPMI2Oy8XSZ:-KKiMZOUVgPMI2Oy8XS_:1418894816) ([raw file](https://github.com/maidsafe/routing/blob/b0b283d89cab2696a3eb996e1867ac6374ae4d48/src/core.rs#L879)):*
> ```Rust
>                         let priority = 0;
> 
>                         try!(self.send_user_message(routing_msg.dst.clone(),
> ```

This probably needs to be `Authority::ManagedNode(our_name)`, as we are not responding as part of the group.

---

*[src/core.rs, line 1030 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKiN3EMGXfXIENNvbFZ:-KKiN3EN8Ahbx5QWahey:-1731374676) ([raw file](https://github.com/maidsafe/routing/blob/b0b283d89cab2696a3eb996e1867ac6374ae4d48/src/core.rs#L1030)):*
> ```Rust
>                         self.stats.count_request(&request);
> 
>                         // TOTO: get response from the response cache.
> ```

Is that necessary? Maybe if we're the recipient anyway, we shouldn't even care for the cache: We can just raise the event as usual.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1059)*

madadam:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/core.rs, line 879 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKiMZOUVgPMI2Oy8XSZ:-KKlXoa_B9Urxn2RAqcH:-1281074240) ([raw file](https://github.com/maidsafe/routing/blob/b0b283d89cab2696a3eb996e1867ac6374ae4d48/src/core.rs#L879)):*
> This probably needs to be `Authority::ManagedNode(our_name)`, as we are not responding as part of the group.
</details>
OK. What about when we are Client?

---

*[src/core.rs, line 1030 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKiN3EMGXfXIENNvbFZ:-KKlWudCw7cBx1XpC1fP:470762868) ([raw file](https://github.com/maidsafe/routing/blob/b0b283d89cab2696a3eb996e1867ac6374ae4d48/src/core.rs#L1030)):*
> Is that necessary? Maybe if we're the recipient anyway, we shouldn't even care for the cache: We can just raise the event as usual.
</details>
Nah, I just forgot to delete this comment.  Same for line 1040. Caching is handled elsewhere.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1059)*

afck:
Reviewed 2 of 4 files at r1, 5 of 5 files at r2, 1 of 1 files at r3.
Review status: 6 of 7 files reviewed at latest revision, 8 unresolved discussions.

---

*[src/core.rs, line 879 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKiMZOUVgPMI2Oy8XSZ:-KKn8dxJOVlqqw8w9q9O:-1896695975) ([raw file](https://github.com/maidsafe/routing/blob/b0b283d89cab2696a3eb996e1867ac6374ae4d48/src/core.rs#L879)):*
> OK. What about when we are Client?
</details>
As discussed, clients don't need to cache.

---

*[src/core.rs, line 442 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKn8H8kx6cZfCULLjBF:-KKn8H8lD8Di3vfJO7_G:-641195444) ([raw file](https://github.com/maidsafe/routing/blob/ccbb4ce9f03901de8d46f1e8a0b42221fff668ee/src/core.rs#L442)):*
> ```Rust
>                 if let UserMessage::Response(ref response) = content {
>                     if response.is_cacheable() {
>                         self.response_cache.put(response.clone());
> ```

I'm not sure whether it's even worth bothering the cache at this point, since we're the ones who have the data anyway.

In fact, we need to be careful not to undo the bandwidth savings from the group message deduplication: If every member of a group thinks it hit its cache instead of responding as a group member, it will send the full `GetSuccess` ...

---

*[src/core.rs, line 894 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKn9tFLuKNfCzt0G2yq:-KKn9tFLuKNfCzt0G2yr:1550413) ([raw file](https://github.com/maidsafe/routing/blob/ccbb4ce9f03901de8d46f1e8a0b42221fff668ee/src/core.rs#L894)):*
> ```Rust
>                                                     dst,
>                                                     UserMessage::Response(response),
>                                                     priority));
> ```

We also need to `send_ack`, so that the requester doesn't resend its request.

---

*[src/core_tests.rs, line 257 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKnBnxScFgPv8LKNvY8:-KKnBnxScFgPv8LKNvY9:406226604) ([raw file](https://github.com/maidsafe/routing/blob/ccbb4ce9f03901de8d46f1e8a0b42221fff668ee/src/core_tests.rs#L257)):*
> ```Rust
>     }
> 
>     fn unwrap(self) -> TestNode {
> ```

Not sure what would be the most idiomatic method names for the builder pattern, especially in Rust, but looking at the `DirBuilder` example, maybe `create` would be better here.
Maybe I'd also call the constructor for the builder `builder()` instead of `build()`. Or make it `TestNodeBuilder::new()` instead? Not sure.

---

*[src/core_tests.rs, line 1338 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKnCmHKiGHV6F0I6heq:-KKnCmHKiGHV6F0I6her:-1770583837) ([raw file](https://github.com/maidsafe/routing/blob/ccbb4ce9f03901de8d46f1e8a0b42221fff668ee/src/core_tests.rs#L1338)):*
> ```Rust
>     );
> 
>     // Second request should hit the cache, thus the nodes should not raise any
> ```

Maybe clarify: `the proxy node's cache`. I had to stop and think here for a moment why it's guaranteed to hit any node's cache. :-)

---

*[src/core_tests.rs, line 1345 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKnD1AqGXDkpcSr3QPa:-KKnD1AqGXDkpcSr3QPb:-251974984) ([raw file](https://github.com/maidsafe/routing/blob/ccbb4ce9f03901de8d46f1e8a0b42221fff668ee/src/core_tests.rs#L1345)):*
> ```Rust
>                              .send_get_request(dst, data_id, message_id));
> 
>     poll_all(&mut nodes, &mut clients);
> ```

To test the `Ack`s, we could `poll_and_resend` here and then verify that the client got only one response and that the cache was only hit once.

---

*[src/messages.rs, line 761 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKnDbktGeUGoxPVVqvq:-KKnDbktGeUGoxPVVqvr:-1156215147) ([raw file](https://github.com/maidsafe/routing/blob/ccbb4ce9f03901de8d46f1e8a0b42221fff668ee/src/messages.rs#L761)):*
> ```Rust
> }
> 
> /// This assembles UserMessages from UserMessageParts.
> ```

Clippy probably wants backticks here.

---

*[src/node.rs, line 75 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKnDww3TdKAxWOz2xA_:-KKnDww3TdKAxWOz2xAa:-2040450581) ([raw file](https://github.com/maidsafe/routing/blob/ccbb4ce9f03901de8d46f1e8a0b42221fff668ee/src/node.rs#L75)):*
> ```Rust
>     /// Create a new `Node` given a cache instance.
>     #[cfg(not(feature = "use-mock-crust"))]
>     pub fn with_cache(event_sender: Sender<Event>, first_node: bool, cache: Box<Cache>)
> ```

It might make the API slightly simpler to box the `Cache` inside and make `with_cache<T: Cache>()` generic instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1059)*

madadam:
Review status: 6 of 7 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/messages.rs, line 761 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKnDbktGeUGoxPVVqvq:-KKnGlw3HS-quV-3v4Zy:1758359711) ([raw file](https://github.com/maidsafe/routing/blob/ccbb4ce9f03901de8d46f1e8a0b42221fff668ee/src/messages.rs#L761)):*
> Clippy probably wants backticks here.
</details>
Fixed.

---

*[src/node.rs, line 75 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKnDww3TdKAxWOz2xA_:-KKnFFiHice5Z7wcZJx1:-1282587862) ([raw file](https://github.com/maidsafe/routing/blob/ccbb4ce9f03901de8d46f1e8a0b42221fff668ee/src/node.rs#L75)):*
> It might make the API slightly simpler to box the `Cache` inside and make `with_cache<T: Cache>()` generic instead.
</details>
I had it that way originally, but then changed it because it was actually more difficult to use in tests. I don't feel too strongly about it either way, so I wouldn't mind changing it back.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1059)*

afck:
Review status: 6 of 7 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/node.rs, line 75 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKnDww3TdKAxWOz2xA_:-KKnHXQiib7v0BSTqsbG:249872768) ([raw file](https://github.com/maidsafe/routing/blob/ccbb4ce9f03901de8d46f1e8a0b42221fff668ee/src/node.rs#L75)):*
> I had it that way originally, but then changed it because it was actually more difficult to use in tests. I don't feel too strongly about it either way, so I wouldn't mind changing it back.
</details>
Ah, okay, if it's easier to use this way in tests, let's keep it as it is. :thumbsup: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1059)*

afck:
Reviewed 1 of 1 files at r4, 1 of 2 files at r5, 1 of 1 files at r6.

---

*[src/core_tests.rs, line 1332 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKraiJ3gO06HgLKz5DV:-KKraiJ3gO06HgLKz5DW:1729589471) ([raw file](https://github.com/maidsafe/routing/blob/4046081ee6320fdfb67ea821255e19fed6becc3e/src/core_tests.rs#L1332)):*
> ```Rust
>         clients[0],
>         Event::Response {
>             response: Response::GetSuccess(ref res_data, res_message_id), ..
> ```

Here you could match `src: Authority::NaeManager(name)` and assert that that's the data name ...

---

*[src/core_tests.rs, line 1341 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKrb1hz4R2ZwOiV9cfE:-KKrb1i-twZL59sYx12X:-349536894) ([raw file](https://github.com/maidsafe/routing/blob/4046081ee6320fdfb67ea821255e19fed6becc3e/src/core_tests.rs#L1341)):*
> ```Rust
>     let message_id = MessageId::new();
> 
>     // Cache hit.
> ```

Maybe explain why this is expected, e.g.: `Cache hit: the proxy node should have cached the data`

---

*[src/core_tests.rs, line 1350 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKrbbf7lmwHm0vTtjXk:-KKrbbf7lmwHm0vTtjXl:-660825097) ([raw file](https://github.com/maidsafe/routing/blob/4046081ee6320fdfb67ea821255e19fed6becc3e/src/core_tests.rs#L1350)):*
> ```Rust
>         clients[0],
>         Event::Response {
>             response: Response::GetSuccess(ref res_data, res_message_id), ..
> ```

... and here you could expect `src: Authority::ManagedNode(name)` with the proxy node's name.

---

*[src/core_tests.rs, line 1355 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1059#-KKrbqcWDf-7oaodmjRl:-KKrbqcWDf-7oaodmjRm:328093505) ([raw file](https://github.com/maidsafe/routing/blob/4046081ee6320fdfb67ea821255e19fed6becc3e/src/core_tests.rs#L1355)):*
> ```Rust
> 
>     // Make sure we received the response only once.
>     expect_no_event!(clients[0]);
> ```

At this point, the client should have received the `Ack`. We should `poll_and_resend` again here and _then_ assert that it hasn't resent the request.

As discussed, however, the client might not actually raise another event even if it _did_ resend, because of the filters.
Is it feasible to assert in mock Crust that it didn't resend the request?

Alternatively, the client could ask its `Core` at this point whether there are unacknowledged messages left and assert that there aren't.

In either case, it would be good to verify that the test actually _does_ fail without the `send_ack` above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1059)*

afck:
Reviewed 5 of 5 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1059)*

PR #1062, opened 2016-06-24
feat/core: send Terminate if there is a beacon on the network

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1062)


maidsafe-highfive:

PR #1063, opened 2016-06-24
feat/mock_crust: function to find peers on LAN

ustulation:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1063)


maidsafe-highfive:
r? @afck

PR #1066, opened 2016-06-27
fix/core_tests: fix node sorting

afck:
Sorting nodes by name causes the `name()` method to be called, which
causes mock Crust event polling. Sorting should therefore only be done
when there are no events to process.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1066)


maidsafe-highfive:

Viv-Rajkumar:
Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1066)*

PR #1069, opened 2016-06-28
fix/core: disconnect bootstrap node on rebootstrap

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1069)


maidsafe-highfive:

PR #1076, opened 2016-07-05
fix/tests: re-enable ci tests, and fix a peer manager bug

afck:
This enables the ci test and makes the "ci_test" example work again by
making the one-node-per-local-network limit configurable when
constructing a node.

Also fixes a bug in `PeerManager` that caused valid connection info tokens to
be considered expired.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1076)


Viv-Rajkumar:
Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1076)*

PR #1079, opened 2016-07-06
chore/peer_manager: move routing_table from core into peer_manager

maqi:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1079)

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -27,11 +27,15 @@ use mock_crust::crust::{self, ConnectionInfoResult, CrustError, PeerId, PrivConn
 
 use itertools::Itertools;
 use kademlia_routing_table::{AddedNodeDetails, ContactInfo, DroppedNodeDetails};
+#[cfg(feature = "use-mock-crust")]
+use kademlia_routing_table::RoutingTable;
 use lru_time_cache::LruCache;
 use maidsafe_utilities::{self, serialisation};
 use maidsafe_utilities::event_sender::MaidSafeEventCategory;
 use message_filter::MessageFilter;
-use peer_manager::{ConnectionInfoPreparedResult, ConnectionInfoReceivedResult, PeerManager};
+use peer_manager::{ConnectionInfoPreparedResult, ConnectionInfoReceivedResult, NodeInfo,
+                   PeerManager};
+pub use peer_manager::GROUP_SIZE;
 use sodiumoxide::crypto::{box_, sign};
 use sodiumoxide::crypto::hash::sha256;
 use std::{cmp, fmt, iter};
@@ -56,13 +60,8 @@ use messages::{DEFAULT_PRIORITY, DirectMessage, HopMessage, Message, MessageCont
                RoutingMessage, SignedMessage, UserMessage, UserMessageCache};
 use utils;
 
-/// The group size for the routing table. This is the maximum that can be used for consensus.
-pub const GROUP_SIZE: usize = 8;
 /// The quorum for group consensus.
 pub const QUORUM_SIZE: usize = 5;
-/// The number of entries beyond `GROUP_SIZE` that are not considered unnecessary in the routing
-/// table.
-const EXTRA_BUCKET_ENTRIES: usize = 2;
 /// Time (in seconds) after which bootstrap is cancelled (and possibly retried).
 const BOOTSTRAP_TIMEOUT_SECS: u64 = 20;
 /// Time (in seconds) after which a `GetNodeName` request is resent.
@@ -88,33 +87,6 @@ enum State {
     Client,
     /// We have been Relocated and now a node.
     Node,
-}
-
-/// `RoutingTable` managing `NodeInfo`s.
-pub type RoutingTable = ::kademlia_routing_table::RoutingTable<NodeInfo>;
-
-/// Info about nodes in the routing table.
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub struct NodeInfo {
-    public_id: PublicId,
-    peer_id: PeerId,
-}
-
-impl NodeInfo {
-    fn new(public_id: PublicId, peer_id: PeerId) -> Self {
-        NodeInfo {
-            public_id: public_id,
-            peer_id: peer_id,
-        }
-    }
-}
-
-impl ContactInfo for NodeInfo {
-    type Name = XorName;
-
-    fn name(&self) -> &XorName {
-        self.public_id.name()
-    }
 }
 
 /// The role this `Core` instance intends to act as once it joined the network.
@@ -204,7 +176,6 @@ pub struct Core {
     grp_msg_filter: MessageFilter<RoutingMessage>,
     full_id: FullId,
     state: State,
-    routing_table: RoutingTable,
     get_node_name_timer_token: Option<u64>,
     bucket_refresh_token_and_delay: Option<(u64, u64)>,
     /// The last joining node we have sent a `GetNodeName` response to, and when.
@@ -276,7 +247,6 @@ impl Core {
             grp_msg_filter: MessageFilter::with_expiry_duration(Duration::from_secs(60 * 20)),
             full_id: full_id,
             state: State::Disconnected,
-            routing_table: RoutingTable::new(our_info, GROUP_SIZE, EXTRA_BUCKET_ENTRIES),
             get_node_name_timer_token: None,
             bucket_refresh_token_and_delay: None,
             sent_network_name_to: None,
@@ -287,7 +257,7 @@ impl Core {
             user_msg_cache: UserMessageCache::with_expiry_duration(user_msg_cache_duration),
             cacheable_user_msg_cache:
                 UserMessageCache::with_expiry_duration(user_msg_cache_duration),
-            peer_mgr: Default::default(),
+            peer_mgr: PeerManager::new(our_info),
             bootstrap_blacklist: HashSet::new(),
             response_cache: cache,
         };
@@ -345,9 +315,9 @@ impl Core {
     }
 
     /// Routing table of this node.
-    #[allow(unused)]
-    pub fn routing_table(&self) -> &RoutingTable {
-        &self.routing_table
+    #[cfg(feature = "use-mock-crust")]
+    pub fn routing_table(&self) -> &RoutingTable<NodeInfo> {
+        self.peer_mgr.routing_table()
     }
 
     /// resends all unacknowledged messages.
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @afck

afck:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/core.rs, line 316 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KLzTA0RJHa4KoeXhnLo:-KLzTA0SIWxZYlhSMamA:98455535) ([raw file](https://github.com/maidsafe/routing/blob/ae370d51c40de173221e710f6c4cad8f68c061a4/src/core.rs#L316)):*
> ```Rust
> 
>     /// Routing table of this node.
>     #[allow(unused)]
> ```

Is this really unused? Or is it used only by tests and mock crust? It would be better to have something like `cfg(any(test, features="use-mock-crust"))`.
Also, if it _is_ used, would it make sense to return a `RoutingTable<XorName>` instead of a `RoutingTable<NodeInfo>`?

---

*[src/peer_manager.rs, line 221 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KLzebmRuZG6vjT45HaD:-KLzebmRuZG6vjT45HaE:-2017915142) ([raw file](https://github.com/maidsafe/routing/blob/ae370d51c40de173221e710f6c4cad8f68c061a4/src/peer_manager.rs#L221)):*
> ```Rust
>     }
> 
>     pub fn close_nodes(&self, name: &XorName) -> Option<Vec<NodeInfo>> {
> ```

Since the `routing_table()` getter exists, I wouldn't wrap functions that don't mutate the routing table. I'd remove all wrappers that just call through to the non-`&mut self` `RoutingTable` method of the same name.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1079)*

Fraser999:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/core.rs, line 316 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KLzTA0RJHa4KoeXhnLo:-KLzhMIfvdNI85O-7hov:-2117940137) ([raw file](https://github.com/maidsafe/routing/blob/ae370d51c40de173221e710f6c4cad8f68c061a4/src/core.rs#L316)):*
> Is this really unused? Or is it used only by tests and mock crust? It would be better to have something like `cfg(any(test, features="use-mock-crust"))`.
> Also, if it _is_ used, would it make sense to return a `RoutingTable<XorName>` instead of a `RoutingTable<NodeInfo>`?
</details>
I think we can get away with just `#[cfg(feature = "use-mock-crust")]`

---

*[src/peer_manager.rs, line 52 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KLziwVcTIZzI0yB_WnK:-KLziwVcTIZzI0yB_WnL:-1640163608) ([raw file](https://github.com/maidsafe/routing/blob/ae370d51c40de173221e710f6c4cad8f68c061a4/src/peer_manager.rs#L52)):*
> ```Rust
> 
> /// `RoutingTable` managing `XorName`s.
> pub type RoutingTableNames = ::kademlia_routing_table::RoutingTable<XorName>;
> ```

Not sure that we win much with this alias, but I don't feel strongly about removing or keeping it.

---

*[src/peer_manager.rs, line 221 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KLzebmRuZG6vjT45HaD:-KLzjDpG9dYntkzDPdy4:-774006928) ([raw file](https://github.com/maidsafe/routing/blob/ae370d51c40de173221e710f6c4cad8f68c061a4/src/peer_manager.rs#L221)):*
> Since the `routing_table()` getter exists, I wouldn't wrap functions that don't mutate the routing table. I'd remove all wrappers that just call through to the non-`&mut self` `RoutingTable` method of the same name.
</details>
I agree

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1079)*

maqi:
*[src/core.rs, line 316 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KLzTA0RJHa4KoeXhnLo:-KLzsaZ21z3pcoFU5kkL:2000768809) ([raw file](https://github.com/maidsafe/routing/blob/ae370d51c40de173221e710f6c4cad8f68c061a4/src/core.rs#L316)):*
> I think we can get away with just `#[cfg(feature = "use-mock-crust")]`
</details>
most of these get wrappers are called multiple times in core.rs , and got some lines reduced.
I do left two `just called once` not being wrapped and using routing_table() for access directly.
Also, we don't know what will be the next step to refactor peer_manager further, so maybe a wrapper can prevent a neglectance  of outside access?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1079)*

Fraser999:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/peer_manager.rs, line 221 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KLzebmRuZG6vjT45HaD:-KLzvUAENZMBrv4Xh9q6:-523335197) ([raw file](https://github.com/maidsafe/routing/blob/ae370d51c40de173221e710f6c4cad8f68c061a4/src/peer_manager.rs#L221)):*
@maqi wrote:
> most of these get wrappers are called multiple times in core.rs , and got some lines reduced.
I do left two just called once not being wrapped and using routing_table() for access directly.
Also, we don't know what will be the next step to refactor peer_manager further, so maybe a wrapper can prevent a neglectance of outside access?

I guess your comment is really in reply to this thread?

Anyway - it's the single-line functions that really seem like overkill.  E.g. I'd leave the `close_group()` function as is (or I'd make that a function in core.rs) and the same for `is_tunnel()` maybe, but I don't think the others don't help much in code-reduction or clarity.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1079)*

afck:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/peer_manager.rs, line 52 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KLziwVcTIZzI0yB_WnK:-KM-0sdN-VoCejjS3ftT:1981367359) ([raw file](https://github.com/maidsafe/routing/blob/ae370d51c40de173221e710f6c4cad8f68c061a4/src/peer_manager.rs#L52)):*
> Not sure that we win much with this alias, but I don't feel strongly about removing or keeping it.
</details>
I'd be in favour of removing it, especially since we are now using `RoutingTable<XorName>` in some places.

---

*[src/peer_manager.rs, line 221 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KLzebmRuZG6vjT45HaD:-KM-11XTfm1BS-RUSt06:-585582428) ([raw file](https://github.com/maidsafe/routing/blob/ae370d51c40de173221e710f6c4cad8f68c061a4/src/peer_manager.rs#L221)):*
> @maqi wrote:
> > most of these get wrappers are called multiple times in core.rs , and got some lines reduced.
> I do left two just called once not being wrapped and using routing_table() for access directly.
> Also, we don't know what will be the next step to refactor peer_manager further, so maybe a wrapper can prevent a neglectance of outside access?
> 
> I guess your comment is really in reply to this thread?
> 
> Anyway - it's the single-line functions that really seem like overkill.  E.g. I'd leave the `close_group()` function as is (or I'd make that a function in core.rs) and the same for `is_tunnel()` maybe, but I don't think the others don't help much in code-reduction or clarity.
</details>
I also don't think a wrapper offers much protection if the `routing_table()` getter is there anyway.
Which is fine!
I think the idea that you can get an _immutable_ reference to the routing table, but that you can only mutate it via the `PeerManager`'s methods, makes sense.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1079)*

afck:
Reviewed 2 of 2 files at r3.

---

*[src/peer_manager.rs, line 210 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KM-UxT7FJRb4u7Xstv1:-KM-UxT8t9w4Fw5g7BsS:1680277922) ([raw file](https://github.com/maidsafe/routing/blob/033fd93a54b4c27435d7039cdad3b40d78d01029/src/peer_manager.rs#L210)):*
> ```Rust
>     }
> 
>     pub fn set_routing_table(&mut self, our_info: NodeInfo) {
> ```

Maybe `reset_routing_table` or `init_routing_table` would be a better name. (`set` suggests that we would pass in the routing table.)

---

*[src/peer_manager.rs, line 350 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KM-VCSPusFgtLFrWr7N:-KM-VCSQWx_NlndTzhPT:-273744449) ([raw file](https://github.com/maidsafe/routing/blob/033fd93a54b4c27435d7039cdad3b40d78d01029/src/peer_manager.rs#L350)):*
> ```Rust
>     }
> 
>     /// Marks the given peer as "Tunnelling to".
> ```

Ah, good catch! :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1079)*

maqi:
*[src/peer_manager.rs, line 52 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KLziwVcTIZzI0yB_WnK:-KM3Q_OzycwL_k7Q20ZX:999000130) ([raw file](https://github.com/maidsafe/routing/blob/ae370d51c40de173221e710f6c4cad8f68c061a4/src/peer_manager.rs#L52)):*
> I'd be in favour of removing it, especially since we are now using `RoutingTable<XorName>` in some places.
</details>
RoutingTableNames has been removed.
but core.rs prefers a RoutingTable alias for mock_crust, so better keep this one.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1079)*

afck:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/peer_manager.rs, line 52 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KLziwVcTIZzI0yB_WnK:-KM3W8QbrRaLuG_-37Cc:-1340866976) ([raw file](https://github.com/maidsafe/routing/blob/ae370d51c40de173221e710f6c4cad8f68c061a4/src/peer_manager.rs#L52)):*
> RoutingTableNames has been removed.
> but core.rs prefers a RoutingTable alias for mock_crust, so better keep this one.
</details>
Couldn't core.rs just use `RoutingTable<NodeInfo>` explicitly?

---

*[src/peer_manager.rs, line 229 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1079#-KM3TpH20iNYcJ6NIIkr:-KM3TpH3IFZTXlQIvJ2q:1550020410) ([raw file](https://github.com/maidsafe/routing/blob/9a1d34c5f9956205a106864508ada067df77d598/src/peer_manager.rs#L229)):*
> ```Rust
>     }
> 
>     pub fn get_peer(&self, peer_id: &PeerId) -> Option<&NodeInfo> {
> ```

These two method names are a bit confusing, as we're not looking into the peer map but into the routing table. Should probably be something like `get_routing_peer` or `get_routing_table_peer`?

(I'd even prefer removing these entirely and calling `peer_manager.routing_table().iter().find...` - your call, though.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1079)*

afck:
Reviewed 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1079)*

afck:
Reviewed 2 of 2 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1079)*

Fraser999:
Reviewed 2 of 2 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1079)*

PR #1080, opened 2016-07-06
Refactoring - split Core into Client and Node states

madadam:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1080)


maidsafe-highfive:
r? @afck

afck:
Reviewed 2 of 13 files at r1, 14 of 19 files at r3.
Review status: 16 of 20 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

afck:
Reviewed 3 of 19 files at r3.
Review status: 19 of 20 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/state_machine.rs, line 242 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcIj69q9RmNA5pVYHZ:-KMcIj6ARM9fy-Xnj_AG:-2001097048) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/state_machine.rs#L242)):*
> ```Rust
>                           dst: Authority) {
>         if let State::Node(_) = self.state {
>             unreachable!();
> ```

Is that special case necessary? There's already an `unreachable!()` in `into_node`.

---

*[src/state_machine.rs, line 247 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcJ-GdqDjfAkbjC5g8:-KMcJ-GdqDjfAkbjC5g9:817341391) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/state_machine.rs#L247)):*
> ```Rust
>         let prev_state = mem::replace(&mut self.state, State::Transitioning);
>         let next_state = prev_state.into_node(close_group_ids, dst);
>         let _ = mem::replace(&mut self.state, next_state);
> ```

This could probably just be a regular assignment now, since we drop the `Transitioning` value anyway: `self.state = prev_state.into_node(close_group_ids, dst);`

---

*[src/state_machine.rs, line 251 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcJJARQA_1i0o1WthM:-KMcJJARQA_1i0o1WthN:1294505008) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/state_machine.rs#L251)):*
> ```Rust
> 
>     fn terminate(&mut self) {
>         let _ = mem::replace(&mut self.state, State::Terminated);
> ```

Could also just be `self.state = State::Terminated`.

---

*[src/states/client.rs, line 62 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcJcCBUoCtWThGkZNP:-KMcJcCCODQd4eW6ez4F:-1041843542) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L62)):*
> ```Rust
>     bootstrap_blacklist: HashSet<SocketAddr>,
>     bootstrap_state: BootstrapState,
>     cache: Box<Cache>,
> ```

This is only to not forget the `Cache` instance until we become a node, is it?
I wonder whether this should go into `StateMachine` instead then?

---

*[src/states/client.rs, line 160 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcKTG45ZEQN0SSMh3o:-KMcKTG45ZEQN0SSMh3p:-1471216429) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L160)):*
> ```Rust
>                 // extend the MessageAccumulator's API with a method to retrieve it.
>                 result_tx.send(0).is_ok()
>             }
> ```

I wonder whether we are actually using that `Action` anywhere.
In any case, `Action` will probably need to be split up into `NodeAction` and `ClientAction`. (In a later PR, I guess.)

---

*[src/states/client.rs, line 226 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcL0BNVZJOZtccIwm1:-KMcL0BO2WTnaRob8q7E:-979544174) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L226)):*
> ```Rust
>     pub fn routing_table(&self) -> &RoutingTable<NodeInfo> {
>         self.peer_mgr.routing_table()
>     }
> ```

That would always be empty, wouldn't it?
I wonder whether `Client` needs a `PeerManager` at all?

---

*[src/states/client.rs, line 228 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcLbJL-WrlDUv6E52f:-KMcLbJL-WrlDUv6E52g:1758193401) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L228)):*
> ```Rust
>     }
> 
>     /// resends all unacknowledged messages.
> ```

`Resends` uppercase.

---

*[src/states/client.rs, line 314 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcM0yg1B8G536go4aa:-KMcM0yg1B8G536go4ab:661804972) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L314)):*
> ```Rust
>                     Err(RoutingError::InvalidDestination)
>                 }
>             }
> ```

A client can't receive `TunnelDirect` or `TunnelHop` messages - 16 more lines to cut down! :-)

---

*[src/states/client.rs, line 694 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcOLXbne5GzHH_2Geo:-KMcOLXbne5GzHH_2Gep:-491736882) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L694)):*
> ```Rust
>                                                   &self.full_id));
>             (peer_id, bytes)
>         } else if let Some(&tunnel_id) = self.tunnels.tunnel_for(&peer_id) {
> ```

This branch can also go: We are always directly connected to our proxy node. Tunnels shouldn't concern clients at all.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

maqi:
*[src/states/mod.rs, line 23 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcUx1bdTKqZz2oWQHo:-KMcUx1bdTKqZz2oWQHp:2055092715) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/mod.rs#L23)):*
> ```Rust
> 
> pub use self::client::Client;
> pub use self::node::Node;
> ```

requires a `line end feed` here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

maqi:
*[src/states/client.rs, line 722 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcVSEmizQc7kKvs8s9:-KMcVSEnB2JbVyIu7Ex6:1852574758) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L722)):*
> ```Rust
>     }
> 
>     fn send_ack(&mut self, routing_msg: &RoutingMessage, route: u8) {
> ```

does client needs to `ack` a message on routing level?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

maqi:
*[src/states/client.rs, line 662 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcVnwhU7xmv9IltLvQ:-KMcVnwiOa96dpGRfNyF:-131477191) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L662)):*
> ```Rust
>         if let Authority::Client { .. } = routing_msg.dst {
>             if self.is_recipient(&routing_msg.dst) {
>                 return Ok(()); // Message is for us.
> ```

i am not sure, but `client sends a msg to himself` sounds weird

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

maqi:
*[src/states/client.rs, line 659 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcW-dWlWh0KJPax9Ee:-KMcW-dWlWh0KJPax9Ef:-1763825447) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L659)):*
> ```Rust
>     }
> 
>     fn send_routing_message_via_route(&mut self, routing_msg: RoutingMessage, route: u8) -> Result<(), RoutingError> {
> ```

does client only sends msg through its proxy?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

maqi:
*[src/states/client.rs, line 342 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcWcjKKrKjuk-64XMA:-KMcWcjLs-C9cFOwiCrQ:2062099608) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L342)):*
> ```Rust
>     }
> 
>     fn handle_hop_message(&mut self,
> ```

as client won't forwarding any msg, so this `hop_message` actually means `new_message` to client?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

maqi:
*[src/states/client.rs, line 342 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcWcjKKrKjuk-64XMA:-KMcWyeI7eZEI8S1-YZV:-1341861863) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L342)):*
> as client won't forwarding any msg, so this `hop_message` actually means `new_message` to client?
</details>
and then, in which scenario, will receive this `hop_message` but not `direct_message`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

madadam:
Review status: 19 of 20 files reviewed at latest revision, 13 unresolved discussions.

---

*[src/state_machine.rs, line 247 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcJ-GdqDjfAkbjC5g8:-KMcShTyyJTBcmP_wU68:-2097291628) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/state_machine.rs#L247)):*
> This could probably just be a regular assignment now, since we drop the `Transitioning` value anyway: `self.state = prev_state.into_node(close_group_ids, dst);`
</details>
Unfortunately not, because we would be moving out of `self.state`, which means moving out of `self`, but  we can't do that because `self` is borrowed.

---

*[src/states/client.rs, line 62 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcJcCBUoCtWThGkZNP:-KMcTaj1npqwb9Mfc5iJ:-1145181977) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L62)):*
> This is only to not forget the `Cache` instance until we become a node, is it?
> I wonder whether this should go into `StateMachine` instead then?
</details>
Yes, it's just to pass it to `Node` when we transition to it. I can move it to `StateMachine` instead, but then I'd have to keep it in an `Option` to be able to move it to `Node` when we switch to it. Not sure which of these two possibilities is uglier :)

---

*[src/states/client.rs, line 160 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcKTG45ZEQN0SSMh3o:-KMcUHrNOiSUgXtup9iP:1490501607) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L160)):*
> I wonder whether we are actually using that `Action` anywhere.
> In any case, `Action` will probably need to be split up into `NodeAction` and `ClientAction`. (In a later PR, I guess.)
</details>
Yes, would be good to check whether it's actually used (in vault, safe_core, ...). If it is, returning 0 might cause problems.

---

*[src/states/client.rs, line 226 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcL0BNVZJOZtccIwm1:-KMcUYN6BfRhzN5RMVzq:930078361) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L226)):*
> That would always be empty, wouldn't it?
> I wonder whether `Client` needs a `PeerManager` at all?
</details>
Client doesn't use routing table at all, but returning an empty one here makes the `#[cfg(feature = "use-mock-crust")]`-only api much simpler. An alternative would be to return `Option<&RoutingTable<...>>` and handle the `None` case everywhere.

As for `PeerManager`, the client uses it to manage the proxy node connection.

---

*[src/states/client.rs, line 314 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcM0yg1B8G536go4aa:-KMcVEAH9YGHb--BJ9mk:-129443389) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L314)):*
> A client can't receive `TunnelDirect` or `TunnelHop` messages - 16 more lines to cut down! :-)
</details>
That's good to know. I love deleting code! :)

---

*[src/states/client.rs, line 342 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcWcjKKrKjuk-64XMA:-KMcX56LKjA3j5WO0tjo:-172580125) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L342)):*
> and then, in which scenario, will receive this `hop_message` but not `direct_message`?
</details>
A client does not forward hop messages, but can still send and receive them (the get/put/post/delete requests/responses are hop messages, aren't they?)

---

*[src/states/client.rs, line 659 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcW-dWlWh0KJPax9Ee:-KMcWQ88nzlWNAKj6aI3:-889088566) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L659)):*
> does client only sends msg through its proxy?
</details>
I believe so.

---

*[src/states/client.rs, line 662 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcVnwhU7xmv9IltLvQ:-KMcWDnM1vWiZQ04wB_Y:-691286530) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L662)):*
> i am not sure, but `client sends a msg to himself` sounds weird
</details>
I agree, but is it guaranteed that I can never happen?  Or that it wouldn't case any issues if it does happen but is not handled here?

---

*[src/states/client.rs, line 722 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcVSEmizQc7kKvs8s9:-KMcXPPJrYSjSKfnpMms:-624890241) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L722)):*
> does client needs to `ack` a message on routing level?
</details>
I'd say it does, for example when it receives a reponse to a request it sent, it need to ack that response so it won't be sent again. Not 100% sure about this though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

maqi:
*[src/states/client.rs, line 662 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcVnwhU7xmv9IltLvQ:-KMcZqPqFYgqcLZkAouO:-2072194484) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L662)):*
> I agree, but is it guaranteed that I can never happen?  Or that it wouldn't case any issues if it does happen but is not handled here?
</details>
I think if it happens, it means some bug in the code. :)
better have a try to remove this part had see what's happening

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

maqi:
*[src/states/client.rs, line 659 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcW-dWlWh0KJPax9Ee:-KMc_MLrCGFShnS_lpra:1628446610) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L659)):*
> I believe so.
</details>
then, can i say this `via_route` option is not required for `client`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

afck:
Reviewed 1 of 13 files at r1, 1 of 19 files at r3.
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[src/state_machine.rs, line 247 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcJ-GdqDjfAkbjC5g8:-KMcb4QrGwzndMrsoo9X:1672156220) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/state_machine.rs#L247)):*
> Unfortunately not, because we would be moving out of `self.state`, which means moving out of `self`, but  we can't do that because `self` is borrowed.
</details>
Yes, that's why the _first_ `mem::replace` is necessary. But the second one could just be an assignment, couldn't it?

---

*[src/states/client.rs, line 62 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcJcCBUoCtWThGkZNP:-KMcbE0xEBGzF0rO1Uq2:452532321) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L62)):*
> Yes, it's just to pass it to `Node` when we transition to it. I can move it to `StateMachine` instead, but then I'd have to keep it in an `Option` to be able to move it to `Node` when we switch to it. Not sure which of these two possibilities is uglier :)
</details>
Ugh. Agreed, let's leave it as is for now. :+1: 

---

*[src/states/client.rs, line 226 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcL0BNVZJOZtccIwm1:-KMcbSk_xOJyael8LMjw:2115073362) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L226)):*
> Client doesn't use routing table at all, but returning an empty one here makes the `#[cfg(feature = "use-mock-crust")]`-only api much simpler. An alternative would be to return `Option<&RoutingTable<...>>` and handle the `None` case everywhere.
> 
> As for `PeerManager`, the client uses it to manage the proxy node connection.
</details>
Hmm ... I guess we just need to reevaluate how and whether the peer manager should handle proxy node. But definitely in a separate PR. :+1: 

---

*[src/states/client.rs, line 659 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcW-dWlWh0KJPax9Ee:-KMcbiVPbZt28GiymAQz:2009459540) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L659)):*
> then, can i say this `via_route` option is not required for `client`?
</details>
The `route` is still required: The routes are still different, although all of them go through the proxy node.

---

*[src/states/node.rs, line 310 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcTQauJBbthmH8-K9f:-KMcTQav-n8v_Sf6ICxX:1161280156) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L310)):*
> ```Rust
>         let result = match action {
>             Action::ClientSendRequest { result_tx, .. } => {
>                 result_tx.send(Err(InterfaceError::InvalidState)).is_ok()
> ```

That should be unreachable, I think. (Same for `NodeSendMessage` in `Client`.)

---

*[src/states/node.rs, line 384 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcXzrAHakfCEygKFK0:-KMcXzrBOo6OTYQyiMHC:1791762431) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L384)):*
> ```Rust
>     fn handle_listener_started(&mut self, port: u16) {
>         // We are the first node.
>         self.is_first_node = true;
> ```

I'm confused: Is this because - due to the anti-beacon - currently only the first node ever starts listening?
And don't we know in the constructor already whether we are the first node? Why not set this there, or in `start_new_network`?

---

*[src/states/node.rs, line 716 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcZCfCgnWuyCpky57b:-KMcZCfDcYj3zzWsco2z:-105513815) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L716)):*
> ```Rust
>                               public_id: PublicId)
>                               -> Result<(), RoutingError>
>     {
> ```

Is this really what rustfmt does by default? (Or at least what it doesn't undo by default?)

---

*[src/states/node.rs, line 1787 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcaRwh5nAA4q7wK7Ux:-KMcaRwh5nAA4q7wK7Uy:596548028) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L1787)):*
> ```Rust
>             // We don't have any contacts in our routing table yet. Keep using
>             // the proxy connection until we do.
>             if let Authority::Client { ref proxy_node_name, .. } = routing_msg.src {
> ```

Why not collapse the `} else if`, as Clippy suggests?

---

*[src/states/node.rs, line 1903 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcavvI2JZgjA39zBG3:-KMcavvI2JZgjA39zBG4:597346426) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L1903)):*
> ```Rust
>                            e);
>                 }
>                 self.reset_bucket_refresh_timer();
> ```

Good catch! :-)

---

*[src/states/node.rs, line 2026 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMc_uTdxrEHy5O53Y7d:-KMc_uTdxrEHy5O53Y7e:-1153798301) ([raw file](https://github.com/maidsafe/routing/blob/1d6d0b323c9565807ce538c844f2a746642ea0f5/src/states/node.rs#L2026)):*
> ```Rust
>                 return self.relay_to_client(signed_msg.clone(), peer_id);
>             } else if self.is_recipient(&routing_msg.dst) {
>                 return Ok(()); // Message is for us as a client.
> ```

We still need to handle messages that were sent to us before we transitioned to `Node`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

madadam:
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---

*[src/state_machine.rs, line 247 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcJ-GdqDjfAkbjC5g8:-KMccoBgBu5HZFm5R8VC:-818361) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/state_machine.rs#L247)):*
> Yes, that's why the _first_ `mem::replace` is necessary. But the second one could just be an assignment, couldn't it?
</details>
I see what you mean now! I'll change it.

---

*[src/states/client.rs, line 662 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcVnwhU7xmv9IltLvQ:-KMch_2YBX7I02SkQGrX:-499811621) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L662)):*
> I think if it happens, it means some bug in the code. :)
> better have a try to remove this part had see what's happening
</details>
I'll remove it if Andreas approves it :)

---

*[src/states/node.rs, line 310 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcTQauJBbthmH8-K9f:-KMcdu-LyJvmKjbmLUrO:841811051) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L310)):*
> That should be unreachable, I think. (Same for `NodeSendMessage` in `Client`.)
</details>
This is to prevent the sender of the action to get stuck trying to receive from the result channel.

---

*[src/states/node.rs, line 384 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcXzrAHakfCEygKFK0:-KMcenn9Xn40T3JHGuU_:-1537825895) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L384)):*
> I'm confused: Is this because - due to the anti-beacon - currently only the first node ever starts listening?
> And don't we know in the constructor already whether we are the first node? Why not set this there, or in `start_new_network`?
</details>
The `handle_listener_started` gets called only for the first node because regular nodes handle it while still in the client state. It has nothing to do with the anti-beacon thing. It's not very clear from the code though. Maybe I should add a comment explaining it? 

That said, the `self.is_first_node = true` assignment can probably go to the constructor (or to `start_new_network`).

---

*[src/states/node.rs, line 716 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcZCfCgnWuyCpky57b:-KMcfdstzxtaQRh-g3Kq:526365186) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L716)):*
> Is this really what rustfmt does by default? (Or at least what it doesn't undo by default?)
</details>
I haven't rustfmt'd the code yet. But just to make sure, are you referring to the opening curly brace being on its own line? I actually prefer it that way when the argument list spans multiple lines, because I think it makes it easier to read. However, I'm not going to fight rustfmt on this.

---

*[src/states/node.rs, line 1787 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcaRwh5nAA4q7wK7Ux:-KMcg_D3Bks6Ftdk9GEZ:1920822033) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L1787)):*
> Why not collapse the `} else if`, as Clippy suggests?
</details>
I think those branches don't semantically belong to the same level. Don't feel strongly about it though.

---

*[src/states/node.rs, line 1903 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcavvI2JZgjA39zBG3:-KMch9e_Wfo_wFZAJIhB:-1361267807) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L1903)):*
> Good catch! :-)
</details>
Not sure what you mean by good catch, but I think this is actually an error - `reset_bucket_refresh_timer` is called again at the end of this function, so I don't think it's necessary to call it here. I probably missed this during the monster rebase. Or am I missing something?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

afck:
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[src/states/client.rs, line 662 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcVnwhU7xmv9IltLvQ:-KMciQKGPlYtnu8uYDW3:-764236720) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L662)):*
> I'll remove it if Andreas approves it :)
</details>
I'd actually opt for keeping it for now. Let's try to make as few behavioral changes as possible in this refactoring. We can try these out in separate PRs later. But let's keep this one in mind!

---

*[src/states/client.rs, line 722 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcVSEmizQc7kKvs8s9:-KMcii8FawJFCZpCacTn:1576595059) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L722)):*
> I'd say it does, for example when it receives a reponse to a request it sent, it need to ack that response so it won't be sent again. Not 100% sure about this though.
</details>
I think it does, yes. We were considering doing this in the proxy node instead, and we should revisit that option later. (Removing all the ack stuff from `Client` would be great!) But for now the client has to ack. :-(

---

*[src/states/node.rs, line 310 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcTQauJBbthmH8-K9f:-KMcj4w9aXF8iWpwlNfa:481703679) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L310)):*
> This is to prevent the sender of the action to get stuck trying to receive from the result channel.
</details>
Ah, sorry, makes sense!

---

*[src/states/node.rs, line 384 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcXzrAHakfCEygKFK0:-KMcjDxbsdGjA2cUlKLh:-1904194703) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L384)):*
> The `handle_listener_started` gets called only for the first node because regular nodes handle it while still in the client state. It has nothing to do with the anti-beacon thing. It's not very clear from the code though. Maybe I should add a comment explaining it? 
> 
> That said, the `self.is_first_node = true` assignment can probably go to the constructor (or to `start_new_network`).
</details>
OK. Is it guaranteed that we don't transition before we handled `ListenerStarted`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

afck:
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[src/states/node.rs, line 716 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcZCfCgnWuyCpky57b:-KMclAeS4gs03y4mBtK3:716712298) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L716)):*
> I haven't rustfmt'd the code yet. But just to make sure, are you referring to the opening curly brace being on its own line? I actually prefer it that way when the argument list spans multiple lines, because I think it makes it easier to read. However, I'm not going to fight rustfmt on this.
</details>
Ah, okay. If rustfmt is fine with either style, I'm too. But otherwise, rustfmt is our style consensus in general.

---

*[src/states/node.rs, line 1903 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcavvI2JZgjA39zBG3:-KMclRHyJdGjOAPGeX5D:945988284) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L1903)):*
> Not sure what you mean by good catch, but I think this is actually an error - `reset_bucket_refresh_timer` is called again at the end of this function, so I don't think it's necessary to call it here. I probably missed this during the monster rebase. Or am I missing something?
</details>
Ah, right, I thought we had just forgotten that.
Well, then _that_ is a good catch. :-D

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

madadam:
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[src/states/node.rs, line 384 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcXzrAHakfCEygKFK0:-KMcl3bmy44S6gh727jZ:-1291341372) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L384)):*
> OK. Is it guaranteed that we don't transition before we handled `ListenerStarted`?
</details>
The transition to `Node` happens when the client receives response to `GetNodeName`. Is it possible that can happen before the client starts listening? I'd say it's not, but I might be missing something.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

afck:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[src/states/node.rs, line 384 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcXzrAHakfCEygKFK0:-KMcpBOFlA-jYCP1jxVf:2077461939) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L384)):*
> The transition to `Node` happens when the client receives response to `GetNodeName`. Is it possible that can happen before the client starts listening? I'd say it's not, but I might be missing something.
</details>
Why not? All communication, including requesting and receiving the node name, happens via the proxy connection which was established as a Crust bootstrap connection.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

madadam:
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[src/states/node.rs, line 384 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcXzrAHakfCEygKFK0:-KMcppcMJLhQgUFf4bHy:-1753169398) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L384)):*
> Why not? All communication, including requesting and receiving the node name, happens via the proxy connection which was established as a Crust bootstrap connection.
</details>
True. But can it happen that a client receives 'GetNodeName` response before it sends the request first? If not, then it's guaranteed that the transition can only happen after the listener is started, because the `GetNodeName` request is send only after that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

afck:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[src/states/node.rs, line 384 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcXzrAHakfCEygKFK0:-KMcqib5ZKdq2i_Y3LZ5:1306585957) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/node.rs#L384)):*
> True. But can it happen that a client receives 'GetNodeName` response before it sends the request first? If not, then it's guaranteed that the transition can only happen after the listener is started, because the `GetNodeName` request is send only after that.
</details>
Ah, right! Then that's fine, thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

afck:
Reviewed 1 of 13 files at r1, 8 of 8 files at r4.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[src/states/client.rs, line 314 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMcM0yg1B8G536go4aa:-KMcv1zQLwdSdzCosqUn:676269983) ([raw file](https://github.com/maidsafe/routing/blob/6be11034d9dac25c6e07fc0271605b1f4acf63a0/src/states/client.rs#L314)):*
> That's good to know. I love deleting code! :)
</details>
:+1:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

Fraser999:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/state_machine.rs, line 55 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMhp_59rFBgRAZAA-hR:-KMhp_5AyGq8_jWVr23-:401553317) ([raw file](https://github.com/maidsafe/routing/blob/d3417db342b4e67c4bbf94c84f58f6531b1295b4/src/state_machine.rs#L55)):*
> ```Rust
> pub enum Transition {
>     // Stay in the current state
>     Stay,
> ```

Only a suggestion for discussion (not implying we should change this PR): would it be worth considering removing the `Stay` variant and using an `Option<Transition>` throughout, with `None` representing "stay in the current state"?  It wouldn't make any difference logically - just maybe more readable?

---

*[src/states/node.rs, line 64 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMhekNMm3LAYTIufM4S:-KMhekNMm3LAYTIufM4T:2141694028) ([raw file](https://github.com/maidsafe/routing/blob/d3417db342b4e67c4bbf94c84f58f6531b1295b4/src/states/node.rs#L64)):*
> ```Rust
> const REFRESH_BUCKET_GROUPS_SECS: u64 = 120;
> 
> /// An interface for clients and nodes that handles routing and connecting to the network.
> ```

I guess this entire comment block should be deleted or moved?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

madadam:
Review status: 15 of 20 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/state_machine.rs, line 55 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMhp_59rFBgRAZAA-hR:-KMhsgzMV4lV99LcqlJW:1475635704) ([raw file](https://github.com/maidsafe/routing/blob/d3417db342b4e67c4bbf94c84f58f6531b1295b4/src/state_machine.rs#L55)):*
> Only a suggestion for discussion (not implying we should change this PR): would it be worth considering removing the `Stay` variant and using an `Option<Transition>` throughout, with `None` representing "stay in the current state"?  It wouldn't make any difference logically - just maybe more readable?
</details>
I slightly prefer the current approach, as it's a bit less verbose.

---

*[src/states/node.rs, line 64 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMhekNMm3LAYTIufM4S:-KMhrwmCou_6xYPoJREL:-1754160691) ([raw file](https://github.com/maidsafe/routing/blob/d3417db342b4e67c4bbf94c84f58f6531b1295b4/src/states/node.rs#L64)):*
> I guess this entire comment block should be deleted or moved?
</details>
Shall I move it to `state_machine.rs` perhaps?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

afck:
Reviewed 1 of 13 files at r1, 5 of 5 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

Fraser999:

---

*[src/state_machine.rs, line 55 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMhp_59rFBgRAZAA-hR:-KMhtH0V-r7r05-77ZeZ:-747852723) ([raw file](https://github.com/maidsafe/routing/blob/d3417db342b4e67c4bbf94c84f58f6531b1295b4/src/state_machine.rs#L55)):*
> I slightly prefer the current approach, as it's a bit less verbose.
</details>
No worries - I don't have a strong preference :)

---

*[src/states/node.rs, line 64 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1080#-KMhekNMm3LAYTIufM4S:-KMhtMo9nACFlXFvATNL:1910806427) ([raw file](https://github.com/maidsafe/routing/blob/d3417db342b4e67c4bbf94c84f58f6531b1295b4/src/states/node.rs#L64)):*
> Shall I move it to `state_machine.rs` perhaps?
</details>
I guess if it's still mostly valid, it should be moved to somewhere public; moving it to state_machine.rs would still keep it hidden from the generated docs.  I think I'd probably delete it since it seems to mainly be covered in the crate-level comments of lib.rs, or cherry pick any parts and move them to the existing comments in lib.rs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

maqi:
Reviewed 2 of 13 files at r1, 9 of 19 files at r3, 5 of 8 files at r4, 5 of 5 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

afck:
Reviewed 1 of 13 files at r1, 2 of 2 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1080)*

PR #1084, opened 2016-07-12
Peer manager refactor

maqi:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1084)

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -222,11 +222,8 @@ impl Core {
         };
 
         let full_id = keys.unwrap_or_else(FullId::new);
-
         let our_info = NodeInfo::new(*full_id.public_id(), crust_service.id());
-
         let user_msg_cache_duration = Duration::from_secs(60 * 20);
-
         let mut core = Core {
             crust_service: crust_service,
             role: role,
@@ -316,7 +313,7 @@ impl Core {
 
     /// Routing table of this node.
     #[cfg(feature = "use-mock-crust")]
-    pub fn routing_table(&self) -> &RoutingTable<NodeInfo> {
+    pub fn routing_table(&self) -> &RoutingTable<XorName> {
         self.peer_mgr.routing_table()
     }
 
@@ -598,10 +595,10 @@ impl Core {
     }
 
     fn find_tunnel_for_peer(&mut self, peer_id: PeerId, pub_id: &PublicId) {
-        for node in self.peer_mgr.set_searching_for_tunnel(peer_id, pub_id) {
-            trace!("{:?} Asking {:?} to serve as a tunnel.", self, node.name());
+        for (name, dst_peer_id) in self.peer_mgr.set_searching_for_tunnel(peer_id, pub_id) {
+            trace!("{:?} Asking {:?} to serve as a tunnel.", self, name);
             let tunnel_request = DirectMessage::TunnelRequest(peer_id);
-            let _ = self.send_direct_message(&node.peer_id, tunnel_request);
+            let _ = self.send_direct_message(&dst_peer_id, tunnel_request);
         }
     }
 
@@ -1378,19 +1375,19 @@ impl Core {
                 if self.peer_mgr.routing_table().len() == 1 {
                     let _ = self.event_sender.send(Event::Connected);
                 }
-                for notify_info in must_notify {
+                for peer_id in self.peer_mgr.get_peer_ids(&must_notify) {
                     let message = DirectMessage::NewNode(public_id);
-                    let _ = self.send_direct_message(&notify_info.peer_id, message);
+                    let _ = self.send_direct_message(&peer_id, message);
                 }
-                for node_info in unneeded {
+                for peer_id in self.peer_mgr.get_peer_ids(&unneeded) {
                     let message = DirectMessage::ConnectionUnneeded(*self.name());
-                    let _ = self.send_direct_message(&node_info.peer_id, message);
+                    let _ = self.send_direct_message(&peer_id, message);
                 }
 
                 self.reset_bucket_refresh_timer();
 
                 if common_groups {
-                    let event = Event::NodeAdded(name, self.peer_mgr.routing_table().to_names());
+                    let event = Event::NodeAdded(name, self.peer_mgr.routing_table().clone());
                     if let Err(err) = self.event_sender.send(event) {
                         error!("{:?} Error sending event to routing user - {:?}", self, err);
                     }
@@ -1579,7 +1576,7 @@ impl Core {
         }
 
         let close_group = match self.peer_mgr.routing_table().close_nodes(&dst_name, GROUP_SIZE) {
-            Some(close_group) => close_group.iter().map(NodeInfo::name).cloned().collect(),
+            Some(close_group) => close_group,
             None => return Err(RoutingError::InvalidDestination),
         };
         let relocated_name = try!(utils::calculate_relocated_name(close_group,
@@ -1626,7 +1623,7 @@ impl Core {
 
         let public_ids = match self.peer_mgr.routing_table().close_nodes(expect_id.name(),
                                                                          GROUP_SIZE) {
-            Some(close_group) => close_group.into_iter().map(|info| info.public_id).collect_vec(),
+            Some(close_group) => self.peer_mgr.get_pub_ids(&close_group),
             None => return Err(RoutingError::InvalidDestination),
         };
 
@@ -1685,7 +1682,7 @@ impl Core {
         }
 
         let public_ids = match self.peer_mgr.routing_table().close_nodes(&dst_name, GROUP_SIZE) {
-            Some(close_group) => close_group.into_iter().map(|info| info.public_id).collect_vec(),
+            Some(close_group) => self.peer_mgr.get_pub_ids(&close_group),
             None => return Err(RoutingError::InvalidDestination),
         };
 
@@ -2141,13 +2138,10 @@ impl Core {
                     .routing_table()
                     .target_nodes(destination, hop, route as usize)
                     .into_iter()
-                    .filter(|target| !sent_to.contains(target.name()))
+                    .filter(|target| !sent_to.contains(target))
                     .collect_vec();
-                let new_sent_to = sent_to.iter()
-                    .chain(targets.iter().map(NodeInfo::name))
-                    .cloned()
-                    .collect_vec();
-                Ok((new_sent_to, targets.into_iter().map(|target| target.peer_id).collect()))
+                let new_sent_to = sent_to.iter().chain(targets.iter()).cloned().collect_vec();
+                Ok((new_sent_to, self.peer_mgr.get_peer_ids(&targets)))
             }
         }
     }
@@ -2308,7 +2302,7 @@ impl Core {
                 .is_in_any_close_group_with(self.name().bucket_index(&name), GROUP_SIZE);
             if common_groups {
                 // If the lost node shared some close group with us, send a NodeLost event.
-                let event = Event::NodeLost(name, self.peer_mgr.routing_table().to_names());
+                let event = Event::NodeLost(name, self.peer_mgr.routing_table().clone());
                 if let Err(err) = self.event_sender.send(event) {
                     error!("{:?} Error sending event to routing user - {:?}", self, err);
                 }
--- a/src/core.rs
+++ b/src/core.rs
@@ -1052,8 +1052,6 @@ impl Core {
             self.dropped_routing_node_connection(&peer_id);
             self.dropped_client_connection(&peer_id);
             self.dropped_tunnel_node(&peer_id);
-        } else {
-            let _ = self.peer_mgr.remove_peer(&peer_id);
         }
         self.dropped_bootstrap_connection(&peer_id);
     }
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @afck

afck:
Review status: 0 of 2 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[src/core.rs, line 1055 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMSy2p01IOhIi0DoSPx:-KMSy2p01IOhIi0DoSPy:-2056584875) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/core.rs#L1055)):*
> ```Rust
>             self.dropped_tunnel_node(&peer_id);
>         } else {
>             let _ = self.peer_mgr.remove_peer(&peer_id);
> ```

Can it actually be anything else than the bootstrap connection? Could this line be omitted entirely?

---

*[src/core.rs, line 1585 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMSzEZkaU2U-i7urQl3:-KMSzEZkaU2U-i7urQl4:-85617957) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/core.rs#L1585)):*
> ```Rust
>             Some(close_group) => close_group,
>             None => return Err(RoutingError::InvalidDestination),
>         };
> ```

You could also avoid the duplicated `close_group` variable:
`let close_group = try!(self.peer_mgr.routing_table().close_nodes(&dst_name, GROUP_SIZE).ok_or(|_| RoutingError::InvalidDestination));`
(Not sure whether that's better - your call.)

---

*[src/core.rs, line 2289 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMT-gMkIp995bVL1ywP:-KMT-gMl0tYJsbKl2cUp:-1995019043) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/core.rs#L2289)):*
> ```Rust
>                     Some(dst_pub_id) => Some((dst_peer_id, *dst_pub_id)),
>                     None => None,
>                 }
> ```

Or:
`self.peer_mgr.get_communicating_peer(&dst_peer_id).map(|dst_pub_id| (dst_peer_id, *dst_pub_id))`?

---

*[src/peer_manager.rs, line 230 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMT0TfV9dEQYnOmsajd:-KMT0TfWUTf3p9UpoV8Q:194346640) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/peer_manager.rs#L230)):*
> ```Rust
>     }
> 
>     pub fn is_tunnel(&self, peer_id: &PeerId, dst_id: &PeerId) -> bool {
> ```

Looks like this is just testing whether both are in the routing table. Maybe it should be renamed - or even just take one argument and check whether that is in the routing table. The caller would then call it twice, with both IDs, if needed.

---

*[src/peer_manager.rs, line 351 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMT0oipHsERJBlc5fMa:-KMT0oiqyOnMPhv3TF1Z:-2069039602) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/peer_manager.rs#L351)):*
> ```Rust
>         self.pub_id_map.get(peer_id).and_then(|pub_id| {
>             match self.get_state(pub_id) {
>                 Some(&PeerState::CrustConnecting) => Some(pub_id),
> ```

... or _would_ it actually make sense (at the call site in core.rs) to consider these other three states here, too?

---

*[src/peer_manager.rs, line 357 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMT14D_D0STwOrSEqkV:-KMT14D_D0STwOrSEqkW:1485630135) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/peer_manager.rs#L357)):*
> ```Rust
>     }
> 
>     /// Return the first peer_id of the peer_node bearing the name.
> ```

I'd use either `peer ID` or `` `PeerId` `` in comments. `peer_id` is just a common variable name, but neither the type nor a natural language term. (And maybe just `node` instead of `peer_node`. Also elsewhere.)

---

*[src/peer_manager.rs, line 372 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMT1ziZi6xEtJy5EWNA:-KMT1ziZi6xEtJy5EWNB:-509987281) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/peer_manager.rs#L372)):*
> ```Rust
>             })
>             .collect()
>     }
> ```

It would be nice to get rid of these two methods: Ideally we should try to always address nodes by their public ID instead of their peer ID, where possible.
But if it _isn't_ possible, and this needs to be done very often, it might make sense to include the peer ID in the `node_map` instead, so we don't need to iterate here and can just look up that entry.

Similarly with `get_pub_ids`: I think in general, we should try to reduce the number of types by which we need to look up peers. And for those types that _need_ to be used as indices, we should consider additional maps, etc.

(I'm fine with leaving all of these as is for this PR, just something to keep in mind.)

---

*[src/peer_manager.rs, line 387 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMT3Qwp7yTYpod8xTut:-KMT3Qwp7yTYpod8xTuu:-1806396749) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/peer_manager.rs#L387)):*
> ```Rust
> 
>     /// Returns the public ID of the given peer, if it is in `Connected` or `Tunnel` state.
>     pub fn get_communicating_peer(&self, peer_id: &PeerId) -> Option<&PublicId> {
> ```

Maybe `get_routing_peer`? Or `get_connected_or_tunnelling_peer`? Or `get_routing_table_entry`? Not sure ...

---

*[src/peer_manager.rs, line 397 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMT3fB_u4Ef_JJfNyQe:-KMT3fB_u4Ef_JJfNyQf:898019980) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/peer_manager.rs#L397)):*
> ```Rust
>     }
> 
>     /// Sets the given peer to state `SearchingForTunnel` or returns `false` if it doesn't exist.
> ```

That comment is wrong now.

---

*[src/peer_manager.rs, line 530 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMT3zT_fkUoAP35qXs5:-KMT3zT_fkUoAP35qXs6:1641074428) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/peer_manager.rs#L530)):*
> ```Rust
>     }
> 
>     /// Removes the given entry.
> ```

That should also describe what the complicated return value means now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1084)*

maqi:
*[src/peer_manager.rs, line 351 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMT0oipHsERJBlc5fMa:-KMTJ-gNFVfpVzh3wZP9:-1084921857) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/peer_manager.rs#L351)):*
> ... or _would_ it actually make sense (at the call site in core.rs) to consider these other three states here, too?
</details>
don't get what you mean.
this function only being called by core once, when there is a `connect_failure`, which the other three states shall not trigger/incur any further handling?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1084)*

afck:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[src/peer_manager.rs, line 351 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMT0oipHsERJBlc5fMa:-KMTJv0e-Qpe8JXeL848:623082494) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/peer_manager.rs#L351)):*
> don't get what you mean.
> this function only being called by core once, when there is a `connect_failure`, which the other three states shall not trigger/incur any further handling?
</details>
Ah, right, that can only happen in `CrustConnecting` state, of course. :+1: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1084)*

maqi:
*[src/core.rs, line 1585 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMSzEZkaU2U-i7urQl3:-KMcMg0fLmgaT-WfnBJa:1453488304) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/core.rs#L1585)):*
> You could also avoid the duplicated `close_group` variable:
> `let close_group = try!(self.peer_mgr.routing_table().close_nodes(&dst_name, GROUP_SIZE).ok_or(|_| RoutingError::InvalidDestination));`
> (Not sure whether that's better - your call.)
</details>
the new approach will have the same lines after rustfmt.
so prefer the current approach as it is more readable   :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1084)*

maqi:
*[src/peer_manager.rs, line 230 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMT0TfV9dEQYnOmsajd:-KMcMzpoFAPSvxpRjoX8:2140413627) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/peer_manager.rs#L230)):*
> Looks like this is just testing whether both are in the routing table. Maybe it should be renamed - or even just take one argument and check whether that is in the routing table. The caller would then call it twice, with both IDs, if needed.
</details>
prefer avoiding force the caller call twice, and handle everything inside peer_manager.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1084)*

maqi:
*[src/core.rs, line 1055 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMSy2p01IOhIi0DoSPx:-KMcP2Us4t-rlA7bURZK:-308380246) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/core.rs#L1055)):*
> Can it actually be anything else than the bootstrap connection? Could this line be omitted entirely?
</details>
not sure about this. if in `client` mode (working as client), does it still have chance of losing a peer?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1084)*

afck:
Reviewed 2 of 2 files at r3.

---

*[src/core.rs, line 1055 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMSy2p01IOhIi0DoSPx:-KMcdRD-s3TEzWd4PRmx:-2070335952) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/core.rs#L1055)):*
> not sure about this. if in `client` mode (working as client), does it still have chance of losing a peer?
</details>
Only the proxy node, I think.
That should be very evident from the code once #1080 is merged.

---

*[src/peer_manager.rs, line 230 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMT0TfV9dEQYnOmsajd:-KMceT7yWbtjF1kHP2cU:-1922612720) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/peer_manager.rs#L230)):*
> prefer avoiding force the caller call twice, and handle everything inside peer_manager.
</details>
OK. But it should probably still have a more descriptive name. `can_tunnel_for`?

---

*[src/peer_manager.rs, line 357 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMT14D_D0STwOrSEqkV:-KMcfFDYhWU98Tpifuht:-198242898) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/peer_manager.rs#L357)):*
> I'd use either `peer ID` or `` `PeerId` `` in comments. `peer_id` is just a common variable name, but neither the type nor a natural language term. (And maybe just `node` instead of `peer_node`. Also elsewhere.)
</details>
(Also below in `get_peer_ids`, `get_pub_ids`, etc.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1084)*

maqi:
*[src/core.rs, line 1055 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1084#-KMSy2p01IOhIi0DoSPx:-KMd4KILhw4pD7aO3y0c:-364590736) ([raw file](https://github.com/maidsafe/routing/blob/4e7189e3c7bc77eb1bb6c18c32c32a41397eb1b8/src/core.rs#L1055)):*
> Only the proxy node, I think.
> That should be very evident from the code once #1080 is merged.
</details>
tested with this line removed and passed all the current tests.
will include the removal in the PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1084)*

afck:
Reviewed 2 of 2 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1084)*

PR #1085, opened 2016-07-12
chore/peer_manager: simplify calling to peer_manager

maqi:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1085)

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -568,12 +568,12 @@ impl Core {
             debug!("{:?} Removing unwanted tunnel for {:?}", self, peer_id);
             let message = DirectMessage::TunnelDisconnect(peer_id);
             let _ = self.send_direct_message(&tunnel_id, message);
-        } else if let Some(node) = self.peer_mgr.routing_table().iter().find(|node| node.peer_id == peer_id) {
+        } else if let Some(pub_id) = self.peer_mgr.get_routing_peer(&peer_id) {
             warn!("{:?} Received ConnectSuccess from {:?}, but node {:?} is already in our \
                    routing table.",
                   self,
                   peer_id,
-                  node.name());
+                  pub_id.name());
             return;
         }
         self.peer_mgr.connected_to(peer_id);
@@ -591,22 +591,17 @@ impl Core {
         }
         if self.role == Role::Client {
             warn!("{:?} Received ConnectFailure event as a client.", self);
-        } else if self.peer_mgr.routing_table().iter().all(|node| node.peer_id != peer_id) {
-            info!("{:?} Failed to connect to peer {:?}.", self, peer_id);
-            if let Some(&pub_id) = self.peer_mgr.get_connecting_peer(&peer_id) {
-                self.find_tunnel_for_peer(peer_id, &pub_id);
-            }
+        } else if let Some(&pub_id) = self.peer_mgr.get_connecting_peer(&peer_id) {
+            info!("{:?} Failed to connect to peer {:?} with pub_id {:?}.", self, peer_id, pub_id);
+            self.find_tunnel_for_peer(peer_id, &pub_id);
         }
     }
 
     fn find_tunnel_for_peer(&mut self, peer_id: PeerId, pub_id: &PublicId) {
-        if self.peer_mgr.set_searching_for_tunnel(peer_id, pub_id) {
-            for node in self.peer_mgr.routing_table()
-                                     .closest_nodes_to(pub_id.name(), GROUP_SIZE, false) {
-                trace!("{:?} Asking {:?} to serve as a tunnel.", self, node.name());
-                let tunnel_request = DirectMessage::TunnelRequest(peer_id);
-                let _ = self.send_direct_message(&node.peer_id, tunnel_request);
-            }
+        for node in self.peer_mgr.set_searching_for_tunnel(peer_id, pub_id) {
+            trace!("{:?} Asking {:?} to serve as a tunnel.", self, node.name());
+            let tunnel_request = DirectMessage::TunnelRequest(peer_id);
+            let _ = self.send_direct_message(&node.peer_id, tunnel_request);
         }
     }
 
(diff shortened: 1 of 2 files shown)


afck:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/core.rs, line 1057 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1085#-KMTL6xx9FmKrNyW1Ndz:-KMTL6xyY_F5uRitrN0j:1070116960) ([raw file](https://github.com/maidsafe/routing/blob/460274af10a369574773b8071cf117bf77d69b61/src/core.rs#L1057)):*
> ```Rust
>             let _ = self.peer_mgr.remove_peer(&peer_id);
>         }
>         self.dropped_bootstrap_connection(&peer_id);
> ```

This whole function looks like it could almost completely (including all the `dropped_` functions) go into `PeerManager`.
But it's probably better do do this step-by-step, so not necessarily in this PR.

---

*[src/core.rs, line 1246 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1085#-KMTNVZW0o0NWHam2mrd:-KMTNVZX7iTla_Co_uSy:1216404721) ([raw file](https://github.com/maidsafe/routing/blob/460274af10a369574773b8071cf117bf77d69b61/src/core.rs#L1246)):*
> ```Rust
>                         return Err(RoutingError::InvalidSource);
>                     }
>                 }
> ```

Do you think it would make sense to include this logic into the `remove_if_unneeded` call?

---

*[src/peer_manager.rs, line 497 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1085#-KMTQDw6sqMAKSl1BssS:-KMTQDw7xdbpxdZV0SrU:697127468) ([raw file](https://github.com/maidsafe/routing/blob/460274af10a369574773b8071cf117bf77d69b61/src/peer_manager.rs#L497)):*
> ```Rust
> 
>     /// Removes the given entry, returns the pair of (peer's public name, the removal result)
>     pub fn remove_peer(&mut self, peer_id: &PeerId) -> (Option<XorName>, Option<DroppedNodeDetails>) {
> ```

Maybe this should return an `Option<(XorName, DroppedNodeDetails)>` instead. I think they will always both be `None` or both `Some` anyway - and at least we don't use the name in `Core` if the second argument is `None`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1085)*

Fraser999:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/peer_manager.rs, line 597 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1085#-KMTU6kKCVoW6qECv7DO:-KMTU6kLgZb1Ts6Yj3YN:-1755378695) ([raw file](https://github.com/maidsafe/routing/blob/460274af10a369574773b8071cf117bf77d69b61/src/peer_manager.rs#L597)):*
> ```Rust
>         let their_connection_info = PubConnectionInfo(PeerId(1), Endpoint(1));
>         // We decide to connect to the peer with `pub_id`:
>         let token = peer_mgr.get_connection_token(node_auth(0), node_auth(1), orig_pub_id).unwrap();
> ```

Couldn't this be using the `unwrap!()` macro?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1085)*

afck:
Reviewed 2 of 2 files at r2.

---

*[src/peer_manager.rs, line 232 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1085#-KMUYrFy7tLvSapgIF7N:-KMUYrFy7tLvSapgIF7O:-494114631) ([raw file](https://github.com/maidsafe/routing/blob/710935c00327faf1d0b0aaeaaa3bf1a4127b84c2/src/peer_manager.rs#L232)):*
> ```Rust
>         if let Some(true) = self.pub_id_map.iter().find(|elt| elt.1.name() == name)
>                                                   .map(|(peer_id, _)| target_id == peer_id) {
>             Some(self.routing_table.remove_if_unneeded(name))
> ```

Probably we should also remove that peer from the maps at this point. Then maybe even the `handle_lost_peer` call above wouldn't be necessary anymore?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1085)*

maqi:
*[src/peer_manager.rs, line 232 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1085#-KMUYrFy7tLvSapgIF7N:-KMUdrp5SPAYcWoOpnE8:848683921) ([raw file](https://github.com/maidsafe/routing/blob/710935c00327faf1d0b0aaeaaa3bf1a4127b84c2/src/peer_manager.rs#L232)):*
> Probably we should also remove that peer from the maps at this point. Then maybe even the `handle_lost_peer` call above wouldn't be necessary anymore?
</details>
peer_manager doesn't have `handle_lost_peer`, and the function in core involves lots of other cleanup operations.
so, better stay with the current operation?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1085)*

afck:

---

*[src/peer_manager.rs, line 232 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1085#-KMUYrFy7tLvSapgIF7N:-KMUfmuhnmqROgpyFntV:-1678270182) ([raw file](https://github.com/maidsafe/routing/blob/8d95a7b2ae3a689d39583c7ca1fbdec75c0a8153/src/peer_manager.rs#L232)):*
> peer_manager doesn't have `handle_lost_peer`, and the function in core involves lots of other cleanup operations.
> so, better stay with the current operation?
</details>
OK, but maybe that's something for one of the next peer manager PRs - it would be really great to get to a state where we know that the routing table is synchronised with the maps.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1085)*

Fraser999:
Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1085)*

PR #1087, opened 2016-07-13
fix/peer_manager: avoid calling connect for proxies or clients

afck:
This sends `NodeIdentify` messages instead of calling
`crust::Service::connect` for proxy and client nodes. The latter would
not result in a `ConnectSuccess` and fail to put the peer into the
routing table.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1087)


maidsafe-highfive:

Viv-Rajkumar:
Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1087)*

PR #1095, opened 2016-07-19
Extract bootstrapping state

madadam:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1095)


maidsafe-highfive:
r? @afck

afck:
Reviewed 12 of 12 files at r1.

---

*[src/core_tests.rs, line 1206 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1W6-OB0KKp1DxvmHl:-KN1W6-Plm2LIz2tJHTn:1106205258) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/core_tests.rs#L1206)):*
> ```Rust
> 
>         // Do not send from the newly added node.
>         if let Some(index) = added_index {
> ```

I'd almost leave these - we shouldn't panic in that case, at least for now.

---

*[src/message_accumulator.rs, line 38 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1SPuuYxhjdPdY85QC:-KN1SPuuYxhjdPdY85QD:-1006780209) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/message_accumulator.rs#L38)):*
> ```Rust
> 
> impl MessageAccumulator {
>     pub fn new() -> Self {
> ```

Doesn't Clippy require a default implementation anymore?

---

*[src/states/bootstrapping.rs, line 84 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1VqoZ-UitBzg9Hlj5:-KN1Vqo_1sAQdgekwvr1:569487476) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/states/bootstrapping.rs#L84)):*
> ```Rust
>             Action::Timeout(token) => self.handle_timeout(token),
>             Action::Terminate => Transition::Terminate,
>             _ => panic!("unhandled action {:?}", action),
> ```

I don't think we should panic here. Let's rather add a TODO that we want to make sure that the upper layers don't try this, and maybe log an error, but not panic.

---

*[src/states/bootstrapping.rs, line 124 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1TfpzCPVmc8LjtFbn:-KN1Tfq-LMXF9cVKFuDF:303695735) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/states/bootstrapping.rs#L124)):*
> ```Rust
>                                    self.full_id,
>                                    self.stats,
>                                    self.timer)
> ```

If the timer can't have any relevant entries at this point anyway, I wonder whether we should even pass it in here, and not just create a new one inside `from_bootstrapping`?
Also, the `Cache` is now passed through several states although it is only used in `Node`. (But maybe it's better to clean that up after the `Client`/`JoiningNode` separation?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1095)*

madadam:

---

*[src/core_tests.rs, line 1206 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1W6-OB0KKp1DxvmHl:-KN1d0vz6qI8zWV93tpk:241380946) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/core_tests.rs#L1206)):*
> I'd almost leave these - we shouldn't panic in that case, at least for now.
</details>
I was getting a consistent test failure without it. The problem  is that we are trying to send a message from a node which is in the `Bootstrapping` state and that state doesn't know how to send routing messages. One possible way to solve this is to buffer the messages while in the `Bootstrapping` state and then pass that buffer to `Client` and `Node` states which would then send them.

---

*[src/message_accumulator.rs, line 38 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1SPuuYxhjdPdY85QC:-KN1dhhaOsZOQ6_Vdwcc:346386571) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/message_accumulator.rs#L38)):*
> Doesn't Clippy require a default implementation anymore?
</details>
I ran it through Clippy and it wasn't complaining.

---

*[src/states/bootstrapping.rs, line 84 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1VqoZ-UitBzg9Hlj5:-KN1eANG4pi8LnrnuENb:-1132748557) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/states/bootstrapping.rs#L84)):*
> I don't think we should panic here. Let's rather add a TODO that we want to make sure that the upper layers don't try this, and maybe log an error, but not panic.
</details>
Problem is some of the actions need a response to be sent via the `Receiver` contained in them. So if we simply ignore those actions, it can cause the whole program to freeze. Figured we can as well panic to save time :)

---

*[src/states/bootstrapping.rs, line 124 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1TfpzCPVmc8LjtFbn:-KN1e1UdR2HounRstQV9:1038199885) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/states/bootstrapping.rs#L124)):*
> If the timer can't have any relevant entries at this point anyway, I wonder whether we should even pass it in here, and not just create a new one inside `from_bootstrapping`?
> Also, the `Cache` is now passed through several states although it is only used in `Node`. (But maybe it's better to clean that up after the `Client`/`JoiningNode` separation?)
</details>
Problem is the timer needs the action sender to be passed to it. So we either pass that, or the timer itself. Not sure which is better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1095)*

afck:

---

*[src/core_tests.rs, line 1206 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1W6-OB0KKp1DxvmHl:-KN1fxtKkiFALkVzEx3W:-1809518898) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/core_tests.rs#L1206)):*
> I was getting a consistent test failure without it. The problem  is that we are trying to send a message from a node which is in the `Bootstrapping` state and that state doesn't know how to send routing messages. One possible way to solve this is to buffer the messages while in the `Bootstrapping` state and then pass that buffer to `Client` and `Node` states which would then send them.
</details>
But the behaviour so far was to just log an error, I think. So I'd prefer to stick to that for now and change the behaviour later.

---

*[src/states/bootstrapping.rs, line 84 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1VqoZ-UitBzg9Hlj5:-KN1gA8WNGFon49ND8Vu:1515081267) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/states/bootstrapping.rs#L84)):*
> Problem is some of the actions need a response to be sent via the `Receiver` contained in them. So if we simply ignore those actions, it can cause the whole program to freeze. Figured we can as well panic to save time :)
</details>
But I'd also do that in a separate change, unless we are absolutely sure that safe_vault and safe_core will never trigger this panic.

---

*[src/states/bootstrapping.rs, line 124 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1TfpzCPVmc8LjtFbn:-KN1gNEVAQoXkCskqX1d:-76000544) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/states/bootstrapping.rs#L124)):*
> Problem is the timer needs the action sender to be passed to it. So we either pass that, or the timer itself. Not sure which is better.
</details>
Ah, right. :-(
Let's pass in the timer then for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1095)*

madadam:

---

*[src/core_tests.rs, line 1206 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1W6-OB0KKp1DxvmHl:-KN1rARKWWkwUJGm36RO:-471350148) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/core_tests.rs#L1206)):*
> But the behaviour so far was to just log an error, I think. So I'd prefer to stick to that for now and change the behaviour later.
</details>
So what happens here is that a node which is still in the `Bootstrapping` state receives a `Action::NodeSendMessage` which it doesn't know how to handle. Then depending on how we handle unknown actions, this either freezes (if we ignore them), or crashes (if we panic on them). 

I think the best thing to do in this case is to report error back to the `Action::NodeSendMessage` sender. So, we would handle the action, but always send an `Err` back through the `result_tx`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1095)*

afck:

---

*[src/core_tests.rs, line 1206 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1W6-OB0KKp1DxvmHl:-KN1w0UE26vt5Qni-S69:628034377) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/core_tests.rs#L1206)):*
> So what happens here is that a node which is still in the `Bootstrapping` state receives a `Action::NodeSendMessage` which it doesn't know how to handle. Then depending on how we handle unknown actions, this either freezes (if we ignore them), or crashes (if we panic on them). 
> 
> I think the best thing to do in this case is to report error back to the `Action::NodeSendMessage` sender. So, we would handle the action, but always send an `Err` back through the `result_tx`.
</details>
I guess that's the way to go for now. Could you check whether that's also the behaviour from before the state machine change? Let's keep the old logic wherever possible in refactoring PRs, and do behavioural changes in separate PRs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1095)*

madadam:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/core_tests.rs, line 1206 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1W6-OB0KKp1DxvmHl:-KN1y0m_HBfsY5oCHEen:-704770547) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/core_tests.rs#L1206)):*
> I guess that's the way to go for now. Could you check whether that's also the behaviour from before the state machine change? Let's keep the old logic wherever possible in refactoring PRs, and do behavioural changes in separate PRs.
</details>
It's not. Previously, we would handle `Action::NodeSendMessage` (and also `Action::ClientSendRequest`) no matter what state we were in. To preserve this behavior, I'd have to either add the logic to send routing messages also to the `Bootstrapping` state, or queue the routing messages and pass that queue to the `Client` and `Node` states to send.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1095)*

afck:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/core_tests.rs, line 1206 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1W6-OB0KKp1DxvmHl:-KN2-3_6qNA4fEJfjT3Z:569188716) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/core_tests.rs#L1206)):*
> It's not. Previously, we would handle `Action::NodeSendMessage` (and also `Action::ClientSendRequest`) no matter what state we were in. To preserve this behavior, I'd have to either add the logic to send routing messages also to the `Bootstrapping` state, or queue the routing messages and pass that queue to the `Client` and `Node` states to send.
</details>
Ah, but that attempt would probably just fail, wouldn't it? Routing table empty -> do nothing and return `Ok(())` -> send back `Ok(())` via `result_tx`?
I agree that that's _wrong_, but if that was the behaviour before, I'd rather also send back an `Ok(())` and add a TODO to change that later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1095)*

madadam:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/core_tests.rs, line 1206 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1W6-OB0KKp1DxvmHl:-KN20eIP6W4h9D2XzvYj:393182744) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/core_tests.rs#L1206)):*
> Ah, but that attempt would probably just fail, wouldn't it? Routing table empty -> do nothing and return `Ok(())` -> send back `Ok(())` via `result_tx`?
> I agree that that's _wrong_, but if that was the behaviour before, I'd rather also send back an `Ok(())` and add a TODO to change that later.
</details>
Right! For some reason, it didn't occur to me that I can return `Ok` in case of failure :grimacing: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1095)*

afck:
Reviewed 3 of 3 files at r3.

---

*[src/states/bootstrapping.rs, line 84 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1VqoZ-UitBzg9Hlj5:-KN29Ewds-Lz4pv9hk_x:974989031) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/states/bootstrapping.rs#L84)):*
> But I'd also do that in a separate change, unless we are absolutely sure that safe_vault and safe_core will never trigger this panic.
</details>
I'm fine with this now, if it is certain that safe_core and safe_vault won't send any other action while in `Bootstrapping` state, and expect a response on it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1095)*

madadam:
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[src/states/bootstrapping.rs, line 84 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1VqoZ-UitBzg9Hlj5:-KN2DAjlSZBXGawmuA_3:-615251637) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/states/bootstrapping.rs#L84)):*
> I'm fine with this now, if it is certain that safe_core and safe_vault won't send any other action while in `Bootstrapping` state, and expect a response on it.
</details>
I don't know it it's certain. Actually, I think it would be better to either panic here (so the bug manifests itself as crash rather than freeze) or explicitly handle all the action types (by responding with errors). What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1095)*

afck:

---

*[src/states/bootstrapping.rs, line 84 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1095#-KN1VqoZ-UitBzg9Hlj5:-KN2ay4Bjas7jgMFWn7g:1991035091) ([raw file](https://github.com/maidsafe/routing/blob/bafc3394189918568e06c8ccd7bc981fa52bb425/src/states/bootstrapping.rs#L84)):*
> I don't know it it's certain. Actually, I think it would be better to either panic here (so the bug manifests itself as crash rather than freeze) or explicitly handle all the action types (by responding with errors). What do you think?
</details>
I just discussed this with Viv: Let's just return an `Err` through the channel before dropping it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1095)*

maqi:
Reviewed 9 of 12 files at r1, 3 of 3 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1095)*

afck:
Reviewed 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1095)*

PR #1100, opened 2016-07-20
fix/peer_manager

maqi:
when add_to_routing_table, check whether shall be added as a tunnel peer

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1100)

### DIFF
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -232,8 +232,12 @@ impl PeerManager {
     pub fn add_to_routing_table(&mut self, info: NodeInfo) -> Option<AddedNodeDetails<XorName>> {
         let result = self.routing_table.add(*info.public_id.name());
         if result.is_some() {
-            let _ = self.node_map.insert(*info.public_id.name(),
-                                         (Instant::now(), PeerState::Connected));
+            let state = match self.node_map.remove(info.public_id.name()) {
+                Some((_, PeerState::SearchingForTunnel)) |
+                Some((_, PeerState::Tunnel)) => PeerState::Tunnel,
+                _ => PeerState::Connected,
+            };
+            let _ = self.node_map.insert(*info.public_id.name(), (Instant::now(), state));
             let _ = self.pub_id_map.insert(info.peer_id, info.public_id);
         }
         result


maidsafe-highfive:
r? @afck

afck:
Reviewed 1 of 1 files at r1.

---

*[src/peer_manager.rs, line 236 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1100#-KN8B8nw4FhRyTGjg1_J:-KN8B8nx9_7UHnMnzwRE:1939550776) ([raw file](https://github.com/maidsafe/routing/blob/0d2a7a2d5f7bcb2cfd1c0ead6d3614a477e2c5d7/src/peer_manager.rs#L236)):*
> ```Rust
>         if result.is_some() {
>             let state = match self.node_map.remove(info.public_id.name()) {
>                 Some((_, PeerState::SearchingForTunnel)) => PeerState::Tunnel,
> ```

This case should probably also include `Tunnel` itself: If we've received a `TunnelSuccess` before, and only now got the `NodeIdentify`, we're probably already in state`Tunnel` here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1100)*

maqi:
*[src/peer_manager.rs, line 236 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1100#-KN8B8nw4FhRyTGjg1_J:-KNC7Gzaj8MPcZ_Hmgui:1461435557) ([raw file](https://github.com/maidsafe/routing/blob/0d2a7a2d5f7bcb2cfd1c0ead6d3614a477e2c5d7/src/peer_manager.rs#L236)):*
> This case should probably also include `Tunnel` itself: If we've received a `TunnelSuccess` before, and only now got the `NodeIdentify`, we're probably already in state`Tunnel` here?
</details>
we shall be already in the tunnel state.
however I don't see the harm to put a guard here to avoid unexpected situation.
we put it in as long as passing all tests

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1100)*

afck:

---

*[src/peer_manager.rs, line 236 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1100#-KN8B8nw4FhRyTGjg1_J:-KNC9UfU9hzm0DYDT1AG:1536850365) ([raw file](https://github.com/maidsafe/routing/blob/0d2a7a2d5f7bcb2cfd1c0ead6d3614a477e2c5d7/src/peer_manager.rs#L236)):*
> we shall be already in the tunnel state.
> however I don't see the harm to put a guard here to avoid unexpected situation.
> we put it in as long as passing all tests
</details>
No, that's what I mean: If we already are in `Tunnel`, we should keep that state and _not_ switch to `Connected` here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1100)*

maqi:
*[src/peer_manager.rs, line 236 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1100#-KN8B8nw4FhRyTGjg1_J:-KNCB_c6c_gDCJP1HtAu:-426804839) ([raw file](https://github.com/maidsafe/routing/blob/0d2a7a2d5f7bcb2cfd1c0ead6d3614a477e2c5d7/src/peer_manager.rs#L236)):*
> No, that's what I mean: If we already are in `Tunnel`, we should keep that state and _not_ switch to `Connected` here.
</details>
sorry, what I mean is that this function shall not be called from node.rs in that case with the current implementation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1100)*

afck:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/peer_manager.rs, line 236 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1100#-KN8B8nw4FhRyTGjg1_J:-KNCCB8A-7cn4pCdGSvf:-1700598034) ([raw file](https://github.com/maidsafe/routing/blob/0d2a7a2d5f7bcb2cfd1c0ead6d3614a477e2c5d7/src/peer_manager.rs#L236)):*
> sorry, what I mean is that this function shall not be called from node.rs in that case with the current implementation.
</details>
Really? What if we first get a `TunnelSuccess`, and then a `NodeIdentify`? Wouldn't we be in `Tunnel` state at this point, then?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1100)*

afck:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1100)*

maqi:
*[src/peer_manager.rs, line 236 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1100#-KN8B8nw4FhRyTGjg1_J:-KNCE9YdXEzeKuf5coBZ:322808089) ([raw file](https://github.com/maidsafe/routing/blob/0d2a7a2d5f7bcb2cfd1c0ead6d3614a477e2c5d7/src/peer_manager.rs#L236)):*
> Really? What if we first get a `TunnelSuccess`, and then a `NodeIdentify`? Wouldn't we be in `Tunnel` state at this point, then?
</details>
as I understand, a normal procedure will be:
node A send TunnelRequest for a tunnel, peer B received it then send back TunnelSuccess, node A receive it then response with NodeIdentity.
So any side shall not receive TunnelSuccess and NodeIdentity together.
But, if A and B requests tunnel at the same time, that might be problematic.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1100)*

afck:

---

*[src/peer_manager.rs, line 236 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1100#-KN8B8nw4FhRyTGjg1_J:-KNCFHW8OpNJjGYfd5wz:-879783981) ([raw file](https://github.com/maidsafe/routing/blob/0d2a7a2d5f7bcb2cfd1c0ead6d3614a477e2c5d7/src/peer_manager.rs#L236)):*
> as I understand, a normal procedure will be:
> node A send TunnelRequest for a tunnel, peer B received it then send back TunnelSuccess, node A receive it then response with NodeIdentity.
> So any side shall not receive TunnelSuccess and NodeIdentity together.
> But, if A and B requests tunnel at the same time, that might be problematic.
> 
</details>
So `A` will put `B` into state `Tunnel` as soon as it receives `TunnelSuccess`. When it receives the `NodeInfo`, it should then _keep_ it in state `Tunnel`.
I agree that it will be rare (or impossible?) that it receives a `NodeIdenfity` while still `SearchingForTunnel`. But it will definitely be common to call `add_to_routing_table` while it is in state `Tunnel`. So I think this code is fine now, until we add additional states to distinguish these.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1100)*

PR #1102, opened 2016-07-21
Extract joining node state

madadam:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1102)


maidsafe-highfive:
r? @afck

maqi:
Review status: 0 of 20 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/client.rs, line 39 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNGb3foePcgrkUnkYh-:-KNGb3fpvpzjN0wDiklV:1225931575) ([raw file](https://github.com/maidsafe/routing/blob/a2da301212d24d91b0d620d0c3a6353c999742ff/src/client.rs#L39)):*
> ```Rust
> use states;
> #[cfg(feature = "use-mock-crust")]
> use states::Testable;
> ```

it looks like this Testable trait is only for states that has bootstrapped?

---

*[src/client.rs, line 120 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNGbokVSrGlaycfk8Ax:-KNGbokWrrDyzvtk_0-Z:742970901) ([raw file](https://github.com/maidsafe/routing/blob/a2da301212d24d91b0d620d0c3a6353c999742ff/src/client.rs#L120)):*
> ```Rust
>                                                                   full_id,
>                                                                   timer))
>         })
> ```

is there dis-alignment or just display error?

---

*[src/client.rs, line 236 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNGcqGB0Ws77QTI4mj6:-KNGcqGCpPQyQzJRvhkF:-1759176406) ([raw file](https://github.com/maidsafe/routing/blob/a2da301212d24d91b0d620d0c3a6353c999742ff/src/client.rs#L236)):*
> ```Rust
>     Transitioning,
>     Terminated,
> }
> ```

two handling branch relates to transitioning are all unreachable, is this state being handled somewhere else?
also, I am not sure if it will be good to have too many unreachable in the production code

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

madadam:
Review status: 0 of 20 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/client.rs, line 39 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNGb3foePcgrkUnkYh-:-KNH-gv9ySIn61QVpgbo:1513397375) ([raw file](https://github.com/maidsafe/routing/blob/a2da301212d24d91b0d620d0c3a6353c999742ff/src/client.rs#L39)):*
> it looks like this Testable trait is only for states that has bootstrapped?
</details>
`Terstable` requires `Bootstrapped`, but I decided to make it a separate trait, for clarity.

---

*[src/client.rs, line 120 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNGbokVSrGlaycfk8Ax:-KNGzAO672wbjbd15v8B:1900416919) ([raw file](https://github.com/maidsafe/routing/blob/a2da301212d24d91b0d620d0c3a6353c999742ff/src/client.rs#L120)):*
> is there dis-alignment or just display error?
</details>
It's bad formatting. I forgot to rustfmt it.

---

*[src/client.rs, line 236 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNGcqGB0Ws77QTI4mj6:-KNGzJQh1hUPRj8elUFH:-2025556039) ([raw file](https://github.com/maidsafe/routing/blob/a2da301212d24d91b0d620d0c3a6353c999742ff/src/client.rs#L236)):*
> two handling branch relates to transitioning are all unreachable, is this state being handled somewhere else?
> also, I am not sure if it will be good to have too many unreachable in the production code
</details>
We need `Transitioning` to be able to move from the current state. See `StateMachine::next` method. An alternative would be to keep the state in `Option`, but I don't think it would be much cleaner.  

I'll see what I can do about some of those `unreachable!`s though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

maqi:
Review status: 0 of 20 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/core_tests.rs, line 515 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNH1z6kTAWZH6d1F9Vl:-KNH1z6lapW7UIFfNrb0:680778917) ([raw file](https://github.com/maidsafe/routing/blob/1b72f862ed613868206843dfbb30d9bc904b4881/src/core_tests.rs#L515)):*
> ```Rust
> 
> fn test_nodes(size: usize) {
>     // use maidsafe_utilities::log;
> ```

guess this is unrequired?

---

*[src/node.rs, line 140 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNH37DIMGzmiuz9TfR1:-KNH37DJ770Yf_adKWVs:-2052004143) ([raw file](https://github.com/maidsafe/routing/blob/1b72f862ed613868206843dfbb30d9bc904b4881/src/node.rs#L140)):*
> ```Rust
>                                                                     event_sender,
>                                                                     full_id,
>                                                                     timer))
> ```

same formatting error here?

---

*[src/node.rs, line 477 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNH49IMs3D1cmHG02Ym:-KNH49IMs3D1cmHG02Yn:-618980078) ([raw file](https://github.com/maidsafe/routing/blob/1b72f862ed613868206843dfbb30d9bc904b4881/src/node.rs#L477)):*
> ```Rust
>     JoiningNode(states::JoiningNode),
>     Node(states::Node),
>     Transitioning,
> ```

This `Tansitioning` state is more like a `programming required state` instead of a real scenario state (which the others are)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

madadam:
Review status: 0 of 20 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/node.rs, line 477 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNH49IMs3D1cmHG02Ym:-KNH8OmcVbQH0NwqLWmT:2131843432) ([raw file](https://github.com/maidsafe/routing/blob/1b72f862ed613868206843dfbb30d9bc904b4881/src/node.rs#L477)):*
> This `Tansitioning` state is more like a `programming required state` instead of a real scenario state (which the others are)
</details>
I removed `Transitioning` and repurposed the `Terminated` state to serve both as terminated and transitioning. It's less code this way, but not sure it's cleaner.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

maqi:
Review status: 0 of 19 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/states/bootstrapping.rs, line 67 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNHOoruNAs-FqjkCC9j:-KNHOorvfPOk2TIR7Dm8:-1849590618) ([raw file](https://github.com/maidsafe/routing/blob/9edafe99324a4fc6c5829efdae78f03b54d141d7/src/states/bootstrapping.rs#L67)):*
> ```Rust
>                timer: Timer)
>                -> Self {
>         let _ = crust_service.start_bootstrap(HashSet::new());
> ```

so the black_list is no longer required?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

madadam:
Review status: 0 of 19 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/states/bootstrapping.rs, line 67 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNHOoruNAs-FqjkCC9j:-KNHX5Izy0ofQX5EMSxg:196247707) ([raw file](https://github.com/maidsafe/routing/blob/9edafe99324a4fc6c5829efdae78f03b54d141d7/src/states/bootstrapping.rs#L67)):*
> so the black_list is no longer required?
</details>
It is, but it's used internally (see the `rebootsrap` method) and doesn't have to be passed through the constructor.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

afck:
(Haven't reviewed everything yet.)
Not sure about all those traits; I find them confusing and hard to read.
I wonder whether some of them could replaced with encapsulation, e.g. having a `DirectMessageSender` field in all the states instead of having them implement the `SendDirectMessage` trait - even if that means e.g. passing the crust service to a lot of the methods? (Not sure - maybe this is a bad idea.)

---

Reviewed 8 of 20 files at r1, 1 of 1 files at r2, 5 of 6 files at r4.
Review status: 14 of 19 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/node.rs, line 477 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNH49IMs3D1cmHG02Ym:-KNHLS028cK4ThLQahcP:556021568) ([raw file](https://github.com/maidsafe/routing/blob/1b72f862ed613868206843dfbb30d9bc904b4881/src/node.rs#L477)):*
> I removed `Transitioning` and repurposed the `Terminated` state to serve both as terminated and transitioning. It's less code this way, but not sure it's cleaner.
</details>
Not sure; the choice between forcing every implementation of `State` to provide a `Transitioning` state, and abusing the `Terminated` state is not a nice one ... :-(

How ugly would it be to actually use an `Option<State>` instead in the `StateMachine`?

Or ... if it saves _a lot_ of code we might also consider going back to not having this as a trait again? And just have all the states as variants of `state_machine::State`, documenting clearly which transitions are possible?

---

*[src/states/bootstrapping.rs, line 54 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNHOwvgtL9-BWQ5IjGA:-KNHOwvhhEx5umRjq3EF:583707695) ([raw file](https://github.com/maidsafe/routing/blob/9edafe99324a4fc6c5829efdae78f03b54d141d7/src/states/bootstrapping.rs#L54)):*
> ```Rust
>     event_sender: Sender<Event>,
>     full_id: FullId,
>     next_state_details: Option<(PeerManager, usize)>,
> ```

Ideally, the client shouldn't need a `PeerManager` - it just needs its proxy's peer ID and public ID.

Also, the client can only join once the network is large enough, so the quorum will always be 5.

---

*[src/states/client.rs, line 66 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNHQgX7pecCOPc13lCZ:-KNHQgX8aaCDAyThGD-o:16104776) ([raw file](https://github.com/maidsafe/routing/blob/9edafe99324a4fc6c5829efdae78f03b54d141d7/src/states/client.rs#L66)):*
> ```Rust
>                               full_id: FullId,
>                               peer_mgr: PeerManager,
>                               quorum_size: usize,
> ```

These two fields should probably be replaced by:
```
proxy_peer_id: PeerId,
proxy_pub_id: PublicId,
```
(See earlier comment.)

---

*[src/states/common/bootstrapped.rs, line 105 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNHV5iyeuomJcuWUqGE:-KNHV5izT48xLZwr4dq5:-447656323) ([raw file](https://github.com/maidsafe/routing/blob/9edafe99324a4fc6c5829efdae78f03b54d141d7/src/states/common/bootstrapped.rs#L105)):*
> ```Rust
>     }
> 
>     fn filter_incoming_signed_msg(&mut self, msg: &SignedMessage) -> usize {
> ```

It might make sense to remove this and just inline it everywhere? It's just one line, which isn't even more complicated than the method signature itself.

---

*[src/states/common/mod.rs, line 97 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNHW-_PpXFU_Etdvv3X:-KNHW-_QUL43yd4kgZYx:458164997) ([raw file](https://github.com/maidsafe/routing/blob/9edafe99324a4fc6c5829efdae78f03b54d141d7/src/states/common/mod.rs#L97)):*
> ```Rust
> }
> 
> pub trait DisconnectPeer {
> ```

Not sure whether this needs its own trait. Only the `Node` state will ever be in the situation that it wants to disconnect from someone but has to check all the possibilities. All the other states only need to check against their proxy node?
I think this might be easier to inline, even if it means a few duplicated lines, and have the `disconnect_peer` function only in `Node`.

---

*[src/states/common/mod.rs, line 107 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNHWSzA1RAmA5m-qloT:-KNHWSzA1RAmA5m-qloU:126059373) ([raw file](https://github.com/maidsafe/routing/blob/9edafe99324a4fc6c5829efdae78f03b54d141d7/src/states/common/mod.rs#L107)):*
> ```Rust
> }
> 
> pub trait GetClientAuthority {
> ```

Maybe this would be less confusing if it were just a utility method instead of a trait? It should just need the public ID and proxy node name as arguments, I think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102#-:-KNHXXTlKtkGArCixyZZ:1418364168)*

madadam:
Review status: 7 of 19 files reviewed at latest revision, 8 unresolved discussions.

---

*[src/node.rs, line 477 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNH49IMs3D1cmHG02Ym:-KNHeJD2gwE9ge2BLb5C:-504035406) ([raw file](https://github.com/maidsafe/routing/blob/1b72f862ed613868206843dfbb30d9bc904b4881/src/node.rs#L477)):*
> Not sure; the choice between forcing every implementation of `State` to provide a `Transitioning` state, and abusing the `Terminated` state is not a nice one ... :-(
> 
> How ugly would it be to actually use an `Option<State>` instead in the `StateMachine`?
> 
> Or ... if it saves _a lot_ of code we might also consider going back to not having this as a trait again? And just have all the states as variants of `state_machine::State`, documenting clearly which transitions are possible?
</details>
Going back to having only one `State` type would certainly save some code. Thinking about it now, maybe the split wasn't really worth the added complexity.

---

*[src/states/bootstrapping.rs, line 54 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNHOwvgtL9-BWQ5IjGA:-KNHebZWT413PJ8wFh93:-290614001) ([raw file](https://github.com/maidsafe/routing/blob/9edafe99324a4fc6c5829efdae78f03b54d141d7/src/states/bootstrapping.rs#L54)):*
> Ideally, the client shouldn't need a `PeerManager` - it just needs its proxy's peer ID and public ID.
> 
> Also, the client can only join once the network is large enough, so the quorum will always be 5.
</details>
Agree about the peer manager, but I'd do that in a separate PR. The client currently uses the peer manager to handle the proxy node and changing that might not be trivial.

About the quorum: the quorum size comes from the `BootstrapIdentify` message. Do you want me to hardcode it instead?

---

*[src/states/client.rs, line 66 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNHQgX7pecCOPc13lCZ:-KNHfedwTIN5FhuThJ4J:364822708) ([raw file](https://github.com/maidsafe/routing/blob/9edafe99324a4fc6c5829efdae78f03b54d141d7/src/states/client.rs#L66)):*
> These two fields should probably be replaced by:
> ```
> proxy_peer_id: PeerId,
> proxy_pub_id: PublicId,
> ```
> (See earlier comment.)
</details>
Sure, when we get rid of peer manager in clients. I'd do that in a separate PR though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

madadam:
`DirectMessageSender` would be tricky to implement using a struct, because it needs to call `send_or_drop` which is also used in other places, not just direct message sending. Using trait here also allow to easily customize the message wrapping - in `Bootstrapping` and `JoiningNode` it just puts it into `Message` directly, but in `Node` it involves tunnels. This would be hard to achieve any other way without some code duplication. 

That said, I agree this method (using traits) is not ideal. It might be possible to separate the code into more logical chunks if we think harder about the different responsibilities routing has. Splitting the code into states is just one step in this whole refactoring effort.

---

Review status: 7 of 19 files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102#-:-KNI-RNkqYU8QGeOrrvw:-1295713476)*

maqi:
Reviewed 8 of 20 files at r1, 1 of 1 files at r2, 1 of 6 files at r4, 8 of 9 files at r5, 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

afck:
Reviewed 3 of 20 files at r1, 1 of 6 files at r4, 8 of 9 files at r5, 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[src/state_machine.rs, line 117 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNVzmGuvMpdtAhjM6jb:-KNVzmGvQi0jHJ5uoVTH:-917228834) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/state_machine.rs#L117)):*
> ```Rust
>             State::Node(ref mut state) => state.handle_crust_event(event),
>             State::Terminated => Transition::Terminate,
>         }
> ```

Could all these `match` statements be simplified by having only one `get_state(&self) -> &mut State` method, returning a fat pointer, so you could e.g. just write `self.get_state().handle_crust_event(event)` here?
(But then we'd need a `Terminated` state implementation, so maybe it's not worth it.)

---

*[src/states/bootstrapping.rs, line 54 \[r4\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNHOwvgtL9-BWQ5IjGA:-KNW-Lja7Aw3qUSzZN6h:-2123319943) ([raw file](https://github.com/maidsafe/routing/blob/9edafe99324a4fc6c5829efdae78f03b54d141d7/src/states/bootstrapping.rs#L54)):*
> Agree about the peer manager, but I'd do that in a separate PR. The client currently uses the peer manager to handle the proxy node and changing that might not be trivial.
> 
> About the quorum: the quorum size comes from the `BootstrapIdentify` message. Do you want me to hardcode it instead?
</details>
I'm pretty sure all the proxy node handling done in `Client` and `JoiningNode` is trivial, so it should be easy to remove the peer manager from there.

Let's leave the quorum as is for now: That will soon become obsolete anyway, with RFC 37.

---

*[src/states/joining_node.rs, line 253 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNW0E5yWBkCOAxH61By:-KNW0E5yWBkCOAxH61Bz:1641283387) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/states/joining_node.rs#L253)):*
> ```Rust
>                           self,
>                           peer_id);
>                     common::disconnect_peer(self, &self.crust_service, &self.peer_mgr, &peer_id);
> ```

`disconnect_peer` should probably be a method of `Node`. In `Client` and `JoiningNode` it only needs to check against our proxy: maybe not even worth a dedicated method there.

---

*[src/states/common/connect.rs, line 192 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNW1cFFj3LIbZ_Bbbdw:-KNW1cFFj3LIbZ_Bbbdx:1513173126) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/states/common/connect.rs#L192)):*
> ```Rust
>                      dst,
>                      Authority::ManagedNode(src_name))
>     }
> ```

These two methods should probably be merged, with the body of the `_from_client` one? (The names should agree for nodes.)
If you're not sure, or prefer to postpone that, please add a TODO.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

afck:
Reviewed 8 of 9 files at r8.
Review status: 19 of 20 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

Fraser999:
Review status: 13 of 20 files reviewed at latest revision, 12 unresolved discussions.

---

*[src/ack_manager.rs, line 64 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWBa0ZKzc4RKz3fLUS:-KNWBa0_R1bFUk8YU43u:1294884657) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/ack_manager.rs#L64)):*
> ```Rust
> 
>     pub fn add_to_pending(&mut self,
>                           ack: u64,
> ```

I wonder if it's worth considering a couple of things related to ack-handling:
1. should we have an `Ack` type (a simple wrapper for `u64`)?
2. should we have a free function in `AckManager` which calculates an `Ack` value from a `RoutingMessage`?
Neither are a big deal really - the ack-handling in bootstrapped.rs is limited to a single function just now anyway, but I was just thinking that moving as much of the ack-related stuff to ack_manager.rs as possible would be a slight improvement.

---

*[src/node.rs, line 63 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWPJRj2npQ8RDeWK5y:-KNWPJRk7a3_Y7karf4W:322668316) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/node.rs#L63)):*
> ```Rust
> 
>     /// Causes node creation to fail if another node on the local network is detected.
>     pub fn deny_other_local_nodes(self) -> NodeBuilder {
> ```

Is this function/setting still needed?

---

*[src/state_machine.rs, line 189 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWRFTn_52iUbS55zgE:-KNWRFTn_52iUbS55zgF:288677268) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/state_machine.rs#L189)):*
> ```Rust
> 
> impl StateMachine {
>     pub fn new<F>(f: F) -> (RoutingActionSender, Self)
> ```

I know it's very common in Rust, but I can't see the win in obfuscating variables and types by giving them single-letter names.  Couldn't this at least have a name like `initialise_state` or something to convey the intent?

---

*[src/states/client.rs, line 62 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWbxeql8o22Q3VjvqX:-KNWbxeql8o22Q3VjvqY:-690244969) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/states/client.rs#L62)):*
> ```Rust
> impl Client {
>     #[cfg_attr(feature = "clippy", allow(too_many_arguments))]
>     pub fn from_bootstrapping(crust_service: Service,
> ```

Are we avoiding just implementing the `From` trait to allow us to keep the member variable private?  Seems like a valid reason, but at the same time, these `from_xxx` function signatures are fairly cumbersome and require corresponding `into_xxx` functions (`into()` would be implied automatically if we implemented the `From` trait instead).

Pity we don't have `friend` in Rust - would suit this scenario I think.

---

*[src/states/common/bootstrapped.rs, line 79 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWEhd9KdAuyGiGe030:-KNWEhd9KdAuyGiGe031:190942816) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/states/common/bootstrapped.rs#L79)):*
> ```Rust
>             // GetNodeName response. Connect requests arent filtered which
>             // should get resolved with peer_mgr completion.
>             // GetNodeName response resends from a node needs to get looked into.
> ```

Are either of these two points fixed?  I guess I'd expect at least the first one to be resolved now.

---

*[src/states/common/proxy_client.rs, line 87 \[r8\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWnn6PwhN79eRGdug7:-KNWnn6Qr2ZHcOU2yhEV:-650874626) ([raw file](https://github.com/maidsafe/routing/blob/06bc8b359a135643f69a35185f7ba5ea837ef1ed/src/states/common/proxy_client.rs#L87)):*
> ```Rust
> 
>         // TODO(adam): remove this but make sure it's handled in JoiningNode.
>         // let _ = self.peer_mgr_mut().remove_peer(&peer_id);
> ```

I'm not really clear on this.  If we're a proxy client, shouldn't we only be connected to a single peer anyway?  What would `JoiningNode` need to handle that's not already handled here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

afck:
Reviewed 7 of 7 files at r9.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[src/states/joining_node.rs, line 486 \[r9\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWpxyH1WQDPWdV6lT-:-KNWpxyH1WQDPWdV6lT0:1257010778) ([raw file](https://github.com/maidsafe/routing/blob/ba17caeca83a72ab9ca51c2207edc2861f46267f/src/states/joining_node.rs#L486)):*
> ```Rust
>         let _ = self.peer_mgr.remove_peer(&peer_id);
> 
>         if *self.proxy_peer_id() == peer_id {
> ```

Will this still be true _after_ `peer_mgr.remove_peer` has been called?

---

*[src/states/joining_node.rs, line 491 \[r9\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWqBQNgf-KPH32Py6c:-KNWqBQNgf-KPH32Py6d:1942508882) ([raw file](https://github.com/maidsafe/routing/blob/ba17caeca83a72ab9ca51c2207edc2861f46267f/src/states/joining_node.rs#L491)):*
> ```Rust
>                    self.proxy_public_id().name(),
>                    peer_id);
>             self.send_event(Event::Terminate);
> ```

Didn't we originally send a `RestartRequired` in this case?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

afck:
Reviewed 3 of 3 files at r10.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

madadam:
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[src/node.rs, line 63 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWPJRj2npQ8RDeWK5y:-KNX9FrhSmTmkXjoeNoC:-553791253) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/node.rs#L63)):*
> Is this function/setting still needed?
</details>
Not sure. @afck?

---

*[src/state_machine.rs, line 117 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNVzmGuvMpdtAhjM6jb:-KNX9QqXKOl5xPXeXLFH:1492973203) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/state_machine.rs#L117)):*
> Could all these `match` statements be simplified by having only one `get_state(&self) -> &mut State` method, returning a fat pointer, so you could e.g. just write `self.get_state().handle_crust_event(event)` here?
> (But then we'd need a `Terminated` state implementation, so maybe it's not worth it.)
</details>
Yeah, I'm not convinced it would be worth it.

---

*[src/states/client.rs, line 62 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWbxeql8o22Q3VjvqX:-KNX5BAvguLwnXqouKRV:-1442246523) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/states/client.rs#L62)):*
> Are we avoiding just implementing the `From` trait to allow us to keep the member variable private?  Seems like a valid reason, but at the same time, these `from_xxx` function signatures are fairly cumbersome and require corresponding `into_xxx` functions (`into()` would be implied automatically if we implemented the `From` trait instead).
> 
> Pity we don't have `friend` in Rust - would suit this scenario I think.
</details>
Yes, I wanted to keep the members private. 

---

*[src/states/joining_node.rs, line 486 \[r9\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWpxyH1WQDPWdV6lT-:-KNXARaL1D4qMTcl1Kfo:1676583839) ([raw file](https://github.com/maidsafe/routing/blob/ba17caeca83a72ab9ca51c2207edc2861f46267f/src/states/joining_node.rs#L486)):*
> Will this still be true _after_ `peer_mgr.remove_peer` has been called?
</details>
Sorry, not sure what you mean here?

---

*[src/states/joining_node.rs, line 491 \[r9\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWqBQNgf-KPH32Py6c:-KNXAIbVtgzCC9e60zdT:1549963639) ([raw file](https://github.com/maidsafe/routing/blob/ba17caeca83a72ab9ca51c2207edc2861f46267f/src/states/joining_node.rs#L491)):*
> Didn't we originally send a `RestartRequired` in this case?
</details>
Just checked in master and no, we do send `Terminate` there.

---

*[src/states/common/bootstrapped.rs, line 79 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWEhd9KdAuyGiGe030:-KNX67aATuW-ZzgP6CSB:448249768) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/states/common/bootstrapped.rs#L79)):*
> Are either of these two points fixed?  I guess I'd expect at least the first one to be resolved now.
</details>
Not sure. @afck? @maqi?

---

*[src/states/common/proxy_client.rs, line 87 \[r8\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWnn6PwhN79eRGdug7:-KNX6eYw-DUqnbRcXKYq:-1284339819) ([raw file](https://github.com/maidsafe/routing/blob/06bc8b359a135643f69a35185f7ba5ea837ef1ed/src/states/common/proxy_client.rs#L87)):*
> I'm not really clear on this.  If we're a proxy client, shouldn't we only be connected to a single peer anyway?  What would `JoiningNode` need to handle that's not already handled here?
</details>
This is already fixed in one of the newer commits.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

afck:
Review status: 13 of 20 files reviewed at latest revision, 12 unresolved discussions.

---

*[src/node.rs, line 63 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWPJRj2npQ8RDeWK5y:-KNXJltydo8HBdY38arJ:-212181484) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/node.rs#L63)):*
> Not sure. @afck?
</details>
Yes: E.g. in the routing ci_test, we want to allow other local nodes, but int safe_vault we don't.
(I'd rather not make this yet another conditionally compiled feature.)

---

*[src/states/joining_node.rs, line 486 \[r9\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWpxyH1WQDPWdV6lT-:-KNXKNkYv6L6wMxzv8-C:-191859300) ([raw file](https://github.com/maidsafe/routing/blob/ba17caeca83a72ab9ca51c2207edc2861f46267f/src/states/joining_node.rs#L486)):*
> Sorry, not sure what you mean here?
</details>
Ah, sorry, it's fine. I thought `remove_peer` would also remove the proxy, but it doesn't.

---

*[src/states/common/bootstrapped.rs, line 79 \[r6\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNWEhd9KdAuyGiGe030:-KNXKwrPVqeD9XirUeH-:-408849883) ([raw file](https://github.com/maidsafe/routing/blob/6e8b9ddd6b6512294647887a5f942fd5ee03c992/src/states/common/bootstrapped.rs#L79)):*
> Not sure. @afck? @maqi?
</details>
Not sure. Especially the second point is rather vague.
I don't see any `Ejected` messages in my current log file, though, so I'm fin with removing the FIXME, until this problem comes up again?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

afck:
Reviewed 7 of 7 files at r11.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

Fraser999:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[src/state_machine.rs, line 321 \[r11\]](https://reviewable.io:443/reviews/maidsafe/routing/1102#-KNaNPGH9zxpf-2yo596:-KNaNPGIwY2qu3N80Vk1:889555000) ([raw file](https://github.com/maidsafe/routing/blob/b522d2c31fddff7ae7164a6f2b943b79157e141a/src/state_machine.rs#L321)):*
> ```Rust
>     }
> 
>     fn into_bootstrapped(&mut self,
> ```

Clippy's complaining here that functions called `into_*` should take `self` by value.  Same for `into_node` just below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

afck:
Reviewed 9 of 10 files at r12.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

Fraser999:
A general point I just noticed about the traits' names - should we maybe make the following name changes:
* `Connect` to `Connected`
* `HandleHopMessage` to `HopMessageHandler`
* `HandleUserMessage` to `UserMessageHandler`
* `SendRoutingMessage` to `RoutingMessageSender`

---

Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102#-:-KNb2KY74Vqt3kEU_L6a:-539718545)*

maqi:
Reviewed 1 of 9 files at r8, 2 of 7 files at r11, 9 of 10 files at r12.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

Fraser999:
Reviewed 4 of 20 files at r1, 2 of 6 files at r4, 2 of 9 files at r5, 1 of 9 files at r8, 2 of 7 files at r11, 9 of 10 files at r12.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1102)*

PR #1115, opened 2016-07-28
WIP feat/routing: use node key as name

maqi:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1115)

### DIFF
--- a/src/state_machine.rs
+++ b/src/state_machine.rs
@@ -56,26 +56,32 @@ use xor_name::XorName;
 /// ## Becoming a node
 ///
 /// If A wants to become a full routing node (`client_restriction == false`), it needs to relocate,
-/// i. e. change its name to a value chosen by the network, and then add its peers to its routing
-/// table and get added to their routing tables.
+/// i. e. change its name to a value that sits in the range chosen by the network, and then add its
+/// peers to its routing table and get added to their routing tables.
 ///
 ///
-/// ### Getting a new network name from the `NaeManager`
+/// ### Getting an identity range from the `NaeManager`
 ///
-/// Once in `Client` state, A sends a `GetNodeName` request to the `NaeManager` group authority X
-/// of A's current name. X computes a new name and sends it in an `ExpectCloseNode` request to  the
-/// `NaeManager` Y of A's new name. Each member of Y caches A's public ID, and Y sends a
-/// `GetNodeName` response back to A, which includes the public IDs of the members of Y.
+/// Once in `Client` state, A sends a `GetNameRange` request to the `NaeManager` group authority X
+/// of A's current name. X computes a new name and sends it in an `ExpectCloseNode` request to the
+/// `NaeManager` Y of A's possible new name. Each member of Y caches node A's ID, sends a `GetNameRange`
+/// response back to A, which includes the public IDs of the members of Y.
 ///
+///
+/// ### Computing the new identity
+///
+/// Once A accumulates the `GetNameRange` response (the quorum_size shall be the same as the previous received
+/// from `BootstrapIdentify`), it starts generating a key pair whose public_key falls into the give range. And such
+/// public_key will then be used as node A's new name address. Such computing must be completed in a given period,
+/// otherwise a new round of connecting shall be understaken.
 ///
 /// ### Connecting to the close group
 ///
-/// To the `ManagedNode` for each public ID it receives from members of Y, A sends its
-/// `ConnectionInfo`. It also caches the ID.
+/// For each public ID it receives from members of Y, A sends its `ConnectionInfo`.
 ///
-/// For each `ConnectionInfo` that a node Z receives from A, it decides whether it wants A in its
-/// routing table. If yes, and if A's ID is in its ID cache, Z sends its own `ConnectionInfo` back
-/// to A and also attempts to connect to A via Crust. A does the same, once it receives the
+/// For each `ConnectionInfo` that a node Z(members of Y) receives from A, it decides whether it wants A in its
+/// routing table. If yes, and if A's ID is in its ID cache(node A's original ID), Z sends its own
+/// `ConnectionInfo` back to A and also attempts to connect to A via Crust. A does the same, once it receives the
 /// `ConnectionInfo`.
 ///
 /// Once the connection between A and Z is established and a Crust `OnConnect` event is raised,
--- a/src/states/joining_node.rs
+++ b/src/states/joining_node.rs
@@ -22,7 +22,7 @@ use kademlia_routing_table::RoutingTable;
 use maidsafe_utilities::serialisation;
 use std::fmt::{self, Debug, Formatter};
 use std::sync::mpsc::Sender;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use ack_manager::{Ack, AckManager};
 use action::Action;
@@ -44,11 +44,12 @@ use super::common::Testable;
 use super::Node;
 use timer::Timer;
 use types::MessageId;
-#[cfg(feature = "use-mock-crust")]
 use xor_name::XorName;
 
-/// Time (in seconds) after which a `GetNodeName` request is resent.
-const GET_NODE_NAME_TIMEOUT_SECS: u64 = 60;
+/// Time (in seconds) after which a `GetNameRange` request is resent.
+const GET_NAME_RANGE_TIMEOUT_SECS: u64 = 60;
+/// Time (in seconds) for the duration to compute a valid new id.
+const COMPUTING_IDENTITY_TIMEOUT_SECS: u64 = 30;
 
 pub struct JoiningNode {
     ack_mgr: AckManager,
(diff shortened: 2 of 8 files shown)


maidsafe-highfive:
r? @afck

afck:
Review status: 0 of 7 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/xor_name.rs, line 126 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1115#-KNpDFlx7zVbqbHScqeY:-KNpDFlx7zVbqbHScqeZ:-1884600603) ([raw file](https://github.com/maidsafe/routing/blob/ac1402619ccd42445e4144879b38041e14ff0805/src/xor_name.rs#L126)):*
> ```Rust
>         let mut distance = XorName([0u8; XOR_NAME_LEN]);
>         for i in 0..XOR_NAME_BITS {
>             distance.0[i] = self.0[i] ^ target.0[i];
> ```

That's not minus, that's xor. But for the interval length computation we'll need minus. Or rather: the _absolute value of the difference_. E.g. compute `self - target` if `self >= target`, otherwise return `target.minus(self)` instead.
And maybe it should be called `abs_diff` then, or `euclidean_distance`.

---

*[src/states/node.rs, line 979 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1115#-KNpCP_nrGU16vLOMb8K:-KNpCP_o8QeA3BJara5F:-1183428144) ([raw file](https://github.com/maidsafe/routing/blob/ac1402619ccd42445e4144879b38041e14ff0805/src/states/node.rs#L979)):*
> ```Rust
>         println!("input nodes {:?}", nodes);
>         let name = XorName([0u8; XOR_NAME_LEN]);
>         nodes.sort_by(|node0, node1| name.cmp_distance(&node0.name(), &node1.name()));
> ```

Doesn't `XorName` implement `Ord` anyway? (If not: it should, I think.) So you could probably just call `nodes.sort()`.

---

*[src/states/node.rs, line 983 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1115#-KNpClgKMFiJ-UfQ1vSH:-KNpClgKMFiJ-UfQ1vSI:-1870617392) ([raw file](https://github.com/maidsafe/routing/blob/ac1402619ccd42445e4144879b38041e14ff0805/src/states/node.rs#L983)):*
> ```Rust
>         let mut max_range = XorName([0u8; XOR_NAME_LEN]);
>         let mut range_index = 0;
>         for i in 0..(nodes.len() - 1) {
> ```

Once Disjoint Groups are in place, this will need to include the lower and upper bound of the group's range (if there are no nodes with exactly those names).
Let's either add a TODO or hold off with merging this PR until after RFC 37 is implemented.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1115)*

afck:
Review status: 0 of 7 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/xor_name.rs, line 126 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1115#-KNpDFlx7zVbqbHScqeY:-KNpJrlUmOuczijSubeM:903638276) ([raw file](https://github.com/maidsafe/routing/blob/ac1402619ccd42445e4144879b38041e14ff0805/src/xor_name.rs#L126)):*
> That's not minus, that's xor. But for the interval length computation we'll need minus. Or rather: the _absolute value of the difference_. E.g. compute `self - target` if `self >= target`, otherwise return `target.minus(self)` instead.
> And maybe it should be called `abs_diff` then, or `euclidean_distance`.
</details>
Actually, ignore the "absolute" part of my comment: Just implementing subtraction should be fine, since you sort the names below anyway, so you always subtract the smaller from the larger one.
I think we should just `impl std::ops::Sub for XorName` then: implement subtraction for 256-bit integers.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1115)*

maqi:
*[src/states/node.rs, line 979 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1115#-KNpCP_nrGU16vLOMb8K:-KO8zvmEnH1-u1nvYAOQ:579645314) ([raw file](https://github.com/maidsafe/routing/blob/ac1402619ccd42445e4144879b38041e14ff0805/src/states/node.rs#L979)):*
> Doesn't `XorName` implement `Ord` anyway? (If not: it should, I think.) So you could probably just call `nodes.sort()`.
</details>
we can call nodes.sort(), but it doesn't change the order (or it sorted in the decreasing order?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1115)*

afck:
Review status: 0 of 7 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/states/node.rs, line 979 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1115#-KNpCP_nrGU16vLOMb8K:-KO9-lMctBduu2HDHh2F:192488956) ([raw file](https://github.com/maidsafe/routing/blob/ac1402619ccd42445e4144879b38041e14ff0805/src/states/node.rs#L979)):*
> we can call nodes.sort(), but it doesn't change the order (or it sorted in the decreasing order?)
</details>
Ah, sorry, it is `PublicId` that needs to implement `Ord` - and that should definitely go by _name_. Maybe it will, if we just make `name` the first field in `PublicId`? If not, `Ord` needs to be implemented by hand.

In either case, please add a test to `id.rs` to make sure that public IDs are ordered by name.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1115)*

src/state_machine.rs
```diff
@@ -56,26 +56,32 @@ use xor_name::XorName;
...
 ///
-/// To the `ManagedNode` for each public ID it receives from members of Y, A sends its
-/// `ConnectionInfo`. It also caches the ID.
+/// For each public ID it receives from members of Y, A sends its `ConnectionInfo`.
 ///
-/// For each `ConnectionInfo` that a node Z receives from A, it decides whether it wants A in its
-/// routing table. If yes, and if A's ID is in its ID cache, Z sends its own `ConnectionInfo` back
-/// to A and also attempts to connect to A via Crust. A does the same, once it receives the
+/// For each `ConnectionInfo` that a node Z(members of Y) receives from A, it decides whether it wants A in its
+/// routing table. If yes, and if A's ID is in its ID cache(node A's original ID), Z sends its own
```
dhardy: I don't follow... why would Z have A's original ID? For the previous line, it would be nicer if your addition was spelled out: "node Z (of `NaeManager` Y)"

src/states/joining_node.rs
```diff
@@ -44,19 +44,20 @@ use super::common::Testable;
...
 use types::MessageId;
-#[cfg(feature = "use-mock-crust")]
 use xor_name::XorName;
 
-/// Time (in seconds) after which a `GetNodeName` request is resent.
-const GET_NODE_NAME_TIMEOUT_SECS: u64 = 60;
+/// Time (in seconds) after which a `GetNameRange` request is resent.
+const GET_NAME_RANGE_TIMEOUT_SECS: u64 = 60;
+/// Time (in seconds) for the duration to compute a valid new id.
+const COMPUTING_IDENTITY_TIMEOUT_SECS: u64 = 30;
```
dhardy: Shorter timeout for a job requiring actual work than for a routing response? What if some network packets get dropped — it has to start again?

src/states/node.rs
```diff
@@ -1487,15 +1513,17 @@ impl Bootstrapped for Node {
...
+        debug!("{:?} Handling NodeIdentify from {:?}. as node",
                self,
                public_id.name());
-
-        if let Some((name, _)) = self.sent_network_name_to {
-            if name == *public_id.name() {
-                self.sent_network_name_to = None;
-            }
-        }
+        // TODO: unless Disjoin Group being implemented, it is not guaranteered cache exists
```
dhardy: Spelling: "Disjoint", "guaranteed"

src/states/joining_node.rs
```diff
@@ -361,6 +360,22 @@ impl JoiningNode {
             let _ = self.crust_service.disconnect(*peer_id);
         }
     }
+
+    fn compute_new_id(&mut self, name_range: &(XorName, XorName)) -> Result<(), RoutingError> {
+        let now = Instant::now();
+        loop {
+            if now.elapsed() > Duration::from_secs(COMPUTING_IDENTITY_TIMEOUT_SECS) {
```
dhardy: Lets hope the client/joining node is not busy, or it might not manage many iterations. Often this type of code allows a maximum number of tries instead of a timeout.

src/state_machine.rs
maqi: the group X has notified group Y about A's original id

Viv-Rajkumar:
closing this for now. lets get this raised and merged when switching the node name pattern with node ageing.

PR #1123, opened 2016-08-04
refactor/node-state: simplify use of peer manager in node state

madadam:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1123)

### DIFF
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -83,9 +83,9 @@ pub enum PeerState {
     /// We are connected - via a tunnel if the field is `true` - and waiting for a `NodeIdentify`.
     AwaitingNodeIdentify(bool),
     /// We are the proxy for the client
-    Client(sign::PublicKey),
+    Client,
     /// We are the proxy for the joining node
-    JoiningNode(sign::PublicKey),
+    JoiningNode,
     /// We are connected and routing to that peer - via a tunnel if the field is `true`.
     Routing(bool),
     /// We are connected to the peer who is our proxy node.
@@ -172,11 +172,11 @@ impl Peer {
             PeerState::SearchingForTunnel => {
                 self.timestamp.elapsed() >= Duration::from_secs(CONNECTION_TIMEOUT_SECS)
             }
-            PeerState::JoiningNode(_) |
+            PeerState::JoiningNode |
             PeerState::Proxy => {
                 self.timestamp.elapsed() >= Duration::from_secs(JOINING_NODE_TIMEOUT_SECS)
             }
-            PeerState::Client(_) |
+            PeerState::Client |
             PeerState::Routing(_) |
             PeerState::AwaitingNodeIdentify(_) => false,
         }
@@ -302,9 +302,7 @@ impl PeerManager {
                                 -> Option<AddedNodeDetails<XorName>> {
         let result = self.routing_table.add(*pub_id.name());
         if result.is_some() {
-            let tunnel = match self.peer_map
-                .remove_by_name(pub_id.name())
-                .map(|peer| peer.state) {
+            let tunnel = match self.peer_map.remove(&peer_id).map(|peer| peer.state) {
                 Some(PeerState::SearchingForTunnel) |
                 Some(PeerState::AwaitingNodeIdentify(true)) => true,
                 Some(PeerState::Routing(tunnel)) => {
@@ -341,6 +339,17 @@ impl PeerManager {
             (Some(&PeerState::Routing(false)), Some(&PeerState::Routing(false))) => true,
             _ => false,
         }
+    }
+
+    /// Returns the public ID of the given peer, if it is in `Routing` state.
+    pub fn get_routing_peer(&self, peer_id: &PeerId) -> Option<&PublicId> {
+        self.peer_map.get(peer_id).and_then(|peer| {
+            if let PeerState::Routing(_) = peer.state {
+                Some(&peer.pub_id)
+            } else {
+                None
+            }
+        })
     }
 
     /// Returns the proxy node, if connected.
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @afck

afck:
Reviewed 2 of 2 files at r1.

---

*[src/peer_manager.rs, line 162 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1123#-KOOonbqtfERqzT3ETQj:-KOOonbrZ7Nvxtj__zHu:bpzgwc4) ([raw file](https://github.com/maidsafe/routing/blob/8be5421d754964387d75d6bc8bfa1a0670ed3157/src/peer_manager.rs#L162)):*
> ```Rust
>         match self.state {
>             PeerState::Client(ref pub_key) => pub_key,
>             PeerState::JoiningNode(ref pub_key) => pub_key,
> ```

Clippy is not amused: `this `match` has identical arm bodies`

---

*[src/peer_manager.rs, line 522 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1123#-KOOpeXRUNBiaFvPRMY9:-KOOpeXRUNBiaFvPRMYA:byj89zk) ([raw file](https://github.com/maidsafe/routing/blob/8be5421d754964387d75d6bc8bfa1a0670ed3157/src/peer_manager.rs#L522)):*
> ```Rust
>     }
> 
>     pub fn get_connected_peer(&self, peer_id: &PeerId) -> Option<&Peer> {
> ```

Let's strive for documenting all new public items, even if they are not exported from the crate.

---

*[src/states/node.rs, line 1651 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1123#-KOOtHQjf3TTxemJNdyN:-KOOtHQkSSqU8iBGUqLS:b5r70qp) ([raw file](https://github.com/maidsafe/routing/blob/8be5421d754964387d75d6bc8bfa1a0670ed3157/src/states/node.rs#L1651)):*
> ```Rust
> 
>         if let Some(details) = dropped_node_details {
>             if !self.dropped_routing_node(*peer.name(), details) {
> ```

So ... `dropped_routing_node` returns `false` if we have to disconnect, does it? A short doc comment for that and `dropped_peer`'s return values would make this a lot easier to understand.

---

*[src/states/node.rs, line 1671 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1123#-KOOuBzC1ANJr2u2FymF:-KOOuBzDCObpTNGqLA3Q:btiqr1s) ([raw file](https://github.com/maidsafe/routing/blob/8be5421d754964387d75d6bc8bfa1a0670ed3157/src/states/node.rs#L1671)):*
> ```Rust
>                        peer.name(),
>                        peer_id);
>                 self.send_event(Event::Terminate);
> ```

If my reading is correct, we'll end up here if we lost our proxy node which hasn't transitioned into our routing table yet - but in a large network the proxy node is likely to never do that and at some point will drop us. I guess this needs a check for whether the routing table has `GROUP_SIZE` entries.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1123)*

madadam:

---

*[src/peer_manager.rs, line 162 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1123#-KOOonbqtfERqzT3ETQj:-KOOwUEwUZN4f3_ghtyw:b-lqzqh5) ([raw file](https://github.com/maidsafe/routing/blob/8be5421d754964387d75d6bc8bfa1a0670ed3157/src/peer_manager.rs#L162)):*
> Clippy is not amused: `this `match` has identical arm bodies`
</details>
Right. Actually, why do we even need to keep those pub_keys as members in there? Are they ever different from `pub_id.signing_public_key`? If not, we can get rid of this method (or at least simplify it to a one-liner).

---

*[src/states/node.rs, line 1671 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1123#-KOOuBzC1ANJr2u2FymF:-KOOx2Iifs7yS8qiadcK:bissvuk) ([raw file](https://github.com/maidsafe/routing/blob/8be5421d754964387d75d6bc8bfa1a0670ed3157/src/states/node.rs#L1671)):*
> If my reading is correct, we'll end up here if we lost our proxy node which hasn't transitioned into our routing table yet - but in a large network the proxy node is likely to never do that and at some point will drop us. I guess this needs a check for whether the routing table has `GROUP_SIZE` entries.
</details>
So terminate only if routing table size is < `GROUP_SIZE`, correct?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1123)*

afck:

---

*[src/peer_manager.rs, line 162 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1123#-KOOonbqtfERqzT3ETQj:-KOOxwMHNk1mZkEkp8m4:bx93oez) ([raw file](https://github.com/maidsafe/routing/blob/8be5421d754964387d75d6bc8bfa1a0670ed3157/src/peer_manager.rs#L162)):*
> Right. Actually, why do we even need to keep those pub_keys as members in there? Are they ever different from `pub_id.signing_public_key`? If not, we can get rid of this method (or at least simplify it to a one-liner).
</details>
You're right, since `Peer` contains that now anyway, it doesn't need to be in `PeerState` anymore. :+1: 
We just need to make sure that if a client joins the routing table, we update their name. (I.e. use the new ID from the `NodeIdentify` message.)

---

*[src/states/node.rs, line 1671 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1123#-KOOuBzC1ANJr2u2FymF:-KOOyWavq21zGwYbYppT:bgzdc3x) ([raw file](https://github.com/maidsafe/routing/blob/8be5421d754964387d75d6bc8bfa1a0670ed3157/src/states/node.rs#L1671)):*
> So terminate only if routing table size is < `GROUP_SIZE`, correct?
</details>
Yes, I think that's a reasonable criterion for now.
(It will probably change with the Disjoint Nodes feature, though.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1123)*

afck:
Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1123)*

PR #1125, opened 2016-08-05
Keep track of peers from whom we're awaiting a node identify

andrebeat:
After receiving a `ConnectSuccess` event the peer is added to a list of pending `NodeIdentify` messages, if after a given amount time this message isn't received we disconnect from the peer.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1125)

### DIFF
--- a/src/states/node.rs
+++ b/src/states/node.rs
@@ -1339,7 +1339,7 @@ impl Node {
         while let Some(&(node_identify_token, peer_id)) = self.pending_node_identify.front() {
             if node_identify_token < token {
                 let _ = self.pending_node_identify.pop_front();
-                debug!("{:?} Unhandled pending `NodeIdentify` timeout for peer {} with timer \
+                debug!("{:?} Unhandled pending `NodeIdentify` timeout for {:?} with timer \
                         token {}.",
                        self,
                        peer_id,
@@ -1347,7 +1347,7 @@ impl Node {
             } else {
                 if node_identify_token == token {
                     let _ = self.pending_node_identify.pop_front();
-                    debug!("{:?} Timed out waiting for `NodeIdentify` from {}.",
+                    debug!("{:?} Timed out waiting for `NodeIdentify` from {:?}.",
                            self,
                            peer_id);
                     self.disconnect_peer(&peer_id);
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -59,6 +60,8 @@ const TICK_TIMEOUT_SECS: u64 = 60;
 const GET_NODE_NAME_TIMEOUT_SECS: u64 = 60;
 /// Time (in seconds) the new close group waits for a joining node it sent a network name to.
 const SENT_NETWORK_NAME_TIMEOUT_SECS: u64 = 30;
+/// Time (in seconds) the node waits for a `NodeIdentify` after a `ConnectSuccess`.
+const NODE_IDENTIFY_TIMEOUT_SECS: u64 = 60;
```
andrebeat: I'm not sure about this value.

andrebeat:
Is [this](https://github.com/andrebeat/routing/blob/59ea51c90a51140e8236a173e57d1f9c537a202f/src/states/node.rs#L356) comment still relevant?

afck:
Sorry, I missed updating this TODO: The `PeerManager` already keeps track of nodes awaiting a `NodeIdentify` message, but I think they currently don't time out. They probably should. However, we can't just remove them from the peer manager and forget about them, since we are already connected: When we remove them, we need to call `crust::Service::disconnect` to drop the connection.

The comment in line 356 is still relevant. The peer manager currently stores peers by name, and if we get a connection from a peer whose public ID (node name and public keys) we don't have yet, we currently just ignore it and hope that we will soon receive a `NodeIdentify` anyway, containing the public ID, so we can put the peer in the map.

Probably `connected_to` shouldn't return anything at all. If it doesn't know the peer yet, it should put it in a new container (maybe `unknown_peers: HashMap<PeerId, Instant>`?).
Then we'll need a new method in the peer manager (`remove_expired_connections(&mut self) -> Vec<PeerId>`?) that removes both timed out peers in `AwaitingNodeIdentify` state and in the `unknown_peers` map, and returns their IDs. `src/states/node.rs` should call that function periodically and then call `crust::Service::disconnect` with all the removed peers.

andrebeat:
No problem! I did a quick read of `peer_manager.rs` the other day but I didn't notice/remember that. On second thought it makes sense that that logic exists in `PeerManager` and not `src/states/node.rs`. I'll close this PR and if you think its appropriate I'd like to implement what you suggested. 👍

afck:
Thank you! Yes, it would be great if you could give it a shot!

PR #1126, opened 2016-08-08
WIP Disjoint groups

maqi:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1126)


maidsafe-highfive:
r? @afck

src/routing_table/routing_table.rs
```diff
@@ -0,0 +1,561 @@
...
+    // Returns the name of the destination, i.e. the node or group address.
+    pub fn name(&self) -> &N {
+        match *self {
+            Destination::Group(ref name, _) |
+            Destination::Node(ref name) => name,
+        }
+    }
+
+    // Returns `true` if the destination os a group, and `false` if it is an individual node.
+    pub fn is_group(&self) -> bool {
```
dirvine: Here it may be worth adding `is_node' although just a negate of `is_group` it makes code easier to read.

src/routing_table/routing_table.rs
```diff
@@ -0,0 +1,561 @@
...
+
+    // Adds a contact to the routing table, or updates it.
+    //
+    // Returns `None` if the contact already existed or was denied (see `allow_connection`).
+    // Otherwise it returns `AddedNodeDetails`.
+    pub fn add(&mut self, info: T) -> Option<AddedNodeDetails<T>> {
+        if info.name() == self.our_name() {
+            return None;
+        }
+        match self.search(info.name()) {
```
dirvine: search->find perhaps

afck:
Reviewed 1 of 8 files at r1, 1 of 6 files at r2.
Review status: 2 of 8 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/xor_name.rs, line 172 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOdN4n5I1z-XAD-xsbm:-KOdN4n5I1z-XAD-xsbn:b-fjkrl1) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/xor_name.rs#L172)):*
> ```Rust
> }
> 
> impl fmt::Binary for XorName {
> ```

Good idea! We should probably also implement `LowerHex` and `UpperHex` at some point, and use those trait implementations in `Display` and `Debug`. (Probably not in this PR, though.)

---

*[src/routing_table/contact_info.rs, line 21 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOdO3RJqBlgGJzfabNs:-KOdO3RJqBlgGJzfabNt:b-iv33xi) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/contact_info.rs#L21)):*
> ```Rust
> 
> /// Contact info about a node in the network.
> pub trait ContactInfo: Clone + Eq {
> ```

I wonder whether we shouldn't just get rid of this, since the routing table now only stores node names anyway.

---

*[src/routing_table/mod.rs, line 41 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOdOT4AOm9bYem5W4qO:-KOdOT4BvbwaH2p9HuzV:bw4x4s3) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/mod.rs#L41)):*
> ```Rust
> #![cfg_attr(feature="clippy", deny(clippy, unicode_not_nfc, wrong_pub_self_convention,
>                                    option_unwrap_used))]
> #![cfg_attr(feature="clippy", allow(use_debug))]
> ```

All these attributes (from docs to clippy) should be unnecessary, since it's a module now and not a `lib.rs`.

---

*[src/routing_table/mod.rs, line 109 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOdP_K78Q3d7mHM25W2:-KOdP_K8hYMBrhYKDkEZ:bd9yo8) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/mod.rs#L109)):*
> ```Rust
> //   address.
> // * There are `bucket_size` different paths along which a message can be sent, to provide
> //   redundancy.
> ```

Let's not forget to update the documentation before we merge this.

---

*[src/routing_table/mod.rs, line 138 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOdPV4RFjzpurNghaE8:-KOdPV4SQJMMQwKAhlBv:b-gk8okf) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/mod.rs#L138)):*
> ```Rust
> 
> // TODO - remove this
> #![allow(unused)]
> ```

Maybe this also isn't necessary anymore.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1126)*

afck:
Review status: 2 of 8 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[src/routing_table/network_test.rs, line 572 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOdyU_xjyZkfLtDpFpB:-KOdyU_yVlj_xwfiOxTv:b-nnaje5) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/network_test.rs#L572)):*
> ```Rust
> 
> #[test]
> fn number_of_nodes_close_to_any_name_is_equal_to_group_size() {
> ```

I guess this test makes still sense, if we just replace it with `greater_or_equal`.

---

*[src/routing_table/network_test.rs, line 601 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOdygQEcvnVF33e20Qe:-KOdygQEcvnVF33e20Qf:b-56n4xs) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/network_test.rs#L601)):*
> ```Rust
> 
> #[test]
> fn nodes_in_close_group_of_any_name_are_connected_to_each_other() {
> ```

In addition, maybe we should test now that the close group of each member of any given address' close group is the same close group.

---

*[src/routing_table/result.rs, line 35 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOdz4ASHgwgxm2q2B7x:-KOdz4AT182nZ5fsYxni:b-n251kt) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/result.rs#L35)):*
> ```Rust
>     // `Some(i)` if the removal of the entry causing a merge into bucket with index `i`
>     pub merged_bucket: Option<usize>,
> }
> ```

If these return values become that simple again - which is great! -, we could get rid of these types again and just use the `Option`/`Vec` itself.

---

*[src/routing_table/routing_table.rs, line 67 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOe-6C_owXsCtBLCsl5:-KOe-6CamxbIw0n_li2w:b-7iw44a) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L67)):*
> ```Rust
>     }
> 
>     // Returns `true` if the destination os a group, and `false` if it is an individual node.
> ```

`os` -> `is`

---

*[src/routing_table/routing_table.rs, line 192 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOe0b8rDU_cse86k6SW:-KOe0b8rDU_cse86k6SX:bd9o2vh) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L192)):*
> ```Rust
>                 self.merge_buckets(merge_index);
>                 result.merged_bucket = Some(merge_index);
>             }
> ```

If this works, I'm happy to merge it like this first. But let's add a TODO:
In a later PR, we need to implement the merge/split flows as updated in the RFC.

---

*[src/routing_table/routing_table.rs, line 204 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOe0tDwU81Ng5rRWmjG:-KOe0tDx8-pnwH_bShLn:b-ks8fn7) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L204)):*
> ```Rust
>     }
> 
>     // Check whether contains the specific name
> ```

`Checks whether the routing table contains the specified name.`

---

*[src/routing_table/routing_table.rs, line 223 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOe1d1V3ym1djQxdL2B:-KOe1d1W30sG5mbV7IIx:b-iodhye) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L223)):*
> ```Rust
>     // actually sufficient! In that case, `true` is returned if and only if we are among the
>     // `group_size` closest node to `name` in the network.
>     pub fn is_close(&self, name: &T::Name, group_size: usize) -> bool {
> ```

We will need to explicitly keep track of group prefixes: our own as well as the prefixes of the groups we are connected to.
Once that is implemented, this function only needs to return whether `name` is an extension of our group prefix.

---

*[src/routing_table/routing_table.rs, line 282 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOe1w2_NoZy8sRECWOF:-KOe1w2_NoZy8sRECWOG:b45ppnd) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L282)):*
> ```Rust
>             .take(n)
>             .cloned()
>             .collect()
> ```

I'm not sure whether this implementation will still work, if bucket `i` isn't the set of the peers with a common prefix of _exactly_ length `i` anymore?

---

*[src/routing_table/routing_table.rs, line 346 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOe2htELrh3m5pd0mFE:-KOe2htELrh3m5pd0mFF:b-oryg0) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L346)):*
> ```Rust
>     pub fn other_close_nodes(&self, name: &T::Name, group_size: usize) -> Option<Vec<T>> {
>         if self.is_close(name, group_size) {
>             Some(self.closest_nodes_to(name, group_size - 1, false))
> ```

Instead of `group_size` (which it doesn't need as a parameter anymore), this should just return our group, excluding ourselves.

---

*[src/routing_table/routing_table.rs, line 365 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOe2yvX_q1uh4F083Gp:-KOe2yvX_q1uh4F083Gq:b-vlczcd) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L365)):*
> ```Rust
>         match self.search(name) {
>             (_, Ok(_)) => true,
>             (_, Err(i)) => i < self.min_bucket_len || self.is_close_to_bucket_of(name),
> ```

I think the connection now must be allowed if and only if the node belongs to a group we are connected to.
The groups we are connected to will change only through group splits.
(If we don't make the group splits explicit in this PR, this might be a bit more complicated: We'd need to check whether the new node will _cause_ a group split, and if it does, we need to check whether the new node's new group prefix differs from ours in at most one bit.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1126)*

afck:
Review status: 2 of 8 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[src/routing_table/routing_table.rs, line 398 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOeCKjeDhw0BmmxfpdU:-KOeCKjeDhw0BmmxfpdV:b3k4dzt) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L398)):*
> ```Rust
>         let mut positions = vec![];
>         for bucket in &self.buckets[..last_bucket_index] {
>             match bucket.iter().position(|ref contact| self.our_name().differs_in_bit(contact.name(), last_bucket_index)) {
> ```

I'm not sure I understand: Why are we looking for the first entry that differs in the `last_bucket_index`-th bit from us?

---

*[src/routing_table/routing_table.rs, line 433 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOeCn1haYsH5f-ljVFV:-KOeCn1iTyrn8UCIPrUw:b-qqfxuw) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L433)):*
> ```Rust
>         if bucket_index + 2 < self.buckets.len() {
>             for bit_index in bucket_index + 1..self.buckets.len() - 1 {
>                 if self.our_name().differs_in_bit(name, bit_index) {
> ```

`name`'s group doesn't necessarily have a prefix of the same length. We need to iterate only to the minimum of our group prefix lengths.

---

*[src/routing_table/routing_table.rs, line 443 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOeD5z_dOVTInPk7ubp:-KOeD5zaRvEjC1dxvTE1:bsiqh0q) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L443)):*
> ```Rust
>     // Returns whether we are `bucket_size`-close to one of `name`'s bucket addresses or to `name`
>     // itself.
>     fn is_close_to_bucket_of(&self, name: &T::Name) -> bool {
> ```

This shouldn't be required anymore, with the new criterion: A node should be in our routing table if and only if its group prefix differs in at most one bit from ours.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1126)*

afck:
Review status: 2 of 9 files reviewed at latest revision, 30 unresolved discussions.

---

*[src/routing_table/routing_table.rs, line 100 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsbqOTVzujWf9HG2IC:-KOsbqOU3NNWSs3kRfsf:b-tyrlx4) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L100)):*
> ```Rust
>     }
> 
>     pub fn is_belong_to(&self, name: &XorName) -> bool {
> ```

`belongs_to`
(http://knowyourmeme.com/memes/all-your-base-are-belong-to-us :-D )

---

*[src/routing_table/routing_table.rs, line 110 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOscCrPIPq7O2AiNkmh:-KOscCrQZQaaEEvQEEbZ:bv0dss1) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L110)):*
> ```Rust
>     pub fn decrease_valid_bit_count(&mut self) {
>         self.valid_bit_count += 1;
>     }
> ```

I think conceptually these two methods are problematic: The fact that we store the prefix as a `XorName` and a bit count is an implementation detail. What it represents is _just the prefix_, i.e. just the first bits of that name. From that point of view it doesn't make sense to extend the prefix without specifying the next bit. For the same reason, I don't understand what `move_one_bit` is for.

I imagine the operations we'll need on the prefix will be something like:

* drop the last bit (i.e. decrease the valid bit count),
* append a 0 (increase the valid bit count and set the next bit to 0),
* append a 1 (increase the valid bit count and set the next bit to 1),
* flip a bit,
* `belongs_to` (already there),
* and probably count the differing bits of two prefixes.

---

*[src/routing_table/routing_table.rs, line 113 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsdeoARpTHBCa2yJ7X:-KOsdeoARpTHBCa2yJ7Y:b-krt0p0) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L113)):*
> ```Rust
> 
>     pub fn max_bucket_index(&self, name: &XorName) -> usize {
>         self.address.max_bucket_index(name)
> ```

That also isn't meaningful if that value is greater than `valid_bit_count`.

---

*[src/routing_table/routing_table.rs, line 141 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOse6kukLrGiksPh4NK:-KOse6kukLrGiksPh4NL:b-e20i9g) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L141)):*
> ```Rust
>     pub fn reset_prefix(&mut self, prefix: Prefix) {
>         self.prefix = prefix;
>     }
> ```

That method seems dangerous: Just changing the prefix without changing the group composition is almost never correct.

Maybe the group should just have `fn merge(self, other: Group) -> Group` and a `fn split(self) -> (Group, Group)` instead, that also merge/split up the members?

---

*[src/routing_table/routing_table.rs, line 143 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsetf6ngEFoZP1sg_M:-KOsetf6ngEFoZP1sg_N:bxkyeab) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L143)):*
> ```Rust
>     }
> 
>     pub fn get_prefix(&self) -> &Prefix {
> ```

It's more idiomatic to omit the `get_`: https://doc.rust-lang.org/style/style/naming/README.html

---

*[src/routing_table/routing_table.rs, line 151 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsez9Ajo-JpDNytZ3Z:-KOsez9Ajo-JpDNytZ3_:b-l3sri9) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L151)):*
> ```Rust
>     }
> 
>     pub fn is_belong_to(&self, name: &XorName) -> bool {
> ```

`belongs_to`

---

*[src/routing_table/routing_table.rs, line 169 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsfVzLOw67oZEiJcp6:-KOsfVzLOw67oZEiJcp7:b-iclnxm) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L169)):*
> ```Rust
>                 true
>             }
>         }
> ```

Maybe this should also check whether the `member` belongs to this group. It feels like it should be impossible to add a member that doesn't match the prefix.

---

*[src/routing_table/routing_table.rs, line 190 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsfpqC8fhxMm7RuEPc:-KOsfpqC8fhxMm7RuEPd:b20jiyu) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L190)):*
> ```Rust
>             Err(_) => false,
>         }
>     }
> ```

Does that method ever do anything? If that name exists, it replaces it with itself?
Unless I'm missing anything, this method should be removed.

---

*[src/routing_table/routing_table.rs, line 198 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsg3LcnG81XxupX3DK:-KOsg3LcnG81XxupX3DL:b7bip7e) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L198)):*
> ```Rust
>             Err(_) => None,
>         }
>     }
> ```

Same here: Without `ContactInfo`s, there's nothing to update anymore.

---

*[src/routing_table/routing_table.rs, line 204 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsg95IRQj20c_NJ9yu:-KOsg95IRQj20c_NJ9yv:b-rqgztz) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L204)):*
> ```Rust
>     }
> 
>     // Returns a new split off group or None if desn't need to split
> ```

`doesn't`

---

*[src/routing_table/routing_table.rs, line 243 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsh5x65uDOUOmLlhUT:-KOsh5x7XmrPILiRb9Nr:b-d1j1vj) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L243)):*
> ```Rust
> 
>     fn search(&self, name: &XorName) -> Result<usize, usize> {
>         self.members.binary_search_by(|member| member.cmp(name))
> ```

Isn't that equivalent with just `self.members.binary_search(name)`?
And this now assumes that the names are sorted by their value, not by their distance to _this node_, does it? Is this sorting useful anywhere? Do we need the sorting by distance from us anywhere in the implementation? If the answer to both is "no", maybe we should make the `members` a `HashSet` instead?

---

*[src/routing_table/routing_table.rs, line 253 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOshl5Y78iIZZ4T8oSI:-KOshl5Y78iIZZ4T8oSJ:buwhzx8) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L253)):*
> ```Rust
>         for (member_index, member) in self.members.iter().enumerate() {
>             let comma = if member_index == self.members.len() - 1 { "" } else { "," };
>             try!(writeln!(formatter, " {:08b} {} ", member, comma));
> ```

The `:08b` is probably temporary, for debugging with `u8`s, is it? Let's not forget to remove it before we merge the PR.

---

*[src/routing_table/routing_table.rs, line 326 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsirXNZGJ6hlooipzq:-KOsirXNZGJ6hlooipzr:b8u6sr) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L326)):*
> ```Rust
>             (Ok(bucket_index), Err(group_index)) => {
>                 let _ = self.buckets[bucket_index].1[group_index].add(info);
>                 self.split_last_bucket();
> ```

Let's not actually change the group prefixes at all here, just return whether we think we should split now - or not even that! We could just have a separate `should_split` method that can be called after this one.

I realised yesterday that the necessity to split can not only arise after adding a node, but also after another split! If we split `0` into `01` and `00` because `01` just got its 9th member, it could happen that `00` can already split further. So it might make sense to just keep that question separate from adding a node.

---

*[src/routing_table/routing_table.rs, line 346 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsjkM0ATYUy-nOqB0o:-KOsjkM0ATYUy-nOqB0p:bpesobn) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L346)):*
> ```Rust
>                 if self.buckets.len() - 1 > 0 {
>                     for i in 0..(self.buckets.len() - 1) {
>                         self.buckets[i].0.increase_valid_bit_count();
> ```

Ah, I understand how you're using that method now!
I still find it questionable, because it really uses an implementation detail of `Prefix`. It might be cleaner to recreate the other bucket's prefixes from our own here every time, even if that's a bit less efficient.
Or to not have them at all: The bucket index of a group (given by _its_ prefix) is easy to compute, as it's simply the number of common leading bits of the group's prefix and our own group's prefix.

---

*[src/routing_table/routing_table.rs, line 455 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsmOcafk3OCC1f4RwZ:-KOsmOcafk3OCC1f4Rw_:by5k93i) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L455)):*
> ```Rust
>                 if common_bits >= max_bucket_index {
>                     max_bucket_index = common_bits;
>                     closest_group_position = (bucket_index, group_index);
> ```

Instead of two indexes, we could just store a reference `&group` itself.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1126)*

maqi:
Review status: 2 of 7 files reviewed at latest revision, 32 unresolved discussions, some commit checks failed.

---

*[src/routing_table/routing_table.rs, line 128 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOcrAhXhjQeACbWuBJX-r2-128:-KOxqsWWYD3_ONsSzhK8:b-r0m2dd) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L128)):*
<details><summary><i>Previously, dirvine (David Irvine) wrote…</i></summary>
> search->find perhaps
</details>
`find` indicates we `will` get it?  :)

---

*[src/routing_table/routing_table.rs, line 282 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOe1w2_NoZy8sRECWOF:-KOxuS94rQCivECM8dVy:b-9up3lt) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L282)):*
> I'm not sure whether this implementation will still work, if bucket `i` isn't the set of the peers with a common prefix of _exactly_ length `i` anymore?
</details>
this function iterate through all the groups in all the buckets to figure out which group is the closest. 
so, the result shall be correct at least from our knowledge?

---

*[src/routing_table/routing_table.rs, line 346 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOe2htELrh3m5pd0mFE:-KOxuvv2gK1BmKMDqwlV:b-8hh56c) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L346)):*
> Instead of `group_size` (which it doesn't need as a parameter anymore), this should just return our group, excluding ourselves.
</details>
group_size is not used (keeping it just in case we may switch back to it later on)
the group in the last bucket `is` our group and myself is not in it.

---

*[src/routing_table/routing_table.rs, line 398 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOeCKjeDhw0BmmxfpdU:-KOxvPen2jvUbrUKBgj2:bn7evk7) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L398)):*
> I'm not sure I understand: Why are we looking for the first entry that differs in the `last_bucket_index`-th bit from us?
</details>
this is from the old implementation and has been updated yet.

---

*[src/routing_table/routing_table.rs, line 433 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOeCn1haYsH5f-ljVFV:-KOxvuw69rdHZkY0Nbi2:b-woiic4) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L433)):*
> `name`'s group doesn't necessarily have a prefix of the same length. We need to iterate only to the minimum of our group prefix lengths.
</details>
this is for `bucket` calculation, not reach `group` yet

---

*[src/routing_table/routing_table.rs, line 443 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOeD5z_dOVTInPk7ubp:-KOxvkE_P_V-FH5FZGRN:ba92gap) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L443)):*
> This shouldn't be required anymore, with the new criterion: A node should be in our routing table if and only if its group prefix differs in at most one bit from ours.
</details>
this is to figure out the `bucket_index` which is from `our` perspective

---

*[src/routing_table/routing_table.rs, line 110 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOscCrPIPq7O2AiNkmh:-KOxjj2z0sARTl4gSTRA:bv5zjz6) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L110)):*
> I think conceptually these two methods are problematic: The fact that we store the prefix as a `XorName` and a bit count is an implementation detail. What it represents is _just the prefix_, i.e. just the first bits of that name. From that point of view it doesn't make sense to extend the prefix without specifying the next bit. For the same reason, I don't understand what `move_one_bit` is for.
> 
> I imagine the operations we'll need on the prefix will be something like:
> 
> * drop the last bit (i.e. decrease the valid bit count),
> * append a 0 (increase the valid bit count and set the next bit to 0),
> * append a 1 (increase the valid bit count and set the next bit to 1),
> * flip a bit,
> * `belongs_to` (already there),
> * and probably count the differing bits of two prefixes.
</details>
when the prefix is used as bucket, i.e. the prefix is a slice of our address, drop/append prefix simply means increas/decreas the valid_bit_count,
as the `address` is already the full copy of our address with one bit flipped (not the last one).

`move_one_bit` is used for group split, whose prefix is inherent from parent group's prefix with the last bit flipped.

---

*[src/routing_table/routing_table.rs, line 141 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOse6kukLrGiksPh4NK:-KOxkyZgwNKUaIdLYOs7:bmczd4z) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L141)):*
> That method seems dangerous: Just changing the prefix without changing the group composition is almost never correct.
> 
> Maybe the group should just have `fn merge(self, other: Group) -> Group` and a `fn split(self) -> (Group, Group)` instead, that also merge/split up the members?
</details>
this method is not being used. i shall remove it.

---

*[src/routing_table/routing_table.rs, line 190 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsfpqC8fhxMm7RuEPc:-KOxnEvHkyxhIKrvJCbE:b-6cm9ef) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L190)):*
> Does that method ever do anything? If that name exists, it replaces it with itself?
> Unless I'm missing anything, this method should be removed.
</details>
well, this method was inherited from previous, i.e. the contact `T` also having additional info other than `name`.
however, with the new refactor, this method is no longer used and can now be removed.

---

*[src/routing_table/routing_table.rs, line 243 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsh5x65uDOUOmLlhUT:-KOxqDcVS1e0UAeynZue:b4jajns) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L243)):*
> Isn't that equivalent with just `self.members.binary_search(name)`?
> And this now assumes that the names are sorted by their value, not by their distance to _this node_, does it? Is this sorting useful anywhere? Do we need the sorting by distance from us anywhere in the implementation? If the answer to both is "no", maybe we should make the `members` a `HashSet` instead?
</details>
when T is not the name only, has to use binary_seach_by :)
i now changed it to binary_search.

yeah, I agree a `hashset` might be more suitable now. I will change it to `hashset` once all stuff has been done.

---

*[src/routing_table/routing_table.rs, line 326 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsirXNZGJ6hlooipzq:-KOxrL4p2o7Yn9_l3TSt:b-ait5jh) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L326)):*
> Let's not actually change the group prefixes at all here, just return whether we think we should split now - or not even that! We could just have a separate `should_split` method that can be called after this one.
> 
> I realised yesterday that the necessity to split can not only arise after adding a node, but also after another split! If we split `0` into `01` and `00` because `01` just got its 9th member, it could happen that `00` can already split further. So it might make sense to just keep that question separate from adding a node.
</details>
To me, relying on the caller to split for us always seems insecure.

regarding the split after split case, first, it can only happen in a quite imbalance network, and second the `00` group can still split once receive more nodes.

---

*[src/routing_table/routing_table.rs, line 346 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsjkM0ATYUy-nOqB0o:-KOxsD7HqwiwheQfoDWK:b-phr62r) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L346)):*
> Ah, I understand how you're using that method now!
> I still find it questionable, because it really uses an implementation detail of `Prefix`. It might be cleaner to recreate the other bucket's prefixes from our own here every time, even if that's a bit less efficient.
> Or to not have them at all: The bucket index of a group (given by _its_ prefix) is easy to compute, as it's simply the number of common leading bits of the group's prefix and our own group's prefix.
</details>
will consider this approach once main features got done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1126)*

maqi:
Review status: 2 of 7 files reviewed at latest revision, 30 unresolved discussions, some commit checks failed.

---

*[src/routing_table/mod.rs, line 138 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOdPV4RFjzpurNghaE8:-KOxwGXC19OkKmR2REbX:b-bqzkr5) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/mod.rs#L138)):*
> Maybe this also isn't necessary anymore.
</details>
still required as this mod is not being used by node.rs yet  :(

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1126)*

afck:
Reviewed 1 of 6 files at r4.
Review status: 2 of 7 files reviewed at latest revision, 26 unresolved discussions, some commit checks failed.

---

*[src/routing_table/routing_table.rs, line 282 \[r2\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOe1w2_NoZy8sRECWOF:-KOxzIvHHWbuLCo2v6TT:b64w4e6) ([raw file](https://github.com/maidsafe/routing/blob/15f7588aded63ac225dbf20677dbff21233ca01e/src/routing_table/routing_table.rs#L282)):*
> this function iterate through all the groups in all the buckets to figure out which group is the closest. 
> so, the result shall be correct at least from our knowledge?
> 
</details>
Ah, you're right, I think.

---

*[src/routing_table/routing_table.rs, line 110 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOscCrPIPq7O2AiNkmh:-KOxx-Qu8eXikDxgTdZn:b-uuyaek) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L110)):*
> when the prefix is used as bucket, i.e. the prefix is a slice of our address, drop/append prefix simply means increas/decreas the valid_bit_count,
> as the `address` is already the full copy of our address with one bit flipped (not the last one).
> 
> `move_one_bit` is used for group split, whose prefix is inherent from parent group's prefix with the last bit flipped.
</details>
I still think it's unnatural for prefixes to behave that way - conceptually, they don't _have_ a "next bit", so the only way to extend them should be to specify that bit.

But I really think that these bucket prefixes are not needed anyway. They basically answer two questions, which should be answered by simple `Prefix` methods applied to the _group's_ and _our group's_ prefixes (not a bucket prefix):

* Do we need this group in our routing table? `our_group_prefix.differing_bits(other_group_prefix) <= 1`.
* Which bucket does the group belong to? `our_group_prefix.bucket_index(other_group_prefix)`.

---

*[src/routing_table/routing_table.rs, line 326 \[r3\]](https://reviewable.io:443/reviews/maidsafe/routing/1126#-KOsirXNZGJ6hlooipzq:-KOxy6uO37Gdvkl3B4-m:b-yqkive) ([raw file](https://github.com/maidsafe/routing/blob/3666223d3916bec0201ee6077dca2f57ec9a8f5d/src/routing_table/routing_table.rs#L326)):*
> To me, relying on the caller to split for us always seems insecure.
> 
> regarding the split after split case, first, it can only happen in a quite imbalance network, and second the `00` group can still split once receive more nodes.
</details>
Yes, that's true: For now it won't be a problem if the new groups split later, when they also get another node. And I'm fine with returning a `bool` from `add` to indicate whether we need to split.

But we still shouldn't actually perform the split inside `add`: Splitting a group is now something that will require a group consensus, so between adding a node and splitting the group, messages need to be exchanged. (On the other hand: Even _that_ is something that we might as well add in another PR, if it's easier to implement it like this.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1126)*

PR #1133, opened 2016-08-31
WIP: implement public appendable data

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1133)


maidsafe-highfive:

maqi:
*[src/pub_appendable_data.rs, line 125 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1133#-KQ_qP6j7rQy6WGmbKCJ:-KQ_qP6krPQk_glPZ6Zd:bmuhllg) ([raw file](https://github.com/maidsafe/routing/blob/1b63980df3638b8459b6f80ba5dbf71030425901/src/pub_appendable_data.rs#L125)):*
> ```Rust
>     /// The `data` will contain the union of the data items, _excluding_ the `deleted_data` as
>     /// given in the update.
>     pub fn update_with_other(&mut self, other: PubAppendableData) -> Result<(), RoutingError> {
> ```

as this is `appendable` data, and `appending` data will update version automatically, is this `update` function really needed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1133)*

afck:
Review status: 0 of 18 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/pub_appendable_data.rs, line 263 \[r1\]](https://reviewable.io:443/reviews/maidsafe/routing/1133#-KQ_sE6qU4zl3A28c7-0:-KQa32iYdgx3nYlBBvw4:b-b6vf6g) ([raw file](https://github.com/maidsafe/routing/blob/1b63980df3638b8459b6f80ba5dbf71030425901/src/pub_appendable_data.rs#L263)):*
> not sure, but as I remember `get_` is preferred not to be used?
</details>
I agree, but most of this is taken over from structured data, and there it's called `get_data`. I'd leave it like this for consistency for now, and make the change for all data types in a separate PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1133)*

src/priv_appendable_data.rs
```diff
@@ -0,0 +1,521 @@
...
+    /// The signatures of the above fields by the previous owners, confirming the last update.
+    pub previous_owner_signatures: Vec<Signature>, // All the above fields
+    /// The collection of appended data items. These are not signed by the owners, as they change
+    /// even between `Post`s.
+    pub data: BTreeSet<PrivAppendedData>, // Unsigned
+}
+
+impl PrivAppendableData {
+    /// Creates a new `PubAppendableData` signed with `signing_key`.
+    pub fn new(name: XorName,
```
ustulation: Shouldn't this function also take `delete_data` parameter instead of initialising with a null (empty set) ? It is a signed field so i can't add it later without invalidating the signature. So if i fetch Pub/PrivAppendableData and want to change a field i will need to create a new one which is the clone of the previous one with changes in version (to re-post) and that field of choice. This will also mean cloning of deleted-data and will need to pass that to function parameter here.

src/lib.rs
```diff
@@ -198,6 +201,9 @@ pub use immutable_data::ImmutableData;
 pub use messages::{Request, Response};
 pub use node::{Node, NodeBuilder};
 pub use plain_data::PlainData;
+pub use append_types::{AppendedData, AppendWrapper, Filter};
+pub use pub_appendable_data::{MAX_PUB_APPENDABLE_DATA_SIZE_IN_BYTES, PubAppendableData};
+pub use priv_appendable_data::{MAX_PRIV_APPENDABLE_DATA_SIZE_IN_BYTES, PrivAppendableData};
```
ustulation: `PrivAppendedData` is not exposed - needs to be exposed too.

src/priv_appendable_data.rs
Fraser999: Can't you just clone the data and mutate the clone?  You don't need to go via `new()` do you?

ustulation: `new()` already signs stuff and deleted-data is included in signature. So it should be exposed as a parameter in `new()`.Adding it later will invalidate signature. It's just like data field of structured data. You take that as a parameter in `new()`. For the same reasons.

Fraser999: Well, as we discussed briefly yesterday, I'd prefer a different API to this altogether (all members private, with mutating functions which would do all that is required to the member fields), but yeah - it probably does make sense to add it for now.

maqi:
*[src/priv_appendable_data.rs, line 101 \[r5\]](https://reviewable.io:443/reviews/maidsafe/routing/1133#-KR-DsL5Hzewi684QByc-r5-101:-KR8ECKpq1i3Rd5Jo576:b-swdu9a) ([raw file](https://github.com/maidsafe/routing/blob/901e8aa9235bef96298802e4e264f02937265fbb/src/priv_appendable_data.rs#L101)):*
> Well, as we discussed briefly yesterday, I'd prefer a different API to this altogether (all members private, with mutating functions which would do all that is required to the member fields), but yeah - it probably does make sense to add it for now.
</details>
does it mean inside delete function, actually we needs to update signature as well?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1133)*

Viv-Rajkumar:
*[src/priv_appendable_data.rs, line 101 \[r5\]](https://reviewable.io:443/reviews/maidsafe/routing/1133#-KR-DsL5Hzewi684QByc-r5-101:-KR8FenaVrMej2bF1pqb:broq2vd) ([raw file](https://github.com/maidsafe/routing/blob/901e8aa9235bef96298802e4e264f02937265fbb/src/priv_appendable_data.rs#L101)):*
> does it mean inside delete function, actually we needs to update signature as well?
</details>
good point @maqi . I guess we have to since we update the `deleted_data` field there. This fn is only used client side right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1133)*

Fraser999:
Review status: 0 of 14 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/priv_appendable_data.rs, line 101 \[r5\]](https://reviewable.io:443/reviews/maidsafe/routing/1133#-KR-DsL5Hzewi684QByc-r5-101:-KR8HBFAdmkudwem1c6F:b2hfmce) ([raw file](https://github.com/maidsafe/routing/blob/901e8aa9235bef96298802e4e264f02937265fbb/src/priv_appendable_data.rs#L101)):*
<details><summary><i>Previously, Viv-Rajkumar (Viv Rajkumar) wrote…</i></summary>
> good point @maqi . I guess we have to since we update the `deleted_data` field there. This fn is only used client side right?
</details>
Not sure if the intent here though is to be able to call `delete()` multiple times before finally incrementing the version and signing everything.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1133)*

maqi:
*[src/priv_appendable_data.rs, line 101 \[r5\]](https://reviewable.io:443/reviews/maidsafe/routing/1133#-KR-DsL5Hzewi684QByc-r5-101:-KR8HpI-IUjQHChKj9v_:b-xqxr33) ([raw file](https://github.com/maidsafe/routing/blob/901e8aa9235bef96298802e4e264f02937265fbb/src/priv_appendable_data.rs#L101)):*
> Not sure if the intent here though is to be able to call `delete()` multiple times before finally incrementing the version and signing everything.
</details>
i guess the intent is to call delete() some times then call add_signature to sign everything.
however, as such usage is not enforced (means chance of misuse), maybe we shall change it it delete(datas: vec, sign_key: SecretKey) to enforce the usage pattern?i think same shall be applied to append.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/routing/1133)*

Viv-Rajkumar:
Closing this as its included with #1136

PR #1143, opened 2016-09-20
feat/routing table: added disjoint groups routing table to crate

Fraser999:
This new routing table is still unused and untested within the crate.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/routing/1143)


maidsafe-highfive:
r? @afck

PR #1145, opened 2016-09-22
Network test routing table

maqi:


maidsafe-highfive:
r? @afck

src/routing_table/network_tests.rs
```diff
@@ -17,297 +17,244 @@
...
 
-// use maidsafe_utilities::SeededRng;
-// use rand::Rng;
-// use std::cmp;
-// use std::collections::{HashMap, HashSet};
-// use std::collections::hash_map::Entry;
-// use std::fmt::{self, Binary, Debug, Formatter};
-// use super::routing_table::{Destination, RoutingTable};
-// use super::xorable::Xorable;
+const GROUP_SIZE: usize = 8;
```
Fraser999: I guess this should be renamed `MIN_GROUP_SIZE` now?

src/routing_table/mod.rs
```diff
@@ -581,14 +581,19 @@ impl<T: Binary + Clone + Copy + Default + Hash + Xorable> RoutingTable<T> {
         keys.sort_by(|&lhs, &rhs| lhs.cmp_distance(rhs, name));
         keys[0]
     }
+
+    #[test]
```
Fraser999: Kinda surprised this works! I thought it was meant to be `#[cfg(test)]`?

afck: It is, otherwise this function will be run _as_ a test, won't it?

src/routing_table/xorable.rs
```diff
@@ -128,6 +162,22 @@ macro_rules! impl_xorable {
                 let pow_i = 1 << (mem::size_of::<Self>() * 8 - 1 - i); // 1 on bit i.
                 self ^ pow_i
             }
+
+            fn binary(&self) -> String {
+                let bits = mem::size_of::<Self>() * 8;
+                let binary = format!("{:1b}", self);
```
afck: If the format string were already `"{:08b}"`, `"{:016b}"`, etc., the string would already be padded with zeros. Maybe it would be easier to just `match mem::size_of::<Self>()` here instead of padding manually?

src/routing_table/network_tests.rs
```diff
@@ -17,297 +17,244 @@
...
-//     nodes: HashMap<u64, RoutingTable<u64>>,
-// }
+impl Network {
+    /// Creates a new empty network with a seeded random number generator.
+    fn new(optional_seed: Option<[u32; 4]>) -> Network {
+        Network {
+            rng: match optional_seed {
+                Some(seed) => SeededRng::from_seed(seed),
+                None => SeededRng::new(),
+            },
```
afck: This could be shortened to `optional_seed.map_or_else(SeededRng::new, SeededRng::from_seed)`.

src/routing_table/network_tests.rs
```diff
@@ -17,297 +17,244 @@
...
+        for node in self.nodes.values_mut() {
+            match node.add(name) {
+                Ok(result) => {
+                    split_prefix.insert(result);
+                }
+                Err(e) => trace!("failed to add node with error {:?}", e),
+            }
+            match new_table.add(*node.our_name()) {
+                Ok(Some(prefix)) => {
+                    let _ = new_table.split(prefix);
```
afck: That should, for now, be equivalent, but in the long run we don't want the new node to decide when to split its groups and which group prefixes to put in its routing table: It should receive that information beforehand. (Maybe just add a TODO.)

src/routing_table/network_tests.rs
```diff
@@ -17,297 +17,244 @@
...
-//     // TODO: Verify that there is only one group.
-// }
+#[test]
+fn merging_groups() {
+    let mut network = Network::new(None);
+    for _ in 0..100 {
+        network.add_node();
+        // if i % 5 == 0 {
+        //     verify_invariant(&mut network);
+        // }
```
afck: Does this currently fail?
(It's okay if it's activated in a follow-up PR, of course, but let's make this a priority: The network must eventually agree on which groups exist. As long as the routing table doesn't make that guarantee, we can't expect anything to work.)

src/routing_table/xorable.rs
maqi: first, I don't see the padded zero being printed out, I only see they have been aligned.
second, the padded of zero will only happen when the value is less than the upper limit.
if we try to print a u64 with {:08b}, there won't be padded zeros.

src/routing_table/network_tests.rs
maqi: In line 62, there is TODO: needs to verify how to broadcasting such info, which including this part to be re-considered as well.

src/routing_table/network_tests.rs
maqi: yes, it is currently failing  :(
I remember I have raised this issue somewhere in the slack previously, it is mainly because when we calculate split, node itself is not included, which means you may not split your group, but others did.
with more node populated, this mismatch will gone quickly, however, we do need to double think about it.

src/routing_table/xorable.rs
afck: It says in the `fmt` module documentation that the `0` after the `:` will make it pad with zeros instead of spaces.

And your second point is what I meant by `match`ing against the size of `Self`: each branch would use the appropriate width, e.g. `{:064b}` for `u64`.

src/routing_table/network_tests.rs
afck: :+1:

PR #1147, opened 2016-10-06
feat/structured_data: support deletion

maqi:


maidsafe-highfive:
r? @afck

src/structured_data.rs
```diff
@@ -96,6 +96,29 @@ impl StructuredData {
         DataIdentifier::Structured(self.name, self.type_tag)
     }
 
+    /// Delete self
+    pub fn delete_self_against_successor(&mut self,
+                                         other: &StructuredData)
+                                         -> Result<(), RoutingError> {
+        match self.validate_self_against_successor(other) {
```
afck: Maybe this should verify that `other` is empty, too, at least `data` and `current_owner_keys`.
(Not sure how realistic that is, but if an attacker were able to intercept a client's valid `Post` request, they could turn it into a valid `Delete` otherwise, without having to change the signature.)

maqi: An attacker being a client's accessing point do have the capbility to replace the Post with Delete.

afck: I'm thinking of a scenario like this: A client `Post`s a change to the structured data, and the attacker controls the relay node. They could then drop the message and send a `Delete` containing the same data (with valid signatures and which is a valid successor).

In either case, it wouldn't hurt to require the new data version to be empty, would it?

PR #1148, opened 2016-10-06
feat/structured_data: support deletion

maqi:


maidsafe-highfive:
r? @afck

src/structured_data.rs
```diff
@@ -96,6 +96,32 @@ impl StructuredData {
         DataIdentifier::Structured(self.name, self.type_tag)
     }
 
+    /// Delete self
+    pub fn delete_self_against_successor(&mut self,
```
afck: Maybe `delete_if_valid_successor` would make more sense as a name.

src/structured_data.rs
```diff
@@ -96,6 +96,32 @@ impl StructuredData {
...
+                                         other: &StructuredData)
+                                         -> Result<(), RoutingError> {
+        if !(other.data.is_empty() && other.current_owner_keys.is_empty()) {
+            return Err(RoutingError::UnknownMessageType);
+        }
+        match self.validate_self_against_successor(other) {
```
afck: Maybe just `try!(self.validate...)`. Then no `match` is needed.

ustulation:
Merged into https://github.com/maidsafe/routing/pull/1149

PR #1149, opened 2016-10-06
Integrate with latest utilities and crust

ustulation:


maidsafe-highfive:
r? @afck

PR #1153, opened 2016-10-12
refactor/all: implements remaining parts of disjoint group logic

Fraser999:
This completes the initial integration of the disjoint group routing table into the remainder of
the production code.  Tests and examples still need to be updated/implemented.

### DIFF
--- a/src/states/node.rs
+++ b/src/states/node.rs
@@ -867,16 +867,17 @@ impl Node {
                        error);
                 self.disconnect_peer(&peer_id);
             }
-            Ok(split_prefix) => {
+            Ok(should_split) => {
                 info!("{:?} Added {:?} to routing table.", self, public_id.name());
                 if self.peer_mgr.routing_table().len() == 1 {
                     self.send_event(Event::Connected);
                 }
 
-                if let Some(prefix) = split_prefix {
+                if should_split {
                     // None of the `peers_to_drop` will have been in our group, so no need to notify
                     // Routing user about them.
-                    let peers_to_drop = self.peer_mgr.split_group(prefix);
+                    let our_group_prefix = *self.peer_mgr.routing_table().our_group_prefix();
+                    let peers_to_drop = self.peer_mgr.split_group(our_group_prefix);
                     let our_new_prefix = *self.peer_mgr.routing_table().our_group_prefix();
                     if let Err(err) = self.event_sender.send(Event::GroupSplit(our_new_prefix)) {
                         error!("{:?} Error sending event to routing user - {:?}", self, err);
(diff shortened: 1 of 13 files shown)


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -1654,6 +1736,64 @@ impl Node {
...
 
+    fn send_own_group_merge(&mut self,
+                            targets: Vec<XorName>,
+                            merge_details: OwnMergeDetails<XorName>,
+                            src: Authority) {
+        for target in &targets {
+            let (connected, needed): (Vec<_>, Vec<_>) = merge_details.groups
+                .clone()
+                .into_iter()
+                .partition(|&(prefix, _)| merge_details.prefix.is_compatible(&prefix));
```
afck: I don't understand this partition: The recipient is not necessarily already connected to everyone in a group compatible with `merge_details.prefix`: E.g. if `00` is merging with the four `01??` groups, `0100` and `0111` won't be connected yet.

On the other hand, the recipient may well already be connected to groups that are not selected here: E.g. `0100` will be connected to `1`, if it exists.

But wouldn't it be easier anyway to only make that distinction on the side of the recipient? The sender could just send every new group member the same information, and they could check for themselves whom they are already connected to?

Also, I think the recipient will need to send another `OwnGroupMerge` to everyone else in the group, because they might know a neighbouring group that the other constituent groups don't know yet. (Or do we currently rely on the `OtherGroupMerge` handling for that?)

Fraser999: I agree with your first point - good catch.  I had been thinking all groups falling under the merged prefix would be fully interconnected, and was trying to save some bandwidth.  I think this would be a valid optimisation if instead of `merge_details.prefix.is_compatible(&prefix)` we did `<target's prefix>.is_compatible(&prefix)`, but I'll just take the approach you suggest and pass the full list of `PublicId`s to each target.

I'm not sure about the second point.  Won't the group triggering the merge always be connected to all the neighbouring groups falling under the new merged prefix?  Any groups outside the new prefix will be handled by the `OtherGroupMerge`s.

afck: I think for that optimisation we should rather check whether `<target's prefix>.is_neighbour(&prefix)` - in that case, it will already be connected.

I don't think the merging group already has all connections. In my example above, `00` won't be connected to `11`, but `0100` is already connected to that, and the new group `0` will need to be.

Fraser999: Yes, I did indeed mean `is_neighbour()`, not `is_compatible()`!

As for the second point, after discussion in Slack the outcome is that I plan to fix this in a separate PR as it involves non-trivial changes to the `RoutingTable`.

The approach will be the recipients of an `OwnGroupMerge` message call a new RT function `prepare_merge(OwnMergeDetails) -> OwnMergeDetails` and send this info to all of the merging groups.  On receipt, these nodes then call `RT::merge_own_group()` which is changed to have an optional return.  This will accumulate the incoming details, returning `None` each time until all expected groups have provided info, at which point it will return `Some` with the required `OtherGroupDetails` as now.

PR #1154, opened 2016-10-12
Routing table tests

dhardy:


maidsafe-highfive:

src/routing_table/mod.rs
```diff
@@ -259,8 +260,8 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
         }
     }
 
-    pub fn our_name(&self) -> &T {
-        &self.our_name
+    pub fn our_name(&self) -> T {
+        self.our_name
```
afck: I think it makes sense to return a reference and avoid copying the name.

dhardy: But `name` is a `u32`, `u256` or similar? Surely it's cheaper to copy than reference?

afck: Outside the tests it's a `[u8; 32]`, which is larger than a reference. Not sure how the overhead of dereferencing itself compares to that, though, but I'd stick to the rule of thumb of using references for anything but primitive types.

src/routing_table/mod.rs
```diff
@@ -300,9 +301,10 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
 
     pub fn is_in_our_group(&self, name: &T) -> bool {
         if self.our_group_prefix.matches(name) {
-            return unwrap!(self.groups.get(&self.our_group_prefix)).contains(name);
+            unwrap!(self.groups.get(&self.our_group_prefix)).contains(name)
```
Fraser999: I'm not sure I see the point in this - we have no official policy, but I normally favour an early `return` over an `else` block.  Should we maybe establish a guideline so we don't end up playing ping-pong over lines like these?

afck: Instead of a _long_ `else` block, I'd also prefer an early `return`, but in this case I don't feel strongly about it. Happy to follow a guideline, if we can agree on one. :-D

dhardy: @Fraser999 I agree about early return when there's much code coming afterwards, but here it seemed simpler. A bit pointless change really though :/

Fraser999: Yeah - I'm happy with whatever.  I guess I'm more concerned with consistency, and _always_ favouring early returns is an easy rule to describe and apply, but I don't feel strongly either.  We could either go for that, or what about we leave it up to the author to decide for short `else` blocks, but don't change existing code to suit our own preference?

PR #1155, opened 2016-10-13
chore/style: fix Clippy lints

afck:


maidsafe-highfive:

src/message_filter.rs
```diff
@@ -18,11 +18,13 @@
 #[cfg(test)]
 extern crate rand;
 
+#[cfg_attr(feature="clippy", allow(useless_attribute))]
```
Fraser999: I normally add a todo if I'm temporarily disabling a lint (e.g. https://github.com/Fraser999/routing/blob/leave_merge/src/message_filter.rs#L23-L24)

afck: Done.

PR #1156, opened 2016-10-13
fix/rustfmt: workaround bug

dhardy:
https://github.com/rust-lang-nursery/rustfmt/issues/1184


maidsafe-highfive:
r? @afck

PR #1157, opened 2016-10-13
Make split order deterministic

dhardy:
Partly Andreas's work. Last commit may be pointless?

### DIFF
--- a/src/routing_table/prefix.rs
+++ b/src/routing_table/prefix.rs
@@ -31,7 +31,7 @@ pub struct Prefix<T: Clone + Copy + Default + Binary + Xorable> {
 }
 
 impl<T: Clone + Copy + Default + Binary + Xorable> Prefix<T> {
-    /// Constructor.
+    /// Creates a new `Prefix` with the first `bit_count` bits of `name`.
     pub fn new(bit_count: usize, name: T) -> Prefix<T> {
         Prefix {
             bit_count: bit_count,
@@ -39,6 +39,7 @@ impl<T: Clone + Copy + Default + Binary + Xorable> Prefix<T> {
         }
     }
 
+    //TODO: remove
     /// Return a copy of `self` with the `bit_count` increased by one and update `self` by also
     /// increasing the `bit_count` by one and flipping bit at the old `bit_count`.
     ///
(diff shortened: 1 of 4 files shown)


maidsafe-highfive:
r? @afck

src/routing_table/prefix.rs
```diff
@@ -28,37 +28,37 @@ pub struct Prefix<T: Clone + Copy + Default + Binary + Xorable> {
     bit_count: usize,
     name: T,
 }
+//TODO: insignificant bits should all be zero!
```
afck: Isn't that already done in line 39?

src/routing_table/prefix.rs
```diff
@@ -94,9 +94,25 @@ impl<T: Clone + Copy + Default + Binary + Xorable> Prefix<T> {
     }
 
     /// Compares the distance of `self` and `other` to `target`. Returns `Less` if `self` is closer,
-    /// `Greater` if `other` is closer, and `Equal` if `self.name == other.name`.
+    /// `Greater` if `other` is closer, and compares the prefix directly if of equal distance
+    /// (this is to make sorting deterministic).
     pub fn cmp_distance(&self, other: &Self, target: &T) -> Ordering {
-        target.cmp_distance(&self.name, &other.name)
+        if self.is_compatible(&other) {
+            Ord::cmp(&self.name, &other.name)
```
afck: Or `bit_count` instead of `name`? Then it wouldn't depend on the canonical name and different prefixes would be guaranteed to not return `Equal` here: Currently `000` and `00` are still considered equal.

src/routing_table/xorable.rs
```diff
@@ -119,6 +141,24 @@ macro_rules! impl_xorable_for_array {
...
                 debug_format(self.binary())
             }
+            
+            fn set_remaining(mut self, n: usize, val: bool) -> Self {
+                let bits = mem::size_of::<$t>() * 8;
+                for i in 0..$l {
+                    if n <= i*bits {
+                        self[i] = if val { !0 } else { 0 };
+                    } else if n < (i+1) * bits {
+                        let mask = !0 >> (n - i*bits);
```
afck: Please run rustfmt: this should probably be `i * bits`.

src/routing_table/xorable.rs
```diff
@@ -44,18 +44,27 @@ pub trait Xorable: Ord {
     /// Returns a copy of `self`, with the `index`-th bit flipped.
     ///
     /// If `index` exceeds the number of bits in `self`, an unmodified copy of `self` is returned.
-    fn with_flipped_bit(&self, i: usize) -> Self;
+    fn with_flipped_bit(self, i: usize) -> Self;
+
+    /// Returns a copy of `self`, with the `index`-th bit flipped.
+    ///
+    /// If `index` exceeds the number of bits in `self`, an unmodified copy of `self` is returned.
+    fn with_bit(self, i: usize, bit: bool) -> Self;
```
afck: Ah, right! That was my fault! :grimacing:

src/routing_table/prefix.rs
dhardy: In a commit after this, yes. I like to write TODO messages for future changes.

src/routing_table/xorable.rs
dhardy: I did set it up as a pre-commit hook...

PR #1158, opened 2016-10-14
chore/examples: re-enable examples

maqi:


maidsafe-highfive:
r? @afck

.travis.yml
```diff
@@ -23,10 +23,10 @@ cache:
...
 script:
   - curl -sSL https://github.com/maidsafe/QA/raw/master/Bash%20Scripts/Travis/build_and_run_tests.sh | bash
-#  - cd $TRAVIS_BUILD_DIR && cargo build --example ci_test
-#  - cd $TRAVIS_BUILD_DIR && cargo build --example key_value_store
-#  - cd $TRAVIS_BUILD_DIR && cargo run --release --example ci_test > output.log
-#  - cat output.log
+  - cd $TRAVIS_BUILD_DIR && cargo build --example ci_test
+  - cd $TRAVIS_BUILD_DIR && cargo build --example key_value_store
+  - cd $TRAVIS_BUILD_DIR && cargo run --release --example ci_test > output.log
+  - cat output.log
```
afck: Those last two lines had already been disabled before. Let's only re-enable the first two (building) and not run the ci_test on Travis for now.

appveyor.yml
```diff
@@ -30,6 +30,6 @@ build_script:
 
 test_script:
   - ps: . ".\Run Tests.ps1"
-        # cd $env:APPVEYOR_BUILD_FOLDER
-        # cargo run --release --example ci_test > output.log
-        # type output.log
+        cd $env:APPVEYOR_BUILD_FOLDER
+        cargo run --release --example ci_test > output.log
+        type output.log
```
afck: Let's keep these three disabled, too. It's only about building for now, and doing that on Travis should suffice.

examples/utils/example_client.rs
```diff
@@ -63,20 +63,19 @@ impl ExampleClient {
...
             }
         }
 
-        ExampleClient {
+        Some(ExampleClient {
             routing_client: routing_client,
             receiver: receiver,
             full_id: full_id,
-        }
+        })
```
afck: Are those changes (returning an option everywhere) related to the disjoint groups implementation?
Otherwise I'd not do that now.

examples/key_value_store.rs
```diff
@@ -49,7 +49,6 @@ extern crate maidsafe_utilities;
 #[macro_use]
 extern crate unwrap;
 extern crate docopt;
-extern crate kademlia_routing_table;
```
afck: This can probably also be removed from Cargo.toml now.

examples/utils/example_client.rs
maqi: those option just to avoid hanging when execute the examples
if ci scripts only checks the compilation, will be ok to change back.

examples/key_value_store.rs
afck: Ah, you're right, sorry. :+1:

examples/utils/example_client.rs
afck: Yes, let's only change what's necessary to make it compile for now.

PR #1159, opened 2016-10-14
Data size validations

ustulation:


maidsafe-highfive:
r? @afck

PR #1160, opened 2016-10-17
feat/data: add NO_OWNER_PUB_KEY to make data effectively immutable

afck:


maidsafe-highfive:

src/structured_data.rs
```diff
@@ -166,10 +166,8 @@ impl StructuredData {
         let data = try!(self.data_to_sign());
         // Count valid previous_owner_signatures and refuse if quantity is not enough
 
-        let check_all_keys = |&sig| {
-            owner_keys.iter()
-                .any(|pub_key| sign::verify_detached(&sig, &data, pub_key))
-        };
+        let check_all_keys =
+            |&sig| owner_keys.iter().any(|pub_key| verify_detached(&sig, &data, pub_key));
```
ustulation: Can this not fail for NO_OWNER_PUB_KEY and still pass overall because 1/2 the vote was ok - or do vaults do something special ?

afck: Yes, but isn't that expected behaviour? Wouldn't it rather be surprising if four valid signatures don't suffice because the fifth one is `NO_OWNER_PUB_KEY`?
I assumed the idea was that to permanently lock a chunk, the user would make this one the only owner.

afck: I added a check to prevent other owners together with that one, as discussed.

PR #1161, opened 2016-10-17
RT test

dhardy:


maidsafe-highfive:
r? @afck

src/routing_table/mod.rs
```diff
@@ -694,4 +701,132 @@ mod tests {
...
+                    assert!(table.iter().any(|u| *u == new_name));
+                    // skip
+                }
+                Err(Error::PeerNameUnsuitable) => {
+                    assert!(table.groups.keys().all(|p| !p.matches(&new_name)));
+                    // We should get a few of these. Save one for tests, but otherwise ignore.
+                    unknown_distant_name = Some(new_name);
+                }
+                Err(_) => {
+                    assert!(false); // no other errors should be possible
```
afck: Maybe just `panic!` and print the unexpected error in the panic message.

src/routing_table/mod.rs
```diff
@@ -694,4 +701,132 @@ mod tests {
...
+    #[test]
+    fn test_routing_groups() {
+        // Use replicable random numbers to initialse a table:
+        use rand::{Rng, SeedableRng, XorShiftRng};
+        let mut rng: XorShiftRng = SeedableRng::from_seed([1315, 30, 61894, 315]);
+        let our_name = rng.next_u32();
+        let mut table = RoutingTable::new(our_name, 8);
+        let mut unknown_distant_name = None;
+
+        for _ in 0..1000 {
```
afck: How long does this test take? Let's just make sure to not hold up the CI tests even more. But probably it's negligible, compared to the mock crust tests?

src/routing_table/mod.rs
```diff
@@ -694,4 +701,132 @@ mod tests {
...
+                    let min_size = table.min_split_size();
+                    if new_group_size >= min_size && group_len - new_group_size >= min_size {
+                        let _ = table.split(group_prefix);  // do the split
+                    }
+                }
+            }
+        }
+
+        let unknown_close_name;
+        loop {
```
afck: Ah yes, if only we'd have https://github.com/rust-lang/rfcs/issues/961 already. :-D

src/routing_table/mod.rs
```diff
@@ -694,4 +701,132 @@ mod tests {
...
+                    break;
+                }
+            }
+        }
+
+        let unknown_distant_name = unwrap!(unknown_distant_name);
+        // These numbers depend on distribution of names
+        let num_known_nodes = 114;
+        let num_groups = 9;
+        let len_our_group = 13;
```
afck: I wonder whether instead of a fixed seed, we shouldn't compute these numbers in the above iteration instead? (And possibly loop until `unknown_distant_name.is_some()`.
Or not assert them at all, but just bounds for them (e.g. we know that our own group must have at least `min_group_size` members, etc.).

src/routing_table/mod.rs
```diff
@@ -694,4 +701,132 @@ mod tests {
...
+                    };
+                    let min_size = table.min_split_size();
+                    if new_group_size >= min_size && group_len - new_group_size >= min_size {
+                        let _ = table.split(group_prefix);  // do the split
+                    }
+                }
+            }
+        }
+
+        let unknown_close_name;
```
afck: Maybe this shouldn't be called `close`, as it's explicitly not in our own group, whereas `close_name` below is.
Maybe `unknown_neighbouring_name`? :grimacing:

src/routing_table/mod.rs
dhardy: `assert!(false)` will give the line number which is enough... but yes, `panic` is probably the correct way to do this.

src/routing_table/mod.rs
dhardy: It's actually a pattern I wasn't aware of before that RFC ;)

src/routing_table/mod.rs
dhardy: I prefer a fixed seed because it's deterministic, and because it forces someone to review this test should the group splitting rules change.

I'm now adding an `invariant` function to check min group size and a few other bits.

src/routing_table/mod.rs
dhardy: Good point.

dhardy:
Next up, after this gets merged: https://github.com/dhardy/routing/commit/24871c7137f3f4c8ff647e0948cef8b15e0f59ca

PR #1162, opened 2016-10-17
chore/test: quick test for churn_test

maqi:


maidsafe-highfive:
r? @afck

src/core_tests.rs
```diff
@@ -43,6 +43,16 @@ use xor_name::XorName;
 // Poll one event per node. Otherwise, all events in a single node are polled before moving on.
 const BALANCED_POLLING: bool = true;
 
+// Toggle iterations for full or quick test according to env_var
+fn iterations() -> usize {
```
dhardy: Quite generic a name to be tied to one specific test. Why not inline in the test or make it return a bool ("do full test")?

maqi: it was thought there will be many tests using it, but turned out there is only one :(
anyway, seems this option doesn't bring down the ci execution time much,
so will just close this PR

PR #1164, opened 2016-10-18
test/example: making example_node handling GroupSplit and GroupMerge events

maqi:


maidsafe-highfive:
r? @afck

src/routing_table/prefix.rs
```diff
@@ -47,6 +47,11 @@ impl<T: Clone + Copy + Default + Binary + Xorable> Prefix<T> {
         self
     }
 
+    /// Returns `name` of the prefix
+    pub fn name(self) -> T {
```
Fraser999: I don't think we want to expose this (the `Prefix` could eventually become a bitset or something and not have an underlying `name` of type `T` I guess).

I think the tentative agreement was to use `Prefix::lower_bound()` as the name to represent a group.

src/routing_table/prefix.rs
```diff
@@ -80,10 +85,12 @@ impl<T: Clone + Copy + Default + Binary + Xorable> Prefix<T> {
         }
     }
 
+    /// Returns the number of common bits with the input name capped with previx length.
```
Fraser999: Typo.  And I guess it should say "common leading bits"?

src/routing_table/prefix.rs
```diff
@@ -80,10 +85,12 @@ impl<T: Clone + Copy + Default + Binary + Xorable> Prefix<T> {
...
     pub fn common_prefix(&self, name: &T) -> usize {
         cmp::min(self.bit_count, self.name.common_prefix(name))
     }
 
+    /// Returns the number of common bits with the input name.
```
Fraser999: Missing "leading" here too.

examples/utils/example_node.rs
```diff
@@ -82,6 +85,18 @@ impl ExampleNode {
...
+                    trace!("{} Received GroupSplit event {:?}",
+                           self.get_debug_name(),
+                           prefix);
+                    self.handle_split(prefix);
+                }
+                Event::GroupMerge(prefix) => {
+                    trace!("{} Received GroupMerge event {:?}",
+                           self.get_debug_name(),
+                           prefix);
+                    self.send_refresh(MessageId::from_lost_node(prefix.upper_bound()));
```
Fraser999: I think we agreed to use `lower_bound()` for this.

examples/utils/example_node.rs
```diff
@@ -177,20 +192,62 @@ impl ExampleNode {
...
+                let refresh_content = RefreshContent::NaeManager {
+                    data_name: *data_name,
+                    data: data.clone(),
+                };
+                let content = unwrap!(serialise(&refresh_content));
+                unwrap!(self.node
+                    .send_refresh_request(Authority::NaeManager(*data_name),
+                                          Authority::NaeManager(*data_name),
+                                          content,
+                                          id));
```
afck: Is there a reason to send these refresh messages at all? The other nodes in the group should already have exactly the same entries and will drop them on split, too.

examples/utils/example_node.rs
```diff
@@ -177,20 +192,62 @@ impl ExampleNode {
...
+            }
+        }
+        for data in &deleted_data {
+            let _ = self.db.remove(data);
+        }
     }
 
+    // fn handle_node_lost(&mut self, name: XorName) {
+    //     self.send_refresh(MessageId::from_lost_node(name));
+    // }
```
afck: Let's remove this instead of commenting it. In this form, it won't be used again.

examples/utils/example_node.rs
maqi: the split is caused by a new joined node. if routing firing both `NodeAdded` and `Split` events, it is save here not to send out those refreshes.
But, if routing only fires `Split` event in this case, then `not to send out refresh` will cause the new node won't receive the copies it wants.

I am OK to assume that routing will fire both events separately, and will update the code based on that assumption.

afck: As discussed, they will both be sent separately, so this code can be removed.

PR #1166, opened 2016-10-19
Backport changes from master

afck:


maidsafe-highfive:

PR #1168, opened 2016-10-19
chore/plain_data: removal of PlainData

maqi:


maidsafe-highfive:
r? @afck

PR #1171, opened 2016-10-20
fix/merge: fix several issues relating to merging groups

Fraser999:
This fixes a couple of bugs and also implements slightly different approaches to various functions/
types relating to mainly to merging groups.

### DIFF
--- a/src/routing_table/mod.rs
+++ b/src/routing_table/mod.rs
@@ -197,17 +197,19 @@ impl<N> Destination<N> {
 
 
 
-// Used when removal of a contact triggers the need to merge two or more groups
-#[derive(Debug)]
+// Used when removal of a contact triggers the need to merge two or more groups.  Sent between all
+// members of all merging groups, but not peers outwith the new group.
+#[derive(Clone, Debug)]
 pub struct OwnMergeDetails<T: Binary + Clone + Copy + Default + Hash + Xorable> {
-    pub prefix: Prefix<T>,
+    pub sender_prefix: Prefix<T>,
+    pub merge_prefix: Prefix<T>,
     pub groups: Groups<T>,
 }
 
 
 
-// Used when merging our own group to send to peers outwith the new group
-#[derive(Debug)]
+// Used once merging our own group has completed to send to peers outwith the new group
+#[derive(Clone, Debug)]
 pub struct OtherMergeDetails<T: Binary + Clone + Copy + Default + Hash + Xorable> {
     pub prefix: Prefix<T>,
     pub group: HashSet<T>,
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @afck

src/routing_table/network_tests.rs
```diff
@@ -118,34 +116,56 @@ impl Network {
...
-            for target in targets {
-                let target_node = unwrap!(self.nodes.get_mut(&target));
-                let contacts = target_node.merge_other_group(&merge_other_details);
-                // add missing contacts
-                for contact in contacts {
-                    let _ = target_node.add(contact);
+            // handle broadcast of merge_other_group
+            let other_info = merge_other_info;
+            merge_other_info = Vec::new();
+            for (target_prefixes, merge_other_details) in other_info {
```
afck: Or just `in merge_other_info.drain()`?

src/states/node.rs
```diff
@@ -1579,7 +1561,13 @@ impl Node {
                    route: u8,
                    sent_to: &[XorName])
                    -> Result<(Vec<XorName>, Vec<PeerId>), RoutingError> {
-        if self.is_proper() {
+        let force_via_proxy = match routing_msg.content {
```
afck: Won't this cause the sending to fail if we don't have a proxy anymore? E.g. when groups are merging, both of the connecting nodes could be without proxy.

src/routing_table/mod.rs
```diff
@@ -499,68 +464,63 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
...
+                if merge_details.sender_prefix == *prefix {
+                    *prefix_entry = true;
+                }
+            }
+            // Add an empty group in the table and cache the corresponding contacts.
             if self.groups.entry(*prefix).or_insert_with(HashSet::new).is_empty() {
-                self.needed.extend(contacts.iter());
-                needed.extend(contacts.into_iter());
+                let needed_group = self.needed.entry(*prefix).or_insert_with(HashSet::new);
+                *needed_group = needed_group.union(contacts).cloned().collect();
```
Fraser999: Yup - it would.  In theory there shouldn't be much difference performance-wise since this _should_ always yield `needed_group` as empty or as identical to `contacts`, but `extend()` is certainly simpler.

src/routing_table/network_tests.rs
Fraser999: Yes - I went a bit mad here I think :)  I think we don't even need the `drain()`, nor the check on line 119 since `merge_other_info.is_empty()` will always be true there!

src/states/node.rs
Fraser999: Good catch.  I should be checking `routing_msg.src` as well so `force_via_proxy` is true where we're sending as a Client and the message is from us (i.e. we're not acting as a proxy ourself).

PR #1172, opened 2016-10-21
MAID-1638 fix/all: add missing handling of group split messages

Fraser999:
This updates the 'RoutingTable' message to be a RoutingMessage (sent from a group) rather than a
'DirectMessage'.  It also adds the sending and handling of 'GroupSplit' messages.

### DIFF
--- a/src/states/node.rs
+++ b/src/states/node.rs
@@ -452,17 +452,6 @@ impl Node {
                 }
                 Ok(())
             }
-            DirectMessage::RoutingTable(public_ids) => {
-                for public_id in &public_ids {
-                    if self.peer_mgr.routing_table().need_to_add(public_id.name()).is_ok() {
-                        let our_name = *self.name();
-                        let _ = self.send_connection_info(*public_id,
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -894,9 +886,21 @@ impl Node {
...
+                        src: Authority::NaeManager(self.peer_mgr
+                            .routing_table()
+                            .our_group_prefix()
+                            .lower_bound()),
+                        dst: Authority::ManagedNode(*public_id.name()),
+                        content: request_content,
+                    };
+                    if let Err(err) = self.send_routing_message(request_msg) {
+                        debug!("{:?} Failed to send RoutingTable: {:?}.", self, err);
+                    }
```
afck: Two things to note for later:
* This will probably fail to accumulate if two nodes join the same group simultaneously. (But we want to disallow that anyway.)
* If the group is about to split, this will still send all the contacts from _before_ the split to the new node. It will connect to them just to disconnect a few seconds later.

PR #1173, opened 2016-10-24
refactor/all: several cleanups

afck:
Make `Authority` `Copy`.
Make the `ExampleNode` store by `DataIdentifier`, not name.


maidsafe-highfive:

PR #1174, opened 2016-10-24
refactor/all: several cleanups

maqi:
Make `Authority` `Copy`.
Make the `ExampleNode` store by `DataIdentifier`, not name.


maidsafe-highfive:
r? @afck

PR #1176, opened 2016-10-24
fix/data: use serialised_size for size checks

afck:


maidsafe-highfive:

PR #1180, opened 2016-10-26
MAID-1638: refactor/examples: fix examples

Fraser999:
This allows the ci_test example to pass again, but with a reduced network size of 10 nodes.  It also
fixes a long-standing bug in the key_value_store example.

### DIFF
--- a/examples/ci_test.rs
+++ b/examples/ci_test.rs
@@ -55,22 +55,18 @@ extern crate unwrap;
 
 mod utils;
 
-
 use docopt::Docopt;
-
 use maidsafe_utilities::SeededRng;
 use maidsafe_utilities::thread::Joiner;
 use maidsafe_utilities::thread::named as thread_named;
-
 use rand::{Rng, ThreadRng, random, thread_rng};
 use rand::distributions::{IndependentSample, Range};
-use routing::{Data, StructuredData};
+use routing::{Data, MIN_GROUP_SIZE, StructuredData};
 use std::{env, io, thread};
 use std::io::Write;
 use std::process::{Child, Command, Stdio};
 use std::sync::{Arc, Condvar, Mutex};
 use std::time::Duration;
-
 use term::color;
 use utils::{ExampleClient, ExampleNode};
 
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @afck

Cargo.toml
```diff
@@ -27,7 +27,7 @@ unwrap = "~1.1.0"
 
 [dev-dependencies]
 docopt = "~0.6.86"
-libc = "~0.2.16"
+libc = "~0.2.17"
```
dhardy: If you did `cargo update`, shouldn't this be committed separately?

Fraser999: I just did the changes manually (I think cargo update would only touch the lock file anyway, but there is a tool for updating the toml file which some of the other guys use I think), but we don't really have an agreed-upon approach.

I generally try and keep Cargo.toml up to date whenever I remember and have been just adding any changes as part of unrelated commits.

dirvine:
@Fraser999  @dhardy `cargo outdated` will automatically show outdated libs in your cargo.toml. Installable via `cargo install cargo-outdated`

examples/ci_test.rs
```diff
@@ -55,33 +55,29 @@ extern crate unwrap;
...
 const CHURN_MIN_WAIT_SEC: u64 = 20;
 const CHURN_MAX_WAIT_SEC: u64 = 30;
 const CHURN_TIME_SEC: u64 = 20;
 const DEFAULT_REQUESTS: usize = 30;
-const DEFAULT_NODE_COUNT: usize = 20;
+const DEFAULT_NODE_COUNT: usize = 10;
 /// The number of churn-get cycles.
 const DEFAULT_BATCHES: usize = 1;
-const GROUP_SIZE: usize = 8;
+const GROUP_SIZE: usize = MIN_GROUP_SIZE;
```
Viv-Rajkumar: why even have this const anymore and not use `MIN_GROUP_SIZE` directly?

Fraser999: Good point.

PR #1181, opened 2016-10-26
MAID-1658 fix RoutingTable::targets when missing connections

afck:


maidsafe-highfive:

src/states/node.rs
```diff
@@ -1740,7 +1740,8 @@ impl Node {
         if let RemovalDetails { targets_and_merge_details: Some((targets, merge_details)), .. } =
                details {
             let our_new_prefix = merge_details.merge_prefix;
-            self.send_own_group_merge(targets, merge_details, Authority::NodeManager(details.name));
+            let src_name = our_new_prefix.lower_bound();
+            self.send_own_group_merge(targets, merge_details, Authority::NodeManager(src_name));
```
Fraser999: Not sure it's worth changing in this PR, but in other places where we use an `Authority` to represent a group, we've used the `NaeManager` type.

afck: Done.

PR #1184, opened 2016-10-28
MAID-1664: fix/node: pass Prefixes as well as PublicIds to joining nodes

Fraser999:
This removes the 'RoutingTable' message type, replacing it by changing the 'GetNodeNameResponse' to
contain the contents of the RT.

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -347,8 +347,8 @@ pub enum MessageContent {
     GetNodeNameResponse {
         /// Supplied `PublicId`, but with the new name
         relocated_id: PublicId,
-        /// Our close group `PublicId`s.
-        close_group_ids: Vec<PublicId>,
+        /// The routing table structure of our group, including the `PublicId`s of our contacts.
+        groups: Vec<(Prefix<XorName>, Vec<PublicId>)>,
         /// The message's unique identifier.
         message_id: MessageId,
     },
@@ -361,8 +361,6 @@ pub enum MessageContent {
         /// The message ID.
         message_id: MessageId,
     },
-    /// Sent to a joining node to allow it to establish connections to the included contacts.
-    RoutingTable(Vec<PublicId>),
     /// Sent to all connected peers when our own group splits
     GroupSplit(Prefix<XorName>),
     /// Sent amongst members of a newly-merged group to allow synchronisation of their routing
@@ -485,12 +483,12 @@ impl Debug for MessageContent {
             MessageContent::GetCloseGroup(id) => write!(formatter, "GetCloseGroup({:?})", id),
             MessageContent::ConnectionInfo { .. } => write!(formatter, "ConnectionInfo {{ .. }}"),
             MessageContent::GetNodeNameResponse { ref relocated_id,
-                                                  ref close_group_ids,
+                                                  ref groups,
                                                   ref message_id } => {
                 write!(formatter,
                        "GetNodeNameResponse {{ {:?}, {:?}, {:?} }}",
-                       close_group_ids,
                        relocated_id,
+                       groups,
                        message_id)
             }
             MessageContent::GetCloseGroupResponse { ref close_group_ids, message_id } => {
@@ -498,9 +496,6 @@ impl Debug for MessageContent {
                        "GetCloseGroupResponse {{ {:?}, {:?} }}",
                        close_group_ids,
                        message_id)
-            }
-            MessageContent::RoutingTable(ref public_ids) => {
-                write!(formatter, "RoutingTable({:?})", public_ids)
             }
             MessageContent::GroupSplit(ref prefix) => write!(formatter, "GroupSplit({:?})", prefix),
             MessageContent::OwnGroupMerge { ref sender_prefix, ref merge_prefix, ref groups } => {
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -286,11 +286,12 @@ impl PeerManager {
                 (*prefix, members.into_iter().map(|pub_id| *pub_id.name()).collect_vec())
             })
             .collect_vec();
-        // Nothing can be done to recover from an error here - use unwrap.
+        // TODO - nothing can be done to recover from an error here - use `unwrap!` for now, but
+        // consider refactoring to return an error which can be used to transition the state
+        // machine to `Terminate`.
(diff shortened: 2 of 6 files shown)


maidsafe-highfive:
r? @afck

src/peer_manager.rs
```diff
@@ -277,9 +279,18 @@ impl PeerManager {
...
-    pub fn reset_routing_table(&mut self, our_public_id: PublicId) {
+    pub fn reset_routing_table(&mut self, our_public_id: PublicId, groups: &[Group]) {
         self.our_public_id = our_public_id;
-        let new_rt = RoutingTable::new(*our_public_id.name(), MIN_GROUP_SIZE);
+        let groups_as_names = groups.into_iter()
+            .map(|&(ref prefix, ref members)| {
+                (*prefix, members.into_iter().map(|pub_id| *pub_id.name()).collect_vec())
+            })
+            .collect_vec();
+        // Nothing can be done to recover from an error here - use unwrap.
```
afck: I'd still prefer to just not reset the routing table and print an error log if we accumulated a message with an inconsistent group structure. It should then be treated like an unsuccessful join attempt, and the node should retry.

src/peer_manager.rs
```diff
@@ -277,9 +279,18 @@ impl PeerManager {
...
+        let new_rt = unwrap!(RoutingTable::new_with_groups(*our_public_id.name(),
+                                                           MIN_GROUP_SIZE,
+                                                           groups_as_names));
+        // RoutingTable::new_with_groups(*our_public_id.name(), MIN_GROUP_SIZE, groups);
```
afck: That can be deleted now.

src/routing_table/mod.rs
```diff
@@ -377,6 +398,37 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
...
+    // the split are returned.  Returns `Err(Error::PeerNameUnsuitable)` if `name` is not within our
+    // group.
+    // TODO - should we check `name` isn't already in RT?
+    pub fn expect_add_to_our_group(&self, name: &T) -> Result<Groups<T>, Error> {
+        if !self.our_group_prefix.matches(name) {
+            return Err(Error::PeerNameUnsuitable);
+        }
+        let mut groups = self.groups.clone();
+        let mut our_group = unwrap!(groups.remove(&self.our_group_prefix));
+        if self.should_split_our_group(&our_group) {
```
afck: Shouldn't `name` be added to `our_group` before this check (and removed after it again)?

src/routing_table/mod.rs
```diff
@@ -377,6 +398,37 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
...
+            let our_prefix_after_split = Prefix::new(self.our_group_prefix.bit_count() + 1,
+                                                     self.our_name);
+            let groups_to_remove = groups.keys()
+                .filter(|&&x| {
+                    x != our_prefix_after_split && !x.is_neighbour(&our_prefix_after_split)
+                })
+                .cloned()
+                .collect_vec();
+            for prefix in groups_to_remove {
+                let _ = groups.remove(&prefix);
```
afck: I wonder whether we should just add those groups as empty instead. Otherwise it will probably make the invariant check in `RoutingTable::new_with_groups` fail if required neighbouring groups are missing.

Alternatively, we'd need to use `our_prefix_after_split` below - but I'm not sure whether that will cause trouble, since the other group members haven't performed the split yet?

src/routing_table/mod.rs
```diff
@@ -869,6 +890,26 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
...
+    #[cfg(any(test, feature = "use-mock-crust"))]
+    pub fn needed(&self) -> &Groups<T> {
+        &self.needed
+    }
+
+    #[cfg(test)]
+    pub fn verify_invariant(&self) {
+        if let Err(_) = self.check_invariant() {
+            let message = format!("Invariant not satisfied for RT: {:?}", self);
+            panic!(message);
```
afck: Let's print the error in this panic message, too.
(I know it's only GitHub highlighting this code as changed, but still. :grin: )

src/routing_table/mod.rs
Fraser999: Yeah, fair point.  I think your first suggestion is safer.  The new node should still receive the split notifications once it actually connects to enough peers, and at that stage the empty group(s) should be removed again.  I'll make that change.

src/routing_table/mod.rs
Fraser999: I'd already noticed this and actually changed it locally to use a single `unwrap!()` statement to achieve this :)  However, the error isn't particularly helpful - it can only be `Error::InvariantViolation` and that doesn't specify _which_ invariant has been broken.  Even adding a `String` to the error variant would help a bit, or I could change the single variant into the six different cases we check for, or I could add a new error enum just for the invariant checks and nest that in `Error::InvariantViolation`?

src/peer_manager.rs
afck: Absolutely! Not sure about that in the first place, since this is an accumulated, trusted message anyway.
We could create a JIRA issue to consider it later.

src/routing_table/mod.rs
afck: Ah, right. Let's also not do that in this PR then.

src/messages.rs
```diff
@@ -347,8 +347,8 @@ pub enum MessageContent {
     GetNodeNameResponse {
         /// Supplied `PublicId`, but with the new name
         relocated_id: PublicId,
-        /// Our close group `PublicId`s.
-        close_group_ids: Vec<PublicId>,
+        /// The routing table structure of our group, including the `PublicId`s of our contacts.
+        groups: Vec<(Prefix<XorName>, Vec<PublicId>)>,
```
maqi: is `groups` is actually the whole routing_table structure of this node, not just `its group` ?

afck: Yes, but it is supposed to be identical among all the nodes in that group.

maqi: sry, I mean the comment indicates `groups` is just about `our group`, which is bit confusing to me

afck: As discussed, let's make the comment clearer in a later PR.

PR #1186, opened 2016-10-31
MAID-1641 Fix some group splitting and merging issues.

afck:


maidsafe-highfive:

PR #1187, opened 2016-10-31
MAID-1669 Defer routing message handling.

afck:
It is confusing and error-prone to handle routing messages inside the
calls to the `send` methods themselves. This adds a message queue and
defers handling, so that messages are processed sequentially, one at a time.


maidsafe-highfive:

PR #1188, opened 2016-10-31
fix/core_tests: make balaced routing do what it says it does

dhardy:


maidsafe-highfive:
r? @afck

src/core_tests.rs
```diff
@@ -264,18 +264,17 @@ fn poll_all(nodes: &mut [TestNode], clients: &mut [TestClient]) -> bool {
     loop {
         let mut n = false;
         if BALANCED_POLLING {
-            nodes.iter_mut().foreach(|node| n = n || node.inner.poll());
+            // handle at most one event from each node:
```
Fraser999: I think [`node.inner.poll()`](https://github.com/Fraser999/routing/blob/master/src/core_tests.rs#L157-L166) is handling _all_ events from each node isn't it?

src/core_tests.rs
```diff
@@ -264,18 +264,17 @@ fn poll_all(nodes: &mut [TestNode], clients: &mut [TestClient]) -> bool {
     loop {
         let mut n = false;
```
Fraser999: Since we're in here anyway, could `n` be changed to something more meaningful, e.g. `events_occurred` or something?

dhardy:
This appears to fail `core_tests::multiple_joining_nodes`.

afck:
It probably makes sense to disable `multiple_joining_nodes` until we can rule out contradictory quora again.

PR #1189, opened 2016-10-31
MAID-1614 re-enable four mock Crust tests

afck:


maidsafe-highfive:

PR #1190, opened 2016-11-01
MAID-1651 test/performance: decrease group and quorum size when using mock crust

dhardy:
messages_accumulate_with_quorum: reduce number of nodes and make dependent on group size.


maidsafe-highfive:
r? @afck

dhardy:
@afck will include these changes in a separate PR.

PR #1191, opened 2016-11-01
MAID-1674 remove old accumulation mechanism

afck:
Accumulating messages in the recipient makes no sense anymore. This will
be replaced by accumulation in the first hop.


maidsafe-highfive:

src/states/node.rs
```diff
@@ -731,17 +700,6 @@ impl Node {
         Ok(false)
     }
 
-    fn dynamic_quorum_size(&self) -> usize {
```
dhardy: Some quorum is still needed. This is all provisional? I wonder if it would have been better to replace this with `network_size / 2 + 1` for now?

PR #1192, opened 2016-11-01
refactor/log,core_tests: adjust logging, impl Debug for Node and Stat…

dhardy:
…eMachine

Test rename is to allow selection via `cargo test ... PATTERN`, which before this would always match at least two other tests.

### DIFF
--- a/src/state_machine.rs
+++ b/src/state_machine.rs
@@ -25,6 +25,7 @@ use routing_table::RoutingTable;
 use states::{Bootstrapping, Client, Node};
 use std::mem;
 use std::sync::mpsc::{self, Receiver};
+use std::fmt::{self, Debug, Formatter};
 use timer::Timer;
 use types::RoutingActionSender;
 #[cfg(feature = "use-mock-crust")]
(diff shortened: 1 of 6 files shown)


maidsafe-highfive:
r? @afck

src/state_machine.rs
```diff
@@ -111,8 +112,8 @@ impl State {
         }
     }
 
-    fn into_bootstrapped(self, proxy_peer_id: PeerId, proxy_public_id: PublicId) -> Self {
-        match self {
+    fn transition_to_final(self, proxy_peer_id: PeerId, proxy_public_id: PublicId) -> Self {
```
afck: I think `into` communicates better that this consumes `self`.
I also like `bootstrapped` better than `final`, as only `Terminated` is really final.

src/state_machine.rs
```diff
@@ -123,6 +124,19 @@ impl State {
                 }
             }
             _ => unreachable!(),
+        };
+        info!("Transitioned to: {:?}", &new_self);
```
afck: Not sure whether we want to have that user-visible in the vault binaries by default. Maybe it should be `trace!`.

src/state_machine.rs
dhardy: I found the name `into_bootstrapped` confusing. If anything, it's converting _from_ the bootstrapped state.

afck: No, it's converting _from_ the `Bootstrapping` state into one of the states that implement the `Bootstrapped` trait (or to `Terminated`.

Fraser999: I guess I agree with @afck on both points really.  Certainly `into` seems to be a standard Rust convention.

dhardy:
Travis shows a failure. But the test passes on my machine on the commit I pushed (c5db457829b28d467d510d814de32d00d4cfc8fe) so I think this is caused by some other ancestor or interaction in 313aa2295c3eae3defccbbd23cbe730b44c3e252.

PR #1194, opened 2016-11-01
MAID-1661: fix/ack_manager: fix bug in AckManager

Fraser999:
This resolves an issue in the AckManager whereby a received Ack was not cached.  Also includes some
minor updates to the ci_test example.

### DIFF
--- a/examples/ci_test.rs
+++ b/examples/ci_test.rs
@@ -64,6 +64,7 @@ use rand::distributions::{IndependentSample, Range};
 use routing::{Data, MIN_GROUP_SIZE, StructuredData};
 use std::{env, io, thread};
 use std::io::Write;
+use std::panic;
 use std::process::{Child, Command, Stdio};
 use std::sync::{Arc, Condvar, Mutex};
 use std::time::Duration;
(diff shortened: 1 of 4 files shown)


maidsafe-highfive:
r? @afck

examples/ci_test.rs
```diff
@@ -139,14 +137,14 @@ fn simulate_churn(mut nodes: Vec<NodeProcess>,
 
         loop {
             {
-                let &(ref lock, ref cvar) = &*stop_flg;
+                let &(ref lock, ref condition_variable) = &*stop_flag;
```
afck: Even the standard library calls it `Condvar` - shall we go with `condvar` as a compromise?

src/states/node.rs
```diff
@@ -1426,6 +1426,17 @@ impl Node {
             }
         }
 
+        let acting_as_proxy = if let Authority::Client { ref proxy_node_name, .. } =
+                                     routing_msg.src {
+            proxy_node_name == self.name()
+        } else {
```
afck: It would be more concise to remove the `else` branch, and just have the `if ... { return Ok(()); }` in the `if let` branch. That would also avoid some unnecessary calls to `is_in_our_group`.

examples/ci_test.rs
Fraser999: Grudgingly, yes :)

src/states/node.rs
Fraser999: I thought using a standalone variable named as `acting_as_proxy` helped self-document the check.  I could move the `!acting_as_proxy` check to the front of the `if` condition to allow it to short-circuit the rest of the conditions, which should achieve the same thing.  Would that be OK?

afck: Sure, that's fine, too.
(I agree with the self-documenting, but in this case I feel it inflates the code a lot. Anyway, I'm okay with just moving the `!acting_as_proxy` to the front.)

PR #1196, opened 2016-11-02
Maid 1681: better output for network invariant failures

dhardy:


maidsafe-highfive:
r? @afck

src/routing_table/network_tests.rs
```diff
@@ -309,33 +309,39 @@ pub fn verify_network_invariant<'a, T, U>(nodes: U)
...
+                continue;
+            }
+            if prefix1.is_compatible(prefix2) {
+                panic!("Group prefixes should be disjoint, but these are not:\n\
+                    Group {:?}: {:?}\n\
+                    Group {:?}: {:?}",
+                       prefix1,
+                       groups[prefix1].1,
+                       prefix2,
+                       groups[prefix2].1);
```
afck: For the tests, it would be extremely helpful to also print the nodes whose routing tables the conflicting group information is from. Maybe `groups` should be a `HashMap<Prefix<T>, (T, HashSet<T>, HashSet<T>)>` to, in addition to its members, include the set of nodes that have that group in that form in their routing table.

The two `node.our_name` in lines 299 and 301 seem to be wrong. That could probably also be addressed if the other node's name were present in the map.

dhardy: This comment addresses different lines of code; these have been changed to avoid printing the same name twice.

PR #1197, opened 2016-11-02
MAID-1683 fix/mock_crust: drop packets sent over a route when it is disconnected

dhardy:
Packet handling is FIFO. The drop happens when the disconnect
message is received, so prior messages on the channel survive. This
may or may not be what's intended.

### DIFF
--- a/src/mock_crust/support.rs
+++ b/src/mock_crust/support.rs
@@ -118,6 +118,17 @@ impl Network {
             .push_back(packet);
     }
 
+    // Drop any pending messages on a specific route (does not automatically
+    // drop packets going the other way).
+    fn drop_pending(&self, sender: Endpoint, receiver: Endpoint) {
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @afck

src/mock_crust/support.rs
```diff
@@ -360,6 +369,12 @@ impl ServiceImpl {
     }
 
     fn handle_disconnect(&mut self, peer_endpoint: Endpoint) {
+        // Drop any remaining messages between these two nodes. We do this now,
+        // not when `disconnect()` is called, to allow the disconnect message
+        // and any prior messages to reach their destinations.
```
afck: But real Crust probably doesn't even do that - I don't think it sends a `Disconnect` message; it probably just closes the connection -, so it might make sense to actually drop these in `disconnect()`.

PR #1198, opened 2016-11-03
MAID-1663: fix/all: fix duplicate handling of messages

Fraser999:
This is part fix and part workaround for the issue where clients and nodes are handling some
'RoutingMessage's more than once.


maidsafe-highfive:
r? @afck

PR #1200, opened 2016-11-05
MAID-1675 Accumulate messages in the first hop

afck:


maidsafe-highfive:

src/messages.rs
```diff
@@ -194,49 +202,97 @@ impl HopMessage {
...
+                .cloned()
+                .collect_vec();
+            for invalid_id in invalid_ids {
+                let _ = self.signatures.remove(&invalid_id);
+            }
+        }
+        self.grp_lists.push(group_list);
+    }
+
+    /// Adds the given signature, if it is valid and new.
```
Fraser999: I think we should validate the signature here, but it looks like we don't?  Or maybe that would be best done in `SignatureAccumulator::add_signature()` and `SignatureAccumulator::add_message()`, in which case this comment should maybe make it clear the signature has already been validated?

src/messages.rs
```diff
@@ -250,15 +306,24 @@ impl SignedMessage {
...
-    }
-
     /// The priority Crust should send this message with.
     pub fn priority(&self) -> u8 {
         self.content.priority()
     }
+
+    /// Returns whether there are enough valid signatures from the sender.
+    pub fn is_fully_signed(&self) -> bool {
+        if self.content.src.is_client() {
```
Fraser999: Why do we have this separate check (with a more stringent condition requiring `len() == 1`) for clients?

src/messages.rs
```diff
@@ -287,22 +352,12 @@ impl RoutingMessage {
...
-            }
-            _ => self.content.clone(),
-        };
-        Ok(RoutingMessage {
-            src: self.src,
-            dst: self.dst,
-            content: content,
-        })
+    /// Returns a `DirectMessage::MessageSignature` for this message.
+    pub fn to_signature(&self, sign_key: &sign::SecretKey) -> Result<DirectMessage, RoutingError> {
```
Fraser999: I guess since "sign" is a verb, I'd have opted for `secret_key` or `signing_key` here, but not a big deal.

src/signature_accumulator.rs
```diff
@@ -0,0 +1,118 @@
...
+use itertools::Itertools;
+use maidsafe_utilities::serialisation;
+use messages::SignedMessage;
+use rust_sodium::crypto::sign;
+use rust_sodium::crypto::hash::sha256;
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
+use std::time::Instant;
+
+/// Time (in seconds) after which a joining node will get dropped from the map of joining nodes.
```
Fraser999: Copy-paste error?

src/states/node.rs
```diff
@@ -472,15 +478,33 @@ impl Node {
             DirectMessage::TunnelDisconnect(dst_id) => {
                 self.handle_tunnel_disconnect(peer_id, dst_id)
             }
-            _ => {
-                debug!("{:?} - Unhandled direct message: {:?}",
-                       self,
-                       direct_message);
+            msg @ DirectMessage::BootstrapIdentify { .. } |
```
Fraser999: Ahhhh - finally :)

src/states/node.rs
```diff
@@ -1443,27 +1464,42 @@ impl Node {
...
         let src = &routing_msg.src;
-        if signed_msg.public_id() == self.full_id.public_id() && hop == self.name() &&
-           src.is_group() &&
-           !self.peer_mgr.routing_table().should_route_full_message(src.name(), route as usize) {
-            return Ok(()); // Not our turn to send the group message. TODO: Send hash.
-        }
-
-        let raw_bytes = try!(self.to_hop_bytes(signed_msg.clone(), route, new_sent_to.clone()));
+        let send_sig =
+            sent_by_us && hop == self.name() && src.is_group() &&
```
Fraser999: `sent_by_us` already checks for `hop == self.name()`.

src/messages.rs
```diff
@@ -194,49 +202,97 @@ impl HopMessage {
...
-        if sign::verify_detached(&self.signature,
-                                 &signed_bytes,
-                                 self.public_id().signing_public_key()) {
-            Ok(())
-        } else {
-            Err(RoutingError::FailedSignature)
+        let signed_bytes = try!(serialise(&self.content));
+        for (pub_id, sig) in &self.signatures {
+            if !sign::verify_detached(sig, &signed_bytes, pub_id.signing_public_key()) {
+                return Err(RoutingError::FailedSignature);
```
afck: Yes, you're right, a bad signature should not stop us from handling the message.

src/messages.rs
```diff
@@ -148,12 +156,12 @@ impl DirectMessage {
 #[derive(RustcEncodable, RustcDecodable)]
 pub struct HopMessage {
     /// Wrapped signed message.
-    content: SignedMessage,
+    pub content: SignedMessage,
```
Fraser999: Nah - no probs - I'll leave as is, but remove the `content()` getter.

src/messages.rs
Fraser999: So you'd prefer something more like `SignatureAccumulator::remove_if_complete()` to be changed to do:
```rust
if !msg.is_fully_signed() {
    return None;
}
msg.remove_invalid_signatures()
if !msg.is_fully_signed() {
    return None;
}
```

afck: You mean because otherwise we _do_ serialise it every time we add a signature?
That's a good point. And maybe that's all premature optimisation. I'd say go with the more elegant solution, if in doubt. We can optimise later.

afck:
Continued in #1213

PR #1201, opened 2016-11-07
MAID-1685 chore/data: move data_types into module

maqi:


maidsafe-highfive:
r? @afck

src/data/mod.rs
```diff
@@ -15,11 +15,19 @@
...
+mod structured_data;
+
+pub use self::append_types::{AppendWrapper, AppendedData, Filter};
+pub use self::immutable_data::{ImmutableData, MAX_IMMUTABLE_DATA_SIZE_IN_BYTES};
+pub use self::pub_appendable_data::{MAX_PUB_APPENDABLE_DATA_SIZE_IN_BYTES, PubAppendableData};
+pub use self::priv_appendable_data::{MAX_PRIV_APPENDABLE_DATA_SIZE_IN_BYTES, PrivAppendableData,
+                                     PrivAppendedData};
 use std::fmt::{self, Debug, Formatter};
-pub use structured_data::StructuredData;
+pub use self::structured_data::{MAX_STRUCTURED_DATA_SIZE_IN_BYTES, StructuredData};
```
dhardy: Would be neater if you reordered so that all `pub use self::...` lines came first, maybe with a blank line before the other imports.

afck: Let's not do anything that rustfmt would _undo_, though. Once rustfmt does this more smartly, we can apply it to all files at once.

PR #1202, opened 2016-11-07
MAID-1687 fix response caching, add print_nodes fn, format fixes

dhardy:
I'm surprised I don't get warnings/errors about unused imports (`maidsafe_utilities` in `core_tests.rs`) and functions (`print_nodes`), but I don't. I'd like these to stay regardless (is there already an `allow(unused)` declaration for this module?).

### DIFF
--- a/src/core_tests.rs
+++ b/src/core_tests.rs
@@ -387,6 +387,7 @@ fn create_connected_nodes_with_cache_till_split(network: &Network) -> Vec<TestNo
     nodes
 }
 
+// Create `size` clients, all of whom are connected to `nodes[0]`.
 fn create_connected_clients(network: &Network,
                             nodes: &mut [TestNode],
                             size: usize)
@@ -1290,23 +1291,23 @@ fn request_during_churn_group_to_group() {
 }
 
 // Generate random immutable data, but make sure the first node in the given
-// node slice (the proxy node) is not the closest to the data. Also sorts
-// the nodes by distance to the data.
-fn gen_immutable_data_not_close_to_first_node<T: Rng>(rng: &mut T, nodes: &mut [TestNode]) -> Data {
+// node slice (the proxy node) is not in the data's group.
+fn gen_immutable_data_not_in_first_node_group<T: Rng>(rng: &mut T, nodes: &[TestNode]) -> Data {
     let first_name = nodes[0].name();
+    // We want to make sure the data is inserted into a different group. Since the
+    // root prefix uses 0 bits, we will have at least one group starting bit 0 and at
+    // least one starting bit 1. If this differs, the groups are guaranteed different.
+    let prefix = Prefix::new(1, first_name);
 
     loop {
         let data = gen_immutable_data(rng, 8);
-        sort_nodes_by_distance_to(nodes, data.name());
-
-        if nodes.iter().take(nodes.len() / 2).all(|node| node.name() != first_name) {
+        if !prefix.matches(data.name()) {
             return data;
         }
     }
 }
 
 #[test]
-#[ignore]
 fn response_caching() {
     let network = Network::new(None);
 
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -551,6 +551,8 @@ impl Node {
         Ok(())
     }
 
+    // This match construct is huge and rustfmt's rules don't yield good results
+    #[cfg_attr(rustfmt, rustfmt_skip)]
```
afck: Let's stick to rustfmt's formatting except in extreme cases (rustfmt bugs, formatted multi-line strings, or after a discussion with the whole team).

src/core_tests.rs
```diff
@@ -178,6 +187,27 @@ impl TestNode {
...
+    }
+    for (prefix, names) in groups {
+        println!("Group {:?}:", prefix);
+        for name in names {
+            println!("Node\t{:?}", name);
+        }
+    }
+    println!("Clients:");
+    for client in clients {
+        println!("\t{:?}", client.name());
```
afck: These should probably be `trace!` (and maybe `log_nodes`).

src/core_tests.rs
```diff
@@ -178,6 +187,27 @@ impl TestNode {
     }
 }
 
+// Prints all nodes in a table, organised by group.
+// Does not check the routing tables (use verify_invariant_for_all_nodes() for that).
+fn print_nodes(nodes: &[TestNode], clients: &[TestClient]) {
```
afck: If it's not used anywhere, let's not add this function.

src/core_tests.rs
```diff
@@ -169,6 +172,12 @@ impl TestNode {
         unwrap!(self.inner.name())
     }
 
+    // For routing nodes, this returns the group prefix. For clients, this
```
Fraser999: This could be a single-line comment.

src/core_tests.rs
```diff
@@ -38,6 +38,9 @@ use std::sync::mpsc;
 use std::thread;
 use types::MessageId;
 use xor_name::XorName;
+use maidsafe_utilities;
+// Add the following line to tests to enable logging:
```
Fraser999: I don't think these two lines should be included - their point is as relevant here as in _any_ source file.

src/core_tests.rs
Fraser999: I prefer a paradigm where we print things like this inside `drop()` functions if the current thread is panicing.  E.g. I was planning to submit this in a PR for RT:
```rust
impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> Drop for RoutingTable<T> {
    fn drop(&mut self) {
        if thread::panicking() {
            trace!("{:?}", self);
        }
    }
}
```
If that's not preferred either, then I agree that we shouldn't merge unused code like this.

src/core_tests.rs
Fraser999: Agree on both points.  Using `trace!` rather than `println!` means the statements yield more info, can be filtered out without recompiling, and can be handled properly by editors which are configured to handle logging syntax.

src/core_tests.rs
dhardy: But they are _commonly_ used to debug tests, and this saves looking up the log-init line.

src/core_tests.rs
dhardy: fixed

src/states/node.rs
```diff
@@ -842,8 +842,8 @@ impl Node {
                             dst: Authority::NaeManager(prefix.lower_bound()),
                             content: MessageContent::GroupSplit(our_group_prefix),
                         };
-                        // We send on route 1 by default to avoid trying to send via the newly joined
-                        // node
+                        // We send on route 1 by default to avoid trying to
+                        // send via the newly joined node
```
afck: We have lots of comments that go beyond column 80, and rustfmt doesn't complain about them; I don't think we gain much by manually reformatting all of them.

dhardy: I'm pretty sure I made that change because `rustfmt` _did_ complain about it. Maybe it doesn't for you; I run it manually on the entire project, not just the files I change.

src/core_tests.rs
Fraser999: Maybe you could create a snippet or alias for this then? (I use a snippet personally)  I don't see why they should be included in this file and not every other source file.

dhardy:
out of sync

PR #1203, opened 2016-11-07
MAID-1693 fix/states/node.rs: resend unacknowledged messages even after group s…

dhardy:
…plit


maidsafe-highfive:
r? @afck

dhardy:
r? afck

You might want to check the second commit. This _might_ be the cause of the test failure on OSX (?).

PR #1204, opened 2016-11-07
MAID-1685 chore/data: move data_types into module 

maqi:


maidsafe-highfive:
r? @afck

PR #1206, opened 2016-11-08
Maid 1687

dhardy:


maidsafe-highfive:
r? @afck

PR #1207, opened 2016-11-08
refactor/logging: give stats messages a custom target; update README

dhardy:
As discussed. This target is implemented as much as an example as anything else.


maidsafe-highfive:
r? @afck

README.md
```diff
@@ -20,6 +20,25 @@ allows the creation of a DHT based PKI and this in turn allows a decentralised n
...
+via `env_logger`. By default this prints messages of level "warn" and higher
+("error"), but not lower levels ("info", "debug", "trace"). The level can be set
+explicitly (any of the above or "off"), e.g.:
+
+    export RUST_LOG=routing=info
+
+Optionally, the following sub-targets can be controlled independently:
+
+*   stats — messages about connections and routing table size
+*   crust — messages from the mock CRust layer (not real CRust)
```
afck: "Crust" isn't written with uppercase R in the Crust readme.

src/mock_crust/crust.rs
```diff
@@ -73,13 +73,13 @@ impl Service {
     /// Start service discovery (beacon).
     /// Note: beacon is not yet implemented in mock.
     pub fn start_service_discovery(&mut self) {
-        trace!("[MOCK] start_service_discovery not implemented in mock");
+        trace!(target: "mock_crust", "[MOCK] start_service_discovery not implemented in mock");
```
afck: This should probably be `crust`. (Or the readme above should list `mock_crust`.)

dhardy: I changed half-way through, to avoid confusion with the real crust. But I guess we'll never use both at the same time.

afck: I agree, let's go with `crust`.

PR #1209, opened 2016-11-08
MAID-1686 chore/data_types: disable multi-sig feature

maqi:


maidsafe-highfive:
r? @afck

PR #1211, opened 2016-11-09
fix/routing_table: update group merge prefixes to remain consistently sorted

Fraser999:
This resolves an issue where some lists of prefixes related to merging groups were not sorted,
causing the merge test to fail sporadically.


maidsafe-highfive:
r? @afck

PR #1214, opened 2016-11-09
[DO NOT MERGE] - MapData POC

madadam:


maidsafe-highfive:
r? @afck

src/data/map_data.rs
```diff
@@ -0,0 +1,174 @@
...
+    sign_key: sign::PublicKey,
+    signature: Signature,
+    operation: PermissionOperation,
+}
+
+#[allow(unused)]
+pub enum PermissionOperation {
+    List,
+    Get {
+        // TODO: shouldn't this be `sign::PublicKey` ?
```
ustulation: yes

src/data/map_data.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// TODO: ordering based on name only. Verify it is sufficient!
+impl Ord for MapData {
+    fn cmp(&self, other: &MapData) -> Ordering {
+        self.name.cmp(&other.name)
+    }
+}
+
+// TODO: hashing based on name only. Verify is is sufficient!
```
afck: I don't think it is: In the data manager, we'll probably exchange hashes in refresh messages to synchronise the state of a data chunk. For that, the `Hash` impl needs to take into account the data and owner fields.

afck: Note that `BTreeMap` implements `Hash`, so with that, `Hash` could be derived.

ustulation:
This is taxing the CI. It is better to not raise it as PR

PR #1215, opened 2016-11-09
MAID-1706 fix/core_tests: don't fail on early GroupSplit events

afck:
The joining node can legitimately receive a `GroupSplit` message as soon as it has connected to a quorum of its new group.


maidsafe-highfive:

PR #1216, opened 2016-11-10
MAID 1694: net invariant output, refactor, doc

dhardy:
These are changes I've made while working on 1694, but not directly related to fixing it.

### DIFF
--- a/src/ack_manager.rs
+++ b/src/ack_manager.rs
@@ -41,10 +41,12 @@ pub struct AckManager {
     received: MessageFilter<Ack>,
 }
 
+/// An identifier for a waiting-to-be-acknowledged message (a hash of the message).
 #[derive(Clone, Copy, Eq, Hash, Ord, PartialEq, PartialOrd, RustcDecodable, RustcEncodable)]
 pub struct Ack(u64);
 
 impl AckManager {
+    /// Creates a new manager, with empty lists.
     pub fn new() -> Self {
         let expiry_duration = Duration::from_secs(EXPIRY_DURATION_SECS);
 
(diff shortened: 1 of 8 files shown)


maidsafe-highfive:
r? @afck

dhardy:
Travis test timed out. AppVeyer failed on `more_than_group_size_nodes` test, but it works for me.

src/ack_manager.rs
```diff
@@ -41,10 +41,12 @@ pub struct AckManager {
     received: MessageFilter<Ack>,
 }
 
+/// An identifier for a waiting-to-be-acknowledged message (a hash of the message).
 #[derive(Clone, Copy, Eq, Hash, Ord, PartialEq, PartialOrd, RustcDecodable, RustcEncodable)]
 pub struct Ack(u64);
 
 impl AckManager {
+    /// Create a new manager, with empty lists.
```
afck: Let's use the third person singular present indicative ("Creates") in method docs:
https://aturon.github.io/style/comments.html#sentence-structure
(I do like the idea of using questions for `-> bool` methods, though, if nobody objects.)

src/routing_table/network_tests.rs
```diff
@@ -316,11 +316,13 @@ pub fn verify_network_invariant<'a, T, U>(nodes: U)
             }
             if prefix1.is_compatible(prefix2) {
                 panic!("Group prefixes should be disjoint, but these are not:\n\
-                    Group {:?}: {:?}\n\
-                    Group {:?}: {:?}",
+                    Group {:?}, according to node {:?}: {:?}\n\
+                    Group {:?}, according to node {:?}: {:?}",
```
afck: Ah, yes, I already missed that bit from the logs yesterday! :+1:

dhardy: This is already my second implementation of this extension. I had nothing to test with when I first wrote this output.

src/ack_manager.rs
```diff
@@ -54,16 +56,20 @@ impl AckManager {
         }
     }
 
-    // Handle received ack.
+    // Handle received ack (remove the corresponding message from the list of
```
afck: That should be `Handles` (third person etc.) - also in the other comments.

src/states/node.rs
```diff
@@ -1344,15 +1332,13 @@ impl Node {
     }
 
     fn connect(&mut self,
-               encrypted_connection_info: Vec<u8>,
-               nonce_bytes: [u8; box_::NONCEBYTES],
-               their_public_id: PublicId,
+               conn_into: ConnectionInfo,
```
Fraser999: Typo in `conn_into`?

PR #1217, opened 2016-11-11
Maid 1708 smplify owners and signatures backport

maqi:


maidsafe-highfive:
r? @afck

src/data/priv_appendable_data.rs
```diff
@@ -19,10 +19,11 @@ use maidsafe_utilities::serialisation::{deserialise, serialise, serialised_size}
 use rust_sodium::crypto::{box_, sealedbox};
 use rust_sodium::crypto::sign::{self, PublicKey, SecretKey, Signature};
 use rustc_serialize::{Decodable, Decoder};
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::{self, Debug, Formatter};
 use xor_name::XorName;
-use super::{AppendWrapper, AppendedData, DataIdentifier, Filter, NO_OWNER_PUB_KEY, verify_detached};
+use super::{AppendWrapper, AppendedData, DataIdentifier, Filter, NO_OWNER_PUB_KEY,
+            verify_signatures};
```
afck: For functions, the policy is to import the module instead. (Probably you don't need to import anything and can just use `super::verify_signatures`.)

src/data/priv_appendable_data.rs
```diff
@@ -323,17 +263,17 @@ impl PrivAppendableData {
...
-        let current_owner_keys: Vec<String> = self.current_owner_keys
+        let current_owner_keys: Vec<String> = self.owners
             .iter()
             .map(|pub_key| ::utils::format_binary_array(&pub_key.0))
             .collect();
-        let previous_owner_signatures: Vec<String> = self.previous_owner_signatures
+        let previous_owner_signatures: Vec<String> = self.signatures
             .iter()
-            .map(|signature| ::utils::format_binary_array(&signature.0[..]))
+            .map(|(_, sig)| ::utils::format_binary_array(&sig.0[..]))
```
afck: Please import `utils` (so that the `::` can be omitted) and update the debug output to reflect the new fields: It should say `owners:` instead of `current_owner_keys:` and just print the list of tuples from the map after `signatures:` instead of `previous_owner_keys:` and `previous_owner_signatures:`.

src/data/priv_appendable_data.rs
```diff
@@ -384,296 +323,256 @@ mod test {
...
+    //                                                          vec![new_owner.0],
+    //                                                          owner_keys,
+    //                                                          BTreeSet::new(),
+    //                                                          Filter::white_list(None),
+    //                                                          encrypt_keys.0,
+    //                                                          Some(&keys1.1)));
+    //     assert_eq!(wrong_name.add_signature(&keys2.1).ok(), Some(0));
+    //     // Shall throw error of UnknownMessageType
+    //     assert!(orig_priv_appendable_data.update_with_other(wrong_name).is_err());
+    // }
```
afck: Let's remove these tests entirely then. By the time we re-enable multisig, the API might change even more and we'll have to rewrite the tests anyway.

src/data/priv_appendable_data.rs
```diff
@@ -384,296 +323,256 @@ mod test {
...
+    //             assert!(priv_appendable_data.verify_previous_owner_signatures(&owner_keys).is_ok());
+    //         }
+    //         Err(error) => panic!("Error: {:?}", error),
+    //     }
+    // }
+
+    // // TODO: this test is disabled as the feature of multi-sig is disabled
+    // #[ignore]
+    // #[test]
+    // fn transfer_owner_attack() {
```
afck: Do we have a test for a valid ownership transfer with a single owner? We probably should.

src/data/priv_appendable_data.rs
maqi: but the testing logic will still be generally the same, and updating API call is much easier than re-implement the testing case.

afck: Not sure about that. It will depend on how we deduplicate the logic and whether our set of data types changes, etc. And we can always retrieve them from git history.

But if you feel that it will be less work to update them than to rewrite them, I'd prefer to update them immediately, whenever we do the changes, and keep them `#[ignore]`d instead of commented.

src/data/priv_appendable_data.rs
```diff
@@ -306,7 +291,7 @@ mod test {
     use maidsafe_utilities::serialisation::serialise;
     use rust_sodium::crypto::{box_, sign};
     use xor_name::XorName;
-    use data::{AppendWrapper, AppendedData, DataIdentifier, Filter};
+    use data::{AppendWrapper, AppendedData, DataIdentifier, Filter, verify_signatures};
```
afck: As above, let's import the module and use `data::verify_signatures`.

PR #1227, opened 2016-11-17
Unit tests for MutableData (MAID-1721)

nbaksalyar:


maidsafe-highfive:

src/data/mutable_data.rs
```diff
@@ -308,3 +318,236 @@ impl Debug for MutableData {
...
+        let (pk1, _) = sign::gen_keypair();
+
+        let mut owners = BTreeSet::new();
+        owners.insert(owner);
+
+        let mut md =
+            unwrap!(MutableData::new(rand::random(), 0, BTreeMap::new(), BTreeMap::new(), owners));
+
+        // Try to do ownership transfer from a non-owner requester
+        assert!(md.change_owner(pk1, pk1).is_err());
```
ustulation: It is better to check for specific error, that way frontend, safe_core, routing and vault always agree on the same error (unlike right now where safe_core expects NoSuchAccount and vault (incorrectly) returns NoSuchData).

src/data/mutable_data.rs
```diff
@@ -308,3 +318,251 @@ impl Debug for MutableData {
...
+
+    use rust_sodium::crypto::sign;
+    use std::collections::{BTreeMap, BTreeSet};
+    use std::iter;
+    use error::RoutingError;
+    use super::*;
+
+    macro_rules! assert_err {
+        ($left: expr, $err: path) => {{
+            assert!(if let Err($err) = $left { true } else { false });
```
ustulation: This will not give much idea on failure - it will only say something asserted to false. If you want to write it that way ^^^ it's better to have something like:
```
match $left {
   Err($err) => (),
   x => panic!("Expected: {:?}  ### Found: {:?}", $err, x),
}
```

src/data/mutable_data.rs
```diff
@@ -308,3 +318,252 @@ impl Debug for MutableData {
                self.owners)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    extern crate rand;
```
Fraser999: I think this crate import should be covered in lib.rs.  Just a `use` statement should work here.

PR #1228, opened 2016-11-17
MAID-1731 test/routing_table: during invariant check, return error when group s…

maqi:
…hall split but doesn't


maidsafe-highfive:
r? @afck

afck:
Let's make that part of the implementation of the multi-split.

PR #1229, opened 2016-11-17
MAID-1734: fix/node: add missing pending acks

Fraser999:
This adds to pending acks beofre sending a signature or accumulating a message at hop 0.  It also
fixes a hanging mock-crust test which curently depends on acks timing out.


maidsafe-highfive:
r? @afck

PR #1230, opened 2016-11-17
MAID-1736 MAID-1730 Fix quorum inequality and ack insertion.

afck:


maidsafe-highfive:

afck:
Merged into #1231

PR #1232, opened 2016-11-18
New Request/Response types & single entry mutation function

nbaksalyar:
Modify entry mutation procedures according to the message format in the [RFC](https://github.com/maidsafe/rfcs/blob/master/text/0047-mutable-data/0047-mutable-data.md). Implement entry version checks.

### DIFF
--- a/src/data/mutable_data.rs
+++ b/src/data/mutable_data.rs
@@ -18,13 +18,14 @@
 use error::RoutingError;
 use maidsafe_utilities::serialisation::serialised_size;
 use rust_sodium::crypto::sign::PublicKey;
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::BTreeSet;
+use std::collections::btree_map::{BTreeMap, Entry};
 use std::fmt::{self, Debug, Formatter};
 use super::DataIdentifier;
 use xor_name::XorName;
 
 /// Maximum allowed size for MutableData (1 MiB)
-pub const MAX_MUTABLE_DATA_SIZE_IN_BYTES: u64 = 1_048_576;
+pub const MAX_MUTABLE_DATA_SIZE_IN_BYTES: u64 = 1024 * 1024;
 
 /// Maximum allowed entries in MutableData
 pub const MAX_MUTABLE_DATA_ENTRIES: u64 = 100;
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:

src/data/mutable_data.rs
```diff
@@ -169,59 +180,117 @@ impl MutableData {
...
+                        let _ = update.insert(key, value);
+                    }
+                    EntryAction::Del(version) => {
+                        let _ = delete.insert(key, version);
+                    }
+                };
+                (insert, update, delete)
+            });
+
+        if (insert.len() > 0 && !self.is_action_allowed(requester, Action::Insert)) ||
```
ustulation: https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.is_empty is more idiomatic i think instead of `len() > 0`

src/data/mutable_data.rs
```diff
@@ -169,59 +180,117 @@ impl MutableData {
...
+           (update.len() > 0 && !self.is_action_allowed(requester, Action::Update)) ||
+           (delete.len() > 0 && !self.is_action_allowed(requester, Action::Delete)) {
             return Err(RoutingError::AccessDenied);
         }
-        if !self.data.contains_key(&key) {
-            return Err(RoutingError::EntryNotFound);
+        if (insert.len() > 0 || update.len() > 0) &&
+           self.data.len() > MAX_MUTABLE_DATA_ENTRIES as usize {
```
ustulation: Not only number of entries - you also need to return error if there is `insert/update` while the total size > 1MiB

src/data/mutable_data.rs
```diff
@@ -98,6 +99,16 @@ impl PermissionSet {
     }
 }
 
+#[derive(Hash, Eq, PartialEq, Clone, PartialOrd, Ord)]
+pub enum EntryAction {
```
ustulation: Is this repeated ? Seems to be same as in `Request/Response` file.

src/messages_new.rs
```diff
@@ -0,0 +1,236 @@
...
+use types::MessageId as MsgId;
+use xor_name::XorName;
+
+pub enum EntryAction {
+    Update(Value),
+    Ins(Value),
+    Del(u64),
+}
+
+#[allow(missing_docs)]
```
ustulation: Let's not leave these for later - it might be missed for a long time

src/messages_new.rs
```diff
@@ -0,0 +1,236 @@
...
+        msg_id: MsgId,
+    },
+    DelAuthKey {
+        key: sign::PublicKey,
+        version: u64,
+        msg_id: MsgId,
+    },
+}
+
+#[allow(missing_docs)]
```
ustulation: Same as previous comment

ustulation:
To be raised once review comments are addressed to prevent commits from hammering CI

PR #1234, opened 2016-11-18
MAID-1729: fix/mock-crust: fix bug in teardown of mock-crust network

Fraser999:
Also re-enabled test which was particularly affected by this bug.


maidsafe-highfive:
r? @afck

PR #1236, opened 2016-11-20
MAID-1775 separate own group and exclude own ID from needed merge contacts

afck:
The routing table always has a group entry that corresponds to the node's
own group. Separating that from the map of groups is in many places more
natural and removes the need for unwrapping.


maidsafe-highfive:

PR #1238, opened 2016-11-21
MAID-1774: fix/test: fix mock crust test utility function

Fraser999:
This updates the test setup which waits for a 'GroupSplit' event to not rely on that event being
triggered by the newest node.  The new node doesn't trigger that event since it now gets constructed
with post-split groups.  This commit also re-enables a couple of tests.


maidsafe-highfive:
r? @afck

PR #1239, opened 2016-11-22
Ran rustfmt

Fraser999:


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -413,7 +413,7 @@ impl Node {
                     self.send_or_drop(&dst, bytes, content.priority())
                 } else {
                     debug!("{:?} Invalid TunnelDirect message received via {:?}: {:?} -> {:?} {:?}",
-                           self,
+                                               self,
```
afck: That looks like a rustfmt bug?

src/states/node.rs
```diff
@@ -494,8 +494,9 @@ impl Node {
                                 peer_id: PeerId)
                                 -> Result<(), RoutingError> {
         if let Some(&pub_id) = self.peer_mgr.get_routing_peer(&peer_id) {
-            if let Some((signed_msg, route)) = self.sig_accumulator
-                .add_signature(digest, sig, pub_id) {
+            if let Some((signed_msg, route)) =
+                self.sig_accumulator
+                    .add_signature(digest, sig, pub_id) {
```
afck: Could these two go on one line now or would rustfmt split them up again?

PR #1240, opened 2016-11-22
various Mutable data improvements

madadam:


maidsafe-highfive:
r? @afck

PR #1245, opened 2016-11-24
MAID-1737_sent_to

maqi:


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -1502,10 +1504,10 @@ impl Node {
             return Ok(());  // Avoid swarming back out to our own group.
         }
 
-        let target_peer_ids = self.get_targets(routing_msg, route, *hop)?;
+        let (new_sent_to, target_peer_ids) = self.get_targets(routing_msg, route, hop, sent_to)?;
```
afck: I wonder whether we can leave `sent_to` empty whenever we're not in the destination group?

maqi: I thought every `hop` shall be recorded in the `sent_to` ?

src/states/node.rs
```diff
@@ -1633,15 +1638,27 @@ impl Node {
...
+            let targets: HashSet<_> = self.peer_mgr
                 .routing_table()
-                .targets(&routing_msg.dst.to_destination(), exclude, route as usize)?;
-            Ok(self.peer_mgr.get_peer_ids(&targets))
+                .targets(&routing_msg.dst.to_destination(), *hop, route as usize)?
+                .into_iter()
+                .filter(|target| !sent_to.contains(target))
+                .collect();
+            let new_sent_to =
+                if self.peer_mgr.routing_table().is_in_our_group(routing_msg.dst.name()) {
```
afck: The target could be the name of a data chunk or something else, other than the name of a node in our group.
This should probably be `our_group_prefix().matches(routing_msg.dst.name())`.

PR #1247, opened 2016-11-26
MAID-1787 merge at most two groups at once, and fix group messages while merging

afck:
This is a - probably temporary - workaround for the problems with
sending a group message while merging or splitting. If we require only
the `MIN_GROUP_SIZE` nodes closest to the source address to sign the
message, it works in both cases, and replaces the other workarounds: we
don't need to store group lists in the ack manager anymore, and we can
stop resending messages when we moved out of the sending group.

### DIFF
--- a/src/states/node.rs
+++ b/src/states/node.rs
@@ -33,7 +33,8 @@ use messages::{ConnectionInfo, DEFAULT_PRIORITY, DirectMessage, GroupList, HopMe
 use peer_manager::{ConnectionInfoPreparedResult, ConnectionInfoReceivedResult, MIN_GROUP_SIZE,
                    PeerManager, PeerState};
 use routing_message_filter::RoutingMessageFilter;
-use routing_table::{OtherMergeDetails, OwnMergeDetails, OwnMergeState, Prefix, RemovalDetails};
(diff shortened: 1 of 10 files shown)


maidsafe-highfive:

src/states/node.rs
```diff
@@ -1810,13 +1793,12 @@ impl Node {
...
-                // node as the route
-                dst: Authority::NaeManager(prefix.substituted_in(!joining_node)),
-                content: MessageContent::GroupSplit(our_prefix),
+                // this way of calculating the source avoids using the joining node as the route
+                src: Authority::NaeManager(our_prefix.substituted_in(!joining_node)),
+                dst: Authority::NaeManager(prefix.substituted_in(joining_node)),
+                content: MessageContent::GroupSplit(our_prefix, joining_node),
             };
-            if let Err(err) = self.send_routing_message(request_msg) {
+            if let Err(err) = self.send_routing_message_via_route(request_msg, 1) {
```
Fraser999: Did you mean to leave this in (i.e. sending via route 1)?

afck: You're right, I removed it again.

PR #1249, opened 2016-11-28
MAID-1801 Log prefixes, rename stats logs, minor cleanup.

afck:


maidsafe-highfive:

PR #1251, opened 2016-11-29
MAID-1828 send refresh from our half of the merged group

afck:


maidsafe-highfive:

PR #1253, opened 2016-11-29
MAID-1830 simplify CI scripts

afck:


maidsafe-highfive:

.travis.yml
```diff
@@ -20,17 +20,22 @@ matrix:
...
+        set -x;
+        cargo fmt -- --write-mode=diff &&
+        cargo rustc --verbose -- --test -Zno-trans &&
+        cargo rustc --verbose --example ci_test -- -Zno-trans &&
+        cargo rustc --verbose --example key_value_store -- -Zno-trans &&
+        cargo rustc --verbose -- -Zno-trans &&
+        cargo test  --verbose --features use-mock-crust
+      );
+    fi
+  - curl -sSL https://github.com/maidsafe/QA/raw/master/Bash%20Scripts/Travis/run_clippy.sh | bash
```
Viv-Rajkumar: this url is in the old format. we can go to [safe_core - travis.yml](https://github.com/maidsafe/safe_core/blob/dev/.travis.yml#L36) format now.

.travis.yml
```diff
@@ -20,17 +20,22 @@ matrix:
...
 before_cache:
-  - curl -sSLO https://github.com/maidsafe/QA/raw/master/Bash%20Scripts/Travis/install_elfutils.sh
-  - . install_elfutils.sh
-after_success:
-  - curl -sSL https://github.com/maidsafe/QA/raw/master/Bash%20Scripts/Travis/after_success.sh | bash
+  - curl -sSL https://github.com/maidsafe/QA/raw/master/Bash%20Scripts/Travis/run_cargo_prune.sh | bash
```
Viv-Rajkumar: this again similar to `safe_core` can just be `cargo prune`

.travis.yml
```diff
@@ -20,17 +20,22 @@ matrix:
...
-  # - cat output.log
+  - if [ "${TRAVIS_RUST_VERSION}" = stable ]; then
+      (
+        set -x;
+        cargo fmt -- --write-mode=diff &&
+        cargo rustc --verbose -- --test -Zno-trans &&
+        cargo rustc --verbose --example ci_test -- -Zno-trans &&
+        cargo rustc --verbose --example key_value_store -- -Zno-trans &&
+        cargo rustc --verbose -- -Zno-trans &&
+        cargo test  --verbose --features use-mock-crust
```
Viv-Rajkumar: I'm guessing running `cargo test` with `use-mock-crust` also includes tests which don't rely on the feature?

afck: Yes, it does. I think the only thing it excludes is `tests/ci.rs`, and we haven't been running that on Travis before this change either.

PR #1254, opened 2016-11-30
MAID-1827 MAID-1831: fixed handling of 'OwnGroupMerge' messages

Fraser999:
Also added a basic regression test for merging three groups into one.

### DIFF
--- a/tests/mock_crust/utils.rs
+++ b/tests/mock_crust/utils.rs
@@ -325,10 +325,18 @@ pub fn create_connected_nodes_with_cache(network: &Network,
     nodes
 }
 
-pub fn create_connected_nodes_with_cache_till_split(network: &Network) -> Vec<TestNode> {
+// This creates new nodes (all with `use_cache` set to `true`) until the specified number of unique
+// split events have happened.  `split_count` should be at least 1.  Note that the final shape of
+// the network is still random, e.g. if `split_count` is 5, then the network could comprise groups
+// [0, 10, 110, 111] or [00, 01, 10, 11].
+pub fn create_connected_nodes_with_cache_until_split(network: &Network,
+                                                     split_count: usize)
+                                                     -> Vec<TestNode> {
+    assert!(split_count > 0);
     let use_cache = true;
     let mut nodes = create_connected_nodes_with_cache(network, MIN_GROUP_SIZE * 2, use_cache);
     let config = Config::with_contacts(&[nodes[0].handle.endpoint()]);
+    let mut split_events = HashSet::new();
 
     'outer: loop {
         let len = nodes.len();
@@ -344,11 +352,21 @@ pub fn create_connected_nodes_with_cache_till_split(network: &Network) -> Vec<Te
                     Event::NodeAdded(..) |
                     Event::Connected |
                     Event::Tick => (),
-                    Event::GroupSplit(..) => break 'outer,
+                    Event::GroupSplit(prefix) => {
+                        let _ = split_events.insert(prefix);
+                        if split_events.len() == split_count {
+                            break 'outer;
+                        }
+                    }
                     event => panic!("Got unexpected event: {:?}", event),
                 }
             }
         }
+    }
+
+    // Clear all event queues
+    for node in &nodes {
+        while node.event_rx.try_recv().is_ok() {}
     }
 
     nodes
--- a/tests/mock_crust_tests.rs
+++ b/tests/mock_crust_tests.rs
@@ -40,6 +40,10 @@
 extern crate itertools;
 #[macro_use]
 extern crate log;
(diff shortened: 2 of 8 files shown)


maidsafe-highfive:
r? @afck

tests/mock_crust/utils.rs
```diff
@@ -325,10 +325,18 @@ pub fn create_connected_nodes_with_cache(network: &Network,
     nodes
 }
 
-pub fn create_connected_nodes_with_cache_till_split(network: &Network) -> Vec<TestNode> {
+// This creates new nodes (all with `use_cache` set to `true`) until the specified number of unique
+// split events have happened.  `split_count` should be at least 1.  Note that the final shape of
+// the network is still random, e.g. if `split_count` is 5, then the network could comprise groups
```
afck: Wouldn't that be 3 instead of 5?

tests/mock_crust_tests.rs
```diff
@@ -40,6 +40,10 @@
 extern crate itertools;
 #[macro_use]
 extern crate log;
+#[cfg_attr(feature="clippy", allow(useless_attribute))]
+#[allow(unused_extern_crates)]
+#[macro_use]
+extern crate maidsafe_utilities;
```
afck: Is that to make the tests compile with and without logging?
I usually just put the two lines
```rust
extern crate maidsafe_utilities;
let _ = maidsafe_utilities::log::init(false);
```
directly into the test function; that's not much more work that just the one line. Not sure about this …

tests/mock_crust/utils.rs
```diff
@@ -344,13 +352,23 @@ pub fn create_connected_nodes_with_cache_till_split(network: &Network) -> Vec<Te
                     Event::NodeAdded(..) |
                     Event::Connected |
                     Event::Tick => (),
-                    Event::GroupSplit(..) => break 'outer,
+                    Event::GroupSplit(prefix) => {
+                        let _ = split_events.insert(prefix);
+                        if split_events.len() == split_count {
+                            break 'outer;
```
afck: Maybe it would be cleaner to just use
```rust
while split_events.len() < split_count
```
above. Also, the set of split prefixes doesn't seem to be needed, so you could just make `split_count` count _down_ to 0 instead?

tests/mock_crust/utils.rs
Fraser999: It took me a few shots to get this right, but I think 5 is correct :)  The events notify about the new prefix, so there would be `0`, `1`, `10`, `11`, then finally `110` or `111` takes us to the state in the first example.

tests/mock_crust/utils.rs
Fraser999: I can use `while split_events.len() < split_count` if you prefer, but this won't break us out of the nested loop on line 350.  Not really a big deal - we clear all the rest of the enqueued events just after this anyway, but we'd be calling `split_events.insert(prefix)` more often than required.

Also, it's a set rather than just a counter as we'll receive multiple identical events since we're polling all the nodes here.

tests/mock_crust_tests.rs
Fraser999: I agree, but I added it since [I used `unwrap!()`](https://github.com/maidsafe/routing/pull/1254/files#diff-12e885abc7235cb34d3e197fb87dbb2aR37).

tests/mock_crust/utils.rs
afck: Right, I missed that. Makes sense! :+1:

tests/mock_crust/utils.rs
afck: You're right, I missed that we're putting the _new_ prefix into the event. :+1:

tests/mock_crust_tests.rs
afck: Shouldn't it be `#[cfg(test)]` or `#cfg[cfg_attr(feature="use-mock-crust")]` then?

Fraser999: Um, not sure.  I'd have thought `#[cfg(test)]` would be true for all files inside the /tests folder?  And we already have [`use-mock-crust` set module wide](https://github.com/maidsafe/routing/pull/1254/files#diff-08fe2788cc792d4f0a04808a8c92da53R18), so `#cfg[cfg_attr(feature="use-mock-crust")]` will also always be true here?

Basically, I'd like to just have
```rust
#[macro_use]
extern crate maidsafe_utilities;
```
but that yields a compiler warning about an unused extern crate since we're only using a macro from it.  That forces the use of `#[allow(unused_extern_crates)]`, but clippy flags that as a "useless attribute", hence the need to disable the clippy check for that import line.

PR #1260, opened 2016-12-05
Churn test: add ClientManager src/dst, enable logging

dhardy:
Fails sometimes (see internal discussion), but passed the last six times I ran this test. (The failure may also have been to do with reduced group size.)

### DIFF
--- a/tests/mock_crust/churn.rs
+++ b/tests/mock_crust/churn.rs
@@ -168,17 +168,22 @@ fn churn() {
         let auth_n1 = Authority::ManagedNode(nodes[index1].name());
         let auth_g0 = Authority::NaeManager(rng.gen());
         let auth_g1 = Authority::NaeManager(rng.gen());
+        let auth_c0 = Authority::ClientManager(rng.gen());
 
         let mut expected_gets = ExpectedGets::default();
 
-        // Test messages from a node to itself, another node and a group ...
+        // Test messages from a node to itself, another node, and another group ...
         expected_gets.send_and_expect(data_id, auth_n0, auth_n0, &nodes, min_group_size);
         expected_gets.send_and_expect(data_id, auth_n0, auth_n1, &nodes, min_group_size);
         expected_gets.send_and_expect(data_id, auth_n0, auth_g0, &nodes, min_group_size);
-        // ... and from a group to itself, another group and a node.
+        // ... from a group to itself, another group, a node, and a client authority ...
         expected_gets.send_and_expect(data_id, auth_g0, auth_g0, &nodes, min_group_size);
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @afck

tests/mock_crust/churn.rs
```diff
@@ -151,6 +154,8 @@ const CHURN_ITERATIONS: usize = 100;
 
 #[test]
 fn churn() {
+    unwrap!(maidsafe_utilities::log::init(false));
```
Fraser999: I don't think logging should be enabled in tests except locally to debug.

dhardy:
Testing results: six passes, one failure with seed `[3992484640, 1172583158, 66367514, 3895083731]`: `Unexpected request for node e4d3da`.

tests/mock_crust/churn.rs
```diff
@@ -74,20 +74,20 @@ impl ExpectedGets {
                        dst: Authority,
                        nodes: &[TestNode],
                        min_group_size: usize) {
-        let src_dest = src.to_destination();
-        let dst_dest = dst.to_destination();
+        let src_addr = src.to_destination();
+        let dst_addr = dst.to_destination();
```
afck: Let's not mix random cleanups with actual features/fixes/tests/etc.
As discussed, the `Destination` type will probably go away anyway, and then this confusion will be automatically removed, too, so I'd just not change this for now. If you must, change `src` and `dst` to `src_auth` and `dst_auth`, and make these two here just `src` and `dst`: We sometimes use "address" for a `XorName`, so these names would be just as confusing. (But as a separate commit, together with any other cleanups not related to adding the clients as senders and recipients.)

dhardy:
@afck refactoring changes were deleted

tests/mock_crust/churn.rs
```diff
@@ -168,17 +168,22 @@ fn churn() {
         let auth_n1 = Authority::ManagedNode(nodes[index1].name());
         let auth_g0 = Authority::NaeManager(rng.gen());
         let auth_g1 = Authority::NaeManager(rng.gen());
+        let auth_c0 = Authority::ClientManager(rng.gen());
```
afck: A `ClientManager` is just a group like the other ones and isn't handled differently in routing, so this doesn't really add to the test.

What we'll need to test is `Authority::Client`: Sending from a client to a group (and possibly node) and vice-versa. That will require some modification to the `ExpectedGets`, and adding clients in addition to the nodes here in the first place.

We also need to make sure to either never churn out the clients' proxy nodes, or to somehow deal with reconnecting.

dhardy:
Closing; this needs refactoring

PR #1262, opened 2016-12-07
MAID-1848: Added further mock-crust tests for merging logic

Fraser999:


maidsafe-highfive:
r? @afck

tests/mock_crust/utils.rs
```diff
@@ -33,7 +33,7 @@ use std::sync::mpsc;
 const BALANCED_POLLING: bool = true;
 
 
-// —————  Random number generation  —————
+// -----  Random number generation  -----
```
dhardy: Isn't m-dash better than hypen for line breaks?

Fraser999: It doesn't display correctly in my git gui.

src/node.rs
```diff
@@ -489,6 +489,17 @@ impl Node {
...
+    /// Sets a name to be used when the next node relocation request is received by this node.
+    pub fn set_next_node_name(&self, relocation_name: XorName) {
+        self.machine.borrow_mut().current_mut().set_next_node_name(Some(relocation_name))
+    }
+
+    /// Clears the name to be used when the next node relocation request is received by this node so
+    /// the normal process is followed to calculate the relocated name.
+    pub fn clear_next_node_name(&self) {
+        self.machine.borrow_mut().current_mut().set_next_node_name(None)
+    }
```
afck: All of these methods should probably be `use-mock-crust`-only. (Also in the other files.)

tests/mock_crust/merge.rs
```diff
@@ -15,42 +15,75 @@
...
     verify_invariant_for_all_nodes(&nodes);
 
-    // Drop nodes from the group with the short prefix until we get a merge event
-    'outer: loop {
-        let range = Range::new(0, nodes.len());
-        let index = range.ind_sample(&mut rng);
-        if unwrap!(nodes[index].routing_table()).our_group_prefix().bit_count() != 1 {
-            continue;
+    // Drop nodes from a group with the shortest prefix until we get a merge event for the empty
+    // Prefix.
```
afck: I'd put prefix either uppercase in backticks or lowercase without backticks.

tests/mock_crust/merge.rs
```diff
@@ -15,42 +15,75 @@
...
         let _ = nodes.remove(index);
         poll_and_resend(&mut nodes, &mut []);
         for node in &nodes {
             while let Ok(event) = node.event_rx.try_recv() {
                 match event {
                     Event::NodeLost(..) |
                     Event::Tick => (),
-                    Event::GroupMerge(..) => break 'outer,
-                    event => panic!("Got unexpected event: {:?}", event),
+                    Event::GroupMerge(prefix) => {
```
afck: With #1259 we need to expect `NodeAdded` here, too, I think.

tests/mock_crust/merge.rs
```diff
@@ -15,42 +15,75 @@
...
+    let mut min_prefix = *unwrap!(nodes[0].routing_table()).our_group_prefix();
+    loop {
+        rng.shuffle(&mut nodes);
+        let mut index = nodes.len();
+        for (i, node) in nodes.iter().enumerate() {
+            let this_prefix = *unwrap!(node.routing_table()).our_group_prefix();
+            if this_prefix.bit_count() < min_prefix.bit_count() {
+                min_prefix = this_prefix;
+                index = i;
+                break;
```
afck: So `min_prefix` is not necessarily the shortest prefix now, but just one that is at least shorter than `node[0]`'s?

tests/mock_crust/utils.rs
afck: Let's stick to ASCII if in doubt.

tests/mock_crust/utils.rs
```diff
@@ -329,51 +329,78 @@ pub fn create_connected_nodes_with_cache(network: &Network,
...
+// `prefix_lengths` is an array representing the required `bit_count`s of the group prefixes.  For
+// example passing [1, 2, 3, 3] could yield a network comprising groups [0, 100, 101, 11], or
+// passing [2, 2, 3, 3, 3, 3] could yield [000, 001, 01, 100, 101, 11], while passing [1, 1] will
+// always yield groups [0, 1].
+//
+// The array is sanity checked (e.g. it would be an error to pass [1, 1, 1]), must comprise at
+// least two elements, and every element must be no more than `8`.
 pub fn create_connected_nodes_with_cache_until_split(network: &Network,
-                                                     split_count: usize)
+                                                     mut prefix_lengths: Vec<usize>)
```
afck: I wonder whether it wouldn't be easier to just specify a `Vec<Prefix>` directly. That would also simpler to sanity check.

tests/mock_crust/utils.rs
```diff
@@ -421,3 +448,67 @@ pub fn gen_bytes<R: Rng>(rng: &mut R, size: usize) -> Vec<u8> {
...
+        panic!("The specified prefix lengths {:?} would not cover the entire address space",
+               prefix_lengths);
+    } else if sum > 256 {
+        panic!("The specified prefix lengths {:?} would require overlapping groups",
+               prefix_lengths);
+    }
+}
+
+fn prefixes<T: Rng>(prefix_lengths: &[usize], rng: &mut T) -> Vec<Prefix<XorName>> {
+    let mut prefixes = vec![Prefix::new(prefix_lengths[0], rng.gen())];
```
afck: If the prefix lengths aren't sorted in ascending order, this method could loop forever, I think.
E.g. `[2,2,1]` could make it choose `10` and `01`, unable to find space for a prefix with length 1.

tests/mock_crust/utils.rs
Fraser999: I did consider that approach, but wanted to make it easy to randomise the actual prefixes used.  I realise there _shouldn't_ be any difference between testing with [`0`, `10`, `11`] and [`00`, `01`, `1`], but I'd rather not make that assumption!

I guess it would be trivial to make a new function which does take a `Vec<Prefix>` and have `create_connected_nodes_with_cache_until_split()` call that after generating the random prefixes?

tests/mock_crust/utils.rs
Fraser999: Yes, I am sorting the prefixes before calling this function, but it would be safer to just do that inside too.  Or would you prefer I add a check to ensure the slice is sorted?

tests/mock_crust/utils.rs
afck: OK, makes sense. No, I'm fine with your approach, too. :+1:

src/node.rs
afck: Ah, sorry, I had missed that! :+1:

PR #1263, opened 2016-12-08
feat/node: log prefixes and connected nodes

afck:
Add the necessary log statements to make it possible to compute a node's
current routing table from the logs, i.e. log every time a node is added
to or removed from the routing table, or when the prefixes change.


maidsafe-highfive:

PR #1265, opened 2016-12-08
style/states/node: clean up complex message matching patterns

dhardy:
No, not an assigned issue, but something I decided to try this evening. Glob use patterns only affect the local method and don't trigger any Clippy lints.


maidsafe-highfive:
r? @afck

PR #1271, opened 2016-12-13
MAID-1833: refactor `needed` to `expected_peers`

dhardy:


maidsafe-highfive:
r? @afck

src/routing_table/network_tests.rs
```diff
@@ -378,6 +379,7 @@ fn groups_have_identical_routing_tables() {
 }
 
 #[test]
+#[ignore]
```
afck: As discussed, this should still work, if the test now keeps track of the `expected_peers` per node.

src/peer_manager.rs
```diff
@@ -415,6 +451,17 @@ impl PeerManager {
             group: group.iter().map(|public_id| *public_id.name()).collect(),
         };
         let needed_names = self.routing_table.merge_other_group(merge_details);
+        let mut expected_peers = HashMap::new();
+        mem::swap(&mut self.expected_peers, &mut expected_peers);
+        expected_peers.extend(needed_names.iter()
+            .filter_map(|name| if self.peer_map.get_by_name(name).is_none() {
+                Some((*name, Instant::now()))
+            } else {
+                    #[cfg(test)]
```
dhardy: Are you _sure_ you want this @afck ? I would just not bother filtering here.

src/peer_manager.rs
```diff
@@ -369,9 +379,24 @@ impl PeerManager {
                 }
             }
         }
+
+        let mut old_expected_peers = HashMap::new();
+        mem::swap(&mut self.expected_peers, &mut old_expected_peers);
```
afck: With `mem::replace` this would be a single line. (Also below.)

src/peer_manager.rs
```diff
@@ -401,6 +426,17 @@ impl PeerManager {
             merge_prefix: merge_prefix,
             groups: groups_as_names,
         };
+        let mut expected_peers = HashMap::new();
+        mem::swap(&mut self.expected_peers, &mut expected_peers);
+        expected_peers.extend(own_merge_details.groups
+            .values()
+            .flat_map(|group| group.iter())
+            .filter_map(|name| if self.peer_map.get_by_name(name).is_none() {
```
afck: Since we only remove from `expected_peers` in `add_to_routing_table`, maybe the right criterion here would be whether it's already in `Routing` state. Every peer that isn't probably should go in?

src/routing_table/network_tests.rs
```diff
@@ -166,15 +170,18 @@ impl Network {
...
                                                       (targets, merge_details));
-                            // add needed contacts
-                            let needed = target_node.needed().clone();
-                            for needed_contact in &needed {
-                                let _ = target_node.add(*needed_contact);
+                            // Forcibly add new connections.
+                            for name in node_expected.iter() {
+                                // Try adding each node we should be connected to.
+                                // Ignore failures and ignore splits.
+                                let _ = target_node.add(*name);
```
afck: Should `name ` be removed from `node_expected`, except if `add` returned an error?

src/peer_manager.rs
```diff
@@ -276,19 +282,20 @@ impl PeerManager {
...
+        self.unknown_peers.clear();
+        self.expected_peers.clear();
         let min_group_size = self.routing_table.min_group_size();
         self.our_public_id = our_public_id;
-        let groups_as_names = groups.into_iter()
-            .map(|&(ref prefix, ref members)| {
-                (*prefix, members.into_iter().map(|pub_id| *pub_id.name()).collect_vec())
-            })
-            .collect_vec();
+        self.expected_peers.extend(groups.iter()
```
dhardy: Avoid copying what, the names? But we can't extract those from the `PublicId`s they are contained in without copying, anyway.

src/peer_manager.rs
```diff
@@ -276,19 +282,20 @@ impl PeerManager {
...
+            .flat_map(|&(_, ref members)| members)
+            .map(|id| (*id.name(), Instant::now())));
+        let prefixes = groups.iter().map(|&(ref prefix, _)| *prefix).collect_vec();
```
dhardy: What keys? The type is `&[Group]` where `Group` is a tuple.

PR #1272, opened 2016-12-14
refactor/routing_table: simplify merge logic

afck:


maidsafe-highfive:

PR #1275, opened 2016-12-14
Enable multiple_joining_nodes test, and update Travis script

afck:


maidsafe-highfive:

PR #1277, opened 2016-12-15
MAID-1877: refactorings and tie up a loose end

dhardy:
These don't directly relate to the JIRA issue, but were done while reading code related to this issue.

Best to read these commits separately.

### DIFF
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -319,9 +319,9 @@ impl PeerManager {
                                    our_public_id: &PublicId)
                                    -> Result<Vec<Group>, RoutingTableError> {
         self.routing_table.validate_joining_node(expected_name)?;
-        let groups = self.routing_table.groups();
+        let sections = self.routing_table.all_sections();
         let mut result = vec![];
-        for (prefix, names) in groups {
+        for (prefix, names) in sections {
             let mut public_ids = vec![];
             for name in names {
                 if name == *our_public_id.name() {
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @afck

src/peer_manager.rs
```diff
@@ -284,7 +284,7 @@ impl PeerManager {
     pub fn reset_routing_table(&mut self, our_public_id: PublicId, groups: &[Group]) {
         self.unknown_peers.clear();
         self.expected_peers.clear();
-        let min_group_size = self.routing_table.min_group_size();
+        let min_group_size = self.routing_table.min_size();
```
afck: I'd call that `min_section_size` or `merge_size` or leave it as `min_group_size` for now. To me, `min_size` sounds like it refers to the whole routing table.

(I also wonder whether in general we should use `len` instead of `size`, as that's what the standard library does.)

src/states/node.rs
```diff
@@ -910,11 +906,12 @@ impl Node {
     // Currently we only send this when nodes join and it's only used to add missing members.
     fn send_section_update(&mut self) {
         trace!("{:?} Sending section update", self);
-        let names = self.peer_mgr.routing_table().our_names();
+        // TODO: do we need to clone?
```
afck: Good point! We could just pass `self.peer_mgr.routing_table().our_section()` directly into `get_pub_ids` now.

PR #1278, opened 2016-12-16
MAID-1876: Changed connection info flow into request and response

Fraser999:


maidsafe-highfive:
r? @afck

src/messages.rs
```diff
@@ -475,8 +456,30 @@ pub enum MessageContent {
...
+    /// have the keys.
+    ConnectionInfoRequest {
+        /// Encrypted Crust connection info.
+        encrypted_connection_info: Vec<u8>,
+        /// Nonce used to provide a salt in the encrypted message.
+        nonce_bytes: [u8; box_::NONCEBYTES],
+        /// The sender's public ID.
+        public_id: PublicId,
+        /// The message's unique identifier.
+        message_id: MessageId,
```
maqi: as we already generate a new `nonce` each time, will it already do the job of avoiding filtered out by using a new msg_id?

src/states/node.rs
```diff
@@ -971,63 +1038,131 @@ impl Node {
...
+                info!("{:?} Already sent a connection info request to {:?} ({:?}); resending our \
+                      same details as a response.",
+                      self,
+                      public_id.name(),
+                      peer_id);
+                let _ = self.do_send_connection_info(our_info.to_pub_connection_info(),
+                                                     public_id,
+                                                     dst,
+                                                     src,
+                                                     Some(message_id));
```
maqi: what shall we do if that `send` fails? shall we at least put a TODO comment?

src/states/node.rs
```diff
@@ -971,63 +1038,131 @@ impl Node {
...
 
-    fn handle_connection_info_from_node(&mut self,
-                                        conn_info: ConnectionInfo,
-                                        src_name: XorName,
-                                        dst: Authority<XorName>)
-                                        -> Result<(), RoutingError> {
-        self.peer_mgr.allow_connect(&src_name)?;
-        self.connect(conn_info, dst, Authority::ManagedNode(src_name))
+    fn handle_connection_info_response(&mut self,
+                                       encrypted_connection_info: Vec<u8>,
```
maqi: looks like the only difference between `handle_connection_info_response` and `handle_connection_info_request` is whether sending back a connection_info.
is it possible to merge this two function?

src/states/node.rs
```diff
@@ -971,63 +1038,131 @@ impl Node {
...
+                      same details as a response.",
+                      self,
+                      public_id.name(),
+                      peer_id);
+                let _ = self.do_send_connection_info(our_info.to_pub_connection_info(),
+                                                     public_id,
+                                                     dst,
+                                                     src,
+                                                     Some(message_id));
+                if let Err(error) = self.crust_service.connect(our_info, their_info) {
```
maqi: is it necessary to carry out this connect attempt on receiving a `connection_info_request`?
I though a connect attempt shall only be made in case of receiving a `connection_info_response`?

src/states/node.rs
```diff
@@ -1756,11 +1859,22 @@ impl Node {
...
+            trace!("{:?} Not sending connection info request to {:?}",
+                       self,
+                       their_name);
+            return Ok(());
+        };
+        trace!("{:?} Resending connection info request to {:?}",
                    self,
                    their_name);
-        }
+        let _ = self.do_send_connection_info(our_pub_info, their_public_id, src, dst, None);
```
maqi: any particular reason we are not returning the result of `do_send_connection_info` directly?

src/states/node.rs
```diff
@@ -636,12 +636,43 @@ impl Node {
...
+                                      public_id,
+                                      message_id },
+             ManagedNode(src_name),
+             dst @ ManagedNode(_)) => {
+                self.handle_connection_info_response(encrypted_connection_info,
+                                                     nonce_bytes,
+                                                     public_id,
+                                                     message_id,
+                                                     src_name,
+                                                     dst)
```
afck: Not important, but I wonder whether with names like `enc_conn_info`, `pub_id`, `msg_id` and `nonce` we could save some vertical space here and make this a bit easier on the eye.

src/states/node.rs
```diff
@@ -937,6 +968,51 @@ impl Node {
         }
     }
 
+    // If `msg_id` is `Some` this is sent as a response, otherwise as a request.
+    fn do_send_connection_info(&mut self,
```
afck: As discussed, this could probably be `send_connection_info` now?

src/states/node.rs
afck: Or log it, I guess.

src/states/node.rs
afck: No, that's right: We call connect when we send or receive the response. The response goes only one way, after all, and both sides need to call `connect`.

src/states/node.rs
Fraser999: I did consider that, and abstracted what I could to a new function to cut down on duplication, but I think there are enough differences inside the match (which makes up the bulk of the functions) to justify having them separate.

src/states/node.rs
Fraser999: Simply to maintain the previous behaviour.

src/states/node.rs
afck: That's true. How about `crypt_conn_info`?
Anyway, I'm fine with the verbose version, too, if you prefer it. :+1:

src/messages.rs
afck: As discussed, let's keep that for debugging purposes for now.

src/states/node.rs
```diff
@@ -971,63 +1033,134 @@ impl Node {
                         debug!("{:?} Prepared connection info for {:?}.",
                                self,
                                pub_id.name());
+                        if let Some(our_info) = self.peer_mgr.our_connection_info(pub_id.name()) {
```
afck: Why not just call `our_connection_info.to_pub_connection_info()` above? Then this method, and in particular the unreachable `error!`, wouldn't be necessary.

Fraser999: Yes - that makes more sense.

PR #1279, opened 2016-12-16
fix/routing_table: minor fixes and cleanups

afck:
Never return our own name from `RoutingTable::targets`.
Don't split if other groups are close to merging with us.


maidsafe-highfive:

PR #1282, opened 2016-12-19
MAID-1877 feat/message accumulation: check Section and PrefixSection sources co…

dhardy:
…rrectly

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -315,7 +315,13 @@ impl SignedMessage {
         if !has_enough_sigs(self, min_group_size) {
             return false;
         }
-        // Remove invalid signatures, then check again that we have enough:
+        // Remove invalid signatures, then check again that we have enough.
+        // We also check (again) that all messages are from valid senders, because the message
+        // may have been sent from another node, and we cannot trust that that node correctly
+        // controlled which signatures were added.
+        // TODO (1677): we also need to check that the sending_nodes list corresponds to the
+        // section(s) at some point in recent history; i.e. that it was valid; but we shouldn't
+        // force it to match our own because our routing table may have changed since.
 
         let signed_bytes = match serialise(&self.content) {
             Ok(serialised) => serialised,
@@ -329,11 +335,10 @@ impl SignedMessage {
         let invalid_signatures = self.signatures
             .iter()
             .filter_map(|(pub_id, sig)| {
-                if sign::verify_detached(sig, &signed_bytes, pub_id.signing_public_key()) {
-                    None
-                } else {
-                    Some(*pub_id)
-                }
+                // Remove if not in sending nodes or signature is invalid:
+                let c = self.sending_nodes.contains(pub_id) &&
+                        sign::verify_detached(sig, &signed_bytes, pub_id.signing_public_key());
+                if c { None } else { Some(*pub_id) }
             })
             .collect_vec();
         for invalid_signature in &invalid_signatures {
@@ -1096,7 +1101,7 @@ mod tests {
             unwrap!(SignedMessage::new(routing_message, &full_id_0, sending_nodes));
         assert_eq!(signed_msg.signatures.len(), 1);
 
-        // Try to add a signature which will not correspond to an ID in the first group list.
+        // Try to add a signature which will not correspond to an ID from the sending nodes.
         let irrelevant_sig = match unwrap!(signed_msg.routing_message()
             .to_signature(irrelevant_full_id.signing_private_key())) {
             DirectMessage::MessageSignature(_, sig) => {
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -194,7 +194,7 @@ impl HopMessage {
 
 /// A list of a group's public IDs, together with a list of signatures of a neighbouring group.
 #[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Hash, RustcEncodable, RustcDecodable, Debug)]
-pub struct GroupList {
+pub struct SectionList {
     // TODO(MAID-1677): pub signatures: BTreeSet<(PublicId, sign::Signature)>,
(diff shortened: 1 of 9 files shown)


maidsafe-highfive:
r? @afck

src/messages.rs
```diff
@@ -202,19 +202,26 @@ pub struct GroupList {
 pub struct SignedMessage {
     /// A request or response type message.
     content: RoutingMessage,
-    /// The lists of the groups involved in routing this message, in chronological order.
-    grp_lists: Vec<GroupList>,
+    /// Nodes sending the message (those expected to sign it)
+    sending_nodes: BTreeSet<PublicId>,
```
afck: I wonder whether that should be a set of `SectionList`s instead: In MAID-1677, we will need to validate these public IDs using the first entry of `sec_lists`. If that first entry is the list of members of the accumulating node's section, and `sending_nodes` are section lists _signed_ by that section, that should suffice.

For validating the signatures of the message itself, you would then still use the union of these sections, as you already do.

src/messages.rs
```diff
@@ -347,6 +337,8 @@ impl SignedMessage {
         for invalid_signature in &invalid_signatures {
             let _ = self.signatures.remove(invalid_signature);
         }
+
+        has_enough_sigs(self, min_group_size)
```
afck: Before this, we also need to remove all signatures from nodes that are not among the `sending_nodes`. The `SignedMessage` instance could have been deserialised, not locally constructed, so the check in `add_signature` doesn't guard against this.

src/messages.rs
```diff
@@ -1090,10 +1083,15 @@ mod tests {
...
+        let sending_nodes =
+            vec![*full_id_0.public_id(), *full_id_1.public_id(), *full_id_2.public_id()]
+                .into_iter()
+                .collect();
+        let mut signed_msg =
+            unwrap!(SignedMessage::new(routing_message, &full_id_0, sending_nodes));
         assert_eq!(signed_msg.signatures.len(), 1);
 
-        // Add a signature which will not correspond to an ID in the first group list.
+        // Try to add a signature which will not correspond to an ID in the first group list.
```
afck: → `an ID from the sending nodes`?

src/routing_table/authority.rs
```diff
@@ -22,8 +22,12 @@ use super::{Prefix, Xorable};
...
 /// An entity that can act as a source or destination of a message.
 ///
-/// An `Authority` can be an individual `Client` or `ManagedNode`, or a group of nodes, like a
-/// `NodeManager`, `ClientManager` or `NaeManager`.
+/// `Client` and `ManagedNode` are single-node authorities (i.e. no verification of messages from
+/// additional sources needed); other authorities require agreement by a quorum of some set.
+/// `NodeManager`, `ClientManager` and `NaeManager` use _group_ verification of messages: they
+/// require quorum agreement from the `min_group_size` nodes closest to the source, while `Section`
+/// and `PrefixSection` use _section_ verification: the set from which a quorum is required is all
+/// members of the section (`Section`) or of all sections matching the prefix (`PrefixSection`).
```
afck: Nice catch, we had missed that! :+1:

src/routing_table/mod.rs
```diff
@@ -1141,6 +1141,7 @@ mod tests {
         let known_neighbour = unwrap!(known_neighbour);
         assert!(!table.our_prefix.matches(&known_neighbour));
 
+        // TODO: this can fail since `table.iter()` excludes own_name; maybe it shouldn't?
```
afck: Bart already has a fix in #1280:
https://github.com/maidsafe/routing/pull/1280/commits/7d954ead027ea1cbe86d3518a32979afe1eea017

src/signature_accumulator.rs
```diff
@@ -185,20 +180,24 @@ mod tests {
...
                 .collect();
             Env {
                 _msg_sender_id: msg_sender_id,
                 other_ids: other_ids,
-                group_list: GroupList { pub_ids: pub_ids },
+                senders: pub_ids,
                 msgs_and_sigs: msgs_and_sigs,
             }
         }
+        fn num_nodes(&self) -> usize {
```
afck: Please separate `fn`s by blank lines.

src/states/client.rs
```diff
@@ -337,22 +338,23 @@ impl Bootstrapped for Client {
...
+        let (proxy_peer_id, sending_nodes) =
+            if let Authority::Client { ref proxy_node_name, .. } = routing_msg.src {
+                (if *self.proxy_public_id.name() == *proxy_node_name {
+                     self.proxy_peer_id
+                 } else {
+                     error!("{:?} - Unable to find connection to proxy node in proxy map",
+                            self);
+                     return Err(RoutingError::ProxyConnectionNotFound);
+                 },
+                 iter::once(*self.full_id().public_id()).collect())
```
afck: That doesn't need to be part of that large `if` clause, since it's always only going to be that single ID. You could just assign `sending_nodes` in line 356.

tests/ci.rs
```diff
@@ -564,6 +564,7 @@ fn core() {
 }
 
 #[test]
+#[ignore]
```
afck: Is that already broken on master?

src/messages.rs
dhardy: Maybe; I didn't want to think that far ahead. Can we leave it until we get there?

src/routing_table/mod.rs
dhardy: Right. If his PR gets merged first I'll remove this.

tests/ci.rs
dhardy: Is it not? I thought this test was pretty unreliable? Not actually sure if we want it disabled, but it fails pretty often when I try it.

PR #1284, opened 2016-12-20
MAID-1877

dhardy:
Use `SectionList`s to verify we have enough signatures from the correct sender when accumulating and receiving `SignedMessage`s.
(Does not check the lists are correct yet; that's issue 1677.)

### DIFF
--- a/src/states/node.rs
+++ b/src/states/node.rs
@@ -27,8 +27,8 @@ use id::{FullId, PublicId};
 use itertools::Itertools;
 use log::LogLevel;
 use maidsafe_utilities::serialisation;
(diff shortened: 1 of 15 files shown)


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -2272,14 +2282,24 @@ impl Bootstrapped for Node {
             None => Ok(()),
             Some(target_name) if target_name == *self.name() => {
                 trace!("{:?} Starting message accumulation for {:?}", self, signed_msg);
-                self.accumulate_message(signed_msg, route)
+                let our_name = *self.name();
```
afck: That's `target_name`. (Maybe `target_name` should just be called `our_name` in this match arm.)

PR #1285, opened 2016-12-21
refactor/client: expose `crust::Config`

vinipsmaker:
**WARNING**: This PR needs this other one first: https://github.com/maidsafe/crust/pull/807

And then I need to change `Cargo.toml` to use upstream `crust` again

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,8 @@ version = "0.25.1"
 
 [dependencies]
 clippy = {version = "=0.0.99", optional = true}
-crust = "~0.19.0"
+#crust = "~0.20.0"
+crust = { git = "https://github.com/maidsafe/crust.git", branch = "master" }
 itertools = "~0.5.5"
 log = "~0.3.6"
 lru_time_cache = "~0.5.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -60,7 +60,7 @@
 //!
 //! let (sender, _receiver) = mpsc::channel::<Event>();
 //! let full_id = FullId::new(); // Generate new keys.
-//! let _ = Client::new(sender, Some(full_id.clone())).unwrap();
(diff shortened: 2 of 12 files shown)


maidsafe-highfive:
r? @afck

vinipsmaker:
> Might also need a getter `crust::config()`

Done.

src/node.rs
```diff
@@ -135,7 +135,8 @@ impl NodeBuilder {
                                                                 full_id,
                                                                 timer))
             }
-        })
+        },
+                          None)
```
afck: Weird indentation. Is that what rustfmt does?

vinipsmaker: Yep, that's what `rustfmt` does. CI would complain if `rustfmt` didn't run.

src/lib.rs
```diff
@@ -168,6 +168,9 @@ mod types;
 mod utils;
 mod xor_name;
 
+/// Reexports `crust::Config`
+pub type BootstrapConfig = crust::Config;
```
afck: Why not `CrustConfig`?

Cargo.toml
```diff
@@ -12,7 +12,8 @@ version = "0.25.1"
 
 [dependencies]
 clippy = {version = "=0.0.99", optional = true}
-crust = "~0.19.0"
+#crust = "~0.20.0"
+crust = { git = "https://github.com/maidsafe/crust.git", branch = "master" }
```
afck: This should work with Crust 0.20.2, I think?

src/lib.rs
ustulation: This is presented to the API of safe_core which presents it to frontend - Safe-core has not mention of crust in it and frontend is also agnostic to that term - so a more generic name was chosen to be more meaningful to them to what it is.

Cargo.toml
ustulation: I think yes - we should give it a go.

PR #1287, opened 2016-12-21
MAID-1908 perform missed splits on SectionUpdates

afck:
A joining node could miss a split event in a neighbouring section if it
happens between the `GetNodeNameResponse` and the first connection.

### DIFF
--- a/tests/mock_crust/mod.rs
+++ b/tests/mock_crust/mod.rs
@@ -127,17 +127,21 @@ fn multiple_joining_nodes() {
     let mut nodes = create_connected_nodes(&network, min_group_size);
     let config = Config::with_contacts(&[nodes[0].handle.endpoint()]);
 
-    // Try adding three nodes at once, to the same section. This makes sure one section can handle
-    // this (probably by rejecting some of the nodes).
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:

tests/mock_crust/mod.rs
```diff
@@ -127,17 +127,23 @@ fn multiple_joining_nodes() {
...
+    while nodes.len() < 40 {
+        info!("Size {}", nodes.len());
+
+        // Try adding five nodes at once, possibly to the same section. This makes sure one section
+        // can handle this, either by adding the nodes in sequence or by rejecting some.
+        nodes.push(TestNode::builder(&network).config(config.clone()).create());
+        nodes.push(TestNode::builder(&network).config(config.clone()).create());
+        nodes.push(TestNode::builder(&network).config(config.clone()).create());
+        nodes.push(TestNode::builder(&network).config(config.clone()).create());
+        nodes.push(TestNode::builder(&network).config(config.clone()).create());
```
Fraser999: `for _ in 0..5` here would be less verbose I think.

PR #1291, opened 2016-12-21
chore/many Introduces sha3 plus remove some let _ = 

dirvine:
Where possible remove some `let _ =` statements and log errors where we previously ignored results.
SHA3 used to reduce possibility of Ack collisions.


maidsafe-highfive:
r? @afck

src/ack_manager.rs
```diff
@@ -56,11 +60,15 @@ impl AckManager {
...
-    // pending ones, and remembers that we have received this ack).
+    /// Handles a received ack (removes the corresponding message from the list of
+    /// pending ones, and remembers that we have received this ack).
     pub fn receive(&mut self, ack: Ack) {
-        let _ = self.pending.remove(&ack);
-        let _ = self.received.insert(&ack);
+        let _ack = self.pending
+            .remove(&ack)
+            .ok_or(debug!("received and Ack we were not expecting : {}", ack));
+        // TODO - Should this insert an ack we were not expecting ??
```
afck: Yes. It can happen that the rest of our group already sent the message and we're receiving the acknowledgement before even considering to send it ourselves. I'm not even sure we should log it, but we can try it out and see how noisy it is.

src/ack_manager.rs
```diff
@@ -43,7 +44,10 @@ pub struct AckManager {
 
 /// An identifier for a waiting-to-be-acknowledged message (a hash of the message).
 #[derive(Clone, Copy, Eq, Hash, Ord, PartialEq, PartialOrd, RustcDecodable, RustcEncodable)]
-pub struct Ack(u64);
+pub struct Ack {
+    id: u64,
```
afck: Do we need that at all anymore? We could just use the `m_hash` everywhere.

src/messages.rs
```diff
@@ -261,7 +261,10 @@ impl SignedMessage {
                 .cloned()
                 .collect_vec();
             for irrelevant_id in irrelevant_ids {
-                let _ = self.signatures.remove(&irrelevant_id);
+                let _removed_sig = self.signatures
+                    .remove(&irrelevant_id)
+                    .ok_or(debug!("Tried to remove a signature we did not have : {:?}",
+                                  irrelevant_id));
```
afck: I think that would just always print this `debug!`: It's not in a closure, it just gets executed and the resulting `()` then gets passed into `ok_or`. (Same below.)

src/mock_crust/support.rs
```diff
@@ -278,7 +283,10 @@ impl ServiceImpl {
     }
 
     pub fn whitelist_peer(&mut self, peer_id: PeerId) {
-        let _ = self.whitelist.insert(peer_id);
+        if !self.whitelist.insert(peer_id.clone()) {
+            debug!("Duplicate insert attempt  whitelist for peer : {:?}",
```
afck: Duplicate whitespace insert attempted! :grin: 
(Between `attempt` and `whitelist`.)

PR #1292, opened 2016-12-22
fix/node: proper src/dst when sending connection_info_response

maqi:


maidsafe-highfive:
r? @afck

PR #1294, opened 2016-12-22
test/utils: move CheckError to utils and use log system

afck:


maidsafe-highfive:

afck:
As discussed, we will put the nodes into a wrapper that logs the routing tables when the thread is panicking instead.

PR #1296, opened 2016-12-22
fix/mock_crust: resend in all nodes before polling again

afck:


maidsafe-highfive:

PR #1297, opened 2016-12-22
chore/clippy 

dirvine:
Cleaner approach to run clippy using `cargo +nightly clippy` 
Also removed hyper as dev dependency
Updated travis script


maidsafe-highfive:
r? @afck

PR #1298, opened 2017-01-02
MAID-1789 test/churn: add clients into churn test

maqi:


maidsafe-highfive:
r? @afck

src/states/client.rs
```diff
@@ -111,6 +111,13 @@ impl Client {
             Action::Terminate => {
                 return Transition::Terminate.to_evented();
             }
+            Action::ClientAuthority { result_tx } => {
+                let _ = result_tx.send(Authority::Client {
+                    client_key: *self.full_id.public_id().signing_public_key(),
+                    proxy_node_name: *self.proxy_public_id.name(),
+                    peer_id: self.crust_service.id(),
+                });
```
afck: Is it really necessary to add another `Action` just for this test?
Isn't it even possible to pass the client's key pair into its constructor? Then no change in the production code should be necessary at all.

tests/mock_crust/churn.rs
```diff
@@ -93,6 +94,41 @@ impl ExpectedGets {
         self.messages.insert((data_id, msg_id, src, dst));
     }
 
+    /// Sends a request from the client, and adds the expectation that the nodes belonging to `dst`
+    /// receive the message.
+    fn client_send_and_expect(&mut self,
+                              data_id: DataIdentifier,
+                              client_auth: Authority<XorName>,
+                              dst: Authority<XorName>,
+                              clients: &Vec<TestClient>,
```
afck: That should probably be `&[TestClient]`.

tests/mock_crust/churn.rs
```diff
@@ -210,9 +259,26 @@ fn churn() {
         // expected_gets.send_and_expect(data_id, auth_s0, auth_g0, &nodes, min_group_size);
         // expected_gets.send_and_expect(data_id, auth_s0, auth_n0, &nodes, min_group_size);
 
-        poll_and_resend(&mut nodes, &mut []);
+        // Test messages from a client to a group and a section...
+        let data_id1 = DataIdentifier::Immutable(rng.gen());
+        let auth_g2 = Authority::ClientManager(rng.gen());
+        let auth_s1 = Authority::Section(section_name);
```
afck: Why create a new data ID and section authority? Let's just reuse the ones we already have.
And why does this use a `ClientManager` and not a `Client`?

tests/mock_crust/churn.rs
```diff
@@ -210,9 +259,26 @@ fn churn() {
...
+        // ... and from group to the client
+        let data = gen_immutable_data(&mut rng, 8);
+        let auth_nae = Authority::NaeManager(*data.name());
+        expected_gets.send_to_client(data.clone(),
+                                     auth_nae,
+                                     client_auth,
+                                     &mut nodes,
+                                     min_group_size);
 
+        let _ = poll_all(&mut nodes, &mut clients);
```
afck: Is `poll_and_resend` not necessary anymore? That's not really plausible, if we're dropping nodes, I think?

tests/mock_crust/churn.rs
```diff
@@ -210,9 +259,26 @@ fn churn() {
...
+
+        verify_client(&mut clients, data);
```
afck: Why not add that verification to `ExpectedGets`, too?
In fact, wouldn't it be easiest to treat `clients` just like `nodes` in `ExpectedGets`? Then we wouldn't even need any new method there. We pass in the authority anyway, so it could just know when a message is for or from a client?

maqi: the current `verify` function is already quite big.
also, the msg from nodes to client doesn't leave any trait in a `node`, which means we have to pass down `client, data` into this function as well and treated it separately.

afck: We already pass in and store the data ID in `send_and_expect`, so only the `clients` would be needed as an additional argument.
And if `verify` gets too big, maybe something can be extracted from it.

src/states/client.rs
maqi: the problem is not passing down the key pair, but the Authority::Client requiring PeerId and Name of the proxy that this client connected to . :(

tests/mock_crust/churn.rs
maqi: the msg sent to client is `GetSuccess` which contains `Data` not `DataIdentifier`.
I agree we can deduce identifier from data then record it, but such mapping looks little bit tweaking.
also, it means in the `verify` function we need to match `Event::Response` additional as well.

We may send `request` to client to avoid such hassle, but not sure if it will be good to do so (as it doesn't reflects the normal role of a client).

afck: Yes, let's just also send a `Get` request then. It shouldn't matter at all what kind of message we send: The routing logic is completely unaware of that, apart from caching.

src/states/client.rs
afck: Looks like it should be feasible to get the peer ID out of mock Crust without touching production code?
https://github.com/maidsafe/routing/blob/master/src/mock_crust/support.rs#L213

tests/mock_crust/churn.rs
```diff
@@ -93,8 +93,27 @@ impl ExpectedGets {
         self.messages.insert((data_id, msg_id, src, dst));
     }
 
+    /// Sends a request from the client, and adds the expectation that the nodes belonging to `dst`
+    /// receive the message.
+    fn client_send_and_expect(&mut self,
```
Fraser999: So much of this is identical to `send_and_expect()` that I think the common part should be abstracted to a separate function.

tests/mock_crust/churn.rs
```diff
@@ -93,8 +93,27 @@ impl ExpectedGets {
         self.messages.insert((data_id, msg_id, src, dst));
     }
 
+    /// Sends a request from the client, and adds the expectation that the nodes belonging to `dst`
```
Fraser999: I think the comment should make it clear that the function is sending a `Get` request from only the first client in the slice, or the function should be changed to only take a single `TestClient` rather than a slice of them.

tests/mock_crust/churn.rs
```diff
@@ -131,6 +150,15 @@ impl ExpectedGets {
                 }
             }
         }
+        while let Ok(event) = clients[0].inner.try_next_ev() {
```
Fraser999: There shouldn't really be any harm in testing all the clients in the slice here I think.  Not a big deal, but it would make the test more thorough if we assert that none of the other clients received any messages by mistake.

PR #1300, opened 2017-01-03
MAID-1924 don't consider send failure a lost peer

afck:


maidsafe-highfive:

PR #1302, opened 2017-01-03
MAID-1923: refactor/examples: allow examples to compile with 'use-mock-crust' enabled

Fraser999:
The examples will now build with 'use-mock-crust' enabled, but when run will print a warning and
return an error code.  Also, clippy with 'use-mock-crust' is now enabled in the Travis script.

### DIFF
--- a/examples/ci_test.rs
+++ b/examples/ci_test.rs
@@ -31,8 +31,7 @@
 #![allow(box_pointers, fat_ptr_transmutes, missing_copy_implementations,
          missing_debug_implementations, variant_size_differences)]
 
-
-#![cfg(not(feature = "use-mock-crust"))]
+#![cfg_attr(feature = "use-mock-crust", allow(unused_extern_crates))]
 
 #[macro_use]
 extern crate log;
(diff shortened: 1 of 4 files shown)


maidsafe-highfive:
r? @afck

examples/ci_test.rs
```diff
@@ -50,226 +49,235 @@ extern crate unwrap;
...
-        match self.0.kill() {
-            Ok(()) => println!("Killed Node with Process ID #{}", self.0.id()),
-            Err(err) => {
-                println!("Error killing Node with Process ID #{} - {:?}",
-                         self.0.id(),
-                         err)
+#[cfg(feature = "use-mock-crust")]
+fn main() {
+    println!("This example should be built without `--features=use-mock-crust`.");
+    ::std::process::exit(78);
```
maqi: shall we have a comment explaining why 78 is chosen?

Fraser999: I didn't think it was important, but I can do :)

Fraser999: Done now.

PR #1304, opened 2017-01-03
fix/peer_manager: re-insert removed peer when their_info is not none for PeerState::ConnectionInfoPreparing

maqi:


maidsafe-highfive:
r? @afck

PR #1305, opened 2017-01-04
Fix blocking behaviour of next_ev and propagate non-blocking errors.

michaelsproul:
This ensures that calling `next_ev` will continually block and process all
events, necessary for, e.g. `run` in safe_vault.

Additionally, using the mpsc error types allows users of `try_next_ev` to determine
if an error was returned because there were no events available, or because of a permanent failure.

This change (mainly the first part) is necessary for my refactor of `safe_vault`

### DIFF
--- a/src/event_stream.rs
+++ b/src/event_stream.rs
@@ -1,20 +1,18 @@
-use std::sync::mpsc::TryRecvError;
+use std::sync::mpsc::{RecvError, TryRecvError};
 
 /// Trait to fake a channel.
 pub trait EventStream {
     /// Item produced by this stream.
     type Item;
-    /// Error type produced if something goes wrong or no items are available.
-    type Error;
 
     /// Read the next available event from the stream, blocking until one becomes available.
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @afck

src/event_stream.rs
```diff
@@ -45,26 +42,34 @@ impl<S> EventStream for S
...
         }
-        if let Ok(new_events) = self.produce_events() {
-            self.buffer_items(new_events);
+        match self.produce_events() {
+            Ok(new_events) => {
+                self.buffer_items(new_events);
+                // Recursing here either returns one of the buffered events we just added,
+                // or in the case where `new_events` was empty, causes us to block again waiting
+                // on `produce_events`.
+                self.next_ev()
```
afck: I don't think Rust guarantees to do tail call elimination; couldn't this cause a stack overflow, depending on the implementation of `produce_events`? Might be safer to use a loop.

PR #1307, opened 2017-01-04
chore/stats: log SectionListSignature messages and explicitly match o…

dhardy:
…ther DirectMessages


maidsafe-highfive:
r? @afck

PR #1308, opened 2017-01-05
fix/event: use an explicit loop in `next_ev`

michaelsproul:
This avoids a potential stack overflow discussed in https://github.com/maidsafe/routing/pull/1305. Better safe than sorry!


maidsafe-highfive:
r? @afck

PR #1309, opened 2017-01-05
MAID-1926 chore/node: reduce the amount of SectionUpdate message

maqi:


maidsafe-highfive:
r? @afck

PR #1314, opened 2017-01-12
MAID-1840 Resource proof hashing WIP

maqi:


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -62,15 +63,7 @@ const TICK_TIMEOUT_SECS: u64 = 60;
...
-const RESOURCE_PROOF_DIFFICULTY: u32 = 10;
-
-// in Bytes
-fn resource_proof_target_size(min_size: usize, group_len: usize) -> u32 {
-    let evaluators = cmp::max(min_size, group_len);
-    // Default value: 2 MBytes sharing among the evaluators
-    // TODO: the algorithm need to be measured and refactored against the real network
-    2 * 1024 * 1024 / evaluators as u32
-}
+const RESOURCE_PROOF_DIFFICULTY: u8 = 4;
```
Fraser999: The new node will only need to try on average 16 hash attempts to satisfy this - shouldn't it be somewhat harder?

src/messages.rs
```diff
@@ -135,6 +135,27 @@ pub enum DirectMessage {
...
     /// Sent to a tunnel node to indicate the tunnel is not needed any more.
     TunnelDisconnect(PeerId),
+    /// Request a proof to be provided by the joining node
+    ///
+    /// This is sent from member of Group Y to the joining node
+    ResourceProof {
+        /// seed of proof
+        seed: Vec<u8>,
+        /// size of the proof
+        target_size: usize,
```
Fraser999: I'm never a fan of using variable-sized integers in messages which are passed between machines of differing architectures, but bincode should handle this I suppose...

src/peer_manager.rs
```diff
@@ -325,42 +370,185 @@ impl PeerManager {
...
+                            -> Option<bool> {
+        self.candidate
+            .as_ref()
+            .map_or(None,
+                    |&Candidate { ref name, ref insertion_time, ref target_size, ref seed, .. }| {
+                if *name == node_name &&
+                   insertion_time.elapsed() <
+                   Duration::from_secs(RESOURCE_PROOF_EVALUATE_TIMEOUT_SECS) {
+                    let rp_object = ResourceProof::new(*target_size, difficulty);
+                    return Some(rp_object.validate_all(&seed, &proof, leading_zero_bytes));
```
Fraser999: Don't need the `return` here I think.

src/states/node.rs
maqi: i tried with 8 and it will fail many of our current tests (taking too much time to resolve the challenge)
also, I think this factor may needs to be refactored when test against the testnet, to figure out what's the most suitable value network can handle

src/peer_manager.rs
```diff
@@ -325,42 +370,185 @@ impl PeerManager {
...
+        // TODO: more specific return error
+        Err(RoutingError::InvalidStateForOperation)
+    }
+
+    /// Update peer's state to `Candidate` if it is a node candidate
+    ///
+    /// Returns:
+    ///
+    /// * Ok(None)                                  If the peer is not a node candidate for
+    /// *                                           resource proof or has been approved
```
Fraser999: This line isn't a separate bullet point - it shouldn't have the `*`.

src/peer_manager.rs
```diff
@@ -325,42 +370,185 @@ impl PeerManager {
...
+    /// * Ok(Some(is_tunnel, seed, target_size))    If the peer is a node candidate
```
Fraser999: Should be `Ok(Some(is_tunnel, target_size, seed))`.

src/states/node.rs
```diff
@@ -61,8 +62,8 @@ use xor_name::XorName;
 const TICK_TIMEOUT_SECS: u64 = 60;
 /// Time (in seconds) after which a `GetNodeName` request is resent.
 const GET_NODE_NAME_TIMEOUT_SECS: u64 = 60;
-/// Time (in seconds) the members of a section wait for a joining node they sent a network name to.
-const SENT_NETWORK_NAME_TIMEOUT_SECS: u64 = 30;
+/// The number of required leading zero bits for the resource proof
+const RESOURCE_PROOF_DIFFICULTY: u8 = 4;
```
Fraser999: The new node will only need to try on average 16 hash attempts to satisfy this - shouldn't it be somewhat harder?

maqi: It is not just 16 hash attempts required. the node needs to keeps pre-pending till the hash result having 4 leading zero bits.
I tried with 8 and it will fail many of our current tests (taking too much time to resolve the challenge)
also, I think this factor may needs to be refactored when test against the testnet, to figure out what's the most suitable value network can handle

maqi: and seems even 4 is somehow too hard for the CI machines  :(

src/peer_manager.rs
```diff
@@ -389,8 +401,11 @@ impl PeerManager {
 
         let names = self.routing_table.expect_join_our_section(expected_name)?;
 
-        let mut rng = SeededRng::new();
-        let seed = rng.gen_iter().take(10).collect();
+#[cfg(not(feature = "use-mock-crust"))]
+        let seed = SeededRng::new().gen_iter().take(10).collect();
```
Fraser999: I didn't spot this during the last review, but we shouldn't be using `SeededRng` outside of tests I believe.  Using it like this will result in the same `seed` being generated every single time `expect_join_our_section()` is called.  It will only be different once the process restarts.

I think for non-tests, we should just stick to the `rand` crate.

src/peer_manager.rs
```diff
@@ -325,42 +378,188 @@ impl PeerManager {
...
+            if name == expected_name ||
+               insertion_time.elapsed() < Duration::from_secs(RESOURCE_PROOF_APPROVE_TIMEOUT_SECS) {
+                return Err(RoutingError::AlreadyHandlingJoinRequest);
+            }
+        }
+
+        let names = self.routing_table.expect_join_our_section(expected_name)?;
+
+#[cfg(not(feature = "use-mock-crust"))]
+        let seed = vec![rand::random::<u8>(); 10];
```
Fraser999: This just repeats the same random `u8` 10 times though.  The way that's used by the `resource_proof` crate, it'd be equivalent to `vec![rand::random::<u8>(); 1]`.

Shouldn't this be more like `let seed = thread_rng().gen_iter().take(10).collect();`?

src/messages.rs
```diff
@@ -581,6 +611,20 @@ pub enum MessageContent {
...
     },
+    /// Send among Group Y to vote for Accept or Reject a joining node
+    CandidateApproval(bool),
+    /// Approves the joining node as a routing node.
+    ///
+    /// Sent from Group Y to the joining node.
+    NodeApproval {
+        /// The routing table shared by the nodes in our group, including the `PublicId`s of our
+        /// contacts.
+        sections: Vec<(Prefix<XorName>, Vec<PublicId>)>,
```
afck: Let's make this a `BTreeMap<Prefix<XorName>, BTreeSet<PublicId>>`, so we don't have to sort manually.

afck: Actually, that should be changed in #1261, not here, I think.

afck:
Included in #1319

PR #1316, opened 2017-01-13
MAID-1924 remove Evented where no events can occur

afck:


afck:
Closing and re-opening to trigger Travis.

PR #1317, opened 2017-01-16
Run tests in release mode

michaelsproul:
Using multiple codegen units allows us to compile quickly in release mode, which makes the whole testing process loads faster!

Let's see what Travis makes of it...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ script:
         cargo rustc --verbose --example ci_test -- -Zno-trans &&
         cargo rustc --verbose --example key_value_store -- -Zno-trans &&
         cargo rustc --verbose -- -Zno-trans &&
-        cargo test  --verbose --features use-mock-crust
+        env RUSTFLAGS="-C codegen-units=8" cargo test --release --verbose --features use-mock-crust
       );
     elif [ "${TRAVIS_OS_NAME}" = linux ]; then
       (


maidsafe-highfive:
r? @afck

michaelsproul:
This does seem to be substantially faster on Travis!

## Before

Build: https://travis-ci.org/maidsafe/routing/builds/192296061
Linux/stable: 15 min 13 sec
OSX/stable: 31 min 11 sec

## After

Build: https://travis-ci.org/maidsafe/routing/builds/192296061
Linux/stable: 8 min 25 sec
OSX/stable: 22 min 9 sec

I suspect it'll ripen with age, as the caching of `--release` compiled libs kicks in

.travis.yml
```diff
@@ -37,7 +37,7 @@ script:
         cargo rustc --verbose --example ci_test -- -Zno-trans &&
         cargo rustc --verbose --example key_value_store -- -Zno-trans &&
         cargo rustc --verbose -- -Zno-trans &&
-        cargo test  --verbose --features use-mock-crust
+        env RUSTFLAGS="-C codegen-units=8" cargo test --release --verbose --features use-mock-crust --test mock_crust_tests
```
Fraser999: We don't want to limit this to only run the mock crust tests do we?  I think ` --test mock_crust_tests` should be removed.

michaelsproul: Oh yes, you're right! I forgot about the library tests inside `src`. I mainly added that line so that the examples wouldn't be compiled in release mode, as that causes `src/lib.rs` to be compiled twice. I'll add another line that just runs the library tests?

michaelsproul: Ah, it looks like running the lib tests requires compiling without `--crate_type lib`, so `src/lib.rs` does get compiled twice...

michaelsproul: Ok, I've made it so that it just does the same thing as before, but with `--release` and `-C codegen-units=8`. It looks like double compiling is unavoidable (due to the need to compile the crate as a test harness **and** a lib).

PR #1318, opened 2017-01-17
chore/fmt: run rustfmt 0.7.0

michaelsproul:


maidsafe-highfive:
r? @afck

src/node.rs
```diff
@@ -101,6 +101,9 @@ impl NodeBuilder {
         let full_id = FullId::new();
 
         StateMachine::new(move |crust_service, timer| {
+            let not_allowed_on_lan = self.deny_other_local_nodes &&
```
michaelsproul: This change was necessary to avoid `rustfmt` getting all confused about line 118. It was inserting a space at the end of the line after `} else if` and failing 👿

PR #1320, opened 2017-01-23
MAID-1924 drop peers on send failure

afck:


maidsafe-highfive:

src/peer_manager.rs
```diff
@@ -962,6 +965,10 @@ impl PeerManager {
         true
     }
 
+    pub fn mark_as_failed(&mut self, peer_id: &PeerId) -> bool {
```
Fraser999: I'm not sure there's any point in this returning a `bool` since in the only place it's called we ignore the returned value.

src/states/node.rs
```diff
@@ -2661,6 +2661,17 @@ impl Base for Node {
...
     }
+
+    /// Sends the given `bytes` to the peer with the given Crust `PeerId`. If that results in an
+    /// error, it marks the peer as `Failed`. It will be dropped on the next tick.
+    fn send_or_drop(&mut self, peer_id: &PeerId, bytes: Vec<u8>, priority: u8) {
+        self.stats().count_bytes(bytes.len());
+
+        if let Err(err) = self.crust_service().send(*peer_id, bytes.clone(), priority) {
+            info!("{:?} Connection to {:?} failed: {:?}", self, peer_id, err);
+            let _ = self.peer_mgr.mark_as_failed(peer_id); // Alredy logged in peer manager.
```
Fraser999: I'm not really following the comment here (and there's a typo).

afck:
Done.
`remove_expired_connections` is called periodically in `handle_timeout`.

afck:
Closing this, to address other issues with the peer manager first.

PR #1321, opened 2017-01-23
fix/peer_manager: don't remove without calling disconnect

afck:


maidsafe-highfive:

PR #1322, opened 2017-01-23
fix/node: don't loop calling merge_if_necessary forever in tests

afck:


maidsafe-highfive:

PR #1324, opened 2017-01-24
Check all nodes in merge tests, and don't poll when panicking.

afck:


maidsafe-highfive:

PR #1326, opened 2017-01-24
feat/node: don't require proof from hard-coded contacts

afck:


maidsafe-highfive:

afck:
Note: This can only be merged after Crust 0.20.0 has been published.

src/peer_manager.rs
```diff
@@ -560,7 +542,17 @@ impl PeerManager {
...
+                    let (target_size, seed) = if difficulty == 0 {
+                        (10, vec![5u8; 4])
+                    } else {
+                        let evaluators = cmp::max(self.routing_table.min_section_size(),
+                                                  self.routing_table.our_section().len());
+                        // TODO: the algorithm need to be measured and refactored against the real network
+                        (2 * 1024 * 1024 / evaluators,
+                         rand::thread_rng().gen_iter().take(10).collect())
+                    };
+                    candidate.resource_proof = Some((target_size, difficulty, seed.clone()));
```
maqi: if section 0, having 20 nodes, being asked to merge with section 1 having 7 nodes and min_section_size is 8.
if the joining node receives get_node_name response from section 0, then starts connecting, but section 0 now merged with section 1, calculation here will use a bigger sized own_section, which results about 33% smaller target_size.
This is just a tiny edge case, and the network shall be ok to stay with it.

afck: But that problem is there with or without this PR, isn't it?
And in that scenario, the joining node would then have to reply successfully to _all_ the challenges, instead of just a quorum, so it doesn't actually get easier to get into the network, I think.

maqi: >But that problem is there with or without this PR, isn't it?
in this scenario, the current code calculates target_size before send_out get_node_name_response. which reflects the num of connections the joining node needs to establish.

>have to reply successfully to all the challenges, instead of just a quorum
the current code also requires that, but with a larger target_size, so the new approach is easier compared with the current.

Anyhow, this is really an edge case and the network shall be ok to stay with it and using the new approach.

src/peer_manager.rs
```diff
@@ -440,38 +420,38 @@ impl PeerManager {
...
+            .state = CandidateState::AcceptedAsCandidate;
         let our_section = self.routing_table.our_section();
         Ok(self.get_pub_ids(our_section))
     }
 
     /// Verifies proof of resource.  If the response is too close to the overall resource proof
     /// timeout, or if it's not the current candidate, or if it fails validation, returns `Err`.
-    /// Otherwise returns a tuple containing the candidate's `PublicId`, its client `Authority` and
-    /// the `PublicId`s of all routing table entries.
+    /// Otherwise returns the target size and difficulty.
```
Fraser999: I think "If the response is too close to the overall resource proof timeout" is no longer true (was missed in my last commit).

src/states/node.rs
```diff
@@ -1015,27 +1023,28 @@ impl Node {
...
-                  name);
+        match self.peer_mgr.verify_candidate(&name, proof, leading_zero_bytes) {
+            Err(error) => {
+                debug!("{:?} Failed to verify candidate {}: {:?}",
+                       self,
+                       name,
+                       error);
+                self.candidate_timer_token = None;
+            }
+            Ok((target_size, difficulty)) if difficulty == 0 && target_size < 1000 => {
```
Fraser999: I'm not seeing any harm in using ` && target_size < 1000`, but is it required?

afck: We might want to set `difficulty` to 0 and have a large `target_size`, or have a small `target_size` but a high difficulty.
I think the right criterion for sending the response immediately is whether the challenge is completely trivial (so it doesn't hinder message accumulation) or not. That's why I'd prefer to have both in there.

Fraser999: Fair point 👍

PR #1327, opened 2017-01-24
MAID-1924 drop peers on send failure

afck:


maidsafe-highfive:

afck:
> No I didn't "suggest changes". Why doesn't GH have a "reject change" option?

It has a "Close and comment" button. :smile: 
I'm going to click that now; I agree: I guess the `NodeLost` event will only be sent to the upper layers _after_ the section-level agreement (because it usually requires a coordinated action by the section).
We'll still need to handle the send failure, though, and record that node as "probably offline", so that we vote for removing it.

PR #1328, opened 2017-01-25
fix/messages: remove duplicate Prefix from SectionListSignature

michaelsproul:
I noticed while reviewing the message flow that a `SectionListSignature` currently contains 2 copies of the same prefix, which could be cause for confusion


maidsafe-highfive:
r? @afck

PR #1329, opened 2017-01-25
chore/states/node: make match rules for *Merge messages more specific

dhardy:


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -2708,66 +2662,38 @@ impl Node {
...
-        let request_msg = RoutingMessage {
-            src: Authority::Section(src_name),
-            dst: Authority::PrefixSection(merge_details.merge_prefix),
-            content: request_content.clone(),
-        };
-        if let Err(err) = self.send_routing_message(request_msg) {
-            debug!("{:?} Failed to send OwnSectionMerge: {:?}.", self, err);
+        if let Some((sender_prefix, merge_prefix, sections)) = self.peer_mgr.should_merge() {
+            let content = MessageContent::OwnSectionMerge(sections);
+            let src = Authority::PrefixSection(sender_prefix);
```
michaelsproul: Why is this now a `PrefixSection(sender_prefix)` rather than just a `Section`? We want to send the message just from our section right? Is it because that prefix only includes our section unless our section splits, which can't happen?

even though the PrefixSection of our section should only include our section, unless our section splits, right?).

src/states/node.rs
```diff
@@ -2708,66 +2662,38 @@ impl Node {
...
-            };
-            let request_msg = RoutingMessage {
-                src: src,
-                dst: Authority::PrefixSection(*target),
-                content: request_content,
-            };
-
-            if let Err(err) = self.send_routing_message(request_msg) {
+            let content = MessageContent::OtherSectionMerge(section.clone());
+            let src = Authority::PrefixSection(merge_details.prefix);
```
michaelsproul: Similar question to above ^

src/states/node.rs
dhardy: It's because I wanted to re-use the authorities to convey a prefix, not just an address (`Section` doesn't include the prefix length...). I talked with Andreas and he seemed to think it a good idea not including mostly-redundant prefixes in the messages. The changes to message sending behaviour don't seem very significant.

michaelsproul: Ah yes, I saw you pattern match on it elsewhere but didn't make the connection. OK, seems sensible!

src/peer_manager.rs
```diff
@@ -683,11 +683,20 @@ impl PeerManager {
     }
 
     /// Wraps `RoutingTable::should_merge` with an extra check.
-    pub fn should_merge(&self) -> Option<OwnMergeDetails<XorName>> {
+    pub fn should_merge(&self) -> Option<(Prefix<XorName>, Prefix<XorName>, SectionMap)> {
```
afck: The documentation should say which returned prefix is which.

src/messages.rs
```diff
@@ -610,17 +610,14 @@ pub enum MessageContent {
     SectionSplit(Prefix<XorName>, XorName),
     /// Sent amongst members of a newly-merged section to allow synchronisation of their routing
     /// tables before notifying other connected peers of the merge.
-    OwnSectionMerge {
-        sender_prefix: Prefix<XorName>,
-        merge_prefix: Prefix<XorName>,
-        sections: SectionMap,
-    },
+    ///
+    /// Source authority is the `sender_prefix` and destination authority is the `merge_prefix`.
```
afck: I'd prefer if this explained that it's send from one of the merging sections to the `PrefixSection` including both. Let's not refer to fields that don't exist anymore (`merge_prefix`).

src/messages.rs
```diff
@@ -610,17 +610,14 @@ pub enum MessageContent {
...
+    OwnSectionMerge(SectionMap),
     /// Sent by members of a newly-merged section to peers outwith the merged section to notify them
     /// of the merge.
-    OtherSectionMerge {
-        prefix: Prefix<XorName>,
-        section: BTreeSet<PublicId>,
-    },
+    ///
+    /// Source authority is the `merge_prefix`.
```
afck: Same as above.

src/states/node.rs
```diff
@@ -2061,12 +2036,7 @@ impl Node {
                 prefix: prefix,
                 members: members,
             };
-            let request_msg = RoutingMessage {
-                src: dst,
-                dst: src,
-                content: content,
-            };
-            if let Err(err) = self.send_routing_message(request_msg) {
+            if let Err(err) = self.send_routing_message(src, dst, content) {
```
afck: Shouldn't `src` and `dst` be switched here?

dhardy:
@afck nice catch with the swapped src/dst! Added comments as you asked.

src/states/node.rs
```diff
@@ -2067,12 +2039,8 @@ impl Node {
                 prefix: prefix,
                 members: members,
             };
-            let response_msg = RoutingMessage {
-                src: dst,
-                dst: src,
-                content: content,
-            };
-            if let Err(err) = self.send_routing_message(response_msg) {
+            // We're sending a reply, so swap src and dst are swapped:
```
Fraser999: The comment needs fixed.

PR #1330, opened 2017-01-25
Improve several log statements, split resource proof messages.

afck:


maidsafe-highfive:

src/states/node.rs
```diff
@@ -1940,7 +1940,7 @@ impl Node {
         } else if old_prefix.bit_count() > new_prefix.bit_count() {
             result.add_event(Event::SectionMerge(new_prefix));
         }
-        info!("{:?} Update on RoutingTableResponse completed. Prefixes: {:?}",
+        info!("{:?} Update on RoutingTableResponse completed. Section prefixes: {:?}",
```
maqi: is this consistent with the script (collecting routing_table) ?

src/peer_manager.rs
```diff
@@ -490,12 +490,12 @@ impl PeerManager {
...
                     }
                 } else {
-                    trace!("{:?} doesn't have peer_id for {:?}",
+                    trace!("Node({:?}) No peer ID with name {:?}",
                            self.routing_table.our_name(),
                            candidate_name);
                 }
             } else {
-                trace!("{:?} doesn't have peer for {:?}",
+                trace!("Node({:?}) No peer with name {:?}",
```
maqi: we can now print `self` here, right?

afck: Unfortunately not, because `self` is mutably borrowed in line 484.

src/states/node.rs
afck: Ah, right, it isn't. Good catch!

src/states/node.rs
```diff
@@ -986,6 +996,22 @@ impl Node {
...
+            let direct_message = DirectMessage::ResourceProofResponse {
+                part_index: part_index,
+                part_count: part_count,
+                proof: part,
+                leading_zero_bytes: leading_zero_bytes,
+            };
+            self.send_direct_message(peer_id, direct_message)?
+        }
         let elapsed = start.elapsed();
         trace!("{:?} created proof data in {}.{:06} seconds.  Min section size: {}, Target size: \
```
dhardy: This now gets printed after sending — is that okay?

Fraser999: That's a good point.  This is changing in an upcoming PR to only send the first part here, but I think the elapsed time would be better calculated immediately after creating the proof.

afck: Yes, it turns out sending is pretty much instantaneous (the calls, I mean: sending is asynchronous).
But this part will change again anyway, with Fraser's next PR.

src/states/node.rs
```diff
@@ -1247,20 +1294,18 @@ impl Node {
...
+                           name,
                            error);
                 }
             }
-            Ok(None) => {
-                debug!("{:?} adding {:?} into routing table.",
-                       self,
-                       public_id.name());
+            Ok(false) => {
+                debug!("{:?} adding {:?} into routing table.", self, name);
```
Fraser999: I'm not sure this log statement isn't just spam - we print inside `add_to_routing_table()` too.

afck: Let's remove that in the next PR, as discussed.

PR #1332, opened 2017-01-27
Avoid sending all ResourceProofResponse parts in one go

Fraser999:


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -997,34 +1013,57 @@ impl Node {
...
+                DirectMessage::ResourceProofResponse {
+                    part_index: part_index,
+                    part_count: part_count,
+                    proof: part,
+                    leading_zero_bytes: leading_zero_bytes,
+                }
+            })
+            .collect_vec();
+        // There will always be at least one part here - safe to unwrap.
+        let first_message = unwrap!(messages.pop());
```
afck: Probably not if target size is 0. We need to check that and potentially handle that separately.

src/states/node.rs
```diff
@@ -997,34 +1013,57 @@ impl Node {
...
         Ok(())
     }
 
+    fn handle_resource_proof_response_receipt(&mut self, peer_id: PeerId) {
+        let popped_message = if let Some(messages) = self.resource_proof_response_parts
+            .get_mut(&peer_id) {
+            messages.pop()
+        } else {
+            None
+        };
```
afck: Probably equivalent:
```rust
let popped_message = self.resource_proof_response_parts.get_mut(&peer_id).and_then(Vec::pop);
```

src/states/node.rs
```diff
@@ -2135,8 +2175,10 @@ impl Node {
     fn handle_timeout(&mut self, token: u64) -> Evented<bool> {
         let mut events = Evented::empty();
         if self.get_approval_timer_token == Some(token) {
-            info!("{:?} Failed to get approval from the network.", self);
-            events.add_event(Event::RestartRequired);
+            info!("{:?} Failed to get approval from the network. {}",
+                  self,
+                  self.resource_proof_response_progress());
```
afck: It would be helpful to print the number of message parts that were sent here. (Or even better: The number of receipts received.)

src/states/node.rs
```diff
@@ -2135,8 +2175,10 @@ impl Node {
...
+            events.add_event(Event::Terminate);
```
afck: Depending on how close it was we might want to restart.

src/states/node.rs
Fraser999: That's what `self.resource_proof_response_progress()` does.

src/states/node.rs
Fraser999: Sorry - I forgot we discussed that!  I'll add a new commit.

Fraser999:
The contents of this are already in master via #1333.

PR #1334, opened 2017-01-31
MAID-1940: Avoid sending section updates when Routing invariant not held

Fraser999:


maidsafe-highfive:
r? @afck

PR #1335, opened 2017-01-31
fix/test: fix bug in mock crust test where rejected joining nodes aren't removed

Fraser999:


maidsafe-highfive:
r? @afck

PR #1336, opened 2017-02-01
refactor/node: always terminate on failure to get NodeApproval

Fraser999:
Also adds further log info relating to joining and resource proof process.


maidsafe-highfive:
r? @afck

PR #1344, opened 2017-02-07
feat/node: balance tree before accepting nodes

Fraser999:


maidsafe-highfive:
r? @afck

src/routing_table/mod.rs
```diff
@@ -835,6 +835,28 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
         self.check_invariant(false, false).is_ok()
     }
 
+    /// Returns the prefix of the section in which `name` belongs, or `None` if there is no such
+    /// section in the routing table.
+    pub fn find_section_prefix(&self, name: &T) -> Option<Prefix<T>> {
```
afck: It would be better to return a reference: `Option<&Prefix<T>>`.

src/routing_table/mod.rs
```diff
@@ -835,6 +835,28 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
...
+    /// Returns `name` modified so that it belongs to one of the known prefixes with minimal bit
+    /// length, favouring our own prefix if it is one of the shortest.
+    pub fn assign_to_min_len_prefix(&self, name: &T) -> T {
+        let mut target_prefix = self.sections
+            .keys()
+            .min_by_key(|prefix| prefix.bit_count())
+            .unwrap_or(&self.our_prefix);
+        if self.our_prefix.bit_count() <= target_prefix.bit_count() {
+            target_prefix = &self.our_prefix;
+        }
```
afck: You could just `chain(iter::once(&self.our_prefix))` above instead. `min_by_key` selects the last one among the entries with equal keys.

Viv-Rajkumar:
OS X CI run locally and passed.

PR #1345, opened 2017-02-07
chore/node: log the hop_msg to unknown connection

maqi:


maidsafe-highfive:
r? @afck

Fraser999:
Incorporated into #1344.

PR #1346, opened 2017-02-07
fix/peer_manager: re-insert joining nodes/proxies after removing them…

afck:
… from the routing table


maidsafe-highfive:

PR #1348, opened 2017-02-08
fix/node: when disconnecting tunnel peer, remove from peer_map as well

maqi:


maidsafe-highfive:
r? @afck

tests/mock_crust/utils.rs
```diff
@@ -322,7 +322,7 @@ pub fn create_connected_nodes_with_cache(network: &Network, size: usize, use_cac
     let n = cmp::min(nodes.len(), network.min_section_size()) - 1;
 
     for node in &mut nodes {
-        expect_next_event!(node, Event::Connected);
+        expect_any_event!(node, Event::Connected if true);
```
Fraser999: This function is used by many of the other mock crust tests, so it would be good to not have to downgrade the strictness of this call.  Once your subsequent commit comes in (to avoid dropping tunnelled candidates) can this change be reverted?

src/peer_manager.rs
```diff
@@ -514,9 +514,7 @@ impl PeerManager {
             candidate.state = CandidateState::Approved;
             if let Some(peer) = self.peer_map.get_by_name(&candidate_name) {
                 if let Some(peer_id) = peer.peer_id() {
-                    if peer.state().is_directly_connected() {
-                        return Ok(*peer_id);
-                    }
+                    return Ok(*peer_id);
```
maqi: done

PR #1349, opened 2017-02-08
chore/misc: general cleanup

Viv-Rajkumar:
- Change Candidate Status log format and interval to 60 seconds.
- Accurately indicate dropped routing node log.
- Peer manager debug trait to include the section prefix
- Pruned excessive log messages from RT recovery


maidsafe-highfive:
r? @afck

PR #1351, opened 2017-02-09
Replace Evented with OutBox

dhardy:
First commit adds new module and a couple of example usages.
Second commit replaces all uses of Evented within states/node.rs.
Third commit replaces all remaining uses of Evented.

Original plan was to extend this for sending other messages too; not sure whether to continue that now. But in any case it's quite a bit easier to use than Evented and less prone to bugs.


maidsafe-highfive:
r? @afck

src/node.rs
```diff
@@ -98,11 +99,11 @@ impl NodeBuilder {
     // TODO - remove this `rustfmt_skip` once rustfmt stops adding trailing space at `else if`.
     #[cfg_attr(rustfmt, rustfmt_skip)]
     fn make_state_machine(self,
-                          min_section_size: usize)
-                          -> Evented<(RoutingActionSender, StateMachine)> {
+                          min_section_size: usize, outtray: &mut EventTray)
```
afck: `rustfmt` might want to have a word with you.

src/outtray.rs
```diff
@@ -0,0 +1,76 @@
...
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+//! A system for dispatching `Event`s and possibly in the future other message types.
+//!
+//! The design allows the separate traits to handle dispatching of each type of message to an
+//! object handling the appropriate types of message.
+
+// The original design was like this:
```
afck: This comment can probably be removed.

src/outtray.rs
```diff
@@ -0,0 +1,76 @@
...
+
+/// An event dispatcher. Collects things to deliver and "sends".
+///
+/// For now, this is a struct (to enable static dispatch). The design allows switching this to a
+/// trait, updating only a few functions which use it at the top level.
+///
+/// The API doesn't specify whether objects get sent immediately synchronously or asynchronously,
+/// or collected and sent later.
+#[derive(Default)]
+pub struct EventTray {
```
afck: Should this also implement a `Drop` trait that logs an error if there are any events left?

src/states/node.rs
```diff
@@ -861,41 +856,41 @@ impl Node {
...
+            (Ack(ack, _), _, _) => self.handle_ack_response(ack),
             (UserMessagePart { hash, part_count, part_index, payload, .. }, src, dst) => {
                 if let Some(msg) = self.user_msg_cache.add(hash, part_count, part_index, payload) {
                     self.stats().count_user_message(&msg);
-                    Evented::single(msg.into_event(src, dst), Ok(()))
+                    outtray.send_event(msg.into_event(src, dst));
+                    Ok(())
                 } else {
-                    Ok(()).to_evented()
+                    Ok(())
```
afck: That `else` can be removed, since it's always `Ok(())` now.

src/outtray.rs
dhardy: I thought it may be useful in the future? In any case, this module isn't huge.

src/outtray.rs
dhardy: Done

src/node.rs
dhardy: Doesn't complain to me. I'm also using 0.7.0 so I'm not sure what gives...

dhardy:
@afck I added one commit since your review (just the last one), after noticing how tighter integration is possible.

I was going to implement `EventTray` directly in `routing::Node` hence the switch to a trait, but realised this wasn't the best path. So at this point I could switch back to static dispatch and rename the whole thing to `EventBuf` if preferred... I'm not sure on this.

src/node.rs
michaelsproul: It's skipping this function, see `[cfg_attr(rustfmt, rustfmt_skip)]` above

src/states/node.rs
```diff
@@ -1845,11 +1840,10 @@ impl Node {
     fn handle_get_node_name_response(&mut self,
                                      relocated_id: PublicId,
                                      section: BTreeSet<PublicId>,
-                                     dst: Authority<XorName>)
-                                     -> Evented<()> {
+                                     dst: Authority<XorName>,
+                                     outtray: &mut EventTray) {
         if !self.peer_mgr.routing_table().is_empty() {
             warn!("{:?} Received duplicate GetNodeName response.", self);
-            return Evented::empty();
```
michaelsproul: Should be `return;`!

michaelsproul:
@dhardy: Looks good to me, pending the Clippy fix that Travis is complaining about and the `return` fix I commented on. Overall I'm really pleased with how well this works compared to Evented.

**Bikeshed alert:** I find the `outtray` name (with the double "t"s) somewhat jarring. I think `out_tray` would be better - or even better - `outbox`, which I think carries the right connotations. We could even rename the `EventTray` trait (or the `EventBuf` type if we ditch the trait) to `Outbox`.

I'm also in favour of (eventually) adapting all messaging to use this pattern, as it'll help us split up `src/states/node.rs` (regardless of any other modules we want to add in).

dhardy:
Thanks @michaelsproul . I fixed the missing return and made Clippy happy (with `allow(too_many_arguments)` since refactoring that function is not on my immediate agenda).

I didn't rename but wouldn't be opposed to `outbox / Outbox`. Any more thoughts?

dhardy:
@afck I added one commit to rename as agreed. Other commits are untouched. Should be ready to merge as soon as the CI runs are complete.

src/outtray.rs
afck: In my experience, commented-out code quickly goes out of sync and is just removed eventually. Let's remove it now.

PR #1354, opened 2017-02-13
fix/routing_table: fix several issues with section merges

afck:


maidsafe-highfive:

src/routing_table/mod.rs
```diff
@@ -632,17 +626,15 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
         };
         if bit_count == 0 || self.we_want_to_merge ||
            !self.sections.contains_key(&self.our_prefix.with_flipped_bit(bit_count - 1)) ||
-           (self.our_section.len() >= self.min_section_size &&
+           (!self.they_want_to_merge && self.our_section.len() >= self.min_section_size &&
             self.sections.iter().all(doesnt_need_to_merge_with_us)) {
             return None;
```
dhardy: This could be refactored into 2 separate `if` blocks each with their own `return None` line and a comment as to why: first block for the first three tests, and second block for the last three conditions chained with AND. (Or even a third block, for the "our prefix with last bit flipped doesn't exist.) More verbose, but hopefully more readable.

src/routing_table/mod.rs
```diff
@@ -662,16 +654,15 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
...
-            if merge_details.merge_prefix.is_compatible(prefix) &&
-               !self.sections.contains_key(prefix) && !compatible_with_ours {
-                self.merge(prefix);
-            }
-            // Add an empty section in the table.
-            if !compatible_with_ours {
-                let _ = self.sections.entry(*prefix).or_insert_with(HashSet::new);
+            if *prefix == self.our_prefix || self.sections.contains_key(prefix) {
+                continue; // Already in our routing table.
+            } else if self.our_prefix.is_compatible(prefix) ||
```
dhardy: Should we also check that the prefix is a neighbour, and that after we're done there aren't any missing prefixes? Okay, `finish_merging...` adds missing prefixes, but if there were any that itself should be an error.

afck: We don't remove any prefixes here so I don't think there's a need to check for missing ones.
And the check that every prefix is a neighbour would fail, unfortunately, because of the way we do the merge in two steps: we _do_ insert prefixes here that aren't our neighbour yet, but will be after `finish_merging_own_section`.
We will need to clean up the merge process to fix that, but that's beyond the scope of this PR.

src/routing_table/mod.rs
afck: Done.

PR #1355, opened 2017-02-13
MAID-1957 test/tunnel: tunnel client reconnect

maqi:


maidsafe-highfive:
r? @afck

maqi:
> we could modify the mock crust framework to enable the blocking/dropping of tunnelled connections there, and just allow the production code to take care of handling the generated crust events
> when we remove a connection we can check for a NodeLost event

tests/mock_crust/tunnel.rs
```diff
@@ -0,0 +1,154 @@
...
+        node.inner.set_next_node_name(name1);
+    }
+    nodes.push(TestNode::builder(network).config(config.clone()).create());
+
+    (Endpoint(nodes.len() - 2), Endpoint(nodes.len() - 1))
+}
+
+#[test]
+fn tunnel_clients() {
+    let min_section_size = 5;
```
afck: Maybe we could even use size 3, to speed up the test?

tests/mock_crust/tunnel.rs
```diff
@@ -0,0 +1,148 @@
...
+        .rev()
+        .filter_map(|(index, node)| if node.routing_table().our_prefix().matches(prefix_name) {
+            Some(index)
+        } else {
+            None
+        })
+        .collect();
+    section_indexes.iter()
+        .take(section_indexes.len() - min_section_size + 1)
+        .foreach(|index| { let _ = nodes.remove(*index); });
```
dhardy: You'd best sort the indices from highest to lowest if you want to remove them from the `Vec` like this.

tests/mock_crust/tunnel.rs
```diff
@@ -0,0 +1,148 @@
...
+    poll_and_resend(nodes, &mut []);
+}
+
+// Adds a pair of nodes with specified names into the network. Returns the endpoints of the nodes.
```
dhardy: Also blocks direct connection between these two nodes if `is_tunnel` is set.

tests/mock_crust/utils.rs
```diff
@@ -378,26 +390,26 @@ pub fn create_connected_nodes_with_cache(network: &Network, size: usize, use_cac
 //
 // The array is sanity checked (e.g. it would be an error to pass [1, 1, 1]), must comprise at
 // least two elements, and every element must be no more than `8`.
-pub fn create_connected_nodes_until_split(network: &Network,
-                                          mut prefix_lengths: Vec<usize>,
-                                          use_cache: bool)
-                                          -> Nodes {
+pub fn create_connected_nodes_until_split_with_nodes(network: &Network,
```
dhardy: Should be `add_connected_nodes_until_split`.

tests/mock_crust/tunnel.rs
maqi: it was taken in a reversed order, https://github.com/maidsafe/routing/pull/1355/files#diff-b83639c52df93eae4d3e66ec59a8e66fR69
so I think we are ok here?

dhardy: I guess so. I don't much like this style of coding because it's easy to accidentally break it, but I guess it'll do.

PR #1356, opened 2017-02-13
MAID-1677: signed message relay signatures

dhardy:
Rebased. Since there were a lot of conflicts, I pushed to a new branch. The old branch and PR are still available:
https://github.com/maidsafe/routing/pull/1310

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -243,10 +243,20 @@ pub struct SectionList {
 pub type SectionListSignatures = BTreeMap<PublicId, sign::Signature>;
 
 /// A list of signatures of a section list
-#[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Hash, RustcEncodable, RustcDecodable, Debug)]
+#[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Hash, RustcEncodable, RustcDecodable)]
 pub struct SignedSectionList {
     pub list: SectionList,
     pub signatures: SectionListSignatures,
+}
+
+impl Debug for SignedSectionList {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        // Printing out signatures is not very useful to us, and makes messages a lot longer
+        write!(f,
+               "SignedSectionList {{ list: {:?}, signatures: {} }}",
+               self.list,
+               self.signatures.len())
+    }
 }
 
 impl SectionList {
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -818,9 +818,11 @@ impl Debug for HopMessage {
 impl Debug for SignedMessage {
     fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
         write!(formatter,
-               "SignedMessage {{ content: {:?}, sending nodes: {:?}, signatures: {:?} }}",
+               "SignedMessage {{ content: {:?}, sending sections: {:?}, relaying sections: {:?}, \
+                    signatures: {:?} }}",
                self.content,
                self.src_sections,
+               self.relay_sections,
                self.signatures.keys().collect_vec())
     }
 }
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -236,8 +236,17 @@ impl HopMessage {
 #[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Hash, RustcEncodable, RustcDecodable, Debug)]
 pub struct SectionList {
     pub prefix: Prefix<XorName>,
-    // TODO(MAID-1677): pub signatures: BTreeSet<(PublicId, sign::Signature)>,
     pub_ids: BTreeSet<PublicId>,
+}
+
+/// A collection of signatures of the contents of a `SectionList`.
+pub type SectionListSignatures = BTreeMap<PublicId, sign::Signature>;
+
+/// A list of signatures of a section list
+#[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Hash, RustcEncodable, RustcDecodable, Debug)]
+pub struct SignedSectionList {
+    pub list: SectionList,
+    pub signatures: SectionListSignatures,
 }
 
 impl SectionList {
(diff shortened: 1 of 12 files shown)


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -722,43 +728,79 @@ impl Node {
...
+                         signed_msg: SignedMessage,
                          route: u8,
                          hop_name: XorName,
                          sent_to: &BTreeSet<XorName>) {
         self.send_ack(signed_msg.routing_message(), route);
         // If the destination is our section we need to forward it to the rest of the section
         if signed_msg.routing_message().dst.is_multiple() {
             if let Err(error) = self.send_signed_message(signed_msg, route, &hop_name, sent_to) {
-                debug!("{:?} Failed to send {:?}: {:?}", self, signed_msg, error);
+                debug!("{:?} Failed to send signed message: {:?}", self, error);
```
Fraser999: Why was the actual message removed from the debug output here?  Same question for line 817 below.

src/states/node.rs
```diff
@@ -2402,8 +2447,12 @@ impl Node {
         Ok(())
     }
 
+    // Send signed_msg on route. Hop if the name of the peer we received this from, or our name if
```
Fraser999: Typo in "Hop if the name of the peer" if -> is?

src/states/node.rs
```diff
@@ -2828,6 +2892,11 @@ impl Node {
         }
     }
 
+    /// Shortcut to get our prefix
+    fn our_prefix(&self) -> &Prefix<XorName> {
```
Fraser999: There are quite a few remaining places where `self.peer_mgr.routing_table().our_prefix()` should be replaced by `self.our_prefix()`.

src/states/node.rs
```diff
@@ -2828,6 +2892,11 @@ impl Node {
         }
     }
 
+    /// Shortcut to get our prefix
```
Fraser999: I'm not sure this getter needs a comment?

src/states/node.rs
```diff
@@ -2962,10 +3031,10 @@ impl Node {
     }
 
     pub fn section_list_signatures(&self,
-                                   prefix: Prefix<XorName>)
+                                   prefix: &Prefix<XorName>)
                                    -> Result<BTreeMap<PublicId, sign::Signature>, RoutingError> {
-        if let Some(&(_, ref signatures)) = self.section_list_sigs.get_signatures(prefix) {
-            Ok(signatures.iter().map(|(&pub_id, &sig)| (pub_id, sig)).collect())
+        if let Some(ssl) = self.section_list_sigs.get_signed_list(prefix) {
+            Ok(ssl.signatures.iter().map(|(&pub_id, &sig)| (pub_id, sig)).collect())
```
Fraser999: Can't this just be `Ok(ssl.signatures)`?

src/messages.rs
```diff
@@ -274,20 +293,81 @@ impl SignedMessage {
...
+            if QUORUM * last_verified.pub_ids.len() > 100 * valid_sigs {
+                return Err(RoutingError::NotEnoughSignatures);
+            }
+            Ok(())
+        }
+
+        // -----  first, verify the route  -----
+        if let Some(known_section) = known_section {
+            let mut last_verified = known_section;
+            for prev_hop in self.relay_sections.iter().rev() {
```
dhardy: Yes. We check each against the last verified section signature, then check the next against that... so reverse order is required.

src/messages.rs
```diff
@@ -236,10 +236,29 @@ impl HopMessage {
 #[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Hash, RustcEncodable, RustcDecodable, Debug)]
 pub struct SectionList {
     pub prefix: Prefix<XorName>,
-    // TODO(MAID-1677): pub signatures: BTreeSet<(PublicId, sign::Signature)>,
     pub_ids: BTreeSet<PublicId>,
 }
 
+/// A collection of signatures of the contents of a `SectionList`.
```
dhardy: Fixed

src/states/node.rs
```diff
@@ -722,43 +728,81 @@ impl Node {
...
-        signed_msg.check_integrity(self.min_section_size())?;
+        let hop_prefix = self.peer_mgr
+            .routing_table()
+            .find_section_prefix(&hop_name)
+            .ok_or(RoutingTableError::NoSuchPeer)?;
+        let section_list = if signed_msg.routing_message().src.is_client() ||
+                              self.in_authority(&signed_msg.routing_message().dst) {
+            // TODO: we should still verify the sending nodes when the sender is our own section
+            // or group, but doing so causes a lot of verification failures at the moment.
+            // (In any case don't call add_relaying_section if from our own section.)
```
afck: I saw similar issues, and I think some of those failures might be because as a candidate, the second node expects two signatures in its `NodeApproval`, because it already considers itself part of the section. A simple workaround is to just exclude messages that were sent to us as a client here, i.e. if `dst` is a `Client` with our key. Does that resolve the failures?

Also, I think it might be a good idea to include our own section's list if we are the first node in our section to receive this message, i.e. if the `hop_name` is not in our section:
We need to ensure that either all nor none of the nodes in the section accept the message, and adding our own section list would reduce that problem to all nodes in our section accepting our own list of our own section, as opposed to our list of a neighbouring section. But that probably needs some more discussion.

dhardy:
TODO (probably me):
1. check that when the destination is a prefix section (i.e. some sections both receive the message themselves and forward), that all necessary signatures still get added so all recipients can verify the route
2. handle `OtherSectionMerge`: if receiving a message from a newly merged section from whom we don't have signatures, check whether we do have signatures for any sub-prefix

dhardy:
From Fraser999:
So, https://github.com/Fraser999/routing/blob/master/src/states/node.rs#L1009 is causing a failure because we set `is_approved` _after_ we call each `send_connection_info_request` (line 986) which means the newly-approved node sends a `CandidateIdentify` to its proxy rather than a `NodeIdentify`.  The proxy at least this time doesn't kill the connection (it tries to but correctly identifies the peer as a joining node), but we're still looking at broken RTs since the new node has added its proxy to its RT, while the proxy hasn't added the new node.

dhardy:
Closing (@afck is working on an updated version)

PR #1357, opened 2017-02-14
MAID-1785 More churn

maqi:


maidsafe-highfive:
r? @afck

tests/mock_crust/churn.rs
```diff
@@ -201,64 +315,104 @@ fn verify_section_list_signatures(nodes: &[TestNode]) {
 
 #[test]
 fn churn() {
+    let min_section_size = 5;
+    let mut network = Network::new(min_section_size, None);
+    let mut rng = network.new_rng();
+
+    // TODO: debug failures when starting with a network < min_section_size
```
Fraser999: This todo can be deleted, right?

tests/mock_crust/churn.rs
```diff
@@ -201,64 +315,104 @@ fn verify_section_list_signatures(nodes: &[TestNode]) {
...
+        send_and_receive(&mut rng, &mut nodes, min_section_size, Some(added_index));
+        client_gets(&mut network, &mut nodes, min_section_size);
+    }
+
+    info!("Churn [{} nodes, {} sections]: done",
+          nodes.len(),
+          count_sections(&nodes));
+}
+
+fn bootstrap_from(initial_nodes: usize, optional_seed: Option<[u32; 4]>) {
```
Fraser999: There's no point to having `optional_seed` here I think; we always pass `None`.

maqi:
added simultaneous join and leave test

tests/mock_crust/churn.rs
```diff
@@ -18,49 +18,83 @@
...
+        let num_excess = cmp::max(1,
+                                  cmp::min(nodes[i].routing_table().our_section().len() -
+                                           min_quorum,
+                                           len - min_section_size));
+        assert!(num_excess > 0);
 
-        nodes.insert(index, TestNode::builder(network).config(config).create());
-        Some(index)
+        let mut removed = 0;
+        // Remove as many more as possible:
```
Fraser999: This comment still implies that at least one has _already_ been removed.  I guess I'd just say "Remove nodes" or something like that.

tests/mock_crust/churn.rs
```diff
@@ -18,49 +18,83 @@
...
+        while removed < num_excess {
+            let i = rng.gen_range(0, len - removed - 1);
```
Fraser999: Again, I think this `- 1` dates back to when a node was removed before the loop - we don't want it now.  We don't even really need the `removed` variable at all if we just use `nodes.len()`, but that's a very minor point.

tests/mock_crust/churn.rs
```diff
@@ -201,64 +317,127 @@ fn verify_section_list_signatures(nodes: &[TestNode]) {
...
+        let (added_index, _) = add_random_node(&mut rng, &network, &mut nodes, min_section_size);
+        poll_and_resend(&mut nodes, &mut []);
+        send_and_receive(&mut rng, &mut nodes, min_section_size, Some(added_index));
+        client_gets(&mut network, &mut nodes, min_section_size);
+    }
+
+    info!("Churn [{} nodes, {} sections]: simultaneous adding and dropping nodes",
+          nodes.len(),
+          count_sections(&nodes));
+    for _ in 0..10 {
```
Fraser999: I think this can be changed for `while nodes.len() > min_section_size + 1` since we always remove more nodes than we add.  We should eventually remove so many that `nodes` goes down to `min_section_size` then we add another one.  We can also then remove the 3 following lines.

tests/mock_crust/churn.rs
```diff
@@ -18,49 +18,83 @@
...
+        // ensured they did not all come from the same section we could remove more):
+        let num_excess = cmp::min(min_section_size - min_quorum, len - min_section_size);
+        let mut removed = 0;
+        while num_excess - removed > 0 {
+            let _ = nodes.remove(rng.gen_range(0, len - removed));
+            removed += 1;
+        }
     }
 }
+// Randomly add a node. Returns the index of this node.
```
afck: Let's separate methods with blank lines.

tests/mock_crust/churn.rs
```diff
@@ -201,64 +317,126 @@ fn verify_section_list_signatures(nodes: &[TestNode]) {
...
+    // decrease back to min_section_size, then increase to again.
+    let mut nodes = create_connected_nodes(&network, min_section_size);
+
+    info!("Churn [{} nodes, {} sections]: adding nodes",
+          nodes.len(),
+          count_sections(&nodes));
+    loop {
+        let (added_index, _) = add_random_node(&mut rng, &network, &mut nodes, min_section_size);
+        poll_and_resend(&mut nodes, &mut []);
+        send_and_receive(&mut rng, &mut nodes, min_section_size, Some(added_index));
```
afck: We should also do `client_gets` in this phase.

tests/mock_crust/churn.rs
```diff
@@ -201,64 +317,126 @@ fn verify_section_list_signatures(nodes: &[TestNode]) {
...
+
+#[test]
+fn bootstrap_6() {
+    bootstrap_from(6);
+}
+
+#[test]
+fn bootstrap_7() {
+    bootstrap_from(7);
+}
```
afck: Aren't those `bootstrap_` tests just duplicating the `test_nodes` tests in `tests/mock_crust/mod.rs`?

PR #1359, opened 2017-02-15
Remove unused #[macro_use] and extern crate decls

michaelsproul:
I've been using the nightly compiler a bit (for incremental compilation), and it's smart enough to know that these declarations aren't used.


maidsafe-highfive:
r? @afck

PR #1360, opened 2017-02-16
fix/node_restart test: drop all-but-one node

dhardy:


maidsafe-highfive:
r? @afck

PR #1362, opened 2017-02-17
Node and churn test fixes.

afck:


maidsafe-highfive:

src/states/node.rs
```diff
@@ -871,8 +871,8 @@ impl Node {
                                                      src_name,
                                                      dst)
             }
-            (CandidateApproval { candidate_id, client_auth, sections }, Section(_), Section(_)) => {
-                self.handle_candidate_approval(candidate_id, client_auth, sections, outbox)
+            (CandidateApproval { candidate_id, client_auth, .. }, Section(_), Section(_)) => {
+                self.handle_candidate_approval(candidate_id, client_auth, outbox)
```
dhardy: You should also remove `sections` from `CandidateApproval` and `PeerManager::verified_candidate_info`. Doesn't need to land in this PR necessarily.

afck: I think the plan was to leave that in for safety: If the section can't even accumulate the `CandidateApproval` including the routing table, it's better to not even add the candidate to anyone's routing table.
The problem was that the routing tables could change between sending and handling that message, and then it would be better to put the new routing table into the `NodeApproval`. If _that_ doesn't accumulate, the nodes do put the candidate in their routing tables and have to rely on the candidate disconnecting.

dhardy: True, okay +1. (But this will become irrelevant with our proposed routing changes later.)

Fraser999: I guess we could just use a hash of the sections.  As well as cutting down the message size, it would maybe be more obvious that we're not actually doing anything with the sections on receipt of this message.

Not a big deal though; this isn't a common message compared to many others.

tests/mock_crust/churn.rs
```diff
@@ -328,115 +356,115 @@ fn churn() {
...
+        expected_gets.send_and_expect(data_id, auth_n0, auth_n1, &mut nodes, min_section_size);
+        expected_gets.send_and_expect(data_id, auth_n0, auth_g0, &mut nodes, min_section_size);
+        expected_gets.send_and_expect(data_id, auth_n0, auth_s0, &mut nodes, min_section_size);
+        // ... and from a group to itself, another group, a section and a node...
+        expected_gets.send_and_expect(data_id, auth_g0, auth_g0, &mut nodes, min_section_size);
+        expected_gets.send_and_expect(data_id, auth_g0, auth_g1, &mut nodes, min_section_size);
+        expected_gets.send_and_expect(data_id, auth_g0, auth_s0, &mut nodes, min_section_size);
+        expected_gets.send_and_expect(data_id, auth_g0, auth_n0, &mut nodes, min_section_size);
+        // ... and from a section to itself, another section, a group and a node...
+        // TODO: Enable these once MAID-1920 is fixed.
```
maqi: this part is currently enabled in the master, shall we enable it here as well?

PR #1363, opened 2017-02-17
refactor/logs: downgrade and deduplicate some warning messages

Fraser999:
Also fixed crate-level example code.


maidsafe-highfive:
r? @afck

PR #1367, opened 2017-02-20
fix/resource_proof: includes merge status in NodeApproval

maqi:


maidsafe-highfive:
r? @afck

PR #1368, opened 2017-02-20
fix/routing_table: send merge messages to PrefixSections

afck:
Send `OtherSectionMerge` messages to the neighbouring `PrefixSection`s
of the merge prefix, too, regardless of whether those exist as sections
in the routing table. That way the messages accumulate even if a
neighbouring section is currently merging or splitting.

### DIFF
--- a/src/routing_table/mod.rs
+++ b/src/routing_table/mod.rs
@@ -909,7 +909,14 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
         self.they_want_to_merge = false;
         self.merge(&merge_prefix);
         self.add_missing_prefixes();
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:

src/routing_table/mod.rs
```diff
@@ -909,7 +909,11 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
         self.they_want_to_merge = false;
         self.merge(&merge_prefix);
         self.add_missing_prefixes();
-        let targets = self.sections.keys().cloned().collect();
+        let targets = self.sections
```
dhardy: Include a comment here explaining why this happens please (copy from PR/commit description if need be).

Otherwise looks good.

PR #1371, opened 2017-02-21
fix/routing_table: not replace shorter prefix on merge

maqi:


maidsafe-highfive:
r? @afck

src/routing_table/mod.rs
```diff
@@ -714,13 +714,19 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
...
-        self.merge(&merge_details.prefix);
-
-        // Establish list of provided contacts which are currently missing from our table.
-        merge_details.section
-            .difference(unwrap!(self.sections.get(&merge_details.prefix)))
-            .cloned()
-            .collect()
+        if self.sections.keys().any(|prefix| {
+            prefix.is_compatible(&merge_details.prefix) &&
+            prefix.bit_count() >= merge_details.prefix.bit_count()
```
Fraser999: I think `>=` can be just `>`.

src/routing_table/mod.rs
```diff
@@ -714,13 +714,19 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
...
+        }) {
+            self.merge(&merge_details.prefix);
+            // Establish list of provided contacts which are currently missing from our table.
+            merge_details.section
+                .difference(unwrap!(self.sections.get(&merge_details.prefix)))
+                .cloned()
+                .collect()
+        } else {
+            BTreeSet::new()
```
Fraser999: I guess a comment like "We've already handled this merge." would be helpful here.

PR #1372, opened 2017-02-22
chore/base: simply error handling in Base::send_message

dhardy:
Don't return an error from send_message if serialisation fails; just log an
error and drop the message.

(This is the first of several refactors I have.)


maidsafe-highfive:
r? @afck

PR #1373, opened 2017-02-22
chore/states/node.rs: minor improvements to doc and style

dhardy:


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -716,19 +719,21 @@ impl Node {
...
+                         signed_msg: SignedMessage,
                          route: u8,
                          hop_name: XorName,
                          sent_to: &BTreeSet<XorName>) {
         self.send_ack(signed_msg.routing_message(), route);
         // If the destination is our section we need to forward it to the rest of the section
         if signed_msg.routing_message().dst.is_multiple() {
             if let Err(error) = self.send_signed_message(signed_msg, route, &hop_name, sent_to) {
-                debug!("{:?} Failed to send {:?}: {:?}", self, signed_msg, error);
+                debug!("{:?} Failed to send signed message: {:?}", self, error);
```
Fraser999: I personally wouldn't make this optimisation for now.

src/states/node.rs
```diff
@@ -771,8 +772,8 @@ impl Node {
             return Ok(());
         }
 
-        if let Err(error) = self.send_signed_message(&signed_msg, route, &hop_name, sent_to) {
-            debug!("{:?} Failed to send {:?}: {:?}", self, signed_msg, error);
+        if let Err(error) = self.send_signed_message(signed_msg, route, &hop_name, sent_to) {
+            debug!("{:?} Failed to send signed message: {:?}", self, error);
```
Fraser999: Again, I'd prefer to have the message printed at the cost of cloning it.

src/states/node.rs
```diff
@@ -2377,8 +2378,12 @@ impl Node {
         Ok(())
     }
 
+    // Send signed_msg on route. Hop if the name of the peer we received this from, or our name if
```
Fraser999: Typo: should be "Hop is the name".

Also in comment lines/blocks, we generally wrap variables in ticks.  This applies to a few places in this PR.

src/states/node.rs
```diff
@@ -2388,36 +2393,46 @@ impl Node {
...
+    // Filter, then convert the message to a `Hop` or `TunnelHop` `Message` and serialise.
+    // Send this byte string.
     fn send_signed_msg_to_peer(&mut self,
-                               signed_msg: &SignedMessage,
+                               signed_msg: SignedMessage,
                                target: PeerId,
                                route: u8,
                                sent_to: BTreeSet<XorName>)
                                -> Result<(), RoutingError> {
+        if self.filter_outgoing_routing_msg(signed_msg.routing_message(), &target, route) {
```
Fraser999: Moving this `if` block before the call to `self.crust_service.is_connected()` could now mean we're not calling `self.disconnect_peer(&target);` in some cases.  Is that intentional?

src/states/node.rs
```diff
@@ -2737,11 +2756,13 @@ impl Node {
...
                                 merge_details: OtherMergeDetails<XorName>) {
         let section = self.peer_mgr.get_pub_ids(&merge_details.section);
+        let content = MessageContent::OtherSectionMerge(section);
+        let src = Authority::PrefixSection(merge_details.prefix);
         for target in &targets {
-            let content = MessageContent::OtherSectionMerge(section.clone());
-            let src = Authority::PrefixSection(merge_details.prefix);
             let dst = Authority::PrefixSection(*target);
-            if let Err(err) = self.send_routing_message(src, dst, content) {
+
```
Fraser999: Can at least one blank line be removed please?

src/states/node.rs
dhardy: If the message got filtered, then we already sent an identical version to this client, so were able to send to it then. I also don't think it's critical to immediately disconnect from a peer we don't have a route to (we ought to have noticed before if the route just dropped?).

I'm not 100% sure this would be equivalent in some edge-case, but the differences seem pretty unimportant.

Fraser999: I tend to agree - I don't recall seeing us hit that line in previous tests.  I'm just a bit nervous about changes like these.

PR #1375, opened 2017-02-23
Exclusive matching in dispatch_routing_message and cleanup of handle_timeout

dhardy:
There is one change to behaviour: reject UserMessageParts while a candidate. Not 100% sure on this.

r? @afck

### DIFF
--- a/src/states/node.rs
+++ b/src/states/node.rs
@@ -808,14 +808,22 @@ impl Node {
                 ConnectionInfoRequest { .. } |
                 SectionUpdate { .. } |
                 RoutingTableRequest(..) |
-                RoutingTableResponse { .. } => {
+                RoutingTableResponse { .. } |
+                UserMessagePart { .. } => {
+                    // These messages should not be handled before node approval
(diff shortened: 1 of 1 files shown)


src/states/node.rs
```diff
@@ -2203,24 +2209,17 @@ impl Node {
...
             self.rt_timeout = cmp::min(Duration::from_secs(RT_MAX_TIMEOUT_SECS),
                                        self.rt_timeout * 2);
             trace!("{:?} Scheduling next RT request for {} seconds from now.",
                    self,
                    self.rt_timeout.as_secs());
             self.rt_timer_token = Some(self.timer.schedule(self.rt_timeout));
-            if self.send_rt_request().is_err() {
-                return true;
-            }
+            self.send_rt_request().is_err();
```
afck: Probably `send_rt_request` and `send_candidate_approval` could have return type `()` now? Serialisation errors can just be logged inside them. Then this `is_err()` here could go.

dhardy:
@afck I updated regarding your suggestions (new commit). You happy with this?

PR #1376, opened 2017-02-24
Use SHA-3 everywhere instead of SipHash

michaelsproul:
Main changes:

* `UserMessagePart`s now include the SHA3-256 hash of the full message, rather than just a SipHash.
* Outgoing routing messages are now de-duplicated using SHA3-256 instead of SipHash.

### DIFF
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -813,13 +813,15 @@ impl Debug for MessageContent {
             Ack(ack, priority) => write!(formatter, "Ack({:?}, {})", ack, priority),
             UserMessagePart { hash, part_count, part_index, priority, cacheable, .. } => {
                 write!(formatter,
-                       "UserMessagePart {{ {}/{}, priority: {}, cacheable: {}, {:02x}{:02x}.. }}",
+                       "UserMessagePart {{ {}/{}, priority: {}, cacheable: {}, \
+                        {:02x}{:02x}{:02x}.. }}",
                        part_index + 1,
                        part_count,
                        priority,
                        cacheable,
                        hash[0],
-                       hash[1])
+                       hash[1],
+                       hash[2])
             }
             AcceptAsCandidate { ref expect_id, ref client_auth, ref message_id } => {
                 write!(formatter,
@@ -1175,11 +1177,13 @@ impl UserMessageCache {
         {
             let entry = self.0.entry((hash, part_count)).or_insert_with(BTreeMap::new);
             if entry.insert(part_index, payload).is_some() {
-                debug!("Duplicate UserMessagePart {}/{} with hash {:02x}{:02x}.. added to cache.",
+                debug!("Duplicate UserMessagePart {}/{} with hash {:02x}{:02x}{:02x}.. \
+                        added to cache.",
                        part_index + 1,
                        part_count,
                        hash[0],
-                       hash[1]);
+                       hash[1],
+                       hash[2]);
             }
 
             if entry.len() != part_count as usize {
--- a/src/messages.rs
+++ b/src/messages.rs
@@ -24,7 +24,6 @@ use event::Event;
 use id::{FullId, PublicId};
 use itertools::Itertools;
 use lru_time_cache::LruCache;
-use maidsafe_utilities;
 use maidsafe_utilities::serialisation::{deserialise, serialise};
 #[cfg(feature = "use-mock-crust")]
 use mock_crust::crust::PeerId;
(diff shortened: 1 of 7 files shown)


maidsafe-highfive:
r? @afck

michaelsproul:
**DON'T MERGE YET**

michaelsproul:
@Viv-Rajkumar: I think this is ready to get un-paused. No longer requires any changes to maidsafe_utilities.

michaelsproul:
@Fraser999: Got it working with 1.14.0 now I think. The issue was a few imports I deleted in a test sub-module that weren't being pulled in by `use super::*` on 1.14.0, but seem to be fine on subsequent versions... strange.

I didn't have any issue with 1.14.0 and the `#[allow(unused_extern_crate)]` business, so I've left it deleted.

michaelsproul:
Updated (for 3 byte debug hashes)

src/messages.rs
```diff
@@ -33,11 +32,13 @@ use routing_table::{Prefix, Xorable};
 use routing_table::Authority;
 use rust_sodium::crypto::{box_, sign};
 use rust_sodium::crypto::hash::sha256;
+use sha3;
```
maqi: can we `use sha3::Digest256` directly here?

src/messages.rs
```diff
@@ -33,11 +32,13 @@ use routing_table::{Prefix, Xorable};
...
 use std::collections::{BTreeMap, BTreeSet, HashSet};
 use std::fmt::{self, Debug, Formatter};
 use std::iter;
 use std::time::Duration;
 use super::QUORUM;
+use tiny_keccak::sha3_256;
```
maqi: is sha3_256 a function?
if so, I think `use tiny_keccak` is preferred here, then `tiny_keccak::sha3_256()` in the code?

src/routing_message_filter.rs
```diff
@@ -70,9 +72,15 @@ impl RoutingMessageFilter {
 
     // Filter outgoing `RoutingMessage`. Return whether this specific message has been seen recently
     // (and thus should not be sent, due to deduplication).
+    //
+    // Return `false` if serialisation of the message fails - that can be handled elsewhere.
```
maqi: seems the previous definition of the return value is whether the msg has been inserted previously.

hence a normal meaning of `false` is that this is the first time of insertion
is it really OK to change the definition?

src/sha3.rs
```diff
@@ -15,57 +15,5 @@
...
-        let expected = vec![0x33, 0x38, 0xbe, 0x69, 0x4f, 0x50, 0xc5, 0xf3, 0x38, 0x81, 0x49,
-                            0x86, 0xcd, 0xf0, 0x68, 0x64, 0x53, 0xa8, 0x88, 0xb8, 0x4f, 0x42,
-                            0x4d, 0x79, 0x2a, 0xf4, 0xb9, 0x20, 0x23, 0x98, 0xf3, 0x92];
-
-        assert_eq!(expected, res);
-    }
-
-}
+/// SHA3-256 hash digest.
+pub type Digest256 = [u8; 32];
```
maqi: do we really still need this as a separate file?

michaelsproul: I figured I'd keep it so I could re-use the same type definition in `safe_vault`. The `sha3` module is public for this reason as well.

None of the alternatives I've thought of are very appealing: not using a type definition (even easier to make mistakes), repeating the type definition in different files of routing and safe_vault (needless duplication).

src/routing_message_filter.rs
michaelsproul: Hmm, it's complicated... The previous comment said nothing about the case where the message is too large to be serialised, and the behaviour of `big_endian_sip_hash` was that messages that failed to serialise would be treated as having content `vec![]`, and thus all have the same hash (see: https://github.com/maidsafe/maidsafe_utilities/blob/master/src/big_endian_sip_hash.rs#L26-L32). This would mean the *first* oversized message to be inserted would return `false`, and all subsequent ones would return `true`.

We could return to similar behaviour by returning `true` for oversized messages (as in, don't send!), although in the case of the first oversized message, our return value would be different from the current behaviour.

Or, maybe we should defer handling of the error entirely and return a `Result<bool, ...>` that can be handled by the caller (which is almost always another function returning a `Result`, so we can just use `?`).

On the other hand, I don't think it's too significant what we return, because a filtering is option usually only occurs right before or right after a message is serialised for sending (and we handle that failure). The only slightly complicated case I came across is in `relay_to_client`, which would previously have returned `Ok(())` for all oversized messages without sending them. However, I'm pretty sure getting an oversized `RoutingMessage` here is impossible, as its pulled out of a `SignedMessage` and a `SignedMessage` ensures that the contents are serialisable...

maqi: I will prefer the approach of `return a Result<bool, ...>`.
Also I agree this can be done in a later PR.

dhardy: Well spotted Michael. Qi may be right about returning `Result` but that distinction doesn't seem especially important.

Perhaps it would be sufficient to add a `debug`/`trace`/`warn` message when serialisation fails?

michaelsproul: Added a trace message (initially added an `error!` 'cause I figured it was unreachable, but even if it is reachable, we probably don't want to spam people with an error here).

PR #1377, opened 2017-02-24
MAID-1976: chore/test: refactored RT test

Fraser999:
This refactoring now allows the test to work with any reasonable value for 'SPLIT_BUFFER'.


maidsafe-highfive:
r? @afck

PR #1378, opened 2017-02-24
chore/test: fix bug in RT network test

Fraser999:
The test has been updated to handle an imbalanced network where more than one split is triggered by
the addition of a single new node.  Also 'HashSet's and 'HashMap's have been replaced with
'BTreeSet's and 'BTreeMap's respectively to make the test deterministic, and some unneeded types
and imports removed.


maidsafe-highfive:
r? @afck

PR #1382, opened 2017-03-02
Maid-1959 tunnel node disrupted

maqi:


maidsafe-highfive:
r? @afck

tests/mock_crust/tunnel.rs
```diff
@@ -147,3 +147,74 @@ fn tunnel_clients() {
...
+    let min_section_size = 5;
+    let network = Network::new(min_section_size, None);
+    network.block_connection(Endpoint(2), Endpoint(3));
+    network.block_connection(Endpoint(3), Endpoint(2));
+    let mut nodes = create_connected_nodes(&network, min_section_size);
+
+    network.send_crust_event(Endpoint(2), crust::Event::LostPeer(PeerId(0)));
+    network.send_crust_event(Endpoint(0), crust::Event::LostPeer(PeerId(2)));
+    poll_and_resend(&mut nodes, &mut []);
+    verify_tunnel_switch(&mut nodes, 0, 2, 3);
```
dhardy: I don't really like the style of this test. Can't we instead add a hook to let us check that firstly, 2 and 3 connect via 0, and secondly that they reconnect (it might be enough just to check the invariant at the end, rather than how the nodes are connected)?

maqi: that `hook` means we need access to node.tunnels, which is not that preferred.  :(
Also, as the tunnel_client A and tunnel_node will remove each other from RT, and section_update/ routing_table_request is not preferred to be used in test, the invariant check at the end will fail the test. :(

tests/mock_crust/tunnel.rs
```diff
@@ -147,3 +147,74 @@ fn tunnel_clients() {
...
+    let _ = poll_all(&mut nodes, &mut []);
+    verify_invariant_for_all_nodes(&nodes);
+}
+
+// Verifies tunnel switched.
+fn verify_tunnel_switch(nodes: &mut Vec<TestNode>, node: usize, client_1: usize, client_2: usize) {
+    while let Ok(event) = nodes[client_1].inner.try_next_ev() {
+        match event {
+            Event::NodeLost(name, _) => {
+                assert!(name == nodes[node].name() || name == nodes[client_2].name());
```
Fraser999: I guess we could tighten this up slightly by ensuring that `client_1` receives exactly two `NodeLost`s then one `NodeAdded`, and that `client_2` receives one `NodeLost` followed by one `NodeAdded`?

src/states/node.rs
```diff
@@ -3012,6 +3019,11 @@ impl Node {
         self.peer_mgr.routing_table()
     }
 
+    /// Check whether the node acts ac tunnel_node for the tunnul pair.
```
Fraser999: Couple of typos here.  Maybe "Check whether this node acts as a tunnel node between \`client_1\` and \`client_2\`"?  Whatever you choose, I'd duplicate it exactly for the corresponding comment in src/node.rs.

src/states/node.rs
```diff
@@ -443,7 +443,14 @@ impl Node {
                    self,
                    peer_id,
                    pub_id);
-            self.find_tunnel_for_peer(peer_id, &pub_id);
+            if self.tunnels.tunnel_for(&peer_id).is_none() {
```
Fraser999: I'm not sure this change to the flow was agreed?

maqi: I don't see the objection regarding https://maidsafe.slack.com/archives/routing/p1488466408009790 point 2?

Fraser999: Not sure that's enough to assume agreement though?

Viv-Rajkumar: Just discussed this and should be fine for this to occur here.

tests/mock_crust/tunnel.rs
```diff
@@ -110,6 +107,25 @@ fn add_a_pair(network: &Network,
...
+        })
+        .collect();
+    // There shall be only one tunnel_node for a pair of tunnel_clients across the network
+    // Or None if they are directly connected or one of them are no longer in the network
+    assert!(tunnel_node_indexes.len() <= 1);
+    if tunnel_node_indexes.is_empty() {
+        None
+    } else {
+        Some(tunnel_node_indexes[0])
+    }
```
afck: You could just return `tunnel_node_indexes.into_iter.next()` instead.

tests/mock_crust/tunnel.rs
```diff
@@ -146,4 +169,104 @@ fn tunnel_clients() {
...
+    let network = Network::new(min_section_size, None);
+    network.block_connection(Endpoint(2), Endpoint(3));
+    network.block_connection(Endpoint(3), Endpoint(2));
+    let mut nodes = create_connected_nodes(&network, min_section_size);
+    let tunnel_node_index = unwrap!(locate_tunnel_node(&nodes, PeerId(2), PeerId(3)));
+
+    network.send_crust_event(Endpoint(2),
+                             crust::Event::LostPeer(PeerId(tunnel_node_index)));
+    network.send_crust_event(Endpoint(tunnel_node_index),
+                             crust::Event::LostPeer(PeerId(2)));
```
afck: I'd prefer to fully simulate dropping the connection, i.e. making mock Crust remove the connection, too, in addition to sending the `LostPeer` events. (Also below.)

maqi: network doesn't hold active connections, and trying to access node::crust_service to remove the connection is somehow incorrect?

do you mean we have network.lost_connection(client_1, client_2), which internally sends LostPeer to both?

afck: Yes, I think making that a method of the network itself would make sense. But it is not sufficient to just send the `LostPeer` events: it also needs to remove the connection entries from the `ServiceHandle`s, like here: https://github.com/maidsafe/routing/blob/master/src/mock_crust/support.rs#L452
So it probably needs to call `disconnect` on one of the service handles (which will automatically cause a `LostPeer` event in the other one), and then in addition raise the `LostPeer` on that side: https://github.com/maidsafe/routing/blob/master/src/mock_crust/support.rs#L476

Otherwise if the test e.g. tried to re-establish the connection, mock Crust would refuse to, because the connection already exists.

PR #1384, opened 2017-03-03
MAID-1996 avoid tunnelling when proxying

maqi:


maidsafe-highfive:
r? @afck

tests/mock_crust/tunnel.rs
```diff
@@ -270,3 +270,35 @@ fn tunnel_node_blocked() {
...
+
+#[test]
+fn avoid_tunnelling_when_proxying() {
+    let min_section_size = 5;
+    let network = Network::new(min_section_size, None);
+    network.block_connection(Endpoint(2), Endpoint(3));
+    network.block_connection(Endpoint(3), Endpoint(2));
+    let mut nodes = create_connected_nodes(&network, min_section_size);
+    verify_invariant_for_all_nodes(&nodes);
+    assert!(unwrap!(locate_tunnel_node(&nodes, PeerId(2), PeerId(3))) != 0);
```
dhardy: Why `!= 0`? Maybe I'm misreading this, but 0 is the index of the first node, which is what we'd normally use, right?

tests/mock_crust/tunnel.rs
```diff
@@ -270,3 +270,35 @@ fn tunnel_node_blocked() {
...
+        .config(config.clone())
+        .endpoint(endpoint)
+        .cache(false)
+        .create());
+    network.block_connection(Endpoint(nodes.len() - 1), Endpoint(0));
+    network.block_connection(Endpoint(0), Endpoint(nodes.len() - 1));
+    poll_and_resend(&mut nodes, &mut []);
+    verify_invariant_for_all_nodes(&nodes);
+    assert_eq!(nodes.len() - 2,
+               unwrap!(locate_tunnel_node(&nodes, PeerId(0), PeerId(nodes.len() - 1))));
```
dhardy: You could instead write `assert!(locate_tunnel_node(..) == Some(nodes.len() - 2))`

tests/mock_crust/tunnel.rs
maqi: we use node 0 as proxy, and with this change, proxying node shall not be used as tunnel_node any more, so here is to assert this.

tests/mock_crust/tunnel.rs
maqi: just prefer the value of two sides can be printed in case of failure

PR #1386, opened 2017-03-06
fix/peer_manager: use closest to own nodes as candidates for tunnel_node

maqi:


maidsafe-highfive:
r? @afck

src/peer_manager.rs
```diff
@@ -1217,11 +1219,14 @@ impl PeerManager {
 
         let _ = self.insert_peer(pub_id, Some(peer_id), PeerState::SearchingForTunnel);
 
-        let close_section = self.routing_table.other_close_names(pub_id.name()).unwrap_or_default();
+        let close_nodes = self.routing_table
```
Fraser999: Wouldn't it be better to just return our entire section here (excluding ourself) and below filter the `peer_map` for those nodes which are in our section and are direct-connected to us and use `.take(NUM_TUNNEL_VIA_NODES)` there?

maqi: not quite follow that, but do you mean we switch to : 
let close_nodes = self.routing_table.other_close_names(self.public_id.name()) .. ;
then `.take(NUM_TUNNEL_VIA_NODES)` after peer_map check ?
i.e. we will use our_section or part of it if section_size > NUM_TUNNEL_VIA_NODES ?

Fraser999: I was effectively meaning, of all the peers in our section which are direct-connected, take the 10 closest to us.  If we're concerned about always using the same peers, we could shuffle the peers in our section before filtering, but not sure if that's worthwhile or not.

src/peer_manager.rs
```diff
@@ -1200,6 +1202,25 @@ impl PeerManager {
         let _ = self.set_state(peer_id, PeerState::Routing(RoutingConnection::Direct));
     }
 
+    /// Get candidates of tunnel_node for us.
+    ///
+    /// To be qualified: peer must be directly conneted to us
+    ///                  it must be within NUM_TUNNEL_VIA_NODES closest to us
+    pub fn get_tunnel_candidates(&self) -> Vec<(XorName, PeerId)> {
```
Fraser999: This function should get _all_ peers from our section before filtering, then after filtering, just take `NUM_TUNNEL_VIA_NODES` elements.

PR #1387, opened 2017-03-08
MAID-1958 test/tunnel: tunnel node leaving

maqi:


maidsafe-highfive:
r? @afck

PR #1390, opened 2017-03-09
MAID-1997 simulates lost of connection properly in mock_crust tunnel test

maqi:


maidsafe-highfive:
r? @afck

src/mock_crust/support.rs
```diff
@@ -115,6 +115,21 @@ impl Network {
         let _ = imp.blocked_connections.remove(&(sender, receiver));
     }
 
+    /// Simulates the lost of connection.
```
michaelsproul: Just a nitpick: *Simulate the loss of a connection* might be clearer here

src/mock_crust/support.rs
```diff
@@ -115,6 +115,21 @@ impl Network {
...
+    pub fn lost_connection(&self,
+                           node_1: Endpoint,
+                           node_2: Endpoint,
+                           peer_1: PeerId,
+                           peer_2: PeerId) {
```
michaelsproul: It would be safe to extract the endpoint ID from the peer ID, wouldn't it? It seems that peer IDs are always constructed from endpoint IDs in mock_crust land: https://github.com/maidsafe/routing/blob/9db123e7b28acddb53e125376ae314a8b1f582fa/src/mock_crust/support.rs#L239

PR #1391, opened 2017-03-10
MAID-2010: retain tunnel state of peers in unknown_peer map

Fraser999:


maidsafe-highfive:
r? @afck

PR #1392, opened 2017-03-13
fix/tunnel: additional check …

maqi:
when tunnel_client accepting TunnelSuccess and tunnel_node accepting identity message

### DIFF
--- a/src/states/node.rs
+++ b/src/states/node.rs
@@ -482,6 +482,13 @@ impl Node {
                 } else if self.tunnels.has_clients(src, dst) {
                     self.send_or_drop(&dst, bytes, content.priority());
                     Ok(())
+                } else if !self.peer_mgr.can_tunnel_for(&src, &dst) {
+                    debug!("{:?} mistakenly accepted to act as tunnel_node for {:?} - {:?}",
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -482,22 +482,31 @@ impl Node {
...
                     Ok(())
-                } else if self.tunnels.accept_clients(src, dst) {
-                    debug!("{:?} agreed to act as tunnel_node for {:?} - {:?}",
-                           self,
-                           src,
-                           dst);
-                    self.send_direct_message(dst, DirectMessage::TunnelSuccess(src));
-                    self.send_or_drop(&dst, bytes, content.priority());
-                    Ok(())
+                } else if self.peer_mgr.can_tunnel_for(&src, &dst) {
```
afck: If you move the `!self.peer_mgr.can_tunnel_for(&src, &dst)` (negated) part here, you can do without the nested `if`s,
i.e. continue with `else if self.tunnels.accept_clients(src, dst)` below. That might be a bit easier to read.

PR #1393, opened 2017-03-16
PAUSED - fix/node: handle out of sequence OwnSectionMerge

maqi:


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -2203,6 +2229,44 @@ impl Node {
...
+
+                if let Some((_sender_prefix, merge_prefix, sections)) =
+                    self.peer_mgr.should_merge() {
+                    let senders = self.out_of_seq_own_merge
+                        .get(&(merge_prefix, sections.clone()))
+                        .map_or(BTreeSet::new(), |senders| if senders.len() == 2 {
+                            senders.clone()
+                        } else {
+                            BTreeSet::new()
+                        });
```
afck: If `senders` exists and has length `!= 2`, this allocates two new, empty `BTreeSets` just do do nothing. Maybe:
```rust
if let Some(senders) = self.out_of_seq_own_merge
                        .get(&(merge_prefix, sections.clone()))
                        .iter()
                        .filter(|senders| senders.len() == 2)
                        .cloned()
                        .next() {
```

src/states/node.rs
```diff
@@ -103,6 +104,8 @@ pub struct Node {
     tick_timer_token: u64,
     timer: Timer,
     tunnels: Tunnels,
+    /// Key is (dst, sections), value is BTreeSet<src>
+    out_of_seq_own_merge: LruCache<(Prefix<XorName>, SectionMap), BTreeSet<Prefix<XorName>>>,
```
afck: Why is this _indexed_ by the section map? Won't that be different even for a matching pair of `OwnSectionMerge` messages?

maqi: This is to ensure we don't handle outdated record.
the failing test has the same sections for a matching pair of OwnSectionMerge messages.

We can use `dst` only as key, but then probably needs to narrow the time_out

afck: Isn't that just because there are not many sections in the test?
If there are four balanced sections and `00` and `01` are merging, the section map sent by `01` will include `11` but not `10`, and the one sent by `00` will include `10` but not `11`, won't it?
And even _if_ they include the same sections, one half might have already handled some added or lost node, and the other might not, so the members list can disagree?

maqi: yes, you are correct

PR #1394, opened 2017-03-16
fix/routing_table: fix bug in calculating nodes covered by a PrefixSection

Fraser999:


maidsafe-highfive:
r? @afck

src/routing_table/mod.rs
```diff
@@ -806,8 +806,16 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
...
+                        let is_compatible = |(pfx, section)| if prefix.is_compatible(pfx) {
+                            Some(section)
+                        } else {
+                            None
+                        };
+                        return Ok(self.sections
+                            .iter()
+                            .filter_map(is_compatible)
+                            .flat_map(BTreeSet::iter)
+                            .chain(self.our_section.iter().filter(|name| **name != self.our_name))
```
maqi: do we need to change the comment of If the destination is an Authority::PrefixSection part, as now is not limited to our_prefix?

PR #1397, opened 2017-03-17
feat/node: add a cache for merge messages

afck:
Only call the routing table merge methods once the messages from both
sides have accumulated. This also fixes out-of-sequence merge messages
if several merges occur in quick succession.


maidsafe-highfive:

PR #1398, opened 2017-03-17
feat/bootstrapper-kind: keep hold of the bootstrapper kind

ustulation:
This will help us verify that someone who bootstrapped as client-kind does not send client_restriction as false later in ClientIdentify

### DIFF
--- a/src/states/node.rs
+++ b/src/states/node.rs
@@ -19,14 +19,15 @@ use ::QUORUM;
 use ack_manager::{ACK_TIMEOUT_SECS, Ack, AckManager};
 use action::Action;
 use cache::Cache;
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -541,6 +559,31 @@ impl Node {
                 self.handle_section_list_signature(peer_id, section_list, sig)?
             }
             ClientIdentify { ref serialised_public_id, ref signature, client_restriction } => {
+                let drop = match self.bootstrappers.remove(&peer_id) {
+                    Some(kind) => {
+                        if kind == CrustUser::Client && client_restriction ||
+                           kind == CrustUser::Node && !client_restriction {
```
afck: I think `client_restriction` should just be removed instead of checked here.

PR #1399, opened 2017-03-20
1394, 1395, 1397, 1398 and other fixes

afck:


maidsafe-highfive:

PR #1401, opened 2017-03-20
Added a mock crust test for concurrent merge of two sections

Fraser999:


maidsafe-highfive:
r? @afck

PR #1403, opened 2017-03-21
MAID-2000: chore/all: cleanup/improve CI scripts and README

Fraser999:
As per the details mentioned in MAID-2000:

* fixed some outdated URLs
* removed mention of a version of the Contributor Agreement
* improved AppVeyor and Travis scripts
* applied rustfmt
* fixed clippy warnings

This also has the following changes:

* updated Travis script to use current versions of Rust, Clippy and rustfmt
* updated AppVeyor script to use the relocated QA Powershell script

### DIFF
--- a/src/routing_table/mod.rs
+++ b/src/routing_table/mod.rs
@@ -5,8 +5,8 @@
 // licence you accepted on initial access to the Software (the "Licences").
 //
 // By contributing code to the SAFE Network Software, or to this project generally, you agree to be
-// bound by the terms of the MaidSafe Contributor Agreement, version 1.1.  This, along with the
-// Licenses can be found in the root directory of this project at LICENSE, COPYING and CONTRIBUTOR.
+// bound by the terms of the MaidSafe Contributor Agreement.  This, along with the Licenses can be
+// found in the root directory of this project at LICENSE, COPYING and CONTRIBUTOR.
 //
 // Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
 // under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
--- a/src/routing_table/mod.rs
+++ b/src/routing_table/mod.rs
@@ -1171,6 +1171,7 @@ mod tests {
     use super::SPLIT_BUFFER;
     use itertools::Itertools;
     use std::collections::BTreeSet;
+    use std::str::FromStr;
 
     #[test]
     fn small() {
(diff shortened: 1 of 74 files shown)


maidsafe-highfive:
r? @afck

src/routing_table/network_tests.rs
```diff
@@ -318,16 +321,17 @@ pub fn verify_network_invariant<'a, T, U>(nodes: U)
...
+                           section_content,
+                           "Section with prefix {:?} doesn't agree between nodes {:?} and {:?}\n\
+                            {:?}: {:?}, {:?}: {:?}",
+                           prefix,
+                           node.our_name,
+                           src,
+                           node.our_name,
+                           section_content,
+                           src,
+                           section);
```
afck: The two differing sections will probably be printed anyway by `assert_eq!`?

tests/mock_crust/drop.rs
```diff
@@ -34,7 +34,7 @@ fn drop_node(nodes: &mut Vec<TestNode>, index: usize) {
             match node.try_next_ev() {
                 Ok(Event::NodeLost(lost_name, _)) if lost_name == name => break,
                 Ok(_) => (),
-                Err(_) => panic!("Event::NodeLost({:?}) not received", name),
+                _ => panic!("Event::NodeLost({:?}) not received", name),
```
afck: I find `Err(_)` more explicit and easier to read. Did Clippy demand this change?

src/routing_table/prefix.rs
```diff
@@ -224,18 +226,19 @@ impl<T: Clone + Copy + Default + Binary + Xorable> Debug for Prefix<T> {
...
+    fn from_str(bits: &str) -> Result<Prefix<u8>, String> {
         let mut name = 0u8;
         for (i, bit) in bits.chars().enumerate() {
             if bit == '1' {
                 name |= 1 << (7 - i);
             } else if bit != '0' {
-                panic!("'{}' not allowed - the string must represent a binary number.",
-                       bit);
+                return Err(format!("'{}' not allowed - the string must represent a binary number.",
+                                   bit));
```
maqi: is this change demand by clippy? or it is changing to err handling?

src/routing_table/mod.rs
```diff
@@ -1404,20 +1405,23 @@ mod tests {
...
+                   table.add_prefix(unwrap!(Prefix::from_str("01"))));
         assert_eq!(prefixes_from_strs(vec!["1", "00", "01"]), table.prefixes());
         assert_eq!(vec![0xc0, 0xd0, 0xe0, 0xf0u8],
-                   table.add_prefix(Prefix::from_str("111")).into_iter().sorted());
+                   table.add_prefix(unwrap!(Prefix::from_str("111"))).into_iter().sorted());
         assert_eq!(prefixes_from_strs(vec!["110", "111", "10", "0"]),
                    table.prefixes());
-        assert_eq!(Vec::<u8>::new(), table.add_prefix(Prefix::from_str("0")));
+        assert_eq!(Vec::<u8>::new(),
+                   table.add_prefix(unwrap!(Prefix::from_str("0"))));
```
maqi: is this kind of line break enforced by the rustfmt?   :(

tests/mock_crust/drop.rs
maqi: +1

src/routing_table/prefix.rs
Fraser999: Clippy didn't like us naming that `from_str()` without using the trait.  The trait requires a `Result` be returned.

tests/mock_crust/drop.rs
Fraser999: It emitted a warning about `Err(_)` matching all error cases and that being bad practice.

src/routing_table/network_tests.rs
Fraser999: Yes, they would.  To be honest, I hardly ever use the actual contents of the sections shown in that message - I just run the test with logging and have a look at the two RTs indicated by `node.name` and `src`.  I could remove the last four variables from the formatted string if that suits everyone else?

If not, I think we're possibly better keeping the log message as is, (since it attributes the two sections to the relevant owners).  In that case, I can leave the PR as it is now, or revert back to `assert!()` and allow the clippy lint for that function?

src/routing_table/mod.rs
Fraser999: Probably a result of inserting `unwrap!(...)` to accommodate the change in return type of `Prefix::from_str()`.

src/routing_table/network_tests.rs
afck: Makes sense. Let's leave it as is for now.

PR #1404, opened 2017-03-21
fix/tunnels: fix bug when handling tunnel messages

Fraser999:
This fixes a couple of issues in node when handling 'TunnelDirect' messages - one where a tunnel
client would ignore a message intended for it, and another where _any_ 'DirectMessage' being handled
by a tunnel node from a pending tunnel client would cause the node to accept both clients.

Also fixed a todo in tunnels.rs to clean up 'new_clients' when moving a pair to 'clients'.

### DIFF
--- a/src/message_filter.rs
+++ b/src/message_filter.rs
@@ -82,6 +82,14 @@ impl<Message: Hash> MessageFilter<Message> {
         self.entries.iter().any(|entry| entry.hash_code == hash_code)
     }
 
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @afck

afck:
(Not in this PR, but: That message filter should really contain a `HashMap` instead of a `Vec`. It looks for entries by hash all the time!)

PR #1405, opened 2017-03-22
Try more tunnels for peers in own section.

afck:


maidsafe-highfive:

src/peer_manager.rs
```diff
@@ -1217,32 +1217,35 @@ impl PeerManager {
...
+    /// `client_name`'s section. If those sections are the same, `tunnel_name` is also allowed to
+    /// match our sibling prefix instead.
+    pub fn is_potential_tunnel_node(&self, tunnel_name: &XorName, client_name: &XorName) -> bool {
+        if self.our_public_id.name() == tunnel_name || self.our_public_id.name() == client_name ||
+           !self.get_state_by_name(tunnel_name).map_or(false, PeerState::can_tunnel_for) {
+            return false;
+        }
+        let our_prefix = self.routing_table.our_prefix();
+        if our_prefix.matches(client_name) {
+            our_prefix.popped().matches(tunnel_name)
```
maqi: when tunnel_client is in our_section, our_section shall still be used as tunnel_nodes, `sibling section` is just in addition to it, right?

afck: Yes, that's why I'm using `popped()`, and not `sibling()`.

PR #1407, opened 2017-03-22
fix/node: avoid node reconnecting if it's not yet approved

Fraser999:


maidsafe-highfive:
r? @afck

src/peer_manager.rs
```diff
@@ -537,38 +537,51 @@ impl PeerManager {
...
         }
 
-        self.remove_unapproved_candidates(&candidate_name);
-        let mut candidate = Candidate::new(client_auth);
-        candidate.state = CandidateState::Approved;
-        let _ = self.candidates.insert(candidate_name, candidate);
-        trace!("{:?} No candidate with name {:?}", self, candidate_name);
+        trace!("{:?} No candidate with name {}", self, candidate_name);
+        if let Some(client_authority) = client_auth {
+            self.remove_unapproved_candidates(&candidate_name);
```
afck: Shouldn't we do that even if we don't have the client authority?

tests/mock_crust/churn.rs
```diff
@@ -123,6 +126,9 @@ fn random_churn<R: Rng>(rng: &mut R,
         if nodes.len() > 2 * network.min_section_size() {
             let peer_1 = rng.gen_range(1, len);
             let peer_2 = gen_range_except(rng, 1, len, &iter::once(peer_1).collect());
+            debug!("Lost connection between {:?} and {:?}",
```
afck: This message uses debug, and the one below uses display.

PR #1408, opened 2017-03-23
Use a map to avoid iterating in MessageFilter.

afck:


maidsafe-highfive:

src/message_filter.rs
```diff
@@ -56,92 +62,60 @@ impl<Message: Hash> MessageFilter<Message> {
...
-impl TimestampedMessage {
-    pub fn new(hash_code: u64, time_to_live: Duration) -> TimestampedMessage {
-        TimestampedMessage {
-            hash_code: hash_code,
-            expiry_point: SystemTime::now() + time_to_live,
-            count: 1,
+        let now = Instant::now();
+        while let Some((hash_code, time)) = self.timeout_queue.pop_front() {
+            if now.duration_since(time) <= self.time_to_live {
+                self.timeout_queue.push_front((hash_code, time));
```
dhardy: What's the time complexity of `pop_front()` followed by `push_front()`? It may be fine, but would be nice to know whether this can cause reallocations.

src/message_filter.rs
```diff
@@ -56,92 +62,60 @@ impl<Message: Hash> MessageFilter<Message> {
...
-            let mut timestamped_message = self.entries.remove(index);
-            timestamped_message.update_expiry_point(self.time_to_live);
-            let count = timestamped_message.increment_count();
-            self.entries.push(timestamped_message);
-            count
-        } else {
-            self.entries.push(TimestampedMessage::new(hash_code, self.time_to_live));
-            1
+        let now = Instant::now();
+        self.timeout_queue.push_back((hash_code, now));
```
dhardy: I think we only ever use this time to check expiry? In which case it's probably better to insert `now + time_to_live`.

src/message_filter.rs
afck: It should all be `O(1)`, but I changed it to only pop if the entry is expired.

src/message_filter.rs
```diff
@@ -15,10 +15,11 @@
...
 
-use std::collections::hash_map::DefaultHasher;
+use std::collections::{HashMap, VecDeque};
+use std::collections::hash_map::{DefaultHasher, Entry};
 use std::hash::{Hash, Hasher};
 use std::marker::PhantomData;
-use std::time::{Duration, SystemTime};
+use std::time::{Duration, Instant};
 
 fn hash<T: Hash>(t: &T) -> u64 {
```
michaelsproul: Uhhh.. could this be another source of non-determinism @fizyk20 ?

michaelsproul: (unrelated to this PR really)

src/message_filter.rs
```diff
@@ -29,7 +30,11 @@ fn hash<T: Hash>(t: &T) -> u64 {
 /// A time based message filter that takes any generic type as a key and will drop keys after a
 /// time period (LRU Cache pattern).
 pub struct MessageFilter<Message> {
-    entries: Vec<TimestampedMessage>,
+    /// The number of times each message has been received so far, and the expiry timestamp.
+    count: HashMap<u64, (usize, Instant)>,
```
michaelsproul: Why not BTreeMap?

afck: Because `HashMap` is faster and we don't iterate over it, so the order should never matter in this case.

src/message_filter.rs
afck: Yes, but both `Instant` and `SystemTime` break test determinism unless they are mocked.

michaelsproul: Actually, now that I think about it, I don't think it is (as Andreas says, we never iterate through this map).

src/message_filter.rs
```diff
@@ -56,92 +62,58 @@ impl<Message: Hash> MessageFilter<Message> {
...
+        self.count.contains_key(&hash(message))
     }
 
     /// Remove the entry for `message`, regardless of how many times it was previously inserted.
     pub fn remove(&mut self, message: &Message) {
-        let hash_code = hash(message);
-        if let Some(index) = self.entries.iter().position(|t| t.hash_code == hash_code) {
-            let _old_val = self.entries.remove(index);
-        }
+        let _old_val = self.count.remove(&hash(message));
```
maqi: do we also need to cleanup the timeout_queue, or it doesn't matter?

afck: It doesn't matter and my motivation for this PR was exactly to avoid ever having to iterate through all of them.

PR #1410, opened 2017-03-24
Add route manager

dhardy:


maidsafe-highfive:
r? @afck

dhardy:
The stuff about tunnel nodes shouldn't be in `RouteManager`, but somebody wanted to access the routing table. It might be easier to put the methods back in `PeerManager` and pass the routing table in. I only merged the change responsible yesterday and didn't find a good solution yet.

src/peer_manager.rs
```diff
@@ -653,14 +619,8 @@ impl PeerManager {
...
-    /// Tries to add the given peer to the routing table. If successful, this returns `Ok(true)` if
-    /// the addition should cause our section to split or `Ok(false)` if the addition shouldn't
-    /// cause a split.
-    pub fn add_to_routing_table(&mut self,
-                                pub_id: &PublicId,
-                                peer_id: &PeerId,
-                                want_to_merge: bool)
-                                -> Result<bool, RoutingTableError> {
+    /// Called just before `RouteManager::add_to_routing_table()`.
+    pub fn pre_add_to_table(&mut self, peer_id: &PeerId) {
```
afck: Does `self` need to be mutable here?
If this method doesn't do anything anyway, I'd rather just include these log messages in `set_to_full_node`.

maqi:
Yes, I also feel it's not a good sign of passing peer_manager into route_manager.
Those two functions, `pre_add_to_table` and `set_to_full_node` of peer_manager, is also not a good sign (or at least not a good naming).

dhardy:
@maqi: agreed. But if you think about it, many functions would be a lot cleaner if we just merged all of `peer_manager` into `states::node`. Modularising code usually has some drawbacks.

As to the function names and all the passing, I was trying to find a quick solution which minimises changes, not a "best code" solution.

PR #1411, opened 2017-03-24
fix/node: send section updates even if RT not valid if accumulated enough requests

Fraser999:
This adds an accumulator for 'SectionUpdateRequest' messages to reduce the chance of a single peer
triggering a section-wide response.  Once three such requests are accumulated, the recipients send
a 'SectionUpdate' even if their routing tables show that the invariant is broken.

It also changes how 'SectionUpdates' are handled in that we'll only split neighbouring sections,
but will merge neighbours or our own section.

### DIFF
--- a/src/states/node.rs
+++ b/src/states/node.rs
@@ -210,7 +210,7 @@ impl Node {
            full_id: FullId,
            min_section_size: usize,
            stats: Stats,
-           mut timer: Timer)
+           timer: Timer)
            -> Self {
         let public_id = *full_id.public_id();
         let tick_period = Duration::from_secs(TICK_TIMEOUT_SECS);
@@ -318,7 +318,13 @@ impl Node {
             Action::ClientSendRequest { result_tx, .. } => {
                 let _ = result_tx.send(Err(InterfaceError::InvalidState));
             }
-            Action::NodeSendMessage { src, dst, content, priority, result_tx } => {
+            Action::NodeSendMessage {
+                src,
+                dst,
+                content,
+                priority,
+                result_tx,
+            } => {
                 let result = match self.send_user_message(src, dst, content, priority) {
                     Err(RoutingError::Interface(err)) => Err(err),
                     Err(_) | Ok(()) => Ok(()),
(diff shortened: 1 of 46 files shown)


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -2127,8 +2143,17 @@ impl Node {
...
+                if rt_pfx.bit_count() >= prefix.bit_count() {
+                    break;
+                }
+                debug!("{:?} Splitting {:?} on section update.", self, rt_pfx);
+                let _ = self.handle_section_split(rt_pfx, rt_pfx.lower_bound(), outbox);
+            }
+        }
+
+        if let Some(rt_pfx) = self.peer_mgr.routing_table().find_section_prefix(&pfx_name) {
             if merge && rt_pfx.bit_count() > prefix.bit_count() {
```
maqi: so we won't hit the case of missing more than one level of merge?
say we are still in 000, and knows 001 and 01, and a SectionUpdate from 0 arrived.
the previous code will make us merge into 0, but now we will only merge into 00.

Fraser999: The previous code wouldn't do that for merges, only splits.  I didn't want to change the existing functionality, but we did notice this and wondered if it might have been an oversight.

src/states/node.rs
```diff
@@ -1574,9 +1583,10 @@ impl Node {
     // `SectionUpdateRequest` from a neighbouring section) and it's only used to add missing members
     // or split a section.
     fn send_section_update(&mut self, dst_prefix: Option<Prefix<XorName>>) {
-        if !self.peer_mgr.routing_table().is_valid() {
-            warn!("{:?} Not sending section update since RT invariant not held.",
-                  self);
+        if dst_prefix.is_none() && !self.peer_mgr.routing_table().is_valid() {
+            warn!("{:?} Not sending section update since RT invariant not held.\n{:?}",
+                  self,
+                  self.peer_mgr.routing_table());
```
Fraser999: This was left in by mistake - I'll remove it!

src/states/node.rs
```diff
@@ -2500,18 +2556,24 @@ impl Node {
...
                 targets,
-                merge_details,
+                mut merge_details,
             } => {
                 // TODO - the event should maybe only fire once all new connections have been made?
                 outbox.send_event(Event::SectionMerge(merge_details.prefix));
                 info!("{:?} Own section merge completed. Prefixes: {:?}",
                       self,
-                      self.peer_mgr.routing_table().prefixes());
+                      self.peer_mgr.routing_table().our_section());
```
afck: Shouldn't this keep printing the prefixes?

PR #1412, opened 2017-03-27
feat/timer: make Timer cloneable

dhardy:
Side effect: worker is no longer joined on drop (though it still stops in its own time).

Motivation: this would allow a cheap clone of Timer to be kept in other locations where timers may want to be scheduled (i.e. in the new consensus code, though there may be other uses too).

Pain point: if the arc reference could be decremented in `Timer::drop` we could just derive `Clone` and remove `use_count`. We could do that anyway, but the worker thread _might_ be woken before the other `Arc` count has been decremented, sleep again, then never finish (unless it is woken automatically by something else, e.g. the main thread terminating — `Condvar` isn't clear on this). Of course it probably doesn't matter except for minor memory/resource leakage if the node is restarted in the same process and uses a new timer.

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -21,48 +21,52 @@ pub use self::implementation::Timer;
 mod implementation {
     use action::Action;
     use itertools::Itertools;
-    use maidsafe_utilities::thread::{self, Joiner};
+    use maidsafe_utilities::thread;
     use std::collections::BTreeMap;
     use std::sync::{Arc, Condvar, Mutex};
     use std::time::{Duration, Instant};
     use types::RoutingActionSender;
 
     struct Detail {
+        next_token: u64,
         deadlines: BTreeMap<Instant, Vec<u64>>,
-        cancelled: bool,
+        use_count: u32,
     }
 
-    /// Simple timer.
+    /// A device for scheduling wake-ups, delivered via a message sender. This device can be
+    /// cloned with all wakeups sent via the same sender.
+    ///
+    /// The worker thread terminates when all associated `Timer` devices are destroyed.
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @afck

src/timer.rs
```diff
@@ -21,48 +21,52 @@ pub use self::implementation::Timer;
...
+
             let worker = thread::named("Timer", move || Self::run(sender, detail_and_cond_var));
-            Timer {
-                next_token: 0,
-                detail_and_cond_var: detail_and_cond_var_clone,
-                _worker: worker,
-            }
+            // TODO: confirm whether it is reasonable to disown the worker like this. Rationale is
+            // that it allows `Timer` to be cloned without a separate `Joiner`.
+            worker.detach();
```
ustulation: I am against this

dhardy: I don't see much reason to be; the thread should still clean itself up. It could also be fixed by returning a second object (the joiner). But I can't see it being a problem unless new timers get created faster than the old threads join... and we only create one timer anyway.

ustulation: That is not the point -  it's just bad coding style. Just because it can be proved that a heap allocation of a raw ptr is being properly freed in all places doesn't mean one should use it (ownerless) instead of using a shared-ptr with ref count (shared owner) or a unique_ptr (Box) (one owner).

It's a maintainence hassle and bad coding in general. Also mentioned here: https://github.com/maidsafe/QA/blob/20110d496e1cfece1bfe80018f456aebf7ab511c/Documentation/Rust%20Style.md#threads

PR #1413, opened 2017-03-27
refactor/timer: make timer clonable

ustulation:
Improve timer maintainence by removing manual work which uses cond-var and drop impl. Use std lib features instead and RAII objects (so we don't have to implement drop manually)

### DIFF
--- a/examples/ci_test.rs
+++ b/examples/ci_test.rs
@@ -216,7 +216,8 @@ mod unnamed {
         let mut term = term::stdout().expect("Could not open stdout.");
         term.fg(color).expect("Failed to set color");
         print!("{}", text);
-        term.reset().expect("Failed to restore stdout attributes.");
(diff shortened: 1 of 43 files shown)


maidsafe-highfive:
r? @afck

ustulation:
This PR assumes timer is to be used only in a single threaded context. However the way client.rs::make_state_machine is implemented it creates it in a different thread and then moves it into a different thread. So either that is a design issue there or if that is right and it needs to be moved from one thread to another then https://github.com/maidsafe/routing/pull/1413/files#diff-dde82bd7fc015253d6e2e7f289dec7baR41 must be changed to `Arc<Mutex` instead of `Rc<RefCell`. Can someone pls comment which is the one intended (or pull the PR and modify the above to make PR builld).

PR #1415, opened 2017-03-30
purge out_of_sync peers; on receiving ConnectSuccess, update Routing(tunnel) to Routing(direct) directly

maqi:


maidsafe-highfive:
r? @afck

src/peer_manager.rs
```diff
@@ -1237,44 +1246,49 @@ impl PeerManager {
...
-        self.routing_table
-            .iter()
-            .filter_map(|name| -> Option<(PeerId, XorName, bool)> {
-                let peer = match self.peer_map.get_by_name(name) {
-                    Some(peer) => peer,
-                    None => {
-                        error!("{:?} Have {} in RT, but have no entry in peer_map for it.",
+    /// Returns all syncing peer's `PeerId`s, names and whether connected via a tunnel or not;
+    /// together with all out-of-sync peer's names and peer_ids if have.
+    pub fn get_routing_peer_details(&self) -> PeerDetails {
```
afck: Let's split this up into two methods, one returning the known peers and one returning the unknown routing table names. Let's try to avoid making this epic inline closure even more complex. 😬

maqi: if split into two methods, the code will be quite duplicated as need to go through almost the same checks

afck: Hmm… you're right, let's leave it in one method.

src/peer_manager.rs
```diff
@@ -1528,6 +1544,14 @@ impl PeerManager {
         }
     }
 
+    /// Removes the given entry from the routing_table or peer_map, returns the removal details
+    pub fn purge_out_of_sync_peer(&mut self,
+                                  name: &XorName)
+                                  -> Result<RemovalDetails<XorName>, RoutingTableError> {
+        let _ = self.peer_map.remove_by_name(name);
```
maqi: this is purging the peers which are out_of_sync between peer_man and RT.
hence the state can be not only routing? (like the case we encountered, the state is in AwaitingIdentify)

afck: Yes, why not do the cleanup inside `get_routing_peer_details` (which should probably get a more descriptive name anyway) and only pass the removal details back to `Node`, so that we don't have to call another peer manager method from there to actually remove that peer.

maqi: will have a check tomorrow whether we can do that without expanding that epic inline closure :)

src/node.rs
```diff
@@ -468,6 +468,11 @@ impl Node {
         self.machine.current().has_unacknowledged()
     }
 
+    /// Are there any invalid routing entry?
```
afck: `entry` → `entries`

src/section_list_cache.rs
```diff
@@ -42,19 +42,25 @@ impl SectionListCache {
...
+        let pub_id_opt = self.signed_by
+            .keys()
+            .find(|pub_id| name == pub_id.name())
+            .cloned();
+        if let Some(pub_id) = pub_id_opt {
+            if let Some(lists) = self.signed_by.remove(&pub_id) {
+                for (prefix, list) in lists {
+                    let _ = self.signatures
+                        .get_mut(&prefix)
+                        .map_or(None, |map| {
```
afck: This should probably be `and_then`. (Also below.)

src/peer_manager.rs
```diff
@@ -1148,6 +1150,13 @@ impl PeerManager {
 
     /// Marks the given peer as "connected and waiting for `NodeIdentify`".
     pub fn connected_to(&mut self, peer_id: &PeerId) {
+        // ConnectSuccess may received after established tunnel to peer (and inserted into RT)
```
dhardy: ConnectSuccess may be received after establishing a tunnel to peer (and adding to RT)

src/states/node.rs
```diff
@@ -2749,9 +2750,21 @@ impl Node {
 
     // Drop peers to which we think we have a direct or tunnel connection, but where Crust reports
     // that we're not connected to the peer or tunnel node respectively.
-    fn purge_invalid_rt_entries(&mut self, outbox: &mut EventBox) -> Transition {
+    fn purge_invalid_rt_entries(&mut self, outbox: &mut EventBox) -> (Transition, bool) {
+        let (known_peers, out_of_syc_peers, removal_details) = self.peer_mgr
```
dhardy: syc → sync

src/macros.rs
```diff
@@ -0,0 +1,31 @@
...
+/// will simply log the message at the requested level.
+///
+/// Example usage:
+/// log_or_panic!(warn, "{:?} Bad value: {}", self, value);
+macro_rules! log_or_panic {
+    ($log_level:ident, $($arg:tt)*) => {
+        if cfg!(feature = "use-mock-crust") {
+            panic!($($arg)*);
+        } else {
+            $log_level!("{}", format!($($arg)*));
```
dhardy: I'd suggest you use the `log` macro: `log!($log_level, $($arg)*)`.

https://doc.rust-lang.org/log/log/macro.log.html

PR #1416, opened 2017-03-30
test/tunnel: expand unidirectional block test

maqi:


maidsafe-highfive:
r? @afck

tests/mock_crust/tunnel.rs
```diff
@@ -51,11 +51,11 @@ fn failing_connections_bidirectional() {
...
 #[test]
 fn failing_connections_unidirectional() {
-    let min_section_size = 8;
+    let min_section_size = 10;
     let network = Network::new(min_section_size, None);
-    network.block_connection(Endpoint(1), Endpoint(6));
-    network.block_connection(Endpoint(1), Endpoint(7));
-    network.block_connection(Endpoint(6), Endpoint(7));
+    for i in 3..min_section_size {
+        network.block_connection(Endpoint(2), Endpoint(i));
```
afck: Doesn't this block all packets from `2` to `i`, completely preventing a direct connection?
Maybe it would be better to update mock Crust, such that it only means that `2`'s connect attempt will fail, but `i`'s attempt will succeed.

maqi: you mean only block certain crust message?
need to think about that

afck: Maybe it should only block `Packet::ConnectRequest`, yes.

maqi: ok, I will give it a try tomorrow

PR #1420, opened 2017-04-03
feat/client_error: add invitation errors

afck:


maidsafe-highfive:

PR #1422, opened 2017-04-04
Timer heavy duty test

maqi:


maidsafe-highfive:
r? @afck

src/timer.rs
```diff
@@ -209,6 +209,23 @@ mod implementation {
...
+            let sender = RoutingActionSender::new(action_sender,
+                                                  routing_event_category,
+                                                  category_sender.clone());
+            {
+                let timer = Timer::new(sender);
+
+                for _ in 0..1000 {
+                    let _ = timer.schedule(Duration::new(0, 3000));
+                }
+            }
```
afck: Is that pair of curly braces necessary?

PR #1423, opened 2017-04-04
test/connections_lost_and_unidirectional_block

maqi:


maidsafe-highfive:
r? @afck

src/mock_crust/support.rs
```diff
@@ -98,9 +100,28 @@ impl Network {
...
+        if !handled {
+            while let Some((sender, receiver, packet)) = self.pop_delayed_packet() {
+                self.process_packet(sender, receiver, packet);
+            }
+        }
+    }
+
+    fn pop_delayed_packet(&self) -> Option<(Endpoint, Endpoint, Packet)> {
+        let mut imp = self.0.borrow_mut();
+        imp.delayed_packets.pop_front()
```
afck: I think it would be way less code to just modify `pop_packet` instead, and turn
```rust
let keys: Vec<_> = network_impl.queue.keys().cloned().collect();
```
into something like:
```rust
let keys: Vec<_> = if network_impl.queue.keys().all(|&(ref s, ref r)| self.connection_blocked(s, r)) {
    network_impl.queue.keys().cloned().collect()
} else {
    network_impl.queue.keys().filter(|&(ref s, ref r)| !self.connection_blocked(s, r)).cloned().collect()
};
```
(Or is it important that packets other than `ConnectRequest` are dropped?)

PR #1425, opened 2017-04-05
fix/peer_manager: also consider RT omissions in sync check

afck:


maidsafe-highfive:

PR #1428, opened 2017-04-10
Section versions

afck:


maidsafe-highfive:

src/messages.rs
```diff
@@ -657,7 +636,7 @@ pub enum MessageContent {
     /// of the merge.
     ///
     /// The source authority is a `PrefixSection` conveying the section which just merged.
-    OtherSectionMerge(BTreeSet<PublicId>),
+    OtherSectionMerge(BTreeSet<PublicId>, u64),
```
maqi: a comment stating `u64` is for version?

src/routing_table/mod.rs
```diff
@@ -227,15 +198,17 @@ pub enum OwnMergeState<T: Binary + Clone + Copy + Default + Hash + Xorable> {
...
     our_name: T,
-    // Prefix of our section
+    /// Prefix of our section
     our_prefix: Prefix<T>,
-    // Members of our section, including our own name
+    /// Members of our section, including our own name
     our_section: BTreeSet<T>,
-    // Other sections (excludes our own) (TODO: rename)
+    /// Our prefix version.
+    our_version: u64,
```
maqi: just wondering why don't have `our_versioned_prefix`?

src/routing_table/mod.rs
```diff
@@ -530,39 +508,96 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
...
+        let (prefix, version) = ver_pfx.into();
         if prefix == self.our_prefix {
-            result = self.split_our_section();
+            result = self.split_our_section(version);
             return (result, Some(self.our_prefix));
         }
 
-        if let Some(to_split) = self.sections.remove(&prefix) {
+        if let Some((v, to_split)) = self.sections.remove(&prefix) {
+            if v != version {
```
maqi: not quite sure about this, the version check was carried out after the removal?

src/states/node.rs
```diff
@@ -392,6 +391,11 @@ impl Node {
         Transition::Stay
     }
 
+    /// Routing table of this node.
+    pub fn routing_table(&self) -> &RoutingTable<XorName> {
+        self.peer_mgr.routing_table()
+    }
```
dhardy: This should have been done a long time ago!

src/routing_table/mod.rs
afck: You're right, that's a bit convoluted. I refactored it somewhat.

src/peer_manager.rs
```diff
@@ -758,15 +752,52 @@ impl PeerManager {
                pub_id.name(),
                PeerState::Routing(conn));
 
-        res
+        Ok(())
+    }
+
+    fn remove_name(&mut self, name: XorName) -> Option<(XorName, PeerId)> {
```
afck: Done.

src/routing_table/mod.rs
afck: I tried it and it seemed to make the code more complicated overall.
Maybe I'll give it another go, but I'd rather leave that for a separate PR.

src/states/node.rs
```diff
@@ -3355,10 +3142,12 @@ impl Node {
 
     fn send_other_section_merge(&mut self,
                                 targets: BTreeSet<Prefix<XorName>>,
-                                merge_details: OtherMergeDetails<XorName>) {
-        let section = self.peer_mgr.get_pub_ids(&merge_details.section);
-        let content = MessageContent::OtherSectionMerge(section);
-        let src = Authority::PrefixSection(merge_details.prefix);
+                                ver_pfx: VersionedPrefix<XorName>,
+                                section: BTreeSet<XorName>) {
+        let section = self.peer_mgr.get_pub_ids(&section);
```
Fraser999: `section` is shadowing here now.

src/states/node.rs
```diff
@@ -2541,15 +2370,22 @@ impl Node {
             let our_merged_section: BTreeSet<_> = our_sections
                 .iter()
                 .chain(their_sections.iter())
-                .filter(|&(prefix, _)| prefix.popped() == self.our_prefix().popped())
+                .filter(|&(ver_pfx, _)| ver_pfx.prefix().is_extension_of(&merge_prefix))
                 .flat_map(|(_, peers)| peers)
                 .map(|peer| *peer.name())
                 .collect();
-            self.process_own_section_merge(our_prefix, our_sections, &our_merged_section, outbox)?;
+            let version = our_sections
```
Fraser999: To avoid doing more or less the same iteration/filter twice, I'd have written this as:

```rust
let mut version = 1;
let mut our_merged_section = BTreeSet::new();
for (ver_pfx, peers) in
    our_sections
        .iter()
        .chain(their_sections.iter())
        .filter(|&(ver_pfx, _)| ver_pfx.prefix().extends(&merge_prefix)) {
    version = cmp::max(version, ver_pfx.version() + 1);
    our_merged_section.append(&mut peers
                                       .into_iter()
                                       .map(|peer| *peer.name())
                                       .collect());
}
```
But that's probably not any clearer, so feel free to ignore :)

src/routing_table/mod.rs
```diff
@@ -922,93 +967,77 @@ impl<T: Binary + Clone + Copy + Debug + Default + Hash + Xorable> RoutingTable<T
...
+                let _section_ref = entry.insert((version, section));
             }
             Entry::Occupied(entry) => {
                 error!("{:?} Inserting section {:?}, but already has members {:?}. This is a bug!",
                        self.our_name,
                        prefix,
                        entry.get());
-                entry.into_mut().extend(section);
+                let &mut (ref mut v, ref mut s) = entry.into_mut();
+                if *v > version {
```
afck: I'm not sure about this. I'd rather not rely on the current set of call sites.
And the problem with potentially re-adding a dropped peer is probably not new with this PR? I don't know what's the best solution here…

src/routing_table/prefix.rs
```diff
@@ -73,6 +73,12 @@ impl<T: Clone + Copy + Default + Binary + Xorable> Prefix<T> {
         i >= self.bit_count() || i >= other.bit_count()
     }
 
+    /// Returns `true` if `other` is compatible but strictly shorter than `self`.
+    pub fn extends(&self, other: &Prefix<T>) -> bool {
```
michaelsproul: I think it's weird too, I would have thought `is_prefix_of` with the arguments switched would be the most natural.

afck: Ideally maybe the type should be called `BitSeq` or something, then I'd be happy with `is_prefix_of`.
But I think it's weird to say that a prefix is a prefix of a prefix…

PR #1429, opened 2017-04-12
Name = Hash(Key)

michaelsproul:


maidsafe-highfive:
r? @afck

Viv-Rajkumar:
Closing this so we can deal with this as detailed from https://maidsafe.atlassian.net/browse/MAID-2049

PR #1433, opened 2017-04-17
feat/quorum: allow fractional quorum; use strict greater than

dhardy:
**This requires discussion**

@dirvine requested that we use strict > 50% for quorum, as in his RFC. This implements that, while also making it easy to use exact fractions such as 2/3.

Further changes may be needed for node ageing; however they may not touch exactly the same code; quorum may then be calculated the same way but by weighting votes and voters first.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -193,8 +193,12 @@ pub const TYPE_TAG_SESSION_PACKET: u64 = 0;
 /// Structured Data Tag for DNS Packet Type
 pub const TYPE_TAG_DNS_PACKET: u64 = 5;
 
-/// The quorum, as a percentage of the number of members of the authority.
-pub const QUORUM: usize = 51;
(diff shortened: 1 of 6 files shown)


maidsafe-highfive:
r? @afck

src/lib.rs
```diff
@@ -193,8 +193,12 @@ pub const TYPE_TAG_SESSION_PACKET: u64 = 0;
 /// Structured Data Tag for DNS Packet Type
 pub const TYPE_TAG_DNS_PACKET: u64 = 5;
 
-/// The quorum, as a percentage of the number of members of the authority.
-pub const QUORUM: usize = 51;
+/// Quorum is defined as having strictly greater than `QUORUM_NUM / QUORUM_DENOM` agreement;
+/// using only integer arithmatic a quorum can be checked with
```
afck: Typo: `arithmetic`

Fraser999:
Closed in favour of #1440.

PR #1435, opened 2017-04-19
MAID-2049: feat/relocation: implement new relocation mechanism

Fraser999:
This takes many changes from PR #1429 by @michaelsproul which was closed in favour of this approach.

The main changes include:
* the addition of a new state 'JoiningNode'
* the removal of the ability to mutate a 'PublicId'
* changes to messages and internal states to allow a relocating node to use two different IDs
* changes to the state machine to allow a 'JoiningNode' to transition back into 'Bootstrapping'


maidsafe-highfive:
r? @afck

src/node.rs
```diff
@@ -511,16 +504,23 @@ impl Node {
...
+            .set_next_relocation_interval(interval)
     }
 
     /// Clears the name to be used when the next node relocation request is received by this node so
     /// the normal process is followed to calculate the relocated name.
-    pub fn clear_next_node_name(&mut self) {
-        self.machine.current_mut().set_next_node_name(None)
+    pub fn clear_next_relocation_dst(&mut self) {
+        self.machine.current_mut().set_next_relocation_dst(None)
     }
```
maqi: do we need a function of `clear_next_relocation_interval` as well?

src/id.rs
```diff
@@ -51,6 +51,25 @@ impl FullId {
         }
     }
 
+    /// Construct a `FullId` whose name is in the interval [start, end] (both endpoint inclusive).
```
afck: Typo: `endpoints`

src/peer_manager.rs
```diff
@@ -406,8 +420,8 @@ pub struct PeerManager {
     proxy_peer_id: Option<PeerId>,
     routing_table: RoutingTable<XorName>,
     our_public_id: PublicId,
-    /// Joining nodes which want to join our section
-    candidates: HashMap<XorName, Candidate>,
+    /// Relocating nodes which want to join our section
```
maqi: better comment this is indexing by the candidates' `old` public_id

src/peer_manager.rs
```diff
@@ -462,53 +461,72 @@ impl PeerManager {
...
                    self,
-                   candidate_name,
-                   ongoing_name);
+                   old_pub_id.name(),
+                   ongoing_pub_id.name(),
+                   candidate.fmt_new_name());
             return Err(RoutingError::AlreadyHandlingJoinRequest);
         }
-        self.routing_table
-            .should_join_our_section(&candidate_name)?;
```
maqi: first, why no longer carry out this check?
second, seems `should_join_our_section` is no longer being used, shall it be removed?

src/id.rs
```diff
@@ -51,6 +51,25 @@ impl FullId {
...
+    /// FIXME(Fraser) - time limit this function?  Document behaviour
```
afck: And a duplicated space. 😛

src/peer_manager.rs
```diff
@@ -536,28 +554,39 @@ impl PeerManager {
...
+                           .get_by_name(new_pub_id.name())
+                           .map_or(Err(RoutingError::UnknownCandidate), |_peer| {
+                    Ok((MessageContent::CandidateApproval {
+                            old_public_id: *old_pub_id,
+                            new_public_id: *new_pub_id,
+                            new_client_auth: *new_client_auth,
+                            sections: self.pub_ids_by_section(),
+                        },
+                        *new_pub_id.name()))
+                })
```
maqi: can we not to put this competed stuff inside a return clause?
also seems the result from rustfmt is not good.  :(

src/id.rs
michaelsproul: This was me :P

src/node.rs
Fraser999: Not at the moment.  I guess it's easy enough to add in later if required.

src/peer_manager.rs
Fraser999: The check can't be carried out in exactly the same way as before since we now don't know what the relocated ID will be at this point - the relocating node has still to be told its target range, and this message is sent by the members of `X` who never get told what the new ID will be anyway.

But you're right about the RT function needing removed :)

src/xor_name.rs
```diff
@@ -98,10 +99,39 @@ impl XorName {
...
     fn get_debug_id(&self) -> String {
         format!("{:02x}{:02x}{:02x}..", self.0[0], self.0[1], self.0[2])
     }
+
+    /// Used to construct an XorName from a `BigUint`. This will panic if `value` represents a
+    /// number greater than or equal to `2^XOR_NAME_BITS`.
+    fn from_big_uint(value: BigUint) -> XorName {
+        let little_endian_value = value.to_bytes_le();
+        assert!(little_endian_value.len() <= XOR_NAME_LEN,
+                "This BigUint value exceeds the maximum capable of being held as an XorName.");
```
afck: We could just wrap, i.e. ignore the most significant bits here instead of panicking.

Fraser999: We could, but it would be an error in the way we're using it currently.  I think an `assert!` in this case is reasonable as we already get similar behaviour if we e.g. create an underflow or try to divide by 0.  If we swallow this error and return a probably invalid result, should we also do that for subtraction and division?

src/id.rs
```diff
@@ -111,6 +112,14 @@ impl Debug for PublicId {
     }
 }
 
+impl Deserialize for PublicId {
+    fn deserialize<D: Deserializer>(deserialiser: D) -> Result<Self, D::Error> {
+        let (public_encrypt_key, public_sign_key): (box_::PublicKey, sign::PublicKey) =
+            Deserialize::deserialize(deserialiser)?;
```
afck: Does Serde guarantee that this tuple's serialisation will always be identical to the struct's?

Fraser999: Not that I know of; I'm quite uncomfortable with this too.  I opted for this approach since we're doing something similar in `ImmutableData` and `PrivAppendedData`, although those don't depend on the behaviour of handling tuples.  Instead they depend on the encoded format not adding anything to the member fields.

I created [a repo](https://gitlab.com/Fraser999/Public-ID-Deserialisation#public-id-deserialisation) earlier to test out what options we have.  My preferred choice would be option 2 from there, but I haven't checked to see if that works with serde v0.9 yet.  I was hoping to get a chance to discuss this with the team soon.

afck: Option 3 also looks straightforward and robust to me.

Fraser999: That would be my second choice, and does work with serde v0.9 :)

PR #1437, opened 2017-04-20
Validate routing table nodes.

Fraser999:


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -1527,6 +1536,39 @@ impl Node {
...
+        if !peer_found {
+            // These are peers we do not have in our peer manager yet. Could be because of a
+            // disconnect and reconnect via tunnel. We thus add them to the peer manager and
+            // indicate invalid to wait for group approval before moving to RT
+            self.peer_mgr
+                .insert_pending_approval_node(*peer_id, *public_id, is_tunnel);
+            return;
+        }
+
+        self.peer_mgr.set_peer_id(public_id.name(), *peer_id);
```
afck: If the name doesn't exist in the peer manager yet, this does nothing and returns `false`, but the result is ignored. How is the association between that name and peer ID stored, so that when it does get accepted, the peer is added?

src/states/node.rs
```diff
@@ -2825,22 +2890,55 @@ impl Node {
...
+            .get_peer_by_name(&their_name)
+            .and_then(|peer| peer.peer_id())
+            .and_then(|peer_id| if let Some(&PeerState::AwaitingNodeIdentify(_)) =
+                self.peer_mgr.get_state_by_name(&their_name) {
+                          Some(*peer_id)
+                      } else if self.peer_mgr.unknown_peers().contains_key(peer_id) {
+                Some(*peer_id)
+            } else {
+                None
+            });
```
afck: I guess if you change that indentation, rustfmt will "fix" it? 😬

src/states/node.rs
Viv-Rajkumar: if the peer by name doesnt exist in peer_manager then we would have returned three lines above right? would have been better if this was in L# 1546 but the borrow checker would complain I guess? or make it part of an else block in L# 1564 which is essentially same as what it is now.

afck: Ah, right, makes sense! 👍 
(I guess I'd find it slightly less confusing if it were in an `else` block, yes, but I'm fine with this, too.)

src/states/node.rs
Fraser999: Yes, it does.

PR #1438, opened 2017-04-21
Maid 2051 mock_crust_tests use fake clock

maqi:


maidsafe-highfive:
r? @afck

tests/mock_crust/utils.rs
```diff
@@ -298,14 +299,15 @@ pub fn poll_and_resend(nodes: &mut [TestNode], clients: &mut [TestClient]) {
...
                 call_count += 1;
                 assert_ne!(call_count,
                            MAX_POLL_CALLS,
-                           "Polling and resending unacknowledged has been called {} times.",
+                           "Polling and advance time has been called {} times.",
                            MAX_POLL_CALLS);
             }
-            nodes.iter_mut().foreach(|node| node.inner.clear_state());
+            // ACK_TIMEOUT_SECS = 20s
+            Instant::advance_time(20 * 1000);
```
Viv-Rajkumar: why not just leave this as `FakeClock`?

src/states/node.rs
```diff
@@ -3129,32 +3144,13 @@ impl Node {
...
     }
 
-    pub fn clear_state(&mut self) {
-        self.ack_mgr.clear();
-        self.routing_msg_filter.clear();
-        if self.peer_mgr.remove_connecting_peers() {
-            self.merge_if_necessary();
-        }
-        self.tunnels.clear_new_clients();
+    pub fn poll(&mut self) -> Vec<u64> {
```
ustulation: This function was very hard to guess till i saw the implementation. Maybe call it `poll_timer` or `poll_for_expired_timer_token` or something.

src/states/client.rs
```diff
@@ -388,18 +388,8 @@ impl Bootstrapped for Client {
...
-        for timer_token in &timer_tokens {
-            self.resend_unacknowledged_timed_out_msgs(*timer_token);
-        }
-        !timer_tokens.is_empty()
-    }
-
-    /// Are there any unacknowledged messages?
-    pub fn has_unacknowledged(&self) -> bool {
-        self.ack_mgr.has_pending()
+    pub fn poll(&mut self) -> Vec<u64> {
```
ustulation: pls see the the review at https://github.com/maidsafe/routing/pull/1438/files#diff-c9f7984a0a4f0ccaf63472cce67a9e67R3152

src/state_machine.rs
```diff
@@ -188,6 +169,14 @@ impl State {
             state.set_next_node_name(relocation_name);
         }
     }
+
+    pub fn poll(&mut self) -> Vec<u64> {
```
ustulation: pls see the review at https://github.com/maidsafe/routing/pull/1438/files#diff-c9f7984a0a4f0ccaf63472cce67a9e67R3152

src/state_machine.rs
```diff
@@ -322,6 +351,36 @@ impl StateMachine {
         }
     }
 
+    /// Query for a result, or yield: Err(NothingAvailable), Err(Disconnected).
+    #[cfg(feature = "use-mock-crust")]
+    pub fn try_step(&mut self, outbox: &mut EventBox) -> Result<(), TryRecvError> {
```
ustulation: This function just seems to be adding events in sequentiial order with timer ones always being at the end. I thought it was specifically decided not to do this, or has the requirement changed ?

src/timer.rs
```diff
@@ -228,25 +228,68 @@ mod implementation {
...
-    // The mock timer currently never raises timeout events.
+    struct Inner {
+        next_token: u64,
+        deadlines: BTreeMap<Instant, Vec<u64>>,
+    }
+
     #[derive(Clone)]
     pub struct Timer {
-        next_token: Cell<u64>,
+        inner: Rc<RefCell<Inner>>,
```
afck: Shouldn't a `Cell` still suffice here?

src/state_machine.rs
maqi: it's just picking random position may cause non-deterministic behaviour across execution, so here just append by order.

tests/mock_crust/churn.rs
```diff
@@ -345,13 +346,6 @@ fn send_and_receive<R: Rng>(rng: &mut R,
     poll_and_resend(nodes, &mut []);
 
     expected_gets.verify(nodes, &mut []);
-
-    // Every few iterations, clear the nodes' caches, simulating a longer time between events.
-    if rng.gen_weighted_bool(5) {
-        for node in nodes {
-            node.inner.clear_state();
-        }
-    }
```
afck: Should we instead advance the time by a random duration?

src/states/node.rs
```diff
@@ -2411,8 +2411,13 @@ impl Node {
...
         if self.tick_timer_token == token {
-            let tick_period = Duration::from_secs(TICK_TIMEOUT_SECS);
-            self.tick_timer_token = self.timer.schedule(tick_period);
+            if cfg!(feature = "use-mock-crust") {
+                trace!("{:?} not to carry out ticking during mock_crust test.",
+                       self);
+            } else {
+                let tick_period = Duration::from_secs(TICK_TIMEOUT_SECS);
+                self.tick_timer_token = self.timer.schedule(tick_period);
+            }
```
maqi: the problem of simulating ticking is that we don't know when to stop. it will be easily that couple of nodes ticking at the different pace and keeps the polling busy forever.  :(

src/state_machine.rs
```diff
@@ -322,6 +351,36 @@ impl StateMachine {
...
+            }
+
+            let timed_out_events = self.state
+                .poll()
+                .iter()
+                .map(|token| EventType::Action(Box::new(Action::Timeout(*token))))
+                .collect_vec();
+
+            self.events.extend(events);
+            self.events.extend(timed_out_events);
```
maqi: this is mainly to achieve as what described in the task description 2.c
`pick random index in the events from category_rx to insert timed_out event for fetched token`

tests/mock_crust/churn.rs
maqi: probably not, as the un-acked msg is vital to not complete polling early

src/timer.rs
maqi: probably, will have a try later.

src/states/node.rs
afck: But if we don't keep advancing time in the tests the ticking will stop, so there shouldn't be a loop?

maqi: but we have advance time so the unack can start working :(

afck: Right, that's true…

src/state_machine.rs
ustulation: Given a seed is it not guaranteed this function will be called exactly `n` times ? If true why would there be non-determinism ? So say each time you enter the function you generate a something random based on which you will insert the timer actions in the array. Why would this be non-deterministic if the global seed is known ?

src/timer.rs
maqi: it will require the struct `Inner` to implement `Copy`  :(

src/state_machine.rs
maqi: didn't realise there is `SeededRng::thread_rng()` available. now changed to it.

src/state_machine.rs
```diff
@@ -355,21 +340,46 @@ impl StateMachine {
...
             let timed_out_events = self.state
-                .poll()
+                .get_timed_out_tokens()
                 .iter()
                 .map(|token| EventType::Action(Box::new(Action::Timeout(*token))))
                 .collect_vec();
 
+            if !events.is_empty() {
+                for event in timed_out_events {
+                    let pos = SeededRng::thread_rng().gen_range(0, events.len());
```
afck: This should probably be `events.len() + 1`, so they can also be inserted in last position. Then the distinction between empty and nonempty `events` would not be needed.

What I also find problematic is that this might now insert timer events in the wrong order. I think in production it is guaranteed that if you set a timer for 1 minute and another one for 2 minutes, you will handle the first one first. It's great if the tests pass this way, but I think it could cause surprising failures later, if we ever rely on the order of timer events.

Instead, you could do something like:
```rust
// Randomise where the timer events are inserted (zeroes).
let mut positions = iter::repeat(0)
    .take(timed_out_events.len())
    .chain(iter::repeat(1).take(events.len()))
    .collect_vec());
rng.shuffle(&mut positions);
self.events.extend(positions.map(|i| match i {
    0 => timed_out_events.pop(),
    1 => events.pop(),
}).rev());
```
(Or collect first, if that can't be `rev`ersed as is.)

src/state_machine.rs
```diff
@@ -322,6 +336,75 @@ impl StateMachine {
...
+                .iter()
+                .all(|event| match *event {
+                         EventType::Action(ref action) => {
+                             match *action.borrow() {
+                                 Action::Timeout(_) => true,
+                                 _ => false,
+                             }
+                         }
+                         _ => false,
+                     });
```
afck: Isn't this `true` exactly if `events.is_empty()` was true before the timed out events were added in? If so, that would be easier to check.

src/states/node.rs
```diff
@@ -2482,20 +2482,32 @@ impl Node {
                 debug!("{:?} Disconnecting from timed out peer {:?}", self, peer_id);
                 let _ = self.crust_service.disconnect(peer_id);
             }
-            let transition = self.purge_invalid_rt_entries(outbox);
+
+            let transition = if cfg!(feature = "use-mock-crust") {
+                Transition::Stay
+            } else {
+                self.purge_invalid_rt_entries(outbox)
```
afck: Why not purge in mock Crust anymore? That contained a lot of consistency checks.

src/states/node.rs
```diff
@@ -2482,20 +2482,32 @@ impl Node {
...
         if self.su_timer_token == Some(token) {
-            self.su_timeout = cmp::min(Duration::from_secs(SU_MAX_TIMEOUT_SECS),
-                                       self.su_timeout * 2);
-            trace!("{:?} Scheduling next RT request for {} seconds from now.",
-                   self,
-                   self.su_timeout.as_secs());
-            self.su_timer_token = Some(self.timer.schedule(self.su_timeout));
-            self.send_section_update(None);
+            if cfg!(feature = "use-mock-crust") {
+                trace!("{:?} not to schedule next RT request during mock_crust test.",
```
afck: These are now section updates. Routing table requests have been removed.

tests/mock_crust/churn.rs
```diff
@@ -121,6 +122,9 @@ fn add_node_and_poll<R: Rng>(rng: &mut R,
     let failed_node = nodes.remove(new_node);
     drop(failed_node);
     poll_and_resend(&mut nodes, &mut []);
+    // max(RESOURCE_PROOF_DURATION_SECS + ACCUMULATION_TIMEOUT_SECS,
+    //     CANDIDATE_ACCEPT_TIMEOUT_SECS) = 330s
```
afck: Wouldn't it be safer to calculate that here?

tests/mock_crust/utils.rs
```diff
@@ -295,20 +296,26 @@ pub fn poll_all(nodes: &mut [TestNode], clients: &mut [TestClient]) -> bool {
 /// Polls and processes all events, until there are no unacknowledged messages left and clearing
 /// the nodes' state triggers no new events anymore.
 pub fn poll_and_resend(nodes: &mut [TestNode], clients: &mut [TestClient]) {
+    // Expect advance at least three times, given NODE_CONNECT_TIMEOUT_SECS is 60s and
+    // ACK_TIMEOUT_SECS is 20s
+    let mut num_of_advance = 0;
```
afck: Maybe `clock_advance_count`?

tests/mock_crust/utils.rs
```diff
@@ -295,20 +296,26 @@ pub fn poll_all(nodes: &mut [TestNode], clients: &mut [TestClient]) -> bool {
...
     for _ in 0..MAX_POLL_CALLS {
         if poll_all(nodes, clients) {
             let mut call_count = 1;
-            while resend_unacknowledged(nodes, clients) && poll_all(nodes, clients) {
+            while poll_all(nodes, clients) {
```
afck: Does it still make sense to have two nested loops that both call `poll_all` now?

tests/mock_crust/utils.rs
```diff
@@ -295,20 +296,26 @@ pub fn poll_all(nodes: &mut [TestNode], clients: &mut [TestClient]) -> bool {
...
             }
-            nodes.iter_mut().foreach(|node| node.inner.clear_state());
-        } else {
+        } else if num_of_advance > 3 {
             return;
         }
+
+        FakeClock::advance_time(20 * 1000 + 1);
+        num_of_advance += 1;
+        let _ = poll_all(nodes, clients);
```
afck: Is that call necessary? `poll_all` will be called at the beginning of the next loop again anyway.

src/states/node.rs
maqi: as now the ticking is enabled, this means purge will be called in the process of `connecting` (waiting for the un-acked msg to establish tunnel or else), and will panic directly.

purge will still be called (to carry out consistency check) once network is settled (no more msg exchange)

tests/mock_crust/churn.rs
maqi: most of the consts are not publicly exposed  :(

tests/mock_crust/utils.rs
maqi: still required  :(

tests/mock_crust/utils.rs
maqi: yes, it is necessary.
the timer events doesn't make `poll` return `true`, so if a timed_out triggers a msg to be sent out, the poll itself to handle that timed_out returns false, and that msg will be handled in the next poll and makes it return `true`.

if we don't have this extra poll_all here, the beginning of the next loop only handles timed_out events (returns as false) and terminates the loop

src/state_machine.rs
maqi: self.events may contain events from previous. so even `events` is empty for this call, there is still chance self.events contains non-time-out events.

src/states/node.rs
afck: But isn't any purge panic considered a bug, regardless of whether it happens during connection or while there are still messages in flight?

maqi: depends on the polling order.
if the timed_out event got handled before connecting msg, the purge will panic.
but if connecting msg got handled first, it will be ok.

src/mock_crust/support.rs
```diff
@@ -166,6 +168,13 @@ impl Network {
         self.0.borrow_mut().rng.new_rng()
     }
 
+    /// Return whether handled any message since laster query.
```
afck: `since last query` (or `previous`?)

src/mock_crust/support.rs
```diff
@@ -166,6 +168,13 @@ impl Network {
...
+    pub fn reset_message_handled(&self) -> bool {
```
afck: Maybe it should be `sent`, not `handled`; "handling" sounds more like the node receiving the event, rather than the network putting it in the node's queue.

tests/mock_crust/utils.rs
```diff
@@ -295,20 +297,17 @@ pub fn poll_all(nodes: &mut [TestNode], clients: &mut [TestClient]) -> bool {
...
-            }
-            nodes.iter_mut().foreach(|node| node.inner.clear_state());
-        } else {
+            clock_advance_count = 0;
+        } else if clock_advance_count > 3 {
+            // Expect advance at least three times, given NODE_CONNECT_TIMEOUT_SECS is 60s and
+            // ACK_TIMEOUT_SECS is 20s
             return;
         }
+        FakeClock::advance_time(20 * 1000 + 1);
```
Fraser999: I think I'd be in favour of taking the hit of more conditionally-compiled code to expose `ACK_TIMEOUT_SECS` and `NODE_CONNECT_TIMEOUT_SECS` rather than using the hard coded values of `20` and `3` here.  Same for https://github.com/maidsafe/routing/pull/1438/files#r113455846

src/state_machine.rs
```diff
@@ -346,6 +359,83 @@ impl StateMachine {
...
+                        if let Ok(action) = self.action_rx.try_recv() {
+                            events.push(EventType::Action(Box::new(action)));
+                        } else {
+                            return Ok(self.apply_transition(Transition::Terminate, outbox));
+                        }
+                    }
+                    MaidSafeEventCategory::Crust => {
+                        if let Ok(crust_event) = self.crust_rx.try_recv() {
+                            events.push(EventType::CrustEvent(crust_event));
+                        } else {
```
maqi: yes, I agree.
the current mock_crust does need to be `work asynchronously` to fully simulate the behaviour of production code.

tests/mock_crust/utils.rs
maqi: can we leave this part in a separate PR as this will introduce conditionally-compiled code in lib.rs as well.

Fraser999:
Closed in favour of #1442.

PR #1439, opened 2017-04-21
feat/acct-packet: add account registration packet struct

nbaksalyar:
This structure is used for verification by vaults when a new account is created.


maidsafe-highfive:
r? @afck

afck:
Makes sense to me, @ustulation, since the type tags are also in `lib.rs` currently. 👍

PR #1441, opened 2017-04-24
fix/peer_manager: remove approved candidates from map

Fraser999:


maidsafe-highfive:
r? @afck

src/peer_manager.rs
```diff
@@ -698,10 +680,8 @@ impl PeerManager {
         }
 
         if have_candidate {
```
Viv-Rajkumar: guess that got missed in the diff `if !have_candidate {`

Fraser999: Well spotted.  I had to apply the diff manually - `git apply` failed.

src/peer_manager.rs
```diff
@@ -629,26 +615,26 @@ impl PeerManager {
...
+                                                        target_size: target_size,
+                                                        difficulty: difficulty,
+                                                        seed: seed,
+                                                        proof: VecDeque::new(),
+                                                    });
+                (Ok(true), Some(state))
             }
+        } else if self.peer_map
+                      .get_by_name(pub_id.name())
+                      .map_or(false, |peer| peer.valid()) {
```
afck: `.map_or(false, Peer::valid)`?

PR #1442, opened 2017-04-28
MAID-2051: Use the fake clock in mock-crust tests

Fraser999:


maidsafe-highfive:
r? @afck

tests/mock_crust/utils.rs
```diff
@@ -295,20 +297,17 @@ pub fn poll_all(nodes: &mut [TestNode], clients: &mut [TestClient]) -> bool {
...
-            }
-            nodes.iter_mut().foreach(|node| node.inner.clear_state());
-        } else {
+            clock_advance_count = 0;
+        } else if clock_advance_count > 3 {
+            // Expect advance at least three times, given NODE_CONNECT_TIMEOUT_SECS is 60s and
+            // ACK_TIMEOUT_SECS is 20s
             return;
         }
+        FakeClock::advance_time(20 * 1000 + 1);
```
afck: Maybe we should make those timeout constants `pub` if `use-mock-crust` is enabled.

Fraser999: I was in favour of that [before](https://github.com/maidsafe/routing/pull/1438#discussion_r113705446) and I'm still in favour of that.

PR #1443, opened 2017-05-02
chore/build: update to Rust nightly 2017-04-28 and Clippy 0.0.128

afck:


maidsafe-highfive:

PR #1445, opened 2017-05-16
Mutable data

madadam:
- Add no-validation variants of some mutation operations on `MutableData`
- Add `GetMData` request and response


maidsafe-highfive:
r? @afck

PR #1446, opened 2017-05-16
Mutable data

madadam:
- Add no-validation variants of some mutation operations on `MutableData`
- Add `GetMData` request and response

### DIFF
--- a/src/data/mutable_data.rs
+++ b/src/data/mutable_data.rs
@@ -21,6 +21,7 @@ use rust_sodium::crypto::sign::PublicKey;
 use std::collections::BTreeSet;
 use std::collections::btree_map::{BTreeMap, Entry};
 use std::fmt::{self, Debug, Formatter};
+use std::mem;
 use xor_name::XorName;
 
 /// Maximum allowed size for `MutableData` (1 MiB)
@@ -294,6 +295,11 @@ impl MutableData {
     /// Returns all entries
     pub fn entries(&self) -> &BTreeMap<Vec<u8>, Value> {
         &self.data
+    }
+
+    /// Removes and returns all entries
+    pub fn take_entries(&mut self) -> BTreeMap<Vec<u8>, Value> {
+        mem::replace(&mut self.data, BTreeMap::new())
     }
 
     /// Mutates entries (key + value pairs) in bulk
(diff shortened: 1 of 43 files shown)


maidsafe-highfive:
r? @afck

src/data/mutable_data.rs
```diff
@@ -395,6 +395,42 @@ impl MutableData {
         Ok(())
     }
 
+    /// Mutates entries without performing any validation.
+    ///
+    /// For updates and deletes, the mutation is performed only if he entry version
```
afck: `the entry version`

src/data/mutable_data.rs
```diff
@@ -395,6 +395,42 @@ impl MutableData {
...
+    /// of the action is higher than the current version of the entry.
+    pub fn mutate_entries_without_validation(&mut self, actions: BTreeMap<Vec<u8>, EntryAction>) {
```
afck: It would probably simplify `mutate_entries` if it would call `mutate_entries_without_validation` after the validation.
Or is `mutate_entries` not even needed by vaults anymore now, since validation always happens before the vote and mutation happens afterwards? (And the same goes for permissions and owners, I guess.)

src/data/mutable_data.rs
```diff
@@ -395,6 +395,42 @@ impl MutableData {
...
+                                                     entry_version: new_version,
+                                                 });
+                        }
+                    }
+                }
+            }
+        }
+    }
+
     /// Mutates single entry withou performing any validations, except the version
```
afck: `without`

src/data/mutable_data.rs
madadam: `mutate_entries` is still used in vaults, but actually only to perform the validation before inserting it into pending writes. The mutation itself is thrown away at that point. So yeah, it's bit of a mess :grimacing: 

I think we should have three methods:
1. `validate_mutate_entries(&self, actions: ...) -> Result<(), ClientError>` - only validate, no mutation
2. `mutate_entries_without_validation(&mut self, actions: ...)` - no validation, only mutation
3. `mutate_entries(&mut self, actions: ...)` - do both, but by calling the above two
(the names are up for discussion)

Viv-Rajkumar:
Closing this as @madadam is about to raise just the new commits rebased to latest mutable-data

PR #1447, opened 2017-05-17
Remove PeerId and redundant state across PeerManager and Node

ustulation:


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -1919,16 +1804,11 @@ impl Node {
...
-
-                if let Some((Some(peer_id), valid)) = x {
-                    if valid {
-                        self.send_node_identify(peer_id, false);
-                        self.add_to_routing_table(&pub_id, &peer_id, false, outbox);
-                    }
+                              pub_id);
+                if self.peer_mgr
+                       .get_peer(&pub_id)
+                       .map_or(false, |peer| peer.valid()) {
```
afck: Probably doesn't need a closure: `Peer::valid`?

src/states/node.rs
```diff
@@ -1945,165 +1825,178 @@ impl Node {
...
+                   srd_id,
                    dst_id);
         }
     }
 
-    /// Handle a `TunnelSuccess` response from `peer_id`: It will act as a tunnel to `dst_id`.
-    fn handle_tunnel_success(&mut self, peer_id: PeerId, dst_id: PeerId) {
-        if let Some(tunnel_id) = self.tunnels.tunnel_for(&dst_id) {
-            if *tunnel_id == peer_id {
+    /// Handle a `TunnelSuccess` response from `pub_id`: It will act as a tunnel to `dst_id`.
```
afck: `Handles` (also below)

PR #1448, opened 2017-05-19
Mutable update

maqi:


maidsafe-highfive:
r? @afck

src/messages/request.rs
```diff
@@ -47,6 +47,15 @@ pub enum Request {
     },
 
     // --- MutableData ---
+    /// Fetches whole MutableData from the network.
```
afck: The documentation should probably stress that during churn this is likely to fail to accumulate and will therefore often receive no response at all.

Viv-Rajkumar:
Closing this as @madadam is about to raise just the new commits rebased to latest mutable-data

PR #1449, opened 2017-05-19
Version change to 0.29.0

ustulation:
Give a fresh timestamp for newly connected peers else remove_expired_peers would remove them immediately if the existing timestamp indicated >60 <90secs.
Fix bug where the tunnel node split off from the tunnel client and would never deliver TunnelSuccess message to it.


maidsafe-highfive:
r? @afck

PR #1450, opened 2017-05-22
Mutable data

madadam:


maidsafe-highfive:
r? @afck

src/messages/request.rs
```diff
@@ -47,6 +47,15 @@ pub enum Request {
     },
 
     // --- MutableData ---
+    /// Fetches whole MutableData from the network.
```
afck: Let's document that during churn, this is unlikely to accumulate. Also for the other bulk gets (keys and values), and also for the methods in `src/node.rs`.

madadam: Done.

PR #1454, opened 2017-05-25
refactor/common-types: put type common to client and vault in a module

nbaksalyar:
Create a module and add AccountPacket struct used by vaults and clients in there. It is not required by routing but currently there is no other crate which vault and client can share so it's put into routing for the moment.


maidsafe-highfive:
r? @afck

PR #1456, opened 2017-05-29
only handle own_section nodes' connect req before approved; test and fixes relates to merge issue

maqi:


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -2375,7 +2388,14 @@ impl Node {
         info!("{:?} Other section merge completed. Prefixes: {:?}",
               self,
               self.routing_table().prefixes());
-        self.merge_if_necessary(outbox);
+
+        // This is an optimisation to avoid un-necessary merge in the scenario:
+        //      we only know one of the subsections that got merged, and that subsection contains
+        //      less than min_section_size nodes.
```
afck: This should probably mention that that section doesn't _really_ contain too few nodes: we just didn't learn about all its members because the message was empty?

tests/mock_crust/merge.rs
```diff
@@ -137,3 +140,49 @@ fn concurrent_merge() {
...
+
+    // Choose one sections to drop nodes from.
+    let prefix_to_drop_from = Prefix::new(1, XorName([0; XOR_NAME_LEN]));
+
+    let mut nodes_count = nodes
+        .iter()
+        .filter(|node| node.routing_table().our_prefix() == &prefix_to_drop_from)
+        .count();
+
+    // Drop enough nodes (without polling) from the two section to take them just below
```
afck: `from the two` → `from that`?

src/peer_manager.rs
```diff
@@ -212,15 +212,17 @@ pub struct Peer {
     state: PeerState,
     timestamp: Instant,
     valid: bool,
+    reconnect: bool,
```
afck: The peer manager is getting really confusing. Let's add a comment here to explain why this field exists.

PR #1458, opened 2017-06-02
fix/joining-node: timer-tokens for joining node exposed for mock-crust

ustulation:
Allow mock-crust to access timer-tokens for joining node so that it can be allowed to select alternate routes in-case say the relocation request failed via the 1st route that was taken.


maidsafe-highfive:
r? @afck

PR #1459, opened 2017-06-06
test/mock_crust: poll less often for random message order

afck:


maidsafe-highfive:

src/mock_crust/support.rs
```diff
@@ -343,6 +342,11 @@ impl<UID: Uid> ServiceImpl<UID> {
         self.start(event_sender, uid)
     }
 
+    /// Delivers all messages that are queued in the network.
+    pub fn poll_network(&self) {
```
Viv-Rajkumar: think this is an unused fn that should be safe to remove?

src/mock_crust/support.rs
```diff
@@ -472,12 +476,8 @@ impl<UID: Uid> ServiceImpl<UID> {
...
     }
 
-    fn handle_message(&self, peer_endpoint: Endpoint, data: Vec<u8>) {
-        if let Some(uid) = self.find_uid_by_endpoint(&peer_endpoint) {
-            self.send_event(CrustEvent::NewMessage(uid, data));
-        } else {
-            unreachable!("Received message from non-connected {:?}", peer_endpoint);
-        }
+    fn handle_message(&self, data: Vec<u8>, uid: UID) {
+        self.send_event(CrustEvent::NewMessage(uid, data));
```
Viv-Rajkumar: so if service A disconnects from B in a given poll iteration and in the same iteration B sends a message to A, couldn't this now trigger a NewMessage received from B after A has disconnected to B?

Saying that this might be closer to reality with crust adding recd messages to the routing event queue(which are not yet handled by routing) before a peer was disconnected, so we could essentially end up with a similar behaviour in production, just wondering if that is what we're going for here?

If we are, lets maybe add a comment to explain the intent with this one so we dont accidentally revert it pls.

src/mock_crust/support.rs
```diff
@@ -294,6 +287,12 @@ impl<UID: Uid> ServiceHandle<UID> {
             .is_peer_connected(&unwrap!(handle.0.borrow().uid))
     }
 
+    /// Delivers all messages that are queued in the network.
+    pub fn poll_network(&self) {
```
Viv-Rajkumar: would it maybe help to rename this and the `Network::poll` to something like dispatch / deliver messages instead. Just so we can keep the poll functions exclusive to the handling events part of the mock crust flow.

PR #1460, opened 2017-06-07
fix/node: resend OtherSectionMerge on duplicated OwnSectionMerge and using union of our_merged_section

maqi:


maidsafe-highfive:
r? @afck

src/states/node.rs
```diff
@@ -110,6 +110,8 @@ pub struct Node {
     routing_msg_backlog: Vec<RoutingMessage>,
     /// Cache of `OwnSectionMerge` messages we have received, by sender section prefix.
     merge_cache: LruCache<Prefix<XorName>, SectionMap>,
+    /// Cache of union of our merged section, deduced from multiple `OwnSectionMerge`.
+    our_merged_section_cache: LruCache<Prefix<XorName>, (bool, BTreeSet<XorName>)>,
```
afck: I think the decision from the discussion was to always only keep at most one entry in this cache, i.e. make it an `Option` instead of `LruCache`?

src/states/node.rs
```diff
@@ -2260,12 +2264,48 @@ impl Node {
         Ok(())
     }
 
+    // If multiple nodes dropped from our section, duplicated `OwnSectionMerge` will be sent,
+    // containing different dropped out nodes. This will result in differnt `OtherSectionMerge` to
```
afck: `different`

src/states/node.rs
```diff
@@ -2260,12 +2264,48 @@ impl Node {
...
+    fn union_our_merged_section(&mut self, merge_prefix: Prefix<XorName>, sections: &SectionMap) {
+        let mut our_merged_section = BTreeSet::new();
+        for (ver_pfx, peers) in sections.iter() {
+            if ver_pfx.prefix().is_extension_of(&merge_prefix) {
+                our_merged_section.extend(peers.into_iter().map(|peer| *peer.name()));
+            }
+        }
+        match self.our_merged_section_cache.remove(&merge_prefix) {
+            Some((send_other_section_merge, prev_our_merged_section)) => {
+                our_merged_section.extend(prev_our_merged_section.iter());
```
afck: If you compare the `_merged_section`s before this line, you could omit the `iter()` and consume `prev_our_merged_section`.

src/states/node.rs
```diff
@@ -2298,18 +2338,19 @@ impl Node {
...
             }
+            let (_, merged_section) = unwrap!(self.our_merged_section_cache.remove(&merge_prefix));
             self.process_own_section_merge(their_prefix,
                                            version,
                                            their_sections,
-                                           our_merged_section,
+                                           merged_section.clone(),
                                            outbox);
+            let _ = self.our_merged_section_cache
+                .insert(merge_prefix, (true, merged_section));
```
afck: Why remove and insert it instead of `get_mut` and setting the first component to `true`?

src/states/node.rs
maqi: the new our_merged_section could be just a subset of the union-ed  prev_our_merged_section,
in that case they are different but we shall not re-send the OtherSectionMerge

afck: Ah, right. Then you could just compare `our_merged_section.len()` before and after the `extend` call to avoid the cloning. But maybe that's premature optimisation…

src/states/node.rs
maqi: in case of merge after merge, the out-dated OwnSectoinMerge of first merge may overwrite the current on-going merge, if using Option only.
Is it possible?

src/states/node.rs
```diff
@@ -75,6 +75,70 @@ const MERGE_TIMEOUT_SECS: u64 = 300;
...
+    // be sent out, hence causing non-accumulation on the receiver side.
+    // So here we will take a union of our_merged_section from all the `OwnSectionMerge`, and resend
+    // `OtherSectionMerge` whenever the new union is different to the last one sent out.
+    fn union_our_merged_section(&mut self,
+                                merge_prefix: Prefix<XorName>,
+                                sections: &SectionMap)
+                                -> Option<BTreeSet<XorName>> {
+        let mut version = 1;
+        let mut our_merged_section = BTreeSet::new();
+        for (ver_pfx, peers) in sections.iter() {
```
afck: That `.iter()` should be unnecessary;

src/states/node.rs
```diff
@@ -75,6 +75,70 @@ const MERGE_TIMEOUT_SECS: u64 = 300;
...
 
+#[derive(Default)]
+struct UnionOurMergedSection {
+    merge_prefix: Prefix<XorName>,
+    version: u64,
+    send_other_section_merge: bool,
+    our_merged_section: BTreeSet<XorName>,
+}
+
+impl UnionOurMergedSection {
```
afck: Maybe `CumulativeOwnSectionMerge`? `SectionMergeUnion`? Not sure…

src/states/node.rs
```diff
@@ -75,6 +75,70 @@ const MERGE_TIMEOUT_SECS: u64 = 300;
...
+    // If multiple nodes dropped from our section, duplicated `OwnSectionMerge` will be sent,
+    // containing different dropped out nodes. This will result in different `OtherSectionMerge` to
+    // be sent out, hence causing non-accumulation on the receiver side.
+    // So here we will take a union of our_merged_section from all the `OwnSectionMerge`, and resend
+    // `OtherSectionMerge` whenever the new union is different to the last one sent out.
+    fn union_our_merged_section(&mut self,
```
afck: Since that's already the type name, maybe this should just be `extend`, or `add`?

PR #1462, opened 2017-06-09
catch up with the master branch

maqi:
1, fix broken link
2, less poll for random message order
3, Union our merged section for re-sending OtherSectionMerge in case of duplicated OwnSectionMerge messages


maidsafe-highfive:
r? @afck

PR #1463, opened 2017-06-09
MAID-2078 Mock crypto

madadam:


maidsafe-highfive:
r? @afck

.travis.yml
```diff
@@ -30,7 +30,7 @@ script:
         cargo rustc --verbose --example ci_test -- -Zno-trans &&
         cargo rustc --verbose --example key_value_store -- -Zno-trans &&
         cargo rustc --verbose -- -Zno-trans &&
-        travis_wait 60 cargo test --release --verbose --features use-mock-crust
+        travis_wait 60 cargo test --release --verbose --features \"use-mock-crust use-mock-crypto\"
```
afck: I think we can remove this `travis_wait 60` now.

src/data/immutable_data.rs
```diff
@@ -87,12 +87,14 @@ impl Debug for ImmutableData {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use hex::ToHex;
     use maidsafe_utilities::{SeededRng, serialisation};
     use rand::Rng;
 
     #[test]
+    #[cfg(not(feature = "use-mock-crypto"))]
```
afck: That test should work with or without mock crypto, shouldn't it? It's just the value of `expected_name` below that needs to be conditionally compiled.

src/mock_crypto.rs
```diff
@@ -0,0 +1,241 @@
...
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+//! Mock cryptographic primitives.
+
+/// Mock version of a subset of the rust_sodium crate.
```
afck: Maybe explain that these are not secure but make the tests several times faster.

src/mock_crypto.rs
```diff
@@ -0,0 +1,241 @@
...
+            use rand::Rng;
+
+            /// Number of bytes in a 'PublicKey`.
+            pub const PUBLICKEYBYTES: usize = 32;
+            /// Number of bytes in a `SecretKey`.
+            pub const SECRETKEYBYTES: usize = 32;
+            /// Number of bytes in a `Nonce`.
+            pub const NONCEBYTES: usize = 4;
+
+            /// Mock public key for asymetric encryption/decryption.
```
afck: `asymmetric` (also below)

afck:
My comment got hidden: we can remove the `travis_wait 60` now.

PR #1464, opened 2017-06-12
fix/bootstrapping: Also start listening as JoiningNode before bootstrap

Viv-Rajkumar:
- This will allow external reachability check to succeed when bootstrapping as a CrustUser::Node role.
- Enable Service discovery as first node on startup and in other cases only after NodeApproval completes successfully.
- Also fix log format for node add/remove to RT and RT size msg.
- Remove redundant log/comments about NodeIdentify.


maidsafe-highfive:
r? @afck

PR #1465, opened 2017-06-13
Mutable data

ustulation:
The examples are currently commented and will be addressed as a part of a separate pr.


maidsafe-highfive:
r? @afck

PR #1466, opened 2017-06-14
Uncomment and fix examples

madadam:


maidsafe-highfive:
r? @afck

examples/key_value_store.rs
```diff
@@ -197,35 +197,32 @@ Options:
...
-        pub fn put(&self, put_where: String, put_what: String) {
-            let name = KeyValueStore::calculate_key_name(&put_where);
-            let data = unwrap!(serialise(&(put_where, put_what)));
-            let sd = unwrap!(StructuredData::new(10000, name, 0, data, BTreeSet::new()));
-            if self.example_client.put(Data::Structured(sd)).is_err() {
-                error!("Failed to put data.");
+        pub fn put(&mut self, put_where: String, put_what: String) {
+            let name = Self::calculate_key_name(&put_where);
+
+            let key = put_where.into_bytes();
```
afck: I wonder whether it would be better to just use a constant (empty?) key.
Since this is an example, it should basically do what we would recommend for such a use-case; and I think using an empty key would make more sense than using `put_what` twice?

PR #1467, opened 2017-06-14
fix/node: restrict the message types handled from a client or bootstrapping node

Fraser999:


maidsafe-highfive:
r? @afck

Viv-Rajkumar:
Closing this for now. MAID-2080 should handle the same and a bit more

PR #1468, opened 2017-06-14
Conditionally promote peer from Candidate to Routing state

Viv-Rajkumar:
check if conn info req is sent to existing candidate. In this case, set peer as valid and them to the RT.


maidsafe-highfive:
r? @afck

PR #1469, opened 2017-06-15
Make mutable data validation more strict

madadam:
It should no longer be possible to exceed the limits using the operations that perform validation (`mutate_entries`, `set_user_permissions`, ...).


maidsafe-highfive:
r? @afck

Thierry61:
Some suggestions:

- Why `MAX_MUTABLE_DATA_ENTRIES + 5` in max_entries_limit? It should be `MAX_MUTABLE_DATA_ENTRIES + 1` to test the limit.

- Associated comment should be corrected: `// It must not be possible to create MutableData with more than 100 entries` or even better: `// It must not be possible to create MutableData with more than MAX_MUTABLE_DATA_ENTRIES entries`.

- You could also test success of adding 1 entry on a 99 entries MD (`MAX_MUTABLE_DATA_ENTRIES - 1`) and then test error of adding 1 entry again.

PR #1471, opened 2017-06-15
MAID-2080: update handling of bootstrapping nodes

Fraser999:
This adds a new 'PeerState' variant for bootstrapping peers which contains the type of connection
(client or node).  That's used to replace the 'bootstrappers' container in 'Node'.

There are further restrictions placed on the types of messages clients and bootstrappers can send,
and some trivial renaming changes included in this commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,9 +37,9 @@ script:
       (
         set -x;
         cargo clippy &&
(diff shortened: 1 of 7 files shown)


maidsafe-highfive:
r? @afck

PR #1474, opened 2017-06-20
Client rate limit check

maqi:


src/rate_limiter.rs
```diff
@@ -0,0 +1,141 @@
+// Copyright 2016 MaidSafe.net limited.
```
ustulation: 2017 ?

src/states/node.rs
```diff
@@ -881,6 +886,12 @@ impl Node {
             // return Err(RoutingError::UnknownConnection);
         };
 
+        if client_check &&
+           self.check_valid_client_message(&pub_id, hop_msg.content.routing_message())
+               .is_err() {
+            return Err(RoutingError::InvalidStateForOperation);
```
ustulation: So previously to disconnected from a client sending invalid hop-msg (probably malicious client) - now you don't. Any reason for the change ?

src/rate_limiter.rs
```diff
@@ -0,0 +1,141 @@
...
+                        Request::GetMDataShell { .. } |
+                        Request::ListMDataKeys { .. } |
+                        Request::ListMDataValues { .. } |
+                        Request::ListMDataEntries { .. } |
+                        Request::GetMDataValue { .. } |
+                        Request::ListMDataPermissions { .. } |
+                        Request::ListMDataUserPermissions { .. } |
+                        Request::ListAuthKeysAndVersion { .. } |
+                        Request::GetAccountInfo { .. } |
+                        Request::Refresh(..) => DEFAULT_CLIENT_GET_CHARGE,
```
ustulation: This is not an allowed RPC for clients - probably need to disconnect from it ?

src/rate_limiter.rs
```diff
@@ -0,0 +1,141 @@
...
+                        _ => payload.len() as u64,
+                    }
+                }
+                _ => payload.len() as u64,
```
ustulation: This is not an allowed RPC for clients - probably need to disconnect from it ? Am wondering if all such clients should be kept in an LRUSet<IP> and not allowed to reconnect for a long time.

src/mock_crust/crust.rs
```diff
@@ -100,6 +101,11 @@ impl<UID: Uid> Service<UID> {
         trace!(target: "crust", "[MOCK] set_service_discovery_listen not implemented in mock");
     }
 
+    /// Return ip address of the peer.
+    pub fn get_peer_ip_addr(&self, _peer: &UID) -> Result<IpAddr, CrustError> {
+        Ok(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)))
```
ustulation: This will have problems that peer gets the same value - so writing future tests based on the actual behavior would be difficult. Think it's better in [this PR](https://github.com/maidsafe/routing/pull/1476/files#diff-9d1f471bb4e917b1d0eec91ca3a02ae2R373)

PR #1475, opened 2017-06-20
MAID-2084: handle 'LostPeer' event while bootstrapping

Fraser999:


src/peer_manager.rs
```diff
@@ -43,7 +43,8 @@ use xor_name::XorName;
 const JOINING_NODE_TIMEOUT_SECS: u64 = 900;
 /// Time (in seconds) after which the connection to a peer is considered failed.
 const CONNECTING_PEER_TIMEOUT_SECS: u64 = 90;
-/// Time (in seconds) the node waits for a peer to become valid once connected.
+/// Time (in seconds) the node waits for a peer to become valid once connected or bootstrapped to
+/// it.
```
ustulation: Maybe this comment is better explicity separating the `or` - something along the lines of: `Time (in seconds) the node waits for a peer to either become valid once connected to it or to transition once bootstrapped to it` ? Else it reads as if node in certain scenarios waits for a peer to be valid once the peer bootstraps to it.

Fraser999: Change force pushed (nothing altered except this comment to the suggested text).

PR #1476, opened 2017-06-21
Send signatures when prefixes change during a section update

Fraser999:


Fraser999:
Closed in favour of #1477

PR #1479, opened 2017-06-23
feature/mutable-data: implement hard delete of entries

madadam:
- `mutate_entries` is now only for inserts and updates
- new API `delete_entries` is for hard (permanent) deletes

### DIFF
--- a/src/data/mutable_data.rs
+++ b/src/data/mutable_data.rs
@@ -437,6 +437,10 @@ impl MutableData {
 
         let keys: Vec<_> = keys.into_iter().collect();
 
+        if keys.is_empty() {
+            return Err(ClientError::InvalidOperation);
+        }
+
         if keys.iter().any(|key| !self.data.contains_key(key)) {
             return Err(ClientError::NoSuchEntry);
         }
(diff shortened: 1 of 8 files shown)


src/messages/request.rs
```diff
@@ -134,15 +134,30 @@ pub enum Request {
...
         actions: BTreeMap<Vec<u8>, EntryAction>,
+        /// Current shell version of the data.
+        version: u64,
+        /// Unique message identifier
+        msg_id: MsgId,
+        /// Requester public key
+        requester: sign::PublicKey,
+    },
+    /// Deletes MutableData entries in bulk.
+    DeleteMDataEntries {
```
maqi: rate_limiter needs to be updated to charge for this request.

src/data/mutable_data.rs
```diff
@@ -454,6 +419,53 @@ impl MutableData {
...
+
+        if keys.iter().any(|key| !self.data.contains_key(key)) {
+            return Err(ClientError::NoSuchEntry);
+        }
+
+        for key in keys {
+            let _ = self.data.remove(&key);
+        }
+
+        self.version = version;
```
maqi: when keys is empty, the version will still be updated, is it expected?

ustulation:
Not currently decided - but pls keep this work around for future @madadam

PR #1480, opened 2017-06-23
Added client rate-limiter and temporary bans of malicious clients' IPs

Fraser999:


src/states/node.rs
```diff
@@ -615,6 +639,7 @@ impl Node {
                           self,
                           error,
                           pub_id);
+                    self.ban_client(&pub_id);
```
ustulation: Ah do we also want to do this for nodes ? This will do for crust-user::node as well i think

src/states/node.rs
```diff
@@ -891,6 +915,7 @@ impl Node {
                 Err(RoutingError::ExceedsRateLimit)
             }
             Err(error) => {
+                self.ban_client(&pub_id);
```
ustulation: same as above:
> do we also want to do this for nodes ? This will do for crust-user::node as well i think

src/states/node.rs
Fraser999: Yes - I thought that would be OK for now since we're not catering for malicious nodes.

src/rate_limiter.rs
```diff
@@ -0,0 +1,224 @@
...
+        // which doesn't need its full quota, the unused portion is equally distributed amongst the
+        // others.
+        let leaking_client_count = self.used.len();
+        let mut quota = leaked_units / leaking_client_count as u64;
+        let mut entries: Vec<(u64, IpAddr)> = self.used
+            .iter()
+            .map(|(ip_addr, used)| (*used, *ip_addr))
+            .collect();
+        entries.sort();
+        self.used.clear();
```
ustulation: Optional if you prefer: Can be done in the same step and safer as you cannot forget to call clear later. `let mut entries = mem::replace(&mut self.used, Default::default()).into_iter().map(...).collect();` Also a single iteration instead of two (clear() will iterate to deallocate again)

src/rate_limiter.rs
```diff
@@ -0,0 +1,224 @@
...
+                       online_clients: u64,
+                       client_ip: &IpAddr,
+                       part_count: u32,
+                       part_index: u32,
+                       payload: &[u8])
+                       -> Result<(), RoutingError> {
+        self.update();
+        let total_used: u64 = self.used.values().sum();
+        let used = self.used.get(client_ip).map_or(0, |used| *used);
+        let allowance = cmp::min(CAPACITY - total_used, CAPACITY / online_clients - used);
```
maqi: i think here we need `saturating_sub` for `CAPACITY / online_clients - used` as there is chance online_clients can jump from 1 to 10 directly?
to verify this, we can add
```rust
        FakeClock::advance_time(wait_millis);
        unwrap!(rate_limiter.add_message(2, &client_2, 1, 0, &get_req_payload));
        FakeClock::advance_time(wait_millis);
        match rate_limiter.add_message(10, &client_2, 1, 0, &get_req_payload) {
            Err(RoutingError::ExceedsRateLimit) => {}
            _ => panic!("unexpected result"),
        }
```
at the end of the `add_message` test to verify it.

src/rate_limiter.rs
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        // Wait until enough has drained to allow the second client's request to succeed.
+        let wait_millis = CLIENT_GET_CHARGE * 1000 / RATE as u64;
+        FakeClock::advance_time(wait_millis);
+        unwrap!(rate_limiter.add_message(10, &client_2, 1, 0, &get_req_payload));
+
+        // Wait for the same period, but now try adding invalid messages.
+        FakeClock::advance_time(wait_millis);
+        let all_zero_paylod = vec![0u8; CLIENT_GET_CHARGE as usize];
```
maqi: all_zero_payload

src/rate_limiter.rs
```diff
@@ -0,0 +1,224 @@
...
+        match rate_limiter.add_message(10, &client_2, 2, 0, &all_zero_paylod) {
+            Err(RoutingError::InvalidMessage) => {}
+            _ => panic!("unexpected result"),
+        }
+
+        unwrap!(rate_limiter.add_message(2, &client_2, 1, 0, &get_req_payload));
```
maqi: better have a
```rust
        match rate_limiter.add_message(10, &client_2, 1, 0, &get_req_payload) {
            Err(RoutingError::ExceedsRateLimit) => {}
            _ => panic!("unexpected result"),
        }
```
test first to confirm the per-client usage check has been carried out properly

src/states/node.rs
```diff
@@ -1366,7 +1393,10 @@ impl Node {
     }
 
     /// Returns `Ok` if a client is allowed to send the given message.
-    fn check_valid_client_message(&self, msg: &RoutingMessage) -> Result<(), RoutingError> {
+    fn check_valid_client_message(&mut self,
```
maqi: shall we use part of the work in https://github.com/Fraser999/routing/pull/4/commits/6d4781012265f57d1685da37986a232777da8eb2
to move this `check_valid_client_message` function into RateLimit
so node.rs can be cleaned_up little bit?

src/rate_limiter.rs
Fraser999: I think we should have a lot more unit test coverage for this module, and I guess I'd rather leave that to a separate PR.

src/rate_limiter.rs
Fraser999: Good catch :)  This would be a problem even going from 1 to 2 clients if the first client had taken more than half the capacity!  Again, I'd prefer to leave adding the test case to a separate PR.

src/states/node.rs
Fraser999: Well, I think the decision was to leave it in node.rs for now.

ustulation:
OSX taking for ever to start, tested locally on OSX - that passed. Rest of CI passes, hence merging

PR #1482, opened 2017-06-27
add integrate test and expand unit test

maqi:


src/state_machine.rs
```diff
@@ -170,6 +170,13 @@ impl State {
         }
     }
 
+    pub fn get_clients_usage(&self) -> BTreeMap<::std::net::IpAddr, u64> {
+        match *self {
+            State::Node(ref state) => state.get_clients_usage(),
+            _ => BTreeMap::new(),
```
Fraser999: Shouldn't this panic instead?

tests/mock_crust/mod.rs
```diff
@@ -245,3 +245,111 @@ fn multiple_clients_per_proxy() {
     let _ = poll_all(&mut nodes, &mut clients);
     expect_next_event!(clients[MAX_CLIENTS_PER_PROXY - 1], Event::Connected);
 }
+
+#[test]
+/// Connects multiple clients to the same proxy node and randomly sending get requests.
+/// Expect some reguests will be blocked due to the rate limit.
```
Fraser999: Typo

tests/mock_crust/mod.rs
```diff
@@ -245,3 +245,111 @@ fn multiple_clients_per_proxy() {
...
+
+#[test]
+/// Connect a client to the network then send an invalid message.
+/// Expect the client will be banned (disconnected);
+fn ban_malicious_client() {
+    let min_section_size = 8;
+    let network = Network::new(min_section_size, None);
+    let mut nodes = create_connected_nodes(&network, min_section_size);
+    let mut clients = create_connected_clients(&network, &mut nodes, 1);
+
```
Fraser999: Too many blank lines?

tests/mock_crust/mod.rs
```diff
@@ -245,3 +245,111 @@ fn multiple_clients_per_proxy() {
...
+    let mut rng = network.new_rng();
+    let data_name: XorName = rng.gen();
+    let group_auth = Authority::NaeManager(data_name);
+    let client_auth = Authority::Client {
+        client_id: *clients[0].full_id.public_id(),
+        proxy_node_name: nodes[0].name(),
+    };
+    let msg_id = MessageId::new();
+
+    // Send a get request from NM to client, which makes client sends an ACK to proxy.
```
Fraser999: This seems like quite a roundabout way to make the client send an invalid message!  I guess I'd make it more explicit by having the client send one of each of the variants of `HopMessage` and `DirectMessage` it's not allowed, and also one of each of the `HopMessage`s it _is_ allowed but with an invalid priority applied.  Also it'd be good to have a similar test where the client bootstraps and doesn't send the expected `BootstrapRequest` (either sends an invalidly-signed one or sends a different variant).  There are more malicious scenarios we've catered for - I guess they should all be tested really.

maqi: i add another step of test to verify Refresh from client to group causing banning.

Regarding other invalid sending, this means: have to expose

PR #1483, opened 2017-06-27
chore/rate_limiter: move check_valid_client_message from node into rate_limiter

maqi:


ustulation:
Closing this as it's better to keep client validations independent of rate-limiter.

PR #1485, opened 2017-06-29
Further work on the rate-limiter

Fraser999:


src/states/node.rs
```diff
@@ -3418,6 +3418,13 @@ impl Node {
         self.clients_rate_limiter.get_clients_usage()
     }
 
+    pub fn get_banned_client_ips(&self) -> BTreeSet<IpAddr> {
+        self.banned_client_ips
+            .peek_iter()
+            .map(|(ip, _)| *ip)
+            .collect()
```
ustulation: (Not meant as a review for this PR) Wouldn't it be better if lru-time-cache (our crate) implemented `keys()` function ?

src/peer_manager.rs
```diff
@@ -442,25 +448,15 @@ impl PeerManager {
...
-                peer.state = PeerState::JoiningNode;
-            }
-            _ => {
-                log_or_panic!(LogLevel::Error,
-                              "{} {:?} is not a bootstrapper.",
-                              self_debug,
-                              pub_id);
-            }
-        }
+        let _ = self.peers
```
ustulation: Why are we ignoring instead of a log-or-panic done previously ?

src/peer_manager.rs
```diff
@@ -963,8 +959,16 @@ impl PeerManager {
...
-        self.client_num() < MAX_CLIENTS_PER_PROXY
+    pub fn can_accept_client(&self, ip: &IpAddr) -> bool {
+        let client_ips = self.peers
+            .values()
+            .filter_map(|peer| match *peer.state() {
+                            PeerState::Bootstrapper { ip, .. } |
+                            PeerState::Client(ip) => Some(ip),
+                            _ => None,
+                        })
+            .collect_vec();
```
ustulation: We avoid extra vector allocations:
```
let mut existing_client_count = 0;
self.peers.values().any(|p| match *p.state() {
    PeerState::BootStrapper { ip, .. } | PeerState::Client(ip) => {
            existing_clients += 1;
            given_ip == ip
    }
    _ => false
}) && existing_client < MAX_CLIENTS_PER_PROXY
```

src/peer_manager.rs
```diff
@@ -963,8 +959,16 @@ impl PeerManager {
     }
 
     /// Checks whether we can accept more clients.
-    pub fn can_accept_client(&self) -> bool {
-        self.client_num() < MAX_CLIENTS_PER_PROXY
+    pub fn can_accept_client(&self, ip: &IpAddr) -> bool {
```
ustulation: `IpAddr` is copyable - is it not better to pass copyable items by value instead of reference ? It's more optimised as pass by reference can prevent NRVO. Probably similar reasons why std libs also [choose this paradigm](https://doc.rust-lang.org/std/net/enum.SocketAddr.html#method.ip).

src/states/node.rs
```diff
@@ -479,24 +490,15 @@ impl Node {
                 return;
             }
         } else {
```
ustulation: We can combine the `if` below into a single line `else if` instead of an explicit split done here consuming 2 more lines of code and not buying anything (i.e. has no differing side-effect)

Fraser999: There are two `if`s inside the single `else` block though... maybe I'm missing something?

src/states/node.rs
Fraser999: Well, maybe `peek_keys()` and `keys()`, but can't say I'm too bothered either way.

src/peer_manager.rs
ustulation: This one is still lacking - although the else has been added below with log or panic, it does not cover the case where we don't find `pub_id` in `self.peers` at all - previously we panicked on that, not anymore. So needs done. Better to retain the previous format and just add the `ip` part to it ?

src/states/node.rs
ustulation: Ah no, you are right, i couldn't guage the structure properly in the diff.

PR #1487, opened 2017-07-04
Add few proxy throughput logs

Viv-Rajkumar:
- Proxy throughput per sec
- Total/Current per client throughput amounts

### DIFF
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -157,7 +157,12 @@ pub enum PeerState {
     /// We are connected - via a tunnel if the field is `true`.
     Connected(bool),
     /// We are the proxy for the client
-    Client(IpAddr),
+    Client {
+        /// Client IP
+        ip: IpAddr,
+        /// Traffic relayed for Client
+        traffic: u64,
+    },
     /// We are the proxy for the joining node
     JoiningNode,
     /// We are approved and routing to that peer.
@@ -281,7 +286,7 @@ impl Peer {
             PeerState::Bootstrapper { .. } |
             PeerState::JoiningNode |
             PeerState::Proxy |
-            PeerState::Client(_) => Some(false),
+            PeerState::Client { .. } => Some(false),
             PeerState::Connected(is_tunnel) => Some(is_tunnel),
             PeerState::Candidate(conn) |
             PeerState::Routing(conn) => Some(conn.is_tunnel()),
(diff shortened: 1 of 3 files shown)


src/peer_manager.rs
```diff
@@ -962,14 +967,40 @@ impl PeerManager {
...
+            .get_mut(pub_id)
+            .map(|peer| if let PeerState::Client {
+                            ip,
+                            traffic: old_traffic,
+                        } = *peer.state() {
+                     let new_traffic = old_traffic + added_bytes;
+                     let log_mult = 10 * 1024 * 1024;
+                     let old_rem = old_traffic % log_mult;
+                     let new_rem = new_traffic % log_mult;
+                     if new_rem < old_rem || added_bytes / log_mult > 0 {
```
maqi: will `if new_traffic % log_mult < added_bytes {}` do the same job?

src/peer_manager.rs
```diff
@@ -962,14 +967,40 @@ impl PeerManager {
...
+    /// Updates the given clients total traffic amount.
+    pub fn add_client_traffic(&mut self, pub_id: &PublicId, added_bytes: u64) {
+        let self_pfx = format!("{:?}", self);
+        let _ = self.peers
+            .get_mut(pub_id)
+            .map(|peer| if let PeerState::Client {
+                            ip,
+                            traffic: old_traffic,
+                        } = *peer.state() {
+                     let new_traffic = old_traffic + added_bytes;
```
maqi: i know the chance will be very tiny as we are using u64 for the traffic.
however, shall we still consider using `wrapping_add` ?

src/peer_manager.rs
```diff
@@ -157,7 +157,12 @@ pub enum PeerState {
     /// We are connected - via a tunnel if the field is `true`.
     Connected(bool),
     /// We are the proxy for the client
-    Client(IpAddr),
+    Client {
+        /// Client IP
+        ip: IpAddr,
+        /// Traffic relayed for Client
```
maqi: it's actually the `charge` that rate_limiter charges, for get requests, client will be over charged (even consider the traffic of response)
for put requests, the response traffic is not counted in.

maybe the comment can be modified to make it clear?

Viv-Rajkumar:
Closing this for now. Should come in as part of the whitelist removal PR later on

PR #1491, opened 2017-07-10
Allow mock-crust network to support multiple nodes/clients with the same IPs

Fraser999:


src/mock_crust/tests.rs
```diff
@@ -262,3 +262,19 @@ fn drop() {
...
+#[test]
+fn gen_endpoint_with_ip() {
+    let min_section_size = 8;
+    let network = Network::<PublicId>::new(min_section_size, None);
+    for _ in 0..258 {
+        let handle0 = network.new_service_handle(None, None);
+        let endpoint0 = handle0.endpoint();
+        let ip0 = to_socket_addr(&endpoint0).ip();
+        for _ in 0..10 {
+            let endpoint1 = network.gen_endpoint_with_ip(&ip0);
```
maqi: being a test, do we need to confirm this new generated endpoint1 is unique (i.e. different to endpoint0 and any previous generated endpoint1 for this ip)?

Fraser999: Good point.  I actually thought `network.new_service_handle()` (which we call in the next line) asserted that we have unique endpoints, but it doesn't - it just [prints a debug-level message](https://github.com/Fraser999/routing/blob/master/src/mock_crust/support.rs#L80-L86).  I think it'd be better to change that behaviour and leave this test as is - do you agree?

PR #1493, opened 2017-07-14
(WIP) feat/mutable-data: make mutable operations idempotent

madadam:
**DO NOT MERGE** yet. Not decided whether we want this.

### DIFF
--- a/src/data/mutable_data.rs
+++ b/src/data/mutable_data.rs
@@ -332,53 +332,45 @@ impl MutableData {
         let mut new_data = self.data.clone();
 
         for (key, val) in insert {
-            if new_data.contains_key(&key) {
-                return Err(ClientError::EntryExists);
+            match new_data.entry(key) {
+                Entry::Occupied(entry) => {
+                    if val != *entry.get() {
+                        return Err(ClientError::EntryExists);
+                    }
+                }
+                Entry::Vacant(entry) => {
+                    let _ = entry.insert(val);
+                }
             }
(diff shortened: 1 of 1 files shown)


src/data/mutable_data.rs
```diff
@@ -332,53 +332,45 @@ impl MutableData {
...
-            }
-            let version_valid = if let Entry::Occupied(mut oe) = new_data.entry(key.clone()) {
-                if val.entry_version != oe.get().entry_version + 1 {
-                    false
-                } else {
-                    let _prev = oe.insert(val);
-                    true
+            match new_data.entry(key) {
+                Entry::Occupied(mut entry) => {
+                    if val.entry_version == entry.get().entry_version + 1 {
```
maqi: say client sends a sequence of updates: update(v_1), update(v_2), update(v_1)
it will get `invalidSuccessor` on the second `update(v_1)` request.

what's the `idempotent` for this case?

src/data/mutable_data.rs
```diff
@@ -332,53 +332,45 @@ impl MutableData {
...
+                        let _ = entry.insert(val);
+                    } else if val != *entry.get() {
+                        return Err(ClientError::InvalidSuccessor);
+                    }
```
maqi: the idempotent operation still causes the later on `mem::replace` and `validate` call, can those be skipped?

ustulation:
We are going for a different paradigm that does not affect routing and vaults: MAID-2146 in JIRA.

PR #1498, opened 2017-07-27
Check before block connection

maqi:
feat/peer_manager: update RoutingConnection-(Proxy/JoiningNode) to RoutingConnection::Direct after 900s time_out
fix/aggressive_churn: check demotable peers before block connection

### DIFF
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -981,13 +981,31 @@ impl PeerManager {
             None
         };
 
+        let mut demote_peers = Vec::new();
         let expired_peers = self.peers
             .values()
-            .filter(|peer| peer.is_expired())
+            .filter(|peer| {
+                match peer.state {
+                    PeerState::Routing(RoutingConnection::JoiningNode(timestamp)) |
+                    PeerState::Routing(RoutingConnection::Proxy(timestamp)) => {
+                        if timestamp.elapsed() >= Duration::from_secs(JOINING_NODE_TIMEOUT_SECS) {
+                            demote_peers.push(*peer.pub_id());
+                        }
+                    }
+                    _ => (),
+                }
+                peer.is_expired()
+            })
             .map(Peer::pub_id)
             .cloned()
             .chain(remove_candidate)
             .collect_vec();
+
+        for id in &demote_peers {
+            if let Some(peer) = self.peers.get_mut(id) {
+                peer.state = PeerState::Routing(RoutingConnection::Direct)
+            }
+        }
 
         for id in &expired_peers {
             let _ = self.remove_peer(id);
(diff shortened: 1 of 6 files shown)


src/peer_manager.rs
```diff
@@ -981,14 +982,33 @@ impl PeerManager {
             None
         };
 
+        let mut demote_peers = Vec::new();
         let expired_peers = self.peers
             .values()
-            .filter(|peer| peer.is_expired())
+            .filter(|peer| {
+                if peer.is_expired() {
```
Viv-Rajkumar: wouldn't this always return `false` for `PeerState::Routing`?

ustulation: Will this even work ? If a peer is already in Peer-state of Routing, it will never be flagged as expired (that check should be bypassed else it would get removed below even in previous code) hence will never enter this, thus domotion logic will not work ?

Or does it work in some other way ?

src/peer_manager.rs
```diff
@@ -981,14 +982,33 @@ impl PeerManager {
...
+                    }
+                }
+                false
+            })
             .map(Peer::pub_id)
             .cloned()
             .chain(remove_candidate)
             .collect_vec();
 
+        for id in &demote_peers {
```
ustulation: The term demotion does not sound right - it's neither a demotion or a promotion in the sense that it does not get any more special status (it's already a routing connection). Rather it's slightly a promotion if anything that now it's allowed to act as a tunnel to its (previously) proxy/joining-node. So maybe should be named something different (e.g. `update_to_routing_direct` or something - see if you can come up with a better name).

src/peer_manager.rs
ustulation: eh @Viv-Rajkumar 's already commented on this.

src/peer_manager.rs
```diff
@@ -1620,6 +1639,27 @@ impl PeerManager {
...
+                        return true;
+                    }
+                    _ => return false,
+                }
+            } else {
+                false
+            })
+            .cloned()
+            .collect();
+        !(&updatable_peers - excludes).is_empty()
```
ustulation: Wow this is nice - didn't know such a functionality existed :+1:

tests/mock_crust/churn.rs
```diff
@@ -550,13 +572,19 @@ fn aggressive_churn() {
...
         // Or be rejected when the proxy node's RT is not large enough due to a lost tunnel.
         // In that case, a restart of candidate shall be carried out.
         if let Some(added_index) =
-            add_node_and_poll(&mut rng, &network, &mut nodes, min_section_size)
+            add_node_and_poll(
+                &mut rng,
+                &network,
+                &mut nodes,
+                min_section_size,
+                dropped_nodes,
```
ustulation: this is just for me to understand more: Do we explicitly need to exclude the dropped nodes ? Won't they already disappear from the peer-mgr if dropped ?

maqi: when we carry out check, we haven't poll yet, hence they are still in the RT.

if we don't exclude them, there will be un-necessary advance when only the dropped nodes being the RoutingConnection-proxy/joining

ustulation: ah right ok

PR #1499, opened 2017-07-28
MAID-2099: refactor/config: update config-handling

Fraser999:
This adds a new 'min_section_size' dev config option, and the Vault 'allow_multiple_lan_nodes' dev
option has been moved to Routing in this commit too.

The mock-crust Node and Client now make use of the 'min_section_size' config option rather than
passing this via the service handle.

Finally, a test for the sample config file has been added.


src/states/node.rs
```diff
@@ -1995,8 +1995,16 @@ impl Node {
 
     /// Informs our peers that our section's member list changed. If `dst_prefix` is `Some`, only
     /// tells that section, otherwise tells all connected sections, including our own.
```
maqi: shall we have a comment here for the new `allow_small_sections`?

tests/mock_crust/utils.rs
```diff
@@ -62,6 +62,14 @@ pub fn gen_range_except<T: Rng>(
     x
 }
 
+fn create_config(network: &Network<PublicId>) -> Config {
+    Config {
+        dev: Some(DevConfig {
+            min_section_size: Some(network.min_section_size()),
+            ..DevConfig::default()
```
ustulation: This `..` prefix is new to me - what does it mean ?

src/rate_limiter.rs
```diff
@@ -182,45 +188,51 @@ mod tests {
...
-        let payload = unwrap!(serialisation::serialise(
-            &UserMessage::Request(Request::GetIData {
-                name: rand::random(),
-                msg_id: MessageId::new(),
-            }),
-        ));
-        let hash = sha3_256(&payload);
-        (payload, hash)
-    }
+    static LARGE_MESSAGE: [u8; CLIENT_CAPACITY as usize] = [0; CLIENT_CAPACITY as usize];
```
ustulation: Does this have to ba static instead of a const ?

tests/mock_crust/utils.rs
Fraser999: Means just fill in the rest of the uninitialised fields with those from the thing after the `..`.

src/rate_limiter.rs
Fraser999: I was just trying to save resources slightly and avoid having a bunch of duplicated large arrays (which I think would happen if we used `const` instead, right?)

ustulation: hmm - interesting and also sounds correct to me :+1:

src/states/node.rs
Fraser999: Well there was a comment added to `RoutingTable::check_invariant()` which explains it OK I guess?  Or do you think we should still add a bit more info here too?

PR #1503, opened 2017-08-07
MAID-2218 Reconnects disconnected client

maqi:


tests/mock_crust/client_restrictions.rs
```diff
@@ -79,3 +79,42 @@ fn only_one_client_per_ip() {
...
+
+    // Client will get rejected as network not having enough nodes.
+    let mut clients =
+        vec![
+            TestClient::new_with_full_id(&network, config.clone(), None, full_id.clone()),
+        ];
+    let _ = poll_all(&mut nodes, &mut clients);
+    expect_next_event!(unwrap!(clients.last_mut()), Event::Terminate);
+
+    let _ = clients.remove(0);
```
ustulation: Can't we collect the removed client and re-insert it [here](https://github.com/maidsafe/routing/pull/1503/files#diff-99efcd5d613235a794676b45dbedbfadR112) instead of re-creating a new one ? Or are we required to create a new one with `TestClient::new...` which does a bunch of other things that we requrie ?

Viv-Rajkumar: Yeh might be better just re creating the client here to be safe, as we might have to re-init the state machine and other objects which might be in an incompatible state after firing the terminate event to re-use later on.

PR #1504, opened 2017-08-08
MAID-2197 Resend parts

maqi:


tests/mock_crust/client_restrictions.rs
```diff
@@ -118,3 +120,49 @@ fn reconnect_disconnected_client() {
...
+
+/// Confirming the number of user message parts being sent in case of exceeding limit.
+#[test]
+fn resending_parts_on_exceeding_limit() {
+    let network = Network::new(MIN_SECTION_SIZE, None);
+    let mut rng = network.new_rng();
+    let mut nodes = create_connected_nodes(&network, MIN_SECTION_SIZE);
+    let mut clients = create_connected_clients(&network, &mut nodes, 1);
+
+    let mut datas = vec![
```
ustulation: the plural for data is data itself - singular would be datum but rarely used. Maybe data_coll for collection then `for data in data_coll {` below in the for loop ?

tests/mock_crust/utils.rs
```diff
@@ -261,13 +261,34 @@ impl TestClient {
...
+        network: &Network<PublicId>,
+        bootstrap_config: Option<BootstrapConfig>,
+        endpoint: Option<Endpoint>,
+        duration: Duration,
+    ) -> Self {
+        let full_id = FullId::new();
+        Self::new_with_all(network, bootstrap_config, endpoint, full_id, duration)
+    }
+
+    fn new_with_all(
```
ustulation: better to call this `new_impl` if other ctors are calling it. `new_with_all` is difficult to follow while reading.

PR #1506, opened 2017-08-08
fix/bootstrap: enforce one client per IP only on bootstrap request

ustulation:
Due to the nature of parallel bootstrap attempts by crust it can happen that two clients with same IP successfully trigger a bootstrap accept in the same proxy but one of them decides to choose another proxy. That client will have disconnected from this proxy but the other client, C, could have decided to choose this one. So C gets a bootstrap connect. However if C was the latter of the two for this proxy and if this proxy hasn't yet realised that the first client has disconnected before it got the bootstrap accept for C, it is going to reject C on basis of one-client-per-IP rule. So C will get a disconnect event after it has received a bootstrap connect.
Moving this check to Bootstrap request should prevent this as only the ones who are interested in this proxy will be sending the bootstrap request and hence get promoted to a client.

### DIFF
--- a/src/states/node.rs
+++ b/src/states/node.rs
@@ -543,20 +543,6 @@ impl Node {
                 self.ban_and_disconnect_peer(&pub_id);
                 return;
             }
-            if !self.peer_mgr.can_accept_client(ip) {
-                debug!(
-                    "{:?} Client {:?} rejected: We cannot accept more clients.",
-                    self,
-                    pub_id
-                );
-                self.send_direct_message(
-                    pub_id,
-                    DirectMessage::BootstrapResponse(Err(BootstrapResponseError::ClientLimit)),
-                );
-                self.disconnect_peer(&pub_id, None);
-                let _ = self.dropped_clients.insert(pub_id, ());
-                return;
-            }
         }
         self.peer_mgr.insert_peer(Peer::new(
             pub_id,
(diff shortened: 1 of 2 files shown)


src/states/node.rs
```diff
@@ -1722,13 +1710,37 @@ impl Node {
...
+            debug!(
+                "{:?} Can't get IP address of bootstrapper {:?}.",
+                self,
+                pub_id
+            );
+            self.disconnect_peer(&pub_id, None);
+            return Err(RoutingError::UnknownConnection(pub_id));
+        };
+
+        if peer_kind == CrustUser::Client && !self.peer_mgr.can_accept_client(ip) {
```
maqi: is that the above `fetch ip` operation shall only happen for `peer_kind == CrustUser::Client` ?
and, when disconnecting peer (reject or fail to fetch ip), it shall be added into dropped_clients ?

src/states/node.rs
```diff
@@ -1722,13 +1708,38 @@ impl Node {
             return Err(RoutingError::UnknownConnection(pub_id));
         };
 
+        if peer_kind == CrustUser::Client {
+            let ip = self.crust_service.get_peer_ip_addr(&pub_id).map_err(|e| {
+                debug!(
+                    "{:?} Can't get IP address of bootstrapper {:?} : {:?}",
+                    self,
+                    pub_id,
+                    e
```
maqi: i think `error` or `err` is preferred?

PR #1508, opened 2017-08-09
MAID-2198: improve rate limiter

michaelsproul:
1. Refund on overcharge for get response
1. Soft-limit for clients - `max(MAX_IMMUT_DATA_SIZE, MAX_SOFT_CAP/clients_in_rate_limiter)`

### DIFF
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -625,6 +625,8 @@ pub enum MessageContent {
     UserMessagePart {
         /// The hash of this user message.
         hash: Digest256,
+        /// The unique message ID of this user message.
+        msg_id: MessageId,
         /// The number of parts.
         part_count: u32,
         /// The index of this part.
@@ -904,6 +906,7 @@ impl UserMessage {
     pub fn to_parts(&self, priority: u8) -> Result<Vec<MessageContent>, RoutingError> {
         let payload = serialise(self)?;
         let hash = sha3_256(&payload);
+        let msg_id = *self.message_id();
         let len = payload.len();
         let part_count = (len + MAX_PART_LEN - 1) / MAX_PART_LEN;
 
@@ -911,13 +914,14 @@ impl UserMessage {
             (0..part_count)
                 .map(|i| {
                     MessageContent::UserMessagePart {
-                        hash: hash,
+                        hash,
+                        msg_id,
                         part_count: part_count as u32,
                         part_index: i as u32,
                         cacheable: self.is_cacheable(),
                         payload: payload[(i * len / part_count)..((i + 1) * len / part_count)]
                             .to_vec(),
-                        priority: priority,
+                        priority,
                     }
                 })
                 .collect(),
@@ -963,6 +967,7 @@ impl UserMessage {
         }
     }
 
+    /// The unique message ID of this `UserMessage`.
     pub fn message_id(&self) -> &MessageId {
         match *self {
             UserMessage::Request(ref request) => request.message_id(),
@@ -1212,6 +1217,7 @@ mod tests {
             .map(|(i, msg)| match msg {
                 MessageContent::UserMessagePart {
                     hash,
+                    msg_id,
                     part_count,
                     part_index,
                     payload,
@@ -1219,6 +1225,7 @@ mod tests {
                     cacheable,
                 } => {
                     assert_eq!(msg_hash, hash);
+                    assert_eq!(user_msg.message_id(), &msg_id);
                     assert_eq!(3, part_count);
                     assert_eq!(i, part_index as usize);
                     assert_eq!(42, priority);
--- a/src/node.rs
+++ b/src/node.rs
@@ -629,6 +629,11 @@ impl Node {
     pub fn get_user_msg_parts_count(&self) -> u64 {
         self.machine.current().get_user_msg_parts_count()
     }
+
+    /// Get the rate limiter's bandwidth usage map.
+    pub fn get_clients_usage(&self) -> BTreeMap<IpAddr, u64> {
+        unwrap!(self.machine.current().get_clients_usage())
+    }
 }
 
 #[cfg(feature = "use-mock-crust")]
--- a/src/rate_limiter.rs
+++ b/src/rate_limiter.rs
@@ -21,28 +21,35 @@ use error::RoutingError;
 use fake_clock::FakeClock as Instant;
 use itertools::Itertools;
 use maidsafe_utilities::serialisation::{self, SerialisationError};
-use messages::UserMessage;
+use messages::{MAX_PART_LEN, UserMessage};
 use sha3::Digest256;
 use std::cmp;
 use std::collections::BTreeMap;
+use std::collections::btree_map::Entry;
 use std::mem;
 use std::net::IpAddr;
 #[cfg(not(feature = "use-mock-crust"))]
 use std::time::Instant;
+use types::MessageId;
 
 /// The number of bytes per second the `RateLimiter` will "leak".
 const RATE: f64 = 5.0 * 1024.0 * 1024.0;
-/// The maximum number of bytes a single client is allowed to have in the `RateLimiter`.  This is
-/// slightly larger than `MAX_IMMUTABLE_DATA_SIZE_IN_BYTES` to allow for the extra bytes created by
-/// wrapping the chunk in a `UserMessage`, splitting it into parts and wrapping those in
-/// `RoutingMessage`s.
-const CLIENT_CAPACITY: u64 = MAX_IMMUTABLE_DATA_SIZE_IN_BYTES + 10240;
+/// The number of bytes per second of bandwidth distributed between all clients when demand is low.
+#[cfg(not(feature = "use-mock-crust"))]
+const SOFT_CAPACITY: u64 = 20 * 1024 * 1024;
+// Soft capacity has to be at least 2 * MIN_CLIENT_CAPACITY for the multi-client tests to work.
+#[cfg(feature = "use-mock-crust")]
+const SOFT_CAPACITY: u64 = 2 * MIN_CLIENT_CAPACITY;
+/// The minimum number of bytes per second of bandwidth supplied to any client.
+const MIN_CLIENT_CAPACITY: u64 = MAX_IMMUTABLE_DATA_SIZE_IN_BYTES + 10240;
 
 #[cfg(feature = "use-mock-crust")]
 #[doc(hidden)]
 pub mod rate_limiter_consts {
-    pub const CLIENT_CAPACITY: u64 = super::CLIENT_CAPACITY;
+    pub const SOFT_CAPACITY: u64 = super::SOFT_CAPACITY;
     pub const MAX_PARTS: u32 = ::messages::MAX_PARTS;
+    pub const MAX_PART_LEN: usize = ::messages::MAX_PART_LEN;
+    pub const MIN_CLIENT_CAPACITY: u64 = super::MIN_CLIENT_CAPACITY;
     pub const RATE: f64 = super::RATE;
 }
 
@@ -53,6 +60,9 @@ pub mod rate_limiter_consts {
 pub struct RateLimiter {
     /// Map of client IP address to their total bytes remaining in the `RateLimiter`.
     used: BTreeMap<IpAddr, u64>,
+    /// Initial charge amount by GET request message ID.
+    /// The IP address of the requesting peer is also tracked so that stale entries can be removed.
+    overcharged: BTreeMap<MessageId, (u64, IpAddr)>,
     /// Timestamp of when the `RateLimiter` was last updated.
     last_updated: Instant,
     /// Whether rate restriction is disabled.
@@ -63,6 +73,7 @@ impl RateLimiter {
     pub fn new(disabled: bool) -> Self {
         RateLimiter {
             used: BTreeMap::new(),
+            overcharged: BTreeMap::new(),
             last_updated: Instant::now(),
             disabled: disabled,
         }
@@ -71,18 +82,19 @@ impl RateLimiter {
     /// Try to add a message. If the message is a form of get request,
     /// `MAX_IMMUTABLE_DATA_SIZE_IN_BYTES` or `MAX_MUTABLE_DATA_SIZE_IN_BYTES` bytes will be used,
     /// otherwise the actual length of the `payload` will be used. If adding that amount will cause
-    /// the client to exceed its capacity (`CLIENT_CAPACITY`), then `Err(ExceedsRateLimit)` is
-    /// returned. If the message is invalid, `Err(InvalidMessage)` is returned (this probably
-    /// indicates malicious behaviour).
+    /// the client to exceed its capacity, then `Err(ExceedsRateLimit)` is returned. If the
+    /// message is invalid, `Err(InvalidMessage)` is returned (this probably indicates malicious
+    /// behaviour).
     pub fn add_message(
         &mut self,
         client_ip: &IpAddr,
         hash: &Digest256,
+        msg_id: &MessageId,
         part_count: u32,
         part_index: u32,
         payload: &[u8],
     ) -> Result<u64, RoutingError> {
-        let bytes_to_add = if part_index == 0 {
+        let (bytes_to_add, overcharged) = if part_index == 0 {
             use self::UserMessage::*;
             use Request::*;
             match serialisation::deserialise::<UserMessage>(payload) {
@@ -91,7 +103,7 @@ impl RateLimiter {
                         return Err(RoutingError::InvalidMessage);
                     }
                     match request {
-                        GetIData { .. } => MAX_IMMUTABLE_DATA_SIZE_IN_BYTES,
+                        GetIData { .. } => (MAX_IMMUTABLE_DATA_SIZE_IN_BYTES, true),
                         GetAccountInfo { .. } |
                         GetMData { .. } |
                         GetMDataVersion { .. } |
@@ -102,7 +114,7 @@ impl RateLimiter {
                         GetMDataValue { .. } |
                         ListMDataPermissions { .. } |
                         ListMDataUserPermissions { .. } |
-                        ListAuthKeysAndVersion { .. } => MAX_MUTABLE_DATA_SIZE_IN_BYTES,
+                        ListAuthKeysAndVersion { .. } => (MAX_MUTABLE_DATA_SIZE_IN_BYTES, true),
                         PutIData { .. } |
                         PutMData { .. } |
                         MutateMDataEntries { .. } |
@@ -110,7 +122,7 @@ impl RateLimiter {
                         DelMDataUserPermissions { .. } |
                         ChangeMDataOwner { .. } |
                         InsAuthKey { .. } |
-                        DelAuthKey { .. } => payload.len() as u64,
+                        DelAuthKey { .. } => (payload.len() as u64, false),
                         Refresh(..) => return Err(RoutingError::InvalidMessage),
                     }
                 }
@@ -122,11 +134,11 @@ impl RateLimiter {
                     if part_count == 1 {
                         return Err(RoutingError::InvalidMessage);
                     }
-                    payload.len() as u64
+                    (payload.len() as u64, false)
                 }
             }
         } else {
-            payload.len() as u64
+            (payload.len() as u64, false)
         };
 
         if self.disabled {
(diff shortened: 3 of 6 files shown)


src/node.rs
```diff
@@ -629,6 +629,11 @@ impl Node {
     pub fn get_user_msg_parts_count(&self) -> u64 {
         self.machine.current().get_user_msg_parts_count()
     }
+
+    /// BLAH
```
michaelsproul: I should get rid of this.

src/rate_limiter.rs
```diff
@@ -274,114 +557,116 @@ mod tests {
...
+        ));
+        assert!(sized_message_cannot_be_added(
+            SOFT_CAPACITY / 2,
+            &mut rate_limiter,
+            &client1,
+        ));
+        assert!(small_message_can_be_added(&mut rate_limiter, &client1));
     }
 
+    // FIXME(michael): this is slightly broken
```
michaelsproul: actually, seems to be fixed now, I just need to get rid of this comment

tests/mock_crust/client_restrictions.rs
```diff
@@ -215,3 +219,132 @@ fn resend_over_load() {
...
+    }
+
+    // Max bandwidth used is C + C/2 + C/3 + C/4 + ... + C/2 = C * HarmonicNumber(N)
+    // FIXME(michael): this isn't really the max bandwidth, it can get higher somehow,
+    // but I haven't worked out how yet.
+    let cap = ::std::cmp::max(
+        SOFT_CAPACITY * harmonic_number(clients.len()).ceil() as u64,
+        MIN_CLIENT_CAPACITY * clients.len() as u64,
+    );
+    assert!(*total_usage <= cap, "{} > {}", total_usage, cap);
```
michaelsproul: This is the assertion that fails about 1/100 runs.

tests/mock_crust/client_restrictions.rs
```diff
@@ -215,3 +219,132 @@ fn resend_over_load() {
...
+    total_usage: &mut u64,
+    _per_client_cap: u64,
+) {
+    // `rejected_reqs` contains only the clients whose request got rejected.
+    // Needs to confirm such rejection is valid. However, if it is the total usage reaching the
+    // cap, the usage of each client could be much less than the cap.
+    if (*total_usage + MAX_IMMUTABLE_DATA_SIZE_IN_BYTES) <= RATE as u64 {
+        for ip in rejected_reqs.values() {
+            let client_usage = unwrap!(clients_usage.get(ip));
+            assert!(*client_usage >= MIN_CLIENT_CAPACITY);
```
michaelsproul: This whole verification step is trickier with the new rules. The concept of a "per client cap" is no longer really meaningful because the caps are constantly changing. This assert just says that if a client had a request rejected, it's because its usage was already above the min capacity, even if its actually capacity at the time the request was rejected may have been much larger.

tests/mock_crust/client_restrictions.rs
michaelsproul: It seems to fail on about the 8th iteration each time, which makes me think that the bucket fills up with junk as more iterations run, and because there's no hard cap, the usage keeps growing (potentially) indefinitely.

maqi: could that because there is multiple clients having gets at the same time on top of the max_cap?

michaelsproul: yeah...

tests/mock_crust/client_restrictions.rs
maqi: yes. I think with the new rule, an `expected per-client-cap` shall be calculated in each iteration.

src/rate_limiter.rs
```diff
@@ -136,16 +145,103 @@ impl RateLimiter {
...
+        msg_id: &MessageId,
+        part_count: u32,
+        part_index: u32,
+        payload: &[u8],
+    ) -> Option<u64> {
+        use Response::*;
+
+        // Check that the response isn't a single-part response that we never overcharged for.
+        // This prevents a malicious client from discounting their GET by preceding
+        // it with a PUT with the same `msg_id`.
```
maqi: any other small charged request, say SetMDataUserPermissions or InsAuthKey, can achieve the same effect, right?

src/messages/mod.rs
```diff
@@ -625,6 +625,8 @@ pub enum MessageContent {
     UserMessagePart {
         /// The hash of this user message.
         hash: Digest256,
+        /// The unique message ID of this user message.
+        msg_id: MessageId,
```
maqi: as the msg_id has been moved into the UserMessagePart,
can the msg_id in Response/Request be removed?
or this will be a too big change and shall leave to be done later on?

src/rate_limiter.rs
```diff
@@ -181,89 +277,276 @@ impl RateLimiter {
...
+        }
+    }
+
+    // Send a single `UserMessagePart` for a response to the rate limiter for discounting.
+    fn discount_user_msg_part(
+        rate_limiter: &mut RateLimiter,
+        client: &IpAddr,
+        msg: &MessageContent,
+    ) -> Option<u64> {
+
```
maqi: does this extra line break required?

src/rate_limiter.rs
```diff
@@ -181,89 +277,276 @@ impl RateLimiter {
...
+    }
+
+    #[test]
+    fn overcharge_correction() {
+        let mut rate_limiter = RateLimiter::new(false);
+        let client = IpAddr::from([0, 0, 0, 0]);
+
+        let approx_num_parts = 10;
+        let data_size = approx_num_parts * MAX_PART_LEN;
+        let data = ImmutableData::new(vec![0; data_size]);
```
maqi: better not to create an `exact sized` data, instead something like : 
```rust
let data_size = rng.gen();
let num_parts = data_size / MAX_PART_LEN```
If we can't refund accurately, it shall be reflected in the testing code.

File: src/rate_limiter.rs [hunk 9]
```diff
@@ -181,89 +277,276 @@ impl RateLimiter {
...
+        );
+        // Then the get request.
+        assert!(
+            add_user_msg_part(
+                &mut rate_limiter,
+                &client,
+                unwrap!(get_request_parts.first()),
+            ).is_ok()
+        );
+        // Now if the put request comes back, the proxy *should not* apply a discount for it,
```
maqi: `put response`?

src/rate_limiter.rs
```diff
@@ -181,89 +277,276 @@ impl RateLimiter {
...
+            msg_id,
+        });
+        let get_request_parts = unwrap!(get_request.to_parts(0));
+        let get_response = UserMessage::Response(Response::GetIData {
+            res: Ok(get_data),
+            msg_id,
+        });
+        let get_response_parts = unwrap!(get_response.to_parts(0));
+
+        // Put request hits the rate limiter first.
```
maqi: what will happen if the get request arrives before the put?
looks like the client will ban (return with InvalidMessage).
shall it be confirmed in the test?

src/rate_limiter.rs
```diff
@@ -181,89 +277,276 @@ impl RateLimiter {
...
+        // Now if the put request comes back, the proxy *should not* apply a discount for it,
+        // even though it has the same message ID as the get that we just overcharged for.
+        assert!(
+            discount_user_msg_part(
+                &mut rate_limiter,
+                &client,
+                unwrap!(put_response_parts.first()),
+            ).is_none()
+        );
+        // The discount should still correctly be applied for the get response.
```
maqi: is it a `refund` not `discount`?  same to the other places

src/rate_limiter.rs
```diff
@@ -274,114 +557,116 @@ mod tests {
...
+        println!("num_clients = {}", num_clients);
+        let clients: Vec<_> = (0..num_clients as u8)
+            .map(|i| IpAddr::from([i, i, i, i]))
+            .collect();
 
-        // Client 1 adds a large message then after `wait_millis`, Client 2 does likewise.
-        assert_large_message_can_be_added(&mut rate_limiter, &client1);
-        FakeClock::advance_time(wait_millis);
-        assert_large_message_can_be_added(&mut rate_limiter, &client2);
+        // All clients put a MIN_CLIENT_CAPACITY-sized message.
```
maqi: `MIN_CLIENT_CAPACITY sized` ?  also I think MIN_CLIENT_CAPACITY shall be surrounded by tick?

src/states/node.rs
```diff
@@ -1640,6 +1642,31 @@ impl Node {
         }
     }
 
+    fn correct_rate_limits(&mut self, ip: &IpAddr, msg: &RoutingMessage) -> Option<u64> {
+
```
maqi: extra line break?

tests/mock_crust/client_restrictions.rs
```diff
@@ -215,3 +219,132 @@ fn resend_over_load() {
...
+    let wait_millis = (leaky_rate * 1000) / RATE as u64;
+    FakeClock::advance_time(wait_millis);
+    *total_usage = total_usage.saturating_sub(leaky_rate);
+}
+
+/// Connects random number of clients to the same proxy node and sends get requests.
+/// Expect some requests will be blocked due to the rate limit.
+/// Expects the total capacity of the proxy will never be exceeded.
+#[test]
+fn rate_limit_proxy_random_clients() {
```
maqi: this integration test could be moved as a unit test.

src/rate_limiter.rs
michaelsproul: nope! I'll get rid of it

src/rate_limiter.rs
michaelsproul: yeah, refund is a better term! I'll change it

src/messages/mod.rs
michaelsproul: I think it's too big of a change for now. Having it duplicated in the Request/Response is only an extra 32 bytes per message, and we're already repeating the message ID up to 52 times per message which is ~1.6KB.

src/rate_limiter.rs
michaelsproul: Yeah, I'll clarify that the get/put is just an example. I think the code as written should prevent problems with other small messages too.

src/rate_limiter.rs
michaelsproul: I think the put will just be allowed through in this case. I'll add a test for it tomorrow

src/rate_limiter.rs
```diff
@@ -136,16 +145,103 @@ impl RateLimiter {
...
+
+        let amount_charged = match self.overcharged.remove(msg_id) {
+            Some(amount) => amount,
+            None => return None,
+        };
+
+        let deduction = amount_charged.saturating_sub(part_count as u64 * MAX_PART_LEN as u64);
+
+        self.used.get_mut(client_ip).map(|used| {
+            *used = used.saturating_sub(deduction);
```
ustulation: After the refund do we need to hold the client in rate limiter in all cases ? What if _used_ for client is 0 ? Or for e.g. [here](https://github.com/michaelsproul/routing/blob/944749dedc5a9c32cd958432d854e542e88be916/src/rate_limiter.rs#L271) when _used_ is less than the quota then clients are not re-inserted. So should this follow a similar logic ?

ustulation: Well it seems it might not skew the logic after all as [client allowance](https://github.com/michaelsproul/routing/blob/944749dedc5a9c32cd958432d854e542e88be916/src/rate_limiter.rs#L150) is calculated after the call to update() which would have removed it (sort of a refresh) if necessary - but might be safer to call update after the refund too.

Anyway that brings to two corollaries:
1. Maybe we can simply check for the existence of client in the overcharged container before trying to execute the refund logic - that way we can save deserialising the payload unnecessarily.
1. there can be a potential disconnect between overcharge-container and used-container. If the responses never arrive, things continue to remain in the overcharge-container forever. So each time we [refrain from re-inserting into the used-container](https://github.com/michaelsproul/routing/blob/944749dedc5a9c32cd958432d854e542e88be916/src/rate_limiter.rs#L270), we should additionally get rid of such a client entry from overcharged-container too (so maybe store extra info there to tie it to the client info in used-container ?) ;; (Or make overcharged-container an LRU but not sure it's worth it)

maqi: we don't have to as the update during next add_message will remove the empty/low_quota entry anyway.

also, to remove in case of `less than the quota` means to calculate the leaked bytes since last update and distribute it among entries, which is actually an `update` call it self.

maqi: >but might be safer to call update after the refund too.

ustulation: @maqi ah i ninja'd you on that - but i guess points (1) and (2) above are valid ?

maqi: yes, point 1 and 2 are valid

src/rate_limiter.rs
michaelsproul: @qi.ma, I didn't end up adding a test for this... should be OK though I reckon

src/rate_limiter.rs
```diff
@@ -21,28 +21,35 @@ use error::RoutingError;
...
 
 /// The number of bytes per second the `RateLimiter` will "leak".
 const RATE: f64 = 5.0 * 1024.0 * 1024.0;
-/// The maximum number of bytes a single client is allowed to have in the `RateLimiter`.  This is
-/// slightly larger than `MAX_IMMUTABLE_DATA_SIZE_IN_BYTES` to allow for the extra bytes created by
-/// wrapping the chunk in a `UserMessage`, splitting it into parts and wrapping those in
-/// `RoutingMessage`s.
-const CLIENT_CAPACITY: u64 = MAX_IMMUTABLE_DATA_SIZE_IN_BYTES + 10240;
+/// The number of bytes per second of bandwidth distributed between all clients when demand is low.
+#[cfg(not(any(test, feature = "use-mock-crust")))]
```
maqi: will `#[cfg(not(feature = "use-mock-crust"))]` and `#[cfg(feature = "use-mock-crust")]` do the same?

src/rate_limiter.rs
```diff
@@ -316,72 +648,66 @@ mod tests {
...
         let start = FakeClock::now();
-        for _ in 0..500 {
+        for _ in 0..200 {
             // Each client tries to add a large request and increments its count on success.
             for (client, count) in &mut clients_and_counts {
-                if rate_limiter
-                    .add_message(client, &hash, 1, 0, &payload)
-                    .is_ok()
-                {
+                if add_user_msg_part(&mut rate_limiter, client, &random_payload()).is_ok() {
```
maqi: the last advance of clock shall not be counted in the advanced_secs ?
could this be the reason of the `off by one` sometimes?

I changed the following off-by-one assertion back to assert_eq, and got a failing seed `Some([1052942881, 318284615, 1049359143, 887479472])`
once excluded the last advance of clock from the advanced_secs, the test can then pass properly.

tests/mock_crust/client_restrictions.rs
```diff
@@ -151,17 +152,26 @@ fn resend_parts_on_exceeding_limit() {
     }
     poll_and_resend(&mut nodes, &mut clients);
 
-    // `MAX_PARTS / 2` parts will be resent from client.
-    let expect_sent_parts = (2 * MAX_PARTS) as u64;
-    assert_eq!(
-        clients[0].inner.get_user_msg_parts_count(),
-        expect_sent_parts
+    // NOTE: these calculations are somewhat approximate – in reality, the user message
+    // parts seem to be around 20,363 bytes each, not 20,480, so an extra one can
```
maqi: the user message of immutable_data with max size also have a 48 bytes head.
and divided by MAX_PARTS (52), it means most part will be 20363, with 4 parts to be 20363
once change the calculation to be
```rust
let parts_allowed =
    (SOFT_CAPACITY as f64 /
      ((MAX_IMMUTABLE_DATA_SIZE_IN_BYTES + 48) as f64 /  MAX_PARTS as f64)) as u64;
```
the assert_eq can be put back (abs_diff no longer required)

same to the other test

tests/mock_crust/client_restrictions.rs
```diff
@@ -182,36 +192,45 @@ fn resend_over_load() {
...
+
     // `poll_and_resend` advance clock by 20 seconds (`ACK_TIME_OUT`), hence the message is expired
     // when handling the timeout for re-sending parts.
-    let expect_sent_parts = (2 * MAX_PARTS) as u64;
-    assert_eq!(
-        clients[0].inner.get_user_msg_parts_count(),
-        expect_sent_parts
+    let expect_sent_parts = total_data_parts;
+    assert!(
+        abs_diff(
```
maqi: this shall always be assert_eq

src/rate_limiter.rs
michaelsproul: (1) and (2) are both addressed now

tests/mock_crust/client_restrictions.rs
michaelsproul: Awesome, that works well. I've left a note in case this calculation breaks with changes to serialisation or message content.

tests/mock_crust/client_restrictions.rs
```diff
@@ -151,17 +152,26 @@ fn resend_parts_on_exceeding_limit() {
...
-    assert_eq!(
-        clients[0].inner.get_user_msg_parts_count(),
-        expect_sent_parts
+    // NOTE: these calculations are somewhat approximate – in reality, the user message
+    // parts seem to be around 20,363 bytes each, not 20,480, so an extra one can
+    // sometimes be crammed into `SOFT_CAPACITY`.
+    let total_data_parts = num_immutable_data * MAX_PARTS as u64;
+    let parts_allowed_first_time = SOFT_CAPACITY / MAX_PART_LEN as u64;
+    let parts_retried = total_data_parts - parts_allowed_first_time;
+    assert!(parts_retried > 0);
```
michaelsproul: I removed this assertion entirely as it was somewhat redundant. I wrote it initially just to check that I was calculating a meaningful non-zero value for `parts_retried`.

src/rate_limiter.rs
michaelsproul: Hmm.. I couldn't get this to work. I tried adding a condition to increment the clock on every iteration except the last one, but I found that it still failed occasionally. I also tried to increase the precision of the calculation by using the nanoseconds from the duration, but that also caused occasional failures.

I really don't think it matters if we're off by one in this test, because the leak rate is approximate in its very nature. Guaranteeing that the calculation is correct down to the last byte is probably overkill when all we _really_ want to test is that the rate limiter shares `RATE` approximately fairly between all the clients.

maqi: >I tried adding a condition to increment the clock on every iteration except the last one, but I found that it still failed occasionally

do you have the seed?

And, even we have to stay with the `off by one`, the `exclusion of the last clock advance` shall still be applied as this is definitely a bug.

michaelsproul: I didn't take note of the seed because I wasn't sure how to set the seed for a unit test. I'll try and re-create it now.

michaelsproul: Ok, found a seed: `[1741684212, 3786665780, 3544795958, 749758005]`. This is the branch: https://github.com/michaelsproul/routing/commit/f22ee509eed4f685f530913063e92d3b42333e3c. I only had to run the test in a loop 5 times to get a failure.

michaelsproul: I think it's not worth making it exact because we're dealing with an average rate, and a discrete time simulation of a continuous time process. I don't think leaving off the last clock advance is "right" any more than it is "wrong". If there are enough iterations, 1 second shouldn't make a difference... and if we don't want to worry about iterating 10,000 times, I think an off-by-one allowance is OK? Or am I missing something? I did a sketch of the process as I understand it:

https://sketch.io/render/sk-d3dad899d03169b273d562b70b3829e6.jpeg

src/rate_limiter.rs
```diff
@@ -173,97 +287,305 @@ impl RateLimiter {
...
+        } else {
+            panic!("message is not a UserMessagePart: {:?}", msg);
         }
     }
 
-    /// Checks that a single client cannot exceed its individual limit and that its throughput is
+    // Generate a single `UserMessagePart` for a random `GetIData` request.
+    fn random_payload() -> MessageContent {
+        let user_message = UserMessage::Request(Request::GetIData {
+            name: rand::random(),
```
maqi: better not to use rand::random ? why not use a passed in SeededRng ?

also, try to use SeededRng in other places of rand::random as well?

Fraser999:
Closed in favour of #1510.

PR #1510, opened 2017-08-15
MAID-2198: improve rate limiter

Fraser999:
1. Refund on overcharge for get response
1. Soft-limit for clients - `max(MAX_IMMUT_DATA_SIZE, MAX_SOFT_CAP/clients_in_rate_limiter)`

Incorporates #1508.

### DIFF
--- a/src/rate_limiter.rs
+++ b/src/rate_limiter.rs
@@ -20,6 +20,7 @@ use error::RoutingError;
 #[cfg(feature = "use-mock-crust")]
 use fake_clock::FakeClock as Instant;
 use itertools::Itertools;
+use lru_time_cache::LruCache;
 use maidsafe_utilities::serialisation::{self, SerialisationError};
 use messages::{MAX_PART_LEN, UserMessage};
 use sha3::Digest256;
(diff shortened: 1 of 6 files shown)


src/rate_limiter.rs
```diff
@@ -20,27 +20,39 @@ use error::RoutingError;
...
 const SOFT_CAPACITY: u64 = 20 * 1024 * 1024;
-// Soft capacity has to be at least 2 * MIN_CLIENT_CAPACITY for the multi-client tests to work.
+/// For the mock-crust tests, we want a small `SOFT_CAPACITY` in order to trigger more rate-limited
+/// rejections. This must be at least `2 * MIN_CLIENT_CAPACITY` for the multi-client tests to work.
 #[cfg(feature = "use-mock-crust")]
 const SOFT_CAPACITY: u64 = 2 * MIN_CLIENT_CAPACITY;
-/// The minimum number of bytes per second of bandwidth supplied to any client.
-const MIN_CLIENT_CAPACITY: u64 = MAX_IMMUTABLE_DATA_SIZE_IN_BYTES + 10240;
+/// Duration for which entries are kept in the `overcharged` cache, in seconds.
+const OVERCHARGED_TIMEOUT_SECS: u64 = 3600;
```
maqi: one hour?
I thought ACK_TIMEOUT (or 2x / 3x times of it) shall be enough?

src/rate_limiter.rs
```diff
@@ -813,4 +812,32 @@ mod tests {
...
+        let finish_time = FakeClock::now() + Duration::from_secs(OVERCHARGED_TIMEOUT_SECS + 60);
+        while FakeClock::now() < finish_time {
+            let name = XorName([0; XOR_NAME_LEN]);
+            let msg_id = MessageId::new();
+            let request = UserMessage::Request(Request::GetIData { name, msg_id });
+            let request_parts = unwrap!(request.to_parts(0));
+            let _ = add_user_msg_part(&mut rate_limiter, &client, &request_parts[0]);
+            FakeClock::advance_time(wait_millis);
+        }
+        assert!((rate_limiter.overcharged.len() as u64) <= max_overcharged_entries);
```
maqi: not sure about the config of executing this test with the previous code.
however, the current code shows that the number of entries in overcharged is maxed at the number of add_message_part calls, which is actually :
num_of_entries = (OVERCHARGED_TIMEOUT_SECS + 60) / (MAX_IMMUTABLE_DATA_SIZE_IN_BYTES * 100 / RATE) = (OVERCHARGED_TIMEOUT_SECS + 60) * RATE / (MAX_IMMUTABLE_DATA_SIZE_IN_BYTES * 100)
which will always less than max_overcharged_entries ?

i think max_overcharged_entries shall be : OVERCHARGED_TIMEOUT_SECS / wait_millis  ?

src/rate_limiter.rs
Fraser999: It might be enough, be we have plenty of room to play with here.  The entries are all pretty small and even at maximum throughput with just one client connected, the maximum number of entries in an hour would only be ~18,000.

Edit: @ustulation prefers 20secs too, so I'll go with the majority 😄

src/rate_limiter.rs
Fraser999: ~90% of the `add_user_msg_part()` calls will fail to reach the `overcharged` container though since they'll be rate-limit rejected.

maqi: yes, and that's even worse I think?
the assertion is actually not testing anything. as long as the rate_limiter not to create multiple entries on one add_msg_part call.

michaelsproul: @maqi:

I think your first calculation for `num_of_entries` is off by a factor of 1000 because `wait_millis` wasn't converted into seconds. Here's my working:

```
m = MAX_IMMUTABLE_DATA_SIZE_IN_BYTES
o = OVERCHARGED_TIMEOUT_SECS
r = RATE
w = wait_millis
```

Then the number of calls to `add_user_message_part` is:

```
n = 1000 * (o + 60) / w
n = 1000 * (o + 60) / (100m / r)
n = 10 * r * (o + 60) / m
```

So we have `n < max_overcharged_entries`, because `max_overcharged_entries = o * r / m`.

michaelsproul: I think `OVERCHARGED_TIMEOUT_SECS / wait_secs` is too large, because as Fraser said an entry will really only be added every `W = 10 * wait_secs`. However, even with that correction, I think we need to round up, because you can cram e.g. 5 events separated by waits of 10 seconds into a period that is a tiny bit longer than 40 seconds. I did another crummy drawing:

image

Alternatively, we can round up Fraser's original calculation (the test failed and was off-by-one when I changed the `OVERCHARGED_TIMEOUT_SECS` to 300).

michaelsproul: Haha, I mucked up the inequality in my previous comment. I've edited it now so that it correctly reads `n > max_overcharged_entries`.

michaelsproul: Test passes correctly with either of:

```
// Qi's approach (rounded up)
max_overcharged_entries = 1000 * OVERCHARGED_TIMEOUT_SECS / wait_millis + 1

// Fraser's original approach (rounded up)
max_overcharged_entries = OVERCHARGED_TIMEOUT_SECS * RATE as u64 / MAX_IMMUTABLE_DATA_SIZE_IN_BYTES + 1

```

michaelsproul: Hmm... `100 * OVERCHARGED_TIMEOUT_SECS / wait_millis + 1` seems to overestimate by a significant amount, e.g. 2501 entries when the timeout is 300 seconds. Fraser's calc (rounded-up/+1'd) gets the real value (2377).

maqi: my previous calculation is wrong. so Fraser's calc is correct (just need a round up)

Fraser999:
Closed in favour of #1511.

PR #1518, opened 2017-09-27
fix rate_limiter many client test

maqi:
revert the previous round fix;
the last iteration doesn't require offset


nbaksalyar:
Integrated into #1528

PR #1519, opened 2017-10-12
fix/node: remove un-necessary filter when handle section update

maqi:


Viv-Rajkumar:
Closing this as redundant with changes in fleming branch

PR #1524, opened 2017-12-04
work on vote and block

maqi:
vote:make validate_signature taking PeerId
block: add_vote, add_proof and get_block_state given a set of valid voters


src/block.rs
```diff
@@ -128,6 +138,24 @@ impl<T: Serialize + Clone> Block<T> {
...
+    /// Return the block state given a set of valid voters.
+    pub fn get_block_state(
+        &self,
+        valid_voters: &BTreeSet<PeerId>,
+    ) -> Result<BlockState, RoutingError> {
+        let total_age = valid_voters
+            .iter()
+            .map(|peer_id| peer_id.age() as usize)
+            .sum();
+        if self.proofs.len() == valid_voters.len() {
```
ustulation: You don't need total age calculation for this branch. Just put it on the top and return. Then get the total age followed by the rest of the logic.

src/block.rs
```diff
@@ -146,21 +175,28 @@ mod tests {
...
         let vote0 = Vote::new(&keys0.1, payload.clone()).unwrap();
-        assert!(vote0.validate_signature(&keys0.0));
+        assert!(vote0.validate_signature(&peer_id0));
         let vote1 = Vote::new(&keys1.1, payload.clone()).unwrap();
-        assert!(vote1.validate_signature(&keys1.0));
-        let proof0 = vote0.proof(&PeerId::new(1u8, keys0.0)).unwrap();
+        assert!(vote1.validate_signature(&peer_id1));
+        let peer_id00 = PeerId::new(1u8, keys0.0);
+        let peer_id10 = PeerId::new(rng.gen_range(0, 255), keys1.0);
+        let proof0 = vote0.proof(&peer_id00).unwrap();
```
Fraser999: All the `unwrap()`s should use `unwrap!()` instead.

PR #1529, opened 2018-02-08
Update CI scripts and fix clippy warnings

nbaksalyar:
Rust compiler version has been updated to latest stable (1.23.0) and clippy was upgraded to 0.0.179. This PR also fixes clippy lint checks that resulted in compilation errors.


tests/mock_crust/client_restrictions.rs
```diff
@@ -152,11 +152,12 @@ fn resend_parts_on_exceeding_limit() {
...
     poll_and_resend(&mut nodes, &mut clients);
 
-    let total_data_parts = num_immutable_data * MAX_PARTS as u64;
+    let total_data_parts = num_immutable_data * u64::from(MAX_PARTS);
     // NOTE: this calculation is approximate and relies on some hardcoded knowledge about
     // the size of serialised user messages.
     let user_msg_header = 48;
-    let part_size = (MAX_IMMUTABLE_DATA_SIZE_IN_BYTES + user_msg_header) as f64 / MAX_PARTS as f64;
+    let part_size = (MAX_IMMUTABLE_DATA_SIZE_IN_BYTES + user_msg_header) as f64 /
+        f64::from(MAX_PARTS);
```
maqi: I am still confused with the rule of using `as f64`  or `f64` .
I thought it shall be consistent.
just a complain regarding clippy, not this change itself.

nbaksalyar: @maqi I believe the reasoning is that if we change the type, the conversion would become lossy: https://rust-lang-nursery.github.io/rust-clippy/v0.0.179/index.html#cast_lossless

maqi: Yeah, I know.
I just want to say then why not use `f64::from()` for all

Fraser999: That's the point - it's better to _not_ use `as` since it might be a lossy conversion.  If we see `::from()`, then we know it's not lossy.  This makes it easier to rationalise about the code, e.g.
```rust
let big = u64::max_value();
// let float = f64::from(big); → won't compile as this is a lossy conversion
let float = big as f64;  // lossy conversion
assert_ne!(float as u64, big);

let small = u32::max_value();
let float = f64::from(small);  // guaranteed not lossy
assert_eq!(float as u32, small);
```

maqi: shall we create a separate task to tackle this, i.e. replace all `as` conversion with `from` ?
However, I am just wondering does it mean we will have to use u32 instead of u64, if we do want to convert it into f64 in some situation.

madadam: >  I am just wondering does it mean we will have to use u32 instead of u64, if we do want to convert it into f64 in some situation
It does not. You can convert `u64` to `f64` just fine (via `as`), it's just such conversion is lossy (not exact). The clippy lint is triggered only if you use `as` in situations where there is a lossless conversion via `from` available.

PR #1531, opened 2018-02-09
Merge Prefix and VersionedPrefix (MAID-2451)

madadam:


src/routing_table/mod.rs
```diff
@@ -571,38 +538,29 @@ impl RoutingTable {
...
-                     update is for a different version: {}",
-                    self.our_name,
-                    prefix,
-                    entry.get().0,
-                    version
-                );
-                return (result, None);
-            }
-            Entry::Occupied(entry) => entry.remove(),
+        let to_split = if let Some(section) = self.sections.remove(&prefix) {
```
maqi: why now we can carry out remove directly without a version check ?

src/routing_table/mod.rs
```diff
@@ -968,78 +912,67 @@ impl RoutingTable {
...
-                self.our_version,
                 version
             );
             return Vec::new(); // Wrong version.
         }
         let next_bit = self.our_name.bit(self.our_prefix.bit_count());
-        let other_prefix = self.our_prefix.pushed(!next_bit);
-        self.our_prefix = self.our_prefix.pushed(next_bit);
+        let other_prefix = self.our_prefix.pushed(!next_bit).with_incremented_version();
+        self.our_prefix = self.our_prefix.pushed(next_bit).with_incremented_version();
```
maqi: will this cause our version increased twice?

src/routing_table/mod.rs
```diff
@@ -968,78 +912,67 @@ impl RoutingTable {
...
                     prefix,
                     entry.get()
                 );
-                let &mut (ref mut v, ref mut s) = entry.into_mut();
-                if *v > version {
-                    return; // Wrong version.
-                }
-                *v = version;
-                s.extend(section);
+                entry.get_mut().extend(section);
```
maqi: do we need to carry out version check before extend?

src/routing_table/prefix.rs
```diff
@@ -23,70 +23,51 @@ use std::hash::{Hash, Hasher};
...
+    pub fn with_incremented_version(mut self) -> Self {
+        self.version += 1;
+        self
     }
 
     /// Returns `self` with an appended bit: `0` if `bit` is `false`, and `1` if `bit` is `true`. If
     /// `self.bit_count` is already at the maximum for this type, then an unmodified copy of `self`
     /// is returned.
-    pub fn pushed(mut self, bit: bool) -> Prefix {
+    pub fn pushed(mut self, bit: bool) -> Self {
```
maqi: I am not sure, but maybe a version up by one shall be enforced with pushed/popped method call?

src/routing_table/mod.rs
madadam: The version check is implicit, because version is now part of `Prefix` and the equality comparison of `Prefix` involves version too.

src/routing_table/mod.rs
madadam: I don't think so. On line 926, the version of `self.our_prefix` is not modified, because `with_incremented_version` doesn't modify `self` in place.

src/routing_table/mod.rs
madadam: Possibly. I guess this change might miss the case where an entry with the same prefix bits, but different version exists. Let me look into it...

src/routing_table/prefix.rs
madadam: Yeah, this occurred to me too. Might be a good idea, not sure either.

src/routing_table/mod.rs
maqi: with_incremented_version will modify the prefix in place.
it's not modifying self just because pushed returned a copy.
It might be ok, but I feel it's bit confusing and unsafe (pushed has been called twice modifying our_prefix but version only increased by one).

madadam: No, `with_incremented_version` does not modify the prefix in place. But anyhow, what would you suggest to be a less confusing way to do this?

madadam:
Closing in favor of https://github.com/maidsafe/routing/pull/1537

PR #1532, opened 2018-02-09
fix/clippy: fix clippy warnings

nbaksalyar:


src/rate_limiter.rs
```diff
@@ -248,7 +248,7 @@ impl RateLimiter {
             None => return None,
         };
 
-        let deduction = amount_charged.saturating_sub(part_count as u64 * MAX_PART_LEN as u64);
+        let deduction = amount_charged.saturating_sub(u64::from(part_count) * MAX_PART_LEN as u64);
```
madadam: Shouldn't the `MAX_PART_LEN as u64` be also rewritten to `u64::from(MAX_PART_LEN)`?

src/states/peer.rs
```diff
@@ -180,7 +180,7 @@ impl Peer {
 
     #[cfg_attr(feature = "cargo-clippy", allow(too_many_arguments))]
     pub fn from_bootstrapping(
-        our_section: (Prefix, BTreeSet<PublicId>),
+        our_section: &(Prefix, BTreeSet<PublicId>),
```
madadam: Might be better to change this to `our_section: (Prefix, &BTreeSet<PublicId>)`, so the called doesn't have to clone the `BTreeSet` when passing it in, as it doesn't seem to be consumed inside this function.

src/rate_limiter.rs
maqi: MAX_PART_LEN  is usize   and `u64::from(usize)` is not implemented  :(

src/timer.rs
```diff
@@ -239,6 +239,10 @@ mod implementation {
 
 #[cfg(feature = "use-mock-crust")]
 mod implementation {
+    // This lint requires to make changes that might introduce discrepancies with the actual Crust code,
+    // so we turn it off for mock crust specifically.
+    #![cfg_attr(feature = "cargo-clippy", allow(needless_pass_by_value))]
```
maqi: will it be better to put this into lib.rs,  just as what the current master branch do?

src/rate_limiter.rs
madadam: Got it. All good then.

PR #1535, opened 2018-02-13
MAID-2453 replace id module with FullInfo and PublicInfo

maqi:


src/public_info.rs
```diff
@@ -0,0 +1,107 @@
+// Copyright 2017 MaidSafe.net limited.
```
Fraser999: Wrong date.

src/full_info.rs
```diff
@@ -0,0 +1,158 @@
+// Copyright 2015 MaidSafe.net limited.
```
Fraser999: Wrong date.

src/block.rs
```diff
@@ -73,31 +73,31 @@ pub struct Block<T> {
 }
 
 impl<T: Serialize + Clone> Block<T> {
-    /// A new `Block` requires a valid vote and the `PeerId` of the peer who sent us this. For
+    /// A new `Block` requires a valid vote and the `PublicInfo` of the peer who sent us this. For
     /// this reason the `Vote` will require a `DirectMessage` from a peer to us.
-    pub fn new(vote: &Vote<T>, peer_id: &PeerId) -> Result<Block<T>, RoutingError> {
+    pub fn new(vote: &Vote<T>, peer_info: &PublicInfo) -> Result<Block<T>, RoutingError> {
```
Fraser999: Most if not all instances of `peer` in this PR should be changed to `node`.

src/public_info.rs
```diff
@@ -0,0 +1,107 @@
...
+//
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+// FIXME: remove when this module is finished
+#![allow(dead_code)]
```
Fraser999: This shouldn't be needed now I guess?

src/resource_prover.rs
```diff
@@ -100,17 +100,17 @@ impl ResourceProver {
     /// Start generating a resource proof in a background thread
     pub fn handle_request(
         &mut self,
-        pub_id: PublicId,
+        pub_info: PublicInfo,
         seed: Vec<u8>,
         target_size: usize,
         difficulty: u8,
-        log_ident: String,
+        log_infoent: String,
```
Fraser999: I think this should be left as `log_ident` or changed to `log_info`.  Same in a few places below.

src/routing_table/network_tests.rs
```diff
@@ -397,7 +397,7 @@ pub fn verify_network_invariant<'a, T: IntoIterator<Item = &'a RoutingTable>>(no
 }
 
 #[test]
-fn sections_have_identical_routing_tables() {
+fn sections_have_infoentical_routing_tables() {
```
Fraser999: Ah yes, the old infoentical routing tables 😆

src/full_info.rs
```diff
@@ -0,0 +1,158 @@
...
+// KIND, either express or implied.
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+use public_info::PublicInfo;
+use rust_sodium::crypto::{box_, sign};
+use xor_name::XorName;
+
+/// Network identity component containing name, and public and private keys.
```
Fraser999: We use "secret" now rather than "private".  Also, we're not keen on calling this an identity.  I guess I'd change this comment to something like:
```
/// A container for a peer's cryptographic keys, and in the case of a node, its age too.  The public
/// signing key defines the peer's name on the network.
```

src/public_info.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// Returning peer's age (client always having age of 0)
+    pub fn age(&self) -> u8 {
+        match *self {
+            PublicInfo::Client { .. } => 0,
+            PublicInfo::Node { age, .. } => age,
+        }
+    }
+
+    /// Updating a peer's age (void if peer is client)
+    pub fn set_age(&mut self, new_age: u8) {
```
Fraser999: Looks like this function is only used for tests.  I'd be in favour of removing it if we can, or if it's really required for the tests, then only have it defined for tests.

src/public_info.rs
```diff
@@ -0,0 +1,107 @@
...
+use std::fmt::{self, Debug, Display, Formatter};
+use xor_name::XorName;
+
+/// Public identifier for a Peer, contains age if peer is a node.
+#[derive(Serialize, Deserialize, PartialOrd, Ord, PartialEq, Eq, Clone, Copy, Hash)]
+pub enum PublicInfo {
+    /// Public identifier for a client
+    Client {
+        /// Client's public signing key
+        sign_key: sign::PublicKey,
```
Fraser999: Can we keep these fields in the same order throughout please?  The `FullInfo` (and task spec) has encrypt keys first.

src/public_info.rs
maqi: I think this will be required later on when we need to reset a node's age for rejoin.
I will make it defined for tests only now.

src/public_info.rs
maqi: There is a test `public_info_order` requires `name` (now the sign_key) to be preferred other than encrypt_key.
I can change the keys order for FullInfo to make them consistent.

src/block.rs
maqi: Think this is kind of depends on whether your PR 1533 (task 2467) got merged or this one  :)

src/public_info.rs
Fraser999: I was expecting `within_range` to do the job there, but I guess we still haven't spec'd out the flow yet, so not sure.

src/public_info.rs
nbaksalyar: @Fraser999 this is left there because this module is not exported publicly (only crate-level, i.e. `pub(crate) use public_info;`), so this results in Clippy complaints about the dead code.

If you make it public, Clippy shows errors about the absent documentation. So we can either add documentation and make the module public, or leave this lint turned off in this module just for now.

src/public_info.rs
Fraser999: I don't think that test is particularly useful.  We can't just use the internals of the `PublicInfo` to define the sort order for the case of choosing the oldest node.  We need to accommodate the tie breaker rule.

Either way, I think we should have consistent ordering of the member variables and functions across the two files!

src/block.rs
Fraser999: Yeah - I guess I should have talked to you about that one first.  I left all instances of `peer_id` alone in my PR to try to avoid needless conflicts! 😄

PR #1536, opened 2018-02-13
MAID-2456: rename SectionState to NodeState

Fraser999:


src/vote.rs
```diff
@@ -88,7 +87,7 @@ mod tests {
         unwrap!(rust_sodium::init_with_rng(&mut rng));
         let full_info = FullInfo::node_new(rng.gen_range(0, 255));
         let node_info = *full_info.public_info();
-        let payload = SectionState::Live(node_info);
+        let payload = "Live";
```
maqi: shall we use `NodeState::ElderLive` to demonstrate the way how network votes?
same to couple of other places

Fraser999: I don't think so - we should be able to use a simpler form for the unit tests here since that's what the API allows.

maqi: even in the tests of Block ?

Fraser999: Yes, I think so.  The votes, proofs and blocks should be usable for nodes to vote on arbitrary data - it's only when we start getting to data_chain that restrictions come in.

PR #1537, opened 2018-02-13
Versioned prefix (MAID-2451)

madadam:
Supersedes https://github.com/maidsafe/routing/pull/1531.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -16,7 +16,7 @@
 // relating to use of the SAFE Network Software.
 
 use messages::{Request, Response};
-use routing_table::{Prefix, RoutingTable};
+use routing_table::{RoutingTable, VersionedPrefix};
 use routing_table::Authority;
 use std::fmt::{self, Debug, Formatter};
 use xor_name::XorName;
@@ -54,11 +54,12 @@ pub enum Event {
     NodeAdded(XorName, RoutingTable),
     /// A node has disconnected from us.
     NodeLost(XorName, RoutingTable),
-    /// Our own section has been split, resulting in the included `Prefix` for our new section.
-    SectionSplit(Prefix),
-    /// Our own section requires merged with others, resulting in the included `Prefix` for our new
-    /// section.
-    SectionMerge(Prefix),
+    /// Our own section has been split, resulting in the included `VersionedPrefix`
+    /// for our new section.
+    SectionSplit(VersionedPrefix),
+    /// Our own section requires merged with others, resulting in the included
+    /// `VersionedPrefix` for our new section.
+    SectionMerge(VersionedPrefix),
     /// The client has successfully connected to a proxy node on the network.
     Connected,
     /// Disconnected or failed to connect - restart required.
--- a/src/event.rs
+++ b/src/event.rs
@@ -16,7 +16,7 @@
 // relating to use of the SAFE Network Software.
 
 use messages::{Request, Response};
-use routing_table::{RoutingTable, VersionedPrefix};
+use routing_table::{Prefix, RoutingTable};
 use routing_table::Authority;
 use std::fmt::{self, Debug, Formatter};
 use xor_name::XorName;
@@ -54,12 +54,12 @@ pub enum Event {
     NodeAdded(XorName, RoutingTable),
     /// A node has disconnected from us.
     NodeLost(XorName, RoutingTable),
-    /// Our own section has been split, resulting in the included `VersionedPrefix`
+    /// Our own section has been split, resulting in the included `Prefix`
     /// for our new section.
-    SectionSplit(VersionedPrefix),
+    SectionSplit(Prefix),
     /// Our own section requires merged with others, resulting in the included
-    /// `VersionedPrefix` for our new section.
-    SectionMerge(VersionedPrefix),
+    /// `Prefix` for our new section.
+    SectionMerge(Prefix),
     /// The client has successfully connected to a proxy node on the network.
     Connected,
     /// Disconnected or failed to connect - restart required.
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -31,8 +31,8 @@ use lru_time_cache::LruCache;
 use maidsafe_utilities::serialisation::{deserialise, serialise};
 use peer_manager::SectionMap;
 use public_info::PublicInfo;
-use routing_table::{Prefix, VersionedPrefix};
 use routing_table::Authority;
+use routing_table::VersionedPrefix;
 use rust_sodium::crypto::{box_, sign};
 use sha3::Digest256;
 use std::collections::{BTreeMap, BTreeSet, HashSet};
@@ -235,23 +235,28 @@ impl HopMessage {
 /// A list of a section's public IDs, together with a list of signatures of a neighbouring section.
 #[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Hash, Serialize, Deserialize, Debug)]
 pub struct SectionList {
-    pub prefix: Prefix,
+    prefix: VersionedPrefix,
     // TODO(MAID-1677): pub signatures: BTreeSet<(PublicInfo, sign::Signature)>,
     pub_infos: BTreeSet<PublicInfo>,
 }
 
 impl SectionList {
     /// Create
-    pub fn new(prefix: Prefix, pub_infos: BTreeSet<PublicInfo>) -> Self {
+    pub fn new(prefix: VersionedPrefix, pub_infos: BTreeSet<PublicInfo>) -> Self {
         SectionList {
-            prefix: prefix,
-            pub_infos: pub_infos,
+            prefix: prefix.with_version(0),
+            pub_infos,
         }
     }
 
     /// Create from any object convertable to an iterator
-    pub fn from<I: IntoIterator<Item = PublicInfo>>(prefix: Prefix, pub_infos: I) -> Self {
+    pub fn from<I: IntoIterator<Item = PublicInfo>>(prefix: VersionedPrefix, pub_infos: I) -> Self {
         Self::new(prefix, pub_infos.into_iter().collect())
+    }
+
+    /// Returns the section prefix
+    pub fn prefix(&self) -> &VersionedPrefix {
+        &self.prefix
     }
 }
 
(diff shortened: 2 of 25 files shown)


src/routing_table/prefix.rs
```diff
@@ -213,83 +282,99 @@ impl Prefix {
...
+        .is_some()
+}
+
+/// Remove and return the element at the given prefix, ignoring versions.
+pub fn unversioned_remove(set: &mut BTreeSet<Prefix>, key: &UnversionedPrefix) -> Option<Prefix> {
+    if let Some(prefix) = set.range(key.with_version(0)..key.with_version(u64::MAX))
+        .cloned()
+        .next()
+    {
+        if set.remove(&prefix) {
```
maqi: it looks like when we carry out the removal, the version does matter?
i.e. if the entry returned doesn't have the matched version, it won't get removed?

src/routing_table/prefix.rs
```diff
@@ -213,83 +282,99 @@ impl Prefix {
...
-            }
+/// Find the entry at the given prefix, ignoring versions.
+pub fn unversioned_find<'a, T>(
+    map: &'a BTreeMap<Prefix, T>,
+    key: &UnversionedPrefix,
+) -> Option<(Prefix, &'a T)> {
+    map.range(key.with_version(0)..key.with_version(u64::MAX))
+        .next()
+        .map(|(prefix, value)| (*prefix, value))
+}
```
maqi: just wondering this `unversioned` methods actually indicates the map shall having `UnversionedPrefix` as key?
otherwise, the insertion into map allow same prefix having different version and the `unversioned_get/find` methods only return the first matched one.
maybe we shall have another `unversion_insert` and ensure the map calls unversion_insert as well?

src/routing_table/prefix.rs
```diff
@@ -17,103 +17,222 @@
...
+            }
+        }
+        let mut name = XorName::default();
+        name.0[0] = first_byte;
+        Ok(Self::new(bits.len(), name, 0))
+    }
+}
+
+/// A section prefix, i.e. a sequence of bits specifying the part of the network's name space
+/// consisting of all names that start with this sequence.
```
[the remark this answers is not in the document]
madadam: :+1:

src/routing_table/prefix.rs
madadam: No, it first finds the (versioned) `Prefix` matching the given `UnversionedPrefix` (via `range`). Then it removes it.

src/routing_table/prefix.rs
madadam: The purpose of this method is to allow you to have a map keyed by **versioned** `Prefix`es, but to query it by `UnversionedPrefix`es.

`unversioned_insert` wouldn't make sense in my opinion, because we need the keys to be versioned. It's currently up to the user of the map to make sure no prefix is present in the map more than once with just the version number differing. I believe `RoutingTable` currently takes care of that. We could in theory write a custom collection for this, but it might be overkill. Not sure.

maqi: I don't see  `RoutingTable` carry out any explicit check to ensure that  no prefix is present in the map more than once with just the version number differing.
It's more like totally depends on the events that RoutingTable receives.

another way around, shall those methods using unversioned prefix shall have `Prefix` as input parameter?

src/routing_table/prefix.rs
```diff
@@ -17,103 +17,222 @@
...
+                    max_prefix_len,
+                ) &&
+                 self.pushed(true).is_covered_by_impl(
+                    prefixes,
+                    max_prefix_len,
+                ))
+    }
+}
+
+impl Deref for Prefix {
```
Fraser999: This is a clever move, but I wonder if it may make the calling code more obvious if we just take the more tedious approach of implementing the handful of missing methods in `Prefix`, e.g.
```rust
pub fn is_neighbour(&self, other: &Self) -> bool {
    self.inner.is_neighbour(&other.inner)
}
```

src/routing_table/prefix.rs
```diff
@@ -17,103 +17,222 @@
...
+    }
+}
+
+impl PartialOrd for Prefix {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for Prefix {
```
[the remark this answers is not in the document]
madadam: > were you concerned that the order of the two member variables might get swapped in a later refactor
Mostly that. Also not 100% if rust guarantees the sort order based on the field order. Anyway, maybe we can derive the traits but write explicit tests for the ordering?

src/event.rs
```diff
@@ -54,10 +54,11 @@ pub enum Event {
     NodeAdded(XorName, RoutingTable),
     /// A node has disconnected from us.
     NodeLost(XorName, RoutingTable),
-    /// Our own section has been split, resulting in the included `Prefix` for our new section.
+    /// Our own section has been split, resulting in the included `Prefix`
+    /// for our new section.
```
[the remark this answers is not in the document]
madadam: No, but rustfmt has max comment length and it errors if it is exceeded. Not sure what would be the best way to keep the comments at the optimal length. Maybe rustfmt can be configured to do it?

src/messages/mod.rs
```diff
@@ -602,12 +607,12 @@ pub enum MessageContent {
     SectionUpdate {
         /// Section prefix and version. Included because this message is sent to both the section's
         /// own members and neighbouring sections.
-        versioned_prefix: VersionedPrefix,
+        versioned_prefix: Prefix,
```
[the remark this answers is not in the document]
madadam: :+1:

src/states/node.rs
```diff
@@ -2881,32 +2890,30 @@ impl Node {
...
-                )
+                self.our_merged_section.get_our_merged_section(merge_prefix)
             {
                 our_merged_section = merged_section;
             }
 
             self.process_own_section_merge(
-                their_prefix,
-                version,
+                // their_prefix.popped().with_version(version),
```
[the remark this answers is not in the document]
madadam: :+1:

src/routing_table/prefix.rs
madadam: Yeah, I went back and forth on this. Not really decided. With `Deref`, you can also pass `Prefix` to functions that take `&UnversionedPrefix` which might be useful or confusing, depends on how you look at it.

src/routing_table/prefix.rs
Fraser999: I think that's handled by `insert_new_section()`, but it may be worth ensuring that the invariant check fails if this case ever happens?

src/routing_table/prefix.rs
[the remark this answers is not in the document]
Fraser999: > Anyway, maybe we can derive the traits but write explicit tests for the ordering?

Yup, I was going to suggest that 😄

src/event.rs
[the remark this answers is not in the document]
Fraser999: Hmm, weird.  It doesn't complain for me at these lines.  I guess you're using rustfmt v0.9.0?

src/routing_table/prefix.rs
Fraser999: > might be useful or confusing

Was both for me initially 😄   I think because of the risk of how badly things could go if we accidentally don't consider versions when we really should, I'd go for the more explicit route, e.g. force the user to call `.unversioned()`.  However, it's probably worth getting a few more opinions before you make the change?

src/routing_table/prefix.rs
madadam: Hm, I think I misunderstood what you meant about the `unversioned_insert` function, @maqi . Did you mean something like this:

```rust
fn unversioned_insert<T>(map: &mut BTreeMap<Prefix, T>, key: Prefix, value: T) -> Option<T> {
    let result = unversioned_remove(map, &key.unversioned());
    let _ = map.insert(key, value);
    result
}
```
? Because I think that might be actually useful.

src/event.rs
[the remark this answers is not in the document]
madadam: The line might have been longer at some point (maybe `Prefix` was `VersionedPrefix` or something), rustfmt complained, I broke it into two lines, then it got shorter later again but there was nothing telling me to reformat it back into single line.

Don't feel like sifting through the code, looking for comments that could be reformatted to take fewer lines to be honest. Unless it can be automated in some way.

src/routing_table/prefix.rs
maqi: yes.

I will prefer `not to carry out insertion if there is prefix having diff version exists`.
But not strongly on it.

src/routing_table/prefix.rs
maqi: I will prefer a more explicit way, i.e. calling `.unversioned()`, as the code will be easier to read.

src/event.rs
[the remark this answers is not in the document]
Fraser999: Absolutely!  I just wanted to check you didn't have some specific editor setting which would cause commits to get polluted with needless changes 😄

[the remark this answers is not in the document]
madadam: :+1:

src/states/node.rs
```diff
@@ -1973,11 +1970,11 @@ impl Node {
             self.routing_table().should_split()
         {
             // i.e. the section should split
-            let our_ver_pfx = self.routing_table().our_versioned_prefix();
+            let our_ver_pfx = *self.routing_table().our_prefix();
```
maqi: the `our_ver_pfx` variable can now be `our_pfx` or `our_prefix` as well?

src/peer_manager.rs
```diff
@@ -1637,9 +1628,8 @@ impl PeerManager {
     /// peers, sorted by section.
     pub fn ideal_rt(&self) -> SectionMap {
         let versioned_prefixes = self.routing_table
```
maqi: the `versioned_prefixes` variable can now be `prefixes`?

src/routing_table/prefix.rs
nbaksalyar: Really nice way of using iterators 👍  I'd probably suggest even adding a helper method to replace `.range(pfx.with_version(0)..key.with_version(u64::MAX))` with something like `.all_versions_range(pfx)`, with an impl like
```
impl Iter<Item = T> {
  ...
  fn all_versions_range(&self, pfx: Prefix) {
    self.range(pfx.with_version(0)..pfx.with_version(u64::MAX))
  }
}
```
This might convey the intent in a more clean manner IMO. Something to note for future refactorings.

PR #1538, opened 2018-02-15
MAID-2454 create a votes iterator for Block to transform proofs into votes

maqi:


src/vote.rs
```diff
@@ -44,6 +44,11 @@ impl<T: Serialize + Clone> Vote<T> {
         })
     }
 
+    /// Compose a Vote.
+    pub fn compose(payload: T, signature: Signature) -> Self {
```
madadam: This function allows you to create a `Vote` where the signature doesn't match the payload. I'd consider marking it `unsafe`, but not entirely sure about that. At the very least, do not expose it outside of the crate, so make it `pub(crate)` or something.

src/block.rs
```diff
@@ -156,6 +156,17 @@ impl<T: Serialize + Clone> Block<T> {
         }
     }
 
+    /// Create an iterator over all proofs and transform into votes.
+    pub fn votes_iter<'a>(&'a self) -> Box<Iterator<Item = (&PublicInfo, Vote<T>)> + 'a> {
+        let payload = self.payload.clone();
+        Box::new(self.proofs.iter().map(move |proof| {
+            (
+                proof.node_info(),
+                Vote::compose(payload.clone(), *proof.sig()),
```
Fraser999: Can't we just use `self.payload.clone()` here rather than taking another copy outside the loop?

src/vote.rs
```diff
@@ -44,6 +44,11 @@ impl<T: Serialize + Clone> Vote<T> {
         })
     }
 
+    /// Compose a Vote.
+    pub(crate) fn compose(payload: T, signature: Signature) -> Self {
```
Fraser999: We've spoken about making a module for the data chain stuff before.  If we did that now, we could reduce the visibility of this method to just that module.

madadam: That would be even better.

src/block.rs
maqi: will that `move` the whole `self` ?

Fraser999: No, I don't think so - we only have a reference to `self` there.

PR #1539, opened 2018-02-15
MAID-2455: Add test for Block::votes_iter

madadam:
Tests for the stuff in https://github.com/maidsafe/routing/pull/1538

### DIFF
--- a/src/data_chain/block.rs
+++ b/src/data_chain/block.rs
@@ -195,85 +195,98 @@ mod tests {
         let mut rng = SeededRng::thread_rng();
         unwrap!(rust_sodium::init_with_rng(&mut rng));
 
-        let mut full_info_0 = FullInfo::node_new(0);
-        let mut full_info_1 = FullInfo::node_new(0);
-        let public_info_0 = *full_info_0.public_info();
-        let public_info_1 = *full_info_1.public_info();
+        let full_info_0 = FullInfo::node_new(1);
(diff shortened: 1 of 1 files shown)


src/data_chain/block.rs
```diff
@@ -189,90 +188,129 @@ mod tests {
...
-        let _ = valid_voters.insert(public_info_00);
-        let _ = valid_voters.insert(public_info_10);
-        assert!(b0.add_proof(proof0, &valid_voters).is_err());
-        assert_eq!(b0.num_proofs(), 1);
-        assert!(b0.add_proof(proof1, &valid_voters).is_ok());
-        assert_eq!(b0.num_proofs(), 2);
+        let _ = valid_voters.insert(*full_info_0.public_info());
+        let _ = valid_voters.insert(*full_info_1.public_info());
+
+        assert!(block.add_proof(proof0, &valid_voters).is_err());
```
nbaksalyar: Maybe it'd make sense to make the asserted error more explicit? I.e. something like this:
```
match block.add_proof(proof0, &valid_voters) {
  RoutingError::DuplicateSignatures => {},
  e => panic!("Unexpected result {}", e),
}
```

And also test other error conditions.
(This is more of a comment for further consideration/refactoring though, feel free to ignore)

madadam: :+1:

PR #1540, opened 2018-02-19
MAID-2468: add unit tests for VoteAccumulator

Fraser999:
Also added quorum function and test to data_chain module.

### DIFF
--- a/src/data_chain/mod.rs
+++ b/src/data_chain/mod.rs
@@ -39,11 +39,11 @@ use std::collections::BTreeSet;
 /// of the cumulative age of all members of `valid_nodes_itr`.
 pub fn quorum<'a, 'b, I, J>(voters_itr: I, valid_nodes_itr: J) -> bool
 where
-    I: Iterator<Item = &'a PublicInfo>,
-    J: Iterator<Item = &'b PublicInfo>,
+    I: IntoIterator<Item = &'a PublicInfo>,
+    J: IntoIterator<Item = &'b PublicInfo>,
 {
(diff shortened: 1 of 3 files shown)


src/data_chain/vote_accumulator.rs
```diff
@@ -0,0 +1,366 @@
...
+                valid_voters.clone(),
+            ));
+            let results2 = unwrap!(accumulator.add_vote(
+                vote2,
+                node.public_info(),
+                valid_voters.clone(),
+            ));
+            nodes_added.push(node.public_info());
+
+            if data_chain::quorum(valid_voters.clone(), nodes_added.iter().cloned()) {
```
madadam: Shouldn't the arguments be reversed? That is, `nodes_added` first, `valid_voters` second?

src/data_chain/mod.rs
```diff
@@ -20,9 +20,92 @@ mod chain;
...
+
+use public_info::PublicInfo;
+use std::collections::BTreeSet;
+
+// FIXME - remove
+#[allow(unused)]
+/// Calculates whether a quorum of nodes have voted.  In this case, "quorum" means >50% of the
+/// members in `valid_nodes_itr` are included in `voters_itr` and that their cumulative age is >50%
+/// of the cumulative age of all members of `valid_nodes_itr`.
+pub fn quorum<'a, 'b, I, J>(voters_itr: I, valid_nodes_itr: J) -> bool
```
madadam: Seems we can use this function in `Block::get_block_state`. Possibly other places too. Not sure it should be done in this PR or other, so feel free to dismiss.

PR #1542, opened 2018-02-21
MAID-2469: add unit tests for chain

Fraser999:
Also added some Debug impls and moved test helpers to avoid code duplication.

### DIFF
--- a/src/data_chain/chain.rs
+++ b/src/data_chain/chain.rs
@@ -18,12 +18,12 @@
 // FIXME: remove when this module is finished
 #![allow(dead_code)]
 
-use super::{Block, NodeState, NodesAndAge, Proof, Vote};
+use super::{Block, NodeState, Vote};
 use error::RoutingError;
 use fs2::FileExt;
 use maidsafe_utilities::serialisation;
 use public_info::PublicInfo;
-use std::collections::BTreeSet;
+use std::fmt::{self, Debug, Formatter};
 use std::fs;
 use std::io::{self, Read, Write};
 use std::path::PathBuf;
@@ -40,6 +40,7 @@ pub struct Chain {
     path: Option<PathBuf>,
     valid_nodes: Vec<Block<NodeState>>, // save to aid network catastrophic failure and restart.
     data: Vec<Block<DataIdentifier>>,
+    pending_blocks: Vec<Block<NodeState>>,
 }
 
 impl Chain {
@@ -60,6 +61,7 @@ impl Chain {
             path: Some(path),
             valid_nodes: vec![],
             data: vec![],
+            pending_blocks: vec![],
         })
     }
 
@@ -79,13 +81,18 @@ impl Chain {
     }
 
     /// Create chain in memory from some blocks
-    pub fn from_blocks(blocks: Vec<Block<NodeState>>, group_size: usize) -> Chain {
+    pub fn from_blocks(
+        blocks: Vec<Block<NodeState>>,
+        pending_blocks: Vec<Block<NodeState>>,
+        group_size: usize,
+    ) -> Chain {
         Chain {
             blocks,
             group_size,
             path: None,
             valid_nodes: vec![],
             data: vec![],
+            pending_blocks,
         }
     }
 
@@ -123,39 +130,18 @@ impl Chain {
         }
     }
 
-    fn add_vote(
-        &mut self,
-        vote: &Vote<NodeState>,
-        node_info: &PublicInfo,
-    ) -> Option<(NodeState, NodesAndAge)> {
-        if !vote.validate_signature(node_info) {
-            return None;
-        }
+    // FIXME - re-enable this lint check
+    #[cfg_attr(feature = "cargo-clippy", allow(needless_pass_by_value))]
+    /// Add the given valid `block` to the chain.
+    pub fn add_block(&mut self, _block: Block<NodeState>) {
+        unimplemented!();
+    }
 
-        let pub_key_matches = |x: &Proof| x.node_info().sign_key() == node_info.sign_key();
-        for blk in &mut self.blocks.iter_mut() {
-            if blk.payload() == vote.payload() {
-                if blk.proofs().iter().any(pub_key_matches) {
-                    info!("duplicate proof");
-                    return None;
-                }
-                // TODO: use proper valid voters list instead of the empty list.
-                let _ = blk.add_proof(vote.proof(node_info).unwrap(), &BTreeSet::new())
-                    .unwrap();
-
-                let p_age = NodesAndAge::new(blk.num_proofs(), blk.total_age());
-                return Some((blk.payload().clone(), p_age));
-            }
-        }
-        if let Ok(ref mut blk) = Block::new(vote, node_info) {
-            self.blocks.push(blk.clone());
-            return Some((
-                blk.payload().clone(),
-                NodesAndAge::new(1, usize::from(node_info.age())),
-            ));
-        }
-        info!("Could not find any block for this proof");
-        None
+    // FIXME - re-enable this lint check
+    #[cfg_attr(feature = "cargo-clippy", allow(needless_pass_by_value))]
+    /// Add the given valid `vote` to the chain.
+    pub fn add_vote(&mut self, _vote: Vote<NodeState>, _node_info: PublicInfo) {
+        unimplemented!();
     }
 
     /// Assumes we trust the first `Block`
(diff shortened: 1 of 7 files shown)


src/data_chain/proof.rs
```diff
@@ -52,6 +53,17 @@ impl Proof {
     }
 }
 
+impl Debug for Proof {
+    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
+        write!(
+            formatter,
+            "Proof{{ {}, age: {}, sig: ... }}",
+            self.node_info.name(),
```
maqi: will it be better to use `PublicInfo`'s Debug trait directly?
I agree the current format is neater, but using `Debug` trait will avoid hassle in case the `PublicInfo` struct got changed?

src/data_chain/chain.rs
```diff
@@ -185,3 +171,309 @@ impl Chain {
...
+    use routing_table::Prefix;
+    use rust_sodium;
+
+    #[ignore]
+    #[test]
+    fn pending_blocks() {
+        // Start with the chain having Live(A).  Create blocks for Offline(A), Live(B), and
+        // Relocated(B), Live(C).  Add the four of them in reverse order and assert after each is
+        // added that they're held in the `pending_blocks` until the final one is added and they can
+        // all be added to the chain in the order defined above.
```
maqi: I think this scenario is assuming the chain contains GROUP_SIZE elder blocks already,
otherwise, a `Live(C)` is valid after a `Live(A)` ?
However, the testing code seems only having one live block in the chain?
Will it be better to make it explicit in the comment ?

src/data_chain/mod.rs
```diff
@@ -60,11 +60,64 @@ where
...
+            .collect_vec();
+        nodes.sort_by(|lhs, rhs| lhs.public_info().cmp(rhs.public_info()));
+        for (index, node) in nodes.iter_mut().enumerate() {
+            node.set_age(index as u8 + 5);
+        }
+        nodes
+    }
+
+    /// Create a block.  If `fill` is true, add votes for every member of `nodes`, otherwise only
+    /// add votes until the block is valid.
```
maqi: just a note: when `fill` is false, only the first Q nodes will sign to make the block valid.
we may need `Q random nodes to sign` as well, but this can be done later on.

src/data_chain/chain.rs
```diff
@@ -185,3 +171,309 @@ impl Chain {
...
+        payload = NodeState {
+            state: State::ElderLive,
+            public_key: *node_c.public_info().sign_key(),
+            age: node_c.public_info().age(),
+            section: prefix,
+        };
+        let live_c = tests::create_block(payload, &nodes, false);
+
+        // Create chain containing just Live(A).
+        let mut chain = Chain::from_blocks(vec![live_a.clone()], vec![], nodes.len());
```
maqi: not sure, but I will prefer to use `GROUP_SIZE` instead of `nodes.len()` .
same to couple of other places.

src/data_chain/chain.rs
```diff
@@ -185,3 +171,309 @@ impl Chain {
...
+            public_key: *node_a.public_info().sign_key(),
+            age: node_a.public_info().age(),
+            section: prefix,
+        };
+        let live_a_block = tests::create_block(payload.clone(), &nodes, false);
+        let last_node = &nodes[nodes.len() - 1];
+        assert!(!live_a_block.get_node_infos().contains(
+            last_node.public_info(),
+        ));
+        let live_a_vote = unwrap!(Vote::new(nodes[0].secret_sign_key(), payload));
```
maqi: shall this be last_node.secret_sign_key ?
and same to the one below.

src/data_chain/chain.rs
```diff
@@ -185,3 +171,309 @@ impl Chain {
...
+    fn equality() {
+        // Create one chain comprising:
+        //     Live(A)→Live(B)→Offline(A)→Live(C)→Relocated(B)→Live(D)
+        // and another comprising:
+        //     Live(A)→Live(B)→Relocated(B)→Live(C)→Offline(A)→Live(D)
+        let mut rng = SeededRng::thread_rng();
+        unwrap!(rust_sodium::init_with_rng(&mut rng));
+
+        let prefix = Prefix::default().pushed(rng.gen()).with_version(rng.gen());
+        let nodes = tests::create_full_infos(Some(prefix));
```
maqi: just to confirm this `nodes` won't contain any node of A, B, C, D, which means those node won't be signer for any of the blocks.

src/data_chain/mod.rs
```diff
@@ -60,11 +60,64 @@ where
...
+        } else {
+            FullInfo::node_new(0)
+        };
+
+        let mut nodes = itertools::repeat_call(initialiser)
+            .take(GROUP_SIZE)
+            .collect_vec();
+        nodes.sort_by(|lhs, rhs| lhs.public_info().cmp(rhs.public_info()));
+        for (index, node) in nodes.iter_mut().enumerate() {
+            node.set_age(index as u8 + 5);
```
maqi: shall we avoid using magic number of `5`, instead using a global variable, say `InitialAge` ?

src/data_chain/proof.rs
Fraser999: The main motivation was to avoid the signature getting printed, so no, I think we should avoid deriving the `Debug` trait for items which print long (fairly useless) strings of bytes.

src/data_chain/chain.rs
Fraser999: I think we can update that once we have agreement on how the chain will be getting validated (if at all) inside the constructor and how pruning will work.  Since the blocks are relating to a non-empty prefix, we're out of the startup phase, and hence two `Live` blocks shouldn't be allowed next to each other.

src/data_chain/chain.rs
Fraser999: I was thinking that if we wanted to change the group size value used in tests though, it'd be less code to change doing it this way.

src/data_chain/chain.rs
Fraser999: Yes - well spotted thanks :)

src/data_chain/chain.rs
Fraser999: Yes, that's right, A, B, C, and D don't sign anything.  I thought the test was simpler that way.  I toyed with making the blocks more complete by having A-D signing those blocks which "made sense" for them to sign, but I think it'd make the test way more complex with little benefit from the point of view of testing chain equality.

src/data_chain/mod.rs
Fraser999: I don't really mind here.  Using `5` or `InitialAge` could be slightly misleading since we don't really need it to be either of those things - it could just as well be 10 for the purposes of these tests.

src/data_chain/proof.rs
maqi: PublicInfo's Debug trait has been re-implemented and it only contains name & age?
I mean you can have something like:
("Proof{{ {:?}, sig: ... }}", self.node_info)

src/data_chain/chain.rs
maqi: isn't change the group size value means only need to change the definition of `GROUP_SIZE` ?

src/data_chain/proof.rs
Fraser999: Ah - I see.  That makes it needlessly verbose though since we'd have
```
Proof{ PublicInfo(Node: 7520ea.., age: 55), sig: ... }
```
as opposed to
```
Proof{ 7520ea.., age: 55, sig: ... }
```
Since proofs should only be signed by nodes and not clients, I think it's worth stripping this extra text.

src/data_chain/chain.rs
Fraser999: No, I mean that we can leave the default value `GROUP_SIZE` unchanged in the code, but replace its use in tests with something else more easily.

src/data_chain/proof.rs
maqi: Yeah, I agree it's neater to strip the extra text.
Was just thinking using an existing Debug trait will avoid future hassle when PublicInfo struct got changed.
If you think it will not be an issue, I am ok with this change.

src/data_chain/chain.rs
```diff
@@ -185,3 +171,309 @@ impl Chain {
...
+        let mut chain = Chain::from_blocks(vec![live_a.clone()], vec![], nodes.len());
+        let mut expected_blocks = vec![live_a.clone()];
+        let mut expected_pendings = vec![];
+
+        // Add Live(C) - should be held in `pending_blocks` as there's no corresponding Offline/
+        // Demoted/Relocated in the `pending_blocks`.
+        chain.add_block(live_c.clone());
+        expected_pendings.push(live_c.clone());
+        assert_eq!(chain.blocks, expected_blocks);
+        assert_eq!(chain.pending_blocks, expected_pendings);
```
madadam: I wonder whether it's a good idea to test private fields like this. My suggestion would be to ignore `pending_blocks` (it's an implementation detail, I don't think it should be tested) and only test the contents of `blocks`, but via a public accessor. Something like: `fn blocks(&self) -> &[Block<...>]`.

src/data_chain/chain.rs
maqi: Ah, I see.
Are those places currently passing nodes.len() actually wants `GROUP_SIZE`?
will it make more sense to use `GROUP_SIZE` as much as possible, and only use nodes.len() when really need it?
Or, if we want to have a different configuration of `GROUP_SIZE` for tests, we can define `TEST_GROUP_SIZE` (or `[cfg(test)]`) ?

just want to eliminate the confusion that passing nodes.len() gives the impression that the parameter can be any value (less/greater than GROUP_SIZE).

src/data_chain/chain.rs
```diff
@@ -185,3 +171,309 @@ impl Chain {
...
+        // vote for each and check the vote is added correctly.
+        let mut rng = SeededRng::thread_rng();
+        unwrap!(rust_sodium::init_with_rng(&mut rng));
+
+        let prefix = Prefix::default().pushed(rng.gen()).with_version(rng.gen());
+        let nodes = tests::create_full_infos(Some(prefix));
+
+        // Create Live(A) block and extra vote for Live(A).
+        let node_a = FullInfo::within_range(5, &prefix.lower_bound(), &prefix.upper_bound());
+        let mut payload = NodeState {
```
madadam: Maybe we can reduce the verbosity of these tests by adding a `NodeState` constructor:
```rust
pub fn new(state: State, public_info: &PublicInfo, prefix: Prefix) -> Self
```
.

src/data_chain/chain.rs
Fraser999: We already have a dev config option for this.  How about if I do:
```rust
let group_size = nodes.len();
let mut chain = Chain::from_blocks(vec![live_a.clone()], vec![], group_size);
```

src/data_chain/chain.rs
Fraser999: I'm fine to put an accessor for `blocks`, but I think we'd want one for `pending_blocks` too.  So far we don't have a use for either, but I expect when we come to passing our chain to a joining node, we'll want those (rather than just serialising the full chain).  So, I'll add those two accessors and use them in the tests.

src/data_chain/node_state.rs
```diff
@@ -72,6 +73,18 @@ pub struct NodeState {
     pub section: Prefix,
```
maqi: as there is a construct now, can we make these members not to be public?

src/data_chain/mod.rs
```diff
@@ -60,11 +60,65 @@ where
...
     use maidsafe_utilities::SeededRng;
+    use routing_table::Prefix;
     use rust_sodium;
+    use serde::Serialize;
     use std::iter;
 
+    /// Creates a collection of `GROUP_SIZE` `FullInfo`s for nodes with ages `MIN_ADULT_AGE`,
+    /// `MIN_ADULT_AGE + 1`, `MIN_ADULT_AGE + 2`, etc.  If
+    /// `in_prefix` is `Some`, all nodes' IDs will fit that prefix.
+    pub fn create_full_infos(in_prefix: Option<Prefix>) -> Vec<FullInfo> {
```
nbaksalyar: Probably it'd be also worth to separate these functions into their own `data_chain::test_utils` module in the future.

PR #1543, opened 2018-02-22
MAID-2457: Implement VoteAccumulator

madadam:


src/data_chain/vote_accumulator.rs
```diff
@@ -39,33 +38,136 @@ pub enum AccumulationReturn<T> {
...
+                    .block
+                    .get_block_state(valid_nodes)
+                    .valid_or_full();
+                let valid_after = entry
+                    .get_mut()
+                    .block
+                    .add_vote(vote, node_info, valid_nodes)?
+                    .valid_or_full();
+
+                if valid_after {
```
maqi: looks like we don't need `valid_after` variable ?

src/data_chain/vote_accumulator.rs
```diff
@@ -39,33 +38,136 @@ pub enum AccumulationReturn<T> {
...
+                            valid_after = state.valid_or_full();
+                            new_proofs.push(*proof);
+                        }
+                        /// Duplicate votes are ignored.
+                        Err(RoutingError::DuplicateSignatures) => (),
+                        Err(error) => return Err(error),
+                    }
+                }
+
+                if valid_after {
```
maqi: can we do this:
```
for proof in block.proofs() {
    let _ =  entry.get_mut().block.add_proof(*proof, valid_nodes);
}
if entry.get().block.get_block_state(valid_nodes).valid_or_full() {
   if valid_before ...
}
```

src/data_chain/vote_accumulator.rs
madadam: We don't necessarily, but I think it makes the code slightly more readable, so I'd prefer to keep it as is.

src/data_chain/vote_accumulator.rs
madadam: I still need to check the result of `add_proof` for errors and also to collect the new proofs, so I may as well use it in the `if` too. This also avoids the `get_block_state` call, making this approach slightly more efficient. I'd prefer to do it the way I'm doing it unless you think it's actually incorrect.

maqi: the `new_proofs` only needs to be collected when `valid_before` is true, and you don't need to carry out `state.valid_or_full()` each time adding proof.
Regarding the `error` check, will it make more sense to `restore to the previous state` or `ignore then to continue`, instead of the current `partially updated then break` ?

So, can we do something like:
```
for proof in block.proofs() {
    if valid_before && entry.get_mut().block.add_proof(*proof, valid_nodes).is_ok() {
        results.push(AccumulationReturn::ValidProof(*proof));
    }
}
if !valid_before && entry.get().block.get_block_state(valid_nodes).valid_or_full() {
    results.push(AccumulationReturn::ValidBlock(entry.get().block.clone()));
}
```

src/data_chain/mod.rs
```diff
@@ -35,25 +35,18 @@ use std::collections::BTreeSet;
 /// Calculates whether a quorum of nodes have voted.  In this case, "quorum" means >50% of the
```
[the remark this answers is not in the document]
madadam: :+1:

src/data_chain/mod.rs
```diff
@@ -35,25 +35,18 @@ use std::collections::BTreeSet;
...
-        .filter(|voter| valid_nodes.contains(voter))
-        .collect::<BTreeSet<_>>();
-
+pub fn quorum(voters: &BTreeSet<PublicInfo>, valid_nodes: &BTreeSet<PublicInfo>) -> bool {
     let valid_total_age = valid_nodes.iter().map(|node| usize::from(node.age())).sum();
     let mut running_total_age = 0;
-    for (count, voter) in valid_voters.iter().rev().enumerate() {
+
+    let valid_voters = voters.intersection(valid_nodes);
+    for (count, voter) in valid_voters.enumerate() {
```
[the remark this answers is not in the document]
madadam: Yeah I know. Problem is that the `btree_set::Intersection` iterator is not a `DoubleEndedIterator` so it cannot be reversed. I would have to `collect` the iterator into something reversible and use that, and I'm not sure it would be worth it. Besides, `valid_voters` contains at most `GROUP_SIZE` elements, doesn't it? So the performance difference is going to be negligible I think.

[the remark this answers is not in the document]
Fraser999: Ah - fair enough then!

PR #1545, opened 2018-02-23
MAID-2465: sign all direct messages and add new variants

Fraser999:


src/messages/mod.rs
```diff
@@ -171,6 +176,28 @@ pub enum DirectMessage {
...
+        /// The valid pending blocks relating to elders.
+        pending_blocks: Vec<Block<NodeState>>,
+        // All adults and infants in our section.
+        non_elders: Vec<PublicInfo>,
+        not_yet_valid_blocks: Vec<Block<NodeState>>,
+    },
+    EncryptionKeyRequest { public_sign_key: sign::PublicKey },
+    EncryptionKeyResponse {
+        public_sign_key: sign::PublicKey,
+        public_encrypt_key: Option<box_::PublicKey>,
```
maqi: I guess this request-response is a point-to-point communication?
so a response with None indicates a failure?

src/messages/mod.rs
```diff
@@ -743,6 +770,69 @@ impl Debug for DirectMessage {
...
+                    public_sign_key.0[2]
+                )
+            }
+            EncryptionKeyResponse {
+                ref public_sign_key,
+                ref public_encrypt_key,
+            } => {
+                write!(
+                    formatter,
+                    "EncryptionKeyResponse {{ public_sign_key: {:02x}{:02x}{:02x}.., \
```
maqi: the encryption key belongs to which public_sign_key is not printed out?
this will be difficult to tracing the message flow.

src/states/bootstrapping.rs
```diff
@@ -303,9 +307,15 @@ impl Bootstrapping {
     fn handle_direct_message(
         &mut self,
         direct_message: DirectMessage,
+        signature: sign::Signature,
         pub_info: PublicInfo,
     ) -> Transition {
         use self::DirectMessage::*;
+
+        if !self.verify_direct_message(&direct_message, &signature, &pub_info) {
+            return Transition::Stay;
```
maqi: Do we need to carry out actions on this verification failure?
maybe worth to put a comment here explain why not or why required.
and same to couple of other places carrying out this new verification check.

src/states/common/base.rs
```diff
@@ -85,4 +87,43 @@ pub trait Base: Debug {
             // return self.handle_lost_peer(*pub_info).map(|_| Err(err.into()));
         }
     }
+
+    fn sign_direct_message(
+        &self,
+        direct_message: &DirectMessage,
+    ) -> Result<sign::Signature, RoutingError> {
+        let serialised_content = match serialisation::serialise(direct_message) {
```
maqi: can we make this serialisation a function to reduce code duplication?

src/messages/mod.rs
Fraser999: The encryption key is printed out too if it's not `None`.

src/states/bootstrapping.rs
Fraser999: No, I don't think so - I think the code is self-documenting enough.  We don't need to do anything other than log the failure (which we do inside `verify_direct_message()`).  What would you suggest as a comment?

src/states/common/base.rs
Fraser999: That seems a bit like overkill to me since the two places it'd be used need to return different things.  We actually end up with more code:
```rust
    fn sign_direct_message(
        &self,
        direct_message: &DirectMessage,
    ) -> Result<sign::Signature, RoutingError> {
        let serialised_content = self.serialise_direct_message(direct_message)?;
        Ok(sign::sign_detached(
            &serialised_content,
            self.full_info().secret_sign_key(),
        ))
    }

    fn verify_direct_message(
        &self,
        direct_message: &DirectMessage,
        signature: &sign::Signature,
        pub_info: &PublicInfo,
    ) -> bool {
        let serialised_content = match self.serialise_direct_message(direct_message) {
            Ok(result) => result,
            Err(_) => {
                return false;
            }
        };

        if !sign::verify_detached(signature, &serialised_content, pub_info.sign_key()) {
            info!("Direct message has an invalid signature");
            return false;
        }

        true
    }

    fn serialise_direct_message(
        &self,
        direct_message: &DirectMessage,
    ) -> Result<Vec<u8>, RoutingError> {
        serialisation::serialise(direct_message).map_err(|error| {
            error!("Failed to serialise direct message: {:?}", error);
            error.into()
        })
    }
```

src/states/bootstrapping.rs
maqi: ok, if we don't need to carry out any other action on this failure, then we don't need extra comment.

src/states/common/base.rs
maqi: 3 more lines VS. de-duplication, hard to decide   :)

btw. we do have some places in code that carrying out similar serialisation result check then log, such as:
```rust
            let mut to_sign = match serialisation::serialise(&old_and_new_pub_infos) {
                Ok(result) => result,
                Err(error) => {
                    error!("Failed to serialise public IDs: {:?}", error);
                    return;
                }
            };
```

just wondering, will it be worth to make it a helper macro.
it doesn't need to be in this PR, can be done in a separate one.

Fraser999: Hmm, I try and avoid macros where possible.  We can maybe try it later to see how it looks?

PR #1546, opened 2018-02-26
chore/ci: update ci scripts to latest rust stable

m-cat:


ustulation:
@nbaksalyar / @m-cat  has verified this works with alpha-2

PR #1549, opened 2018-04-03
MAID-2506: use crust dev

povilasb:
Updates crust to latest commit in `dev` branch. 
Latest crust is a total rewrite using Futures and Tokio, but it has compatibility layer, hence minimal changes are needed to integrate it to Routing.


povilasb:
deprecated in favor of https://github.com/maidsafe/routing/pull/1557

PR #1557, opened 2018-07-20
MAID-2886, MAID-2727, MAID-2909: integrate safe_crypto and latest crust

povilasb:
Uses `safe_crypto` crate that abstracts cryptographic primitives.
Crypto changes:
```
FullId --> safe_crypto::SecretKeys
PublicId --> safe_crypto::PublicKeys
rust_sodium::sign::PublicKey --> PublicSignKeys
```

Integrates latest Crust based on Tokio and Futures.
This Crust version supports multiple transport  protocols: uTP and TCP , implements end-to-end encryption, improves hole punching, implements peer bootstrap cache, etc.


examples/key_value_store.rs
```diff
@@ -38,12 +38,14 @@ extern crate maidsafe_utilities;
 #[macro_use]
 extern crate unwrap;
 extern crate docopt;
-extern crate rust_sodium;
 #[macro_use]
 extern crate serde_derive;
+#[cfg(not(feature = "use-mock-crust"))]
+extern crate sha2;
```
ustulation: 1. Why are we doing this ? Does not safe_crypto provide this for both mock and non-mock ?
2. We don't use sha2 anyway having migrated to sha3

src/data/mutable_data.rs
```diff
@@ -39,7 +39,7 @@ pub struct MutableData {
     version: u64,
     /// Contains a set of owners which are allowed to mutate permissions.
     /// Currently limited to one owner to disallow multisig.
-    owners: BTreeSet<PublicKey>,
+    owners: BTreeSet<PublicKeys>,
```
ustulation: Not correct at all - this is what was explicitly noted in the Friday Gossip on safe-crypto that we don't want encryption keys but *_only _* the signing keys and safe_crypto has that facility to hand over only the signing keys

src/messages/request.rs
```diff
@@ -56,7 +56,7 @@ pub enum Request {
         /// Unique message identifier
         msg_id: MsgId,
         /// Requester public key
-        requester: sign::PublicKey,
+        requester: PublicKeys,
```
ustulation: Again, not correct - pls don't rampantly use PublicKeys instead of sign::PublicKey - we don't want to include the encryption keys uselessly in the transmitted data and bloat it. Precisely the point of Friday Gossip on safe_crypto

src/state_machine.rs
```diff
@@ -246,7 +242,7 @@ impl StateMachine {
     // Construct a new StateMachine by passing a function returning the initial state.
     pub fn new<F>(
         init_state: F,
-        pub_id: PublicId,
+        sec_id: SecretKeys,
```
ustulation: Not checking this thoroughly just now as other fundamental changes are required as put in the review comments elsewhere, but do pls go over this and see this is what we are saying we need - a prior pub stuff getting replace by secret stuff is suspicious code. Might be totally legit but pls see what and why are we wanting this.

povilasb: In this case it makes sense cause `crust::Service` requires `SecretKeys` to do data encryption. Before Crust didn't encrypt anything.

src/states/node.rs
```diff
@@ -3052,27 +3052,19 @@ impl Node {
...
-            public_id.encrypting_public_key(),
-            self.full_id.encrypting_private_key(),
-        );
-
-        let serialised_connection_info =
-            decipher_result.map_err(|()| RoutingError::AsymmetricDecryptionFailure)?;
-        Ok(serialisation::deserialise(&serialised_connection_info)?)
+        shared_secret
+            .decrypt(encrypted_connection_info)
+            .map_err(|_e| RoutingError::AsymmetricDecryptionFailure)
```
ustulation: should we be logging `e` at-least, as `debug` instead of ignoring it as `_e` ?

src/mock_crust/support.rs
```diff
@@ -407,34 +417,41 @@ impl<UID: Uid> ServiceImpl<UID> {
...
+        );
         self.send_packet(their_info.endpoint, packet);
     }
 
     pub fn set_accept_bootstrap(&mut self, accept: bool) {
         self.accept_bootstrap = accept;
     }
 
-    pub fn start_listening_tcp(&mut self, port: u16) {
+    pub fn start_listening(&mut self) {
```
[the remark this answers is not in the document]
povilasb: Crust itself picks up listener addresses (with ports) from config. Since this mock doesn't actually listen for any networking connections, seems like this static port makes no difference.

src/mock_crust/support.rs
```diff
@@ -447,25 +464,38 @@ impl<UID: Uid> ServiceImpl<UID> {
...
 
-    fn handle_bootstrap_accept(&mut self, peer_endpoint: Endpoint, uid: UID, kind: CrustUser) {
-        let _ = self.add_connection(uid, peer_endpoint, kind);
+    fn handle_bootstrap_accept(
+        &mut self,
+        peer_endpoint: Endpoint,
+        uid: PublicKeys,
+        kind: CrustUser,
+    ) {
+        let _ = self.add_connection(uid.clone(), peer_endpoint, kind);
```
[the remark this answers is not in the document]
povilasb: image
oh we did, take a look on the left: the only thing that was changed is uid type :)

src/mock_crust/tests.rs
```diff
@@ -132,39 +133,39 @@ fn start_two_services_rendezvous_connect() {
...
     let our_ci_1 = expect_event!(event_rx_1,
-                                 CrustEvent::ConnectionInfoPrepared::<PublicId>(cir) => {
+                                 CrustEvent::ConnectionInfoPrepared(cir) => {
         assert_eq!(cir.result_token, PREPARE_CI_TOKEN);
         unwrap!(cir.result)
     });
 
-    let their_ci_0 = our_ci_0.to_pub_connection_info();
-    let their_ci_1 = our_ci_1.to_pub_connection_info();
+    let their_ci_0 = our_ci_0.clone();
```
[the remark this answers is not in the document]
povilasb: Crust compatibility layer doesn't expose private connection info anymore. See: https://github.com/maidsafe/crust/blob/dev/src/compat/event.rs#L26

src/mock_crust/support.rs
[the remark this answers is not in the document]
ustulation: indeed !

src/states/bootstrapping.rs
```diff
@@ -227,31 +225,30 @@ impl Bootstrapping {
...
                 );
-
-                self.rebootstrap();
+            } else {
+                return;
             }
+        } else {
+            return;
         }
+        self.rebootstrap();
```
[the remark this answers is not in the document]
povilasb: It's because of borrow checker. When we take a reference to `bootstrap_id` which is part of `self.bootstrap_connection`, we borrow `self` and can't call `self.rebootstrap()` in that scope

src/error.rs
```diff
@@ -94,7 +94,7 @@ pub enum RoutingError {
     /// Serialisation Error
     SerialisationError(serialisation::SerialisationError),
     /// Asymmetric Decryption Failure
-    AsymmetricDecryptionFailure,
+    AsymmetricDecryptionFailure(safe_crypto::Error),
```
ustulation: Is implementing `From` also not better ? That way you probably don't have to do `map_err`; simply `?` would convert it for you.

src/messages/mod.rs
```diff
@@ -1145,8 +1145,7 @@ mod tests {
 
         assert!(hop_message.verify(&public_signing_key).is_ok());
 
-        let secret_signing_key = SecretKeys::new();
-        let public_signing_key = secret_signing_key.public_keys().clone();
+        let public_signing_key = SecretKeys::new().public_keys().clone();
```
ustulation: this is not really `public_signing_key` so much as it is `pub_keys/pub_id`

src/states/bootstrapping.rs
```diff
@@ -225,19 +225,20 @@ impl Bootstrapping {
...
+        let mut retry_bootstrap = false;
         if let Some((ref bootstrap_id, bootstrap_token)) = self.bootstrap_connection {
             if bootstrap_token == token {
                 debug!(
                     "{:?} Timeout when trying to bootstrap against {:?}.",
                     self, bootstrap_id
                 );
-            } else {
-                return;
+                retry_bootstrap = true;
```
ustulation: why not just `return self.rebootstrap();` right here ?

povilasb: Because borrow checker won't allow us to use `self` in this place since it's borrowed in this context :/

PR #1572, opened 2019-03-07
CI Test Run Refactor

jacderida:
Hi Jean/Pierre,

This now runs 4 jobs in parallel:

* Clippy on Linux
* Clippy on OSX
* Tests on Linux
* Tests on OSX

As we discussed, the 2 long-running tests have been disabled.

Here is a log with timings for the 2nd run, where the cache takes effect: 
https://travis-ci.com/jacderida/routing/builds/103562433

There is also a small fix for something clippy flagged up.

Cheers,

Chris

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,8 @@ env:
   global:
     - RUST_BACKTRACE=1
     - PATH=$PATH:$HOME/.cargo/bin
-os:
-  - linux
-  - osx
 branches:
-  except:
-  - fleming
+  only: ci_test_run_refactor
 language: rust
 rust:
   - stable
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -2,12 +2,8 @@ env:
   global:
     - RUST_BACKTRACE=1
     - PATH=$PATH:$HOME/.cargo/bin
-os:
-  - linux
-  - osx
 branches:
-  except:
-  - fleming
+  only: ci_test_run_refactor
```
pierrechevalier83: You probably want to remove this, no?

.travis.yml
```diff
@@ -3,7 +3,7 @@ env:
     - RUST_BACKTRACE=1
     - PATH=$PATH:$HOME/.cargo/bin
 branches:
-  only: ci_test_run_refactor
+  only: fleming
```
pierrechevalier83: I think we don't need any restriction on the branch anymore. These scripts are only in this branch anyway.

jacderida: Right ok, no problem. Will remove this.

PR #1573, opened 2019-03-11
New direct message: ParsecPoke

madadam:
The purpose of this messase is that a newly joined node can ask for a gossip request with the same version it is on, even when all other members might have already moved to another version due to split.


pierrechevalier83:
This is now covered by https://github.com/maidsafe/routing/pull/1576 which adds more context to it. Let's close this PR.

PR #1574, opened 2019-03-12
chore/codeowners:added codeowners file

calumcraig:


pierrechevalier83:
@Viv-Rajkumar recommended that @afck is the other code owner instead, if he's OK with it.

calumcraig:
> @Viv-Rajkumar recommended that @afck is the other code owner instead, if he's OK with it.

OK, I'll wait for him to confirm then I can update the codeowners file.

afck:
I'm okay with it.

PR #1576, opened 2019-03-13
Implement mock parsec

madadam:
This PR either supersedes https://github.com/maidsafe/routing/pull/1573 or needs to go after it.


Cargo.toml
```diff
@@ -30,7 +30,7 @@ serde_derive = "1.0.25"
 term = "0.4.6"
 unwrap = "1.2.1"
 # Parsec master branch.
-parsec = {git = "https://github.com/maidsafe/parsec", rev = "29d4b48"}
+parsec = { git = "https://github.com/maidsafe/parsec", rev = "f447952c70b068f0cfec783245d38a65fbddef56" }
```
pierrechevalier83: What changes from Parsec did you need?

madadam: Can't remember anymore :flushed: . But it seems it still works with the original revision, so I'm going to revert this.

src/states/node.rs
```diff
@@ -1541,22 +1550,41 @@ impl Node {
...
+                    full_id,
+                    &genesis_info.our_info.members(),
+                    &genesis_info.latest_info.members(),
+                    consensus_mode,
+                )
+            };
+
+            #[cfg(feature = "mock")]
+            let parsec = {
+                let section_info = genesis_info.our_info.prefix().with_version(genesis_ver);
```
Viv-Rajkumar: ah can you make this not rely on `VersionedPrefix` here pls. That type is just lingering code left behind until the routing_table state gets derived from the Chain itself. If this just took a Digest(Prefix+ver), we're prolly better off here or would have to make that change when we remove that type anyways.

Also maybe name the variable ver_pfx or something pls. `section_info` is an actual type(`SectionInfo`) used in the Chain and might be confusing this called the same.

src/mock_parsec/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+        unimplemented!()
+    }
+
+    #[allow(unused)]
+    pub fn have_voted_for(&self, _observation: &Observation<T, S::PublicId>) -> bool {
+        unimplemented!()
+    }
+
+    pub fn has_unconsensused_observations(&self) -> bool {
+        state::with::<T, S::PublicId, _, _>(self.section_hash, |section_state| {
```
afck: I'd rather call this one `state`, and the one below `observation_state`, since the section state is called `state` in the other methods, too.

src/mock_parsec/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+            Observation::Accusation { ref offender, .. } => {
+                let _ = self.peer_list.remove(offender);
+            }
+            _ => (),
+        }
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug)]
+#[serde(bound = "")]
```
afck: Is that attribute really needed everywhere? Aren't `NetworkEvent` and `PublicId` both serialisable?

src/mock_parsec/observation.rs
```diff
@@ -0,0 +1,128 @@
...
+// Returns whether `small` is more than two thirds of `large`.
+fn is_more_than_two_thirds(small: usize, large: usize) -> bool {
+    3 * small > 2 * large
+}
+
+fn serialise<T: Serialize>(data: &T) -> Vec<u8> {
+    if let Ok(serialised) = serialisation::serialise(data) {
+        serialised
+    } else {
+        vec![]
```
afck: Shouldn't we rather panic in this case?

src/mock_parsec/state.rs
```diff
@@ -0,0 +1,77 @@
...
+            observations: BTreeMap::new(),
+            blocks: Vec::new(),
+        }
+    }
+}
+
+type NetworkState<T, P> = HashMap<Digest256, SectionState<T, P>>;
+
+thread_local! {
+    static STATE: RefCell<Option<Box<dyn Any>>> = RefCell::new(None);
```
afck: Why `Any` and not `NetworkState`?

src/mock_parsec/mod.rs
madadam: Hm. I removed it and it still compiles. I remember having issues with this before until I added that attribute. Maybe `serde` is not smart enough to figure the trait bounds somehow? Anyway, the issue seems to be gone now.

src/mock_parsec/state.rs
madadam: `NetworkState` is generic. I'd have to instantiate it with concrete types before putting it into `thread_local` which I don't think I can, because at least the `P` param is different depending on whether we use mock crypto or not. Alternative would be to use conditional compilation. Not sure which is worse :thinking:

PR #1583, opened 2019-04-02
Document source diagrams

pierrechevalier83:


pierrechevalier83:
I'll rebase later. For now, we should aim to merge JP's PR 1582 first.

PR #1587, opened 2019-04-04
More documentation

pierrechevalier83:


routing_model/index.html
```diff
@@ -684,7 +707,7 @@ <h2>Process Adult/Elder promotion and demotion needed from last check</h2>
       OneVoteConsensused --> WaitComplete
       WaitComplete(("Check?"))
       WaitComplete--"WAITED_VOTES<br/>.is_empty()<br/>(Wait complete)"-->MarkNewElderAdults
-      MarkNewElderAdults["For all nodes:<br/>set_elder_status(node, is_elder)"]
+      MarkNewElderAdults["for node in nodes {<br/>is_elder = new_section_info.contains(node)<br/>set_elder_status(node, is_elder)<br/>}"]
```
jeanphilippeD: You don't have access to nodes.
Maybe you'd be better off creating a function you want:
`update_elder_status(new_section_info)`

routing_model/index.html
```diff
@@ -733,7 +756,7 @@ <h2>ProcessMerge Sub-routine</h2>
       CompleteMerge["Complete_merge()<br/>(Start parsec with new genesis...)"]
       CompleteMerge --> MarkNewElderAdults
 
-      MarkNewElderAdults["For all nodes:<br/>set_elder_status(node, is_elder)<br/>based on new section info"]
+      MarkNewElderAdults["for node in nodes {<br/>is_elder = new_section_info.contains(node)<br/>set_elder_status(node, is_elder)<br/>}"]
```
jeanphilippeD: same

routing_model/index.html
```diff
@@ -714,12 +714,34 @@ <h2>Process Adult/Elder promotion and demotion needed from last check</h2>
...
   </div>
 
-  <h2>ProcessMerge Sub-routine</h2>
+  <h2>Handling merges</h2>
   <div class=description>
   <p>Vote for ParsecOurMerge, and take over handling any ParsecNeighbourMerge.<br/>
      Complete when one merge has completed, and a NewSectionInfo is consensused.<br/>
-     If Multi stage merges are required, they will require calling this function again<br/>
+     If multi-stage merges are required, they will require calling this function again<br/>
+     While in this sanctuary, our SectionInfo shall not be disturbed by elder/adult changes.
```
jeanphilippeD: adult may change: only elder do not change.

routing_model/index.html
```diff
@@ -714,12 +714,34 @@ <h2>Process Adult/Elder promotion and demotion needed from last check</h2>
...
+     This avoids troublesome situations where different nodes may send different SectionInfo's to the neighbour section and cause confusion.
```
jeanphilippeD: No, we are in shared state, so this cannot happen.
What could happen, is we generate a new SectionInfo after the one we said would be the parent of the merged SectionInfo.

So maybe:
"This stops us from changing our SectionInfo after indicating to our neighbour the last SectionInfo before merge"

routing_model/index.html
pierrechevalier83: I'll reword. I meant change as in going online/offline.

jeanphilippeD: no, they may go online offline (including elder), it is ok, they just not loose/gain elder status

pierrechevalier83: Oh. I see. You mean it's Ok if adults change during that time frame. It is. Now with the Online/Offline flow, it could actually happen.

jeanphilippeD: Yes, and it is even ok for an Elder to be voted offline, it will just not be removed from Parsec or the section until after the merge.

routing_model/index.html
```diff
@@ -729,14 +729,16 @@ <h2>Process Adult/Elder promotion and demotion needed from last check</h2>
...
+  <button class="collapsible">update_elder_status</button>
   <div class="content">
-    <p>This function updates the chain to reflect the eldership status of a given node.<br>
-    Input:
-    <lu>
-      <li> node: PublicId
-      <li> is_elder: bool
-    </lu>
+    <p>This function updates the eldership status of each node in the chain based on the new section info.<br>
+    It could look something like this:<br>
```
jeanphilippeD: I'm not sure it is sensible to have pseudo code here since you then have to document the sub function.

Maybe just:
"<p>This function updates the eldership status of each node in the chain based on the new section info: Only nodes with their public id in new_section_info are elders.<br>"
+ document intput parameter of update_elder_status like you did for other.

PR #1588, opened 2019-04-04
Decouple Relocate processing and integrate with merge

jeanphilippeD:
This is an initial attempt at decoupling the relocating flow to integrate better with merging.
Also update the documentation.

Full PR:
https://raw.githack.com/maidsafe/routing/4a51d86d04c9d68529359f532a5696f961906516/routing_model/index.html

### DIFF
--- a/routing_model/index.html
+++ b/routing_model/index.html
@@ -281,7 +281,7 @@
   <div class="mermaid">
       graph TB
 
-      Start["StartTopLevelDst:<br/>No exit - Need Killed"]
+      Start["StartTopLevelDst:<br/>No exit - Needs killed"]
       style Start fill:#f9f,stroke:#333,stroke-width:4px
       Start --> LoopStart
       LoopEnd --> LoopStart
@@ -427,17 +427,18 @@
   This is maintained accross merge/split as for any adult node, so we are ready to resource proof any node in WaitingProofing state<br/>
   <br/>
   This leads us here: to the resource proof.<br/>
-  We only process one resource proof at a time (only schedule a new proofing when completing the current one).<br/>
+  We only process up to one resource proof at a time (i.e: we wait for the current resource proof to reach completion before scheduling a new one).<br/>
   When we periodicaly decide to resource proof a node, we check if any node is ready for it:  WaitingProofing state, and pick the best candidate (There may be mutliple after a merge).<br/>
   As an elder, I will send the candidate a ResourceProof RPC. This gives them the "problem to solve". As they solve it, they will send me ResourceProofResponses. These will be parts of the proof. On receiving valid parts, I must send a ResourceProofReceipt. Once they finally send me the last valid part, they passed their resource proof and I vote for ParsecOnline (essentially accepting them as a member of my section).<br/>
   At any time during this process, they may timeout (The whole process it taking longer than expected), in which case I will decided to reject them and vote for ParsecPurgeCandidate.<br/>
-  This process ends once I reach consensus on either accepting or the candidate (ParsecOnline) or refusing them (ParsecPurgeCandidate).<br/>
+  This process ends once I reach consensus on either accepting the candidate (ParsecOnline) or refusing them (ParsecPurgeCandidate).<br/>
   It is possible that both reach the quorum consensus, in which case the second one will be discarded in the flow above. It won't cause issues as consistency is the only property that matters here: if we accept someone who then went Offline, we will be able to detect they are Offline later with the standard Offline detection mechanism. But it is more likely that they took close to the time limit to complete their proof.
   </p>
   <button class="collapsible">CANDIDATE</button>
   <div class="content">
-    <p>Option&lt(Candidate, Nonce)&gt we are currently ressource proofing.<br/>
-      To handle re-doing resource proof after merge/split so we can fairly measure a node, we use the Nonce to differentiate the tries.
+    <p>Option&lt(Candidate, Nonce)&gt.<br/>
+	  The candidate we are currently ressource proofing, if any.<br/>
+      The nonce is used to distinguish new attempts of resource proofing the same candidate after a merge or a split.
     </p>
   </div>
   <button class="collapsible">VOTED_ONLINE</button>
@@ -446,10 +447,10 @@
     This local variable helps us with this.
     </p>
   </div>
-  <button class="collapsible">RPC: ResourceProofResponse/<br/>CandidateInfo</button>
+  <button class="collapsible">RPC: ResourceProofResponse</button>
   <div class="content">
-  <p>These RPCs may continue to be sent by a node we have not accepted, even if consensus was reached to add it.<br/>
-  It's OK to discard the RPC in this case as they are no longer relevant RPCs.
+  <p>This RPC may continue to be sent by a node we have not accepted, even if consensus was reached to add it.<br/>
+  It's OK to discard the RPC in this case as it is no longer relevant.
   </p>
   </div>
   <button class="collapsible">ParsecOnline<br/>ParsecPurgeCandidate</button>
@@ -460,7 +461,7 @@
   </div>
   <button class="collapsible">ResourceProof_Cancel</button>
   <div class="content">
-    <p>Provides and external entry point to cancel the currently processed nodes: Restart the resource proofing with all involved voters.<br/>
+    <p>Provides an external entry point to cancel the currently processed nodes: Restart the resource proofing with all involved voters.<br/>
        This will be called for example after merge/split as the new nodes would become voters.
     </p>
   </div>
@@ -497,7 +498,7 @@
 
       CheckRequestRP((Check))
       CheckRequestRP --"CANDIDATE.is_none()" --> ReStartCheckResourceProofTimeout
-      ReStartCheckResourceProofTimeout["CANDIDATE=None><br/>VOTED_ONLINE==no<br/><br/>schedule(<br/>CheckResourceProofTimeout)"]
+      ReStartCheckResourceProofTimeout["CANDIDATE=None<br/>VOTED_ONLINE==no<br/><br/>schedule(<br/>CheckResourceProofTimeout)"]
       ReStartCheckResourceProofTimeout --> LoopEnd
 
       CheckRequestRP --"CANDIDATE.is_some()"-->RequestRP
@@ -516,7 +517,7 @@
 
 
       DiscardRPC[Discard RPC]
-      RPC -- ResourceProofResponse --> DiscardRPC
+      RPC -- ResourceProofResponse<br/>from CANDIDATE<br/><br/>VOTED_ONLINE==yes --> DiscardRPC
       DiscardRPC --> LoopEnd
 
       WaitFor --Event--> Event
@@ -545,14 +546,14 @@
       style EndCancel fill:#19f,stroke:#333,stroke-width:4px
 
       Cancel --> CancelCheckResourceProofTimeout
-      CancelCheckResourceProofTimeout["CANDIDATE=None><br/>VOTED_ONLINE==no<br/><br/>schedule(<br/>CheckResourceProofTimeout)"]
+      CancelCheckResourceProofTimeout["CANDIDATE=None<br/>VOTED_ONLINE==no<br/><br/>schedule(<br/>CheckResourceProofTimeout)"]
       CancelCheckResourceProofTimeout --> EndCancel
   </div>
 
   <h1>Source section</h1>
   <div class=description>
   <p>As members of a section, each node must keep track of how many "work units" other nodes have performed.<br/>
-      Once a node has done accumulated enough work units to gain age, the section must work together to relocate that node to a new section where they can become 1 age unit older.<br/>
+      Once a node has accumulated enough work units to gain age, the section must work together to relocate that node to a new section where they can become 1 age unit older.<br/>
       These diagrams detail how this happens.
   </p>
   </div>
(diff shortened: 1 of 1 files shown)


routing_model/index.html
```diff
@@ -263,17 +270,11 @@ <h2>Deciding when to accept an incoming relocation</h2>
...
-    The local variable: PROCESSING_CANDIDATE indicates whether someone is already passing the resource proof process, so we can't accept anyone else for now.<br/>
-    When PROCESSING_CANDIDATE is true and we reach consensus on ParsecExpectCandidate, we send a RefuseCandidate RPC to the would-be-incoming-node so they can try another section or try again later.
-    </p>
-  </div>
-  <button class="collapsible">AcceptAsCandidate</button>
-  <div class="content">
-    <p>If we are ready to accept an incoming node and we reach consensus on ParsecExpectCandidate, we will execute this flow to make them pass resource proof.<br/>
-    See AcceptAsCandidate diagram for details.
+    waiting_proofing function return the list of nodes that we have yet to resource proof, (States from WaitingCandidateInfo until it reach OnlineState).<br/>
+    When waiting_proofing() is not empty and we reach consensus on ParsecExpectCandidate, we send a RefuseCandidate RPC to the would-be-incoming-node so they can try another section or try again later.
```
pierrechevalier83: I would add an explanation for the leftmost branch here: if it's not empty, but it's the same guy, we resend our RelocateResponse.

routing_model/index.html
```diff
@@ -197,14 +199,33 @@ <h2>Node as a member of a section</h2>
...
+
+
+    ConcurentStartElder --> ConcurentStartDst
+    ConcurentStartDst{"Concurrent<br/>start dst"}
+
+    ConcurentStartDst --> StartTopLevelDst
+    style StartTopLevelDst fill:#f9f,stroke:#333,stroke-width:4px
+
+    ConcurentStartDst --> StartRelocatedNodeConnection
+    StartRelocatedNodeConnection[Start RelocatedNodeConnection]
```
pierrechevalier83: `Start RelocatedNodeConnection` and `Start ResourceProof`:
These spaces are inconsistent with all other subroutines

routing_model/index.html
```diff
@@ -319,14 +320,34 @@ <h2>Deciding when to accept an incoming relocation</h2>
...
+  <div class="content">
+    <p>Provides and external entry point to reset the currently processed nodes: Do not reject a node because it took longer than expected.<br/>
+       This will be called for example after merge/split as the new nodes would become voters.
+    </p>
+  </div>
   </div>
   <div class="mermaid">
       graph TB
-      RelocatedNodeConnection["Start Relocated Node Connection"]
+      RelocatedNodeConnection["Start RelocatedNodeConnection"]
```
pierrechevalier83: One more space should go :wink:

routing_model/index.html
```diff
@@ -319,14 +320,34 @@ <h2>Deciding when to accept an incoming relocation</h2>
 
   <h2>Relocated node connection</h2>
   <div class=description>
-  <p>Manage node with NodeState=WaitingConnectionInfo and connection info request/response.<br/>
-    CandidateInfo is sent to NaeManager of the target interval<br/>
-    When we complete, we either stop responding to CandidateInfo RPCs, or we send NodeConnected RPC<br/>
+  <p>Manage node with NodeState=WaitingConnectionInfo and connection info request/response RPCs.<br/>
+    The candidate sends CandidateInfo to NaeManager of the target interval, so it will reach the managing section after split/merge.<br/>
```
pierrechevalier83: "The candidate sends CandidateInfo" -> "The candidate sends its CandidateInfo"

pierrechevalier83: "to NaeManager" -> "to the NaeManager"

pierrechevalier83: " so it will reach the managing section after split/merge." -> "to ensure it reaches a section even in the event of a split or a merge." ?

routing_model/index.html
```diff
@@ -319,14 +320,34 @@ <h2>Deciding when to accept an incoming relocation</h2>
...
+    This target interval address could be the middle address of the interval, but the full interval should not be overlapping two section after a split.<br/>
```
pierrechevalier83: "two section" -> "two sections"

pierrechevalier83: This is a bit unclear to me. Could you explain what you mean by "the full interval shouldn't be overlapping" ?

routing_model/index.html
```diff
@@ -319,14 +320,34 @@ <h2>Deciding when to accept an incoming relocation</h2>
...
+    When we complete, we either stop responding to CandidateInfo RPCs if it failed, or we send NodeConnected RPC on success,<br/>
+    (We could also omit this RPC, the node would continue sending CandidateInfo until ResourceProof or time out).<br/>
+    When the node reach WaitingProofing state, the section becomes responsible for managing communication with this node as it would any of its adults.
```
pierrechevalier83: "When the node reach" -> "When the node reaches"

routing_model/index.html
```diff
@@ -319,14 +320,34 @@ <h2>Deciding when to accept an incoming relocation</h2>
...
   </p>
+  <button class="collapsible">is_valid_waited_info</button>
+  <div class="content">
+    <p>Return true if the given CandidateInfo is a valid CandidateInfo and match a one of our node and it is in WaitingCandidateInfo state<br/>
```
pierrechevalier83: "and match a one" -> "and matches one"
"of our node" -> "of our nodes"

pierrechevalier83: Maybe I would reformulate the sentence a bit:
```
Return true if
<ul>
  <li> The given CandidateInfo is valid,
  <li> It matches one of our nodes that is in the WaitingCandidateInfo state,
  <li> The message_src (from the current RPC) is consistent with the CandidateInfo's new_public_id
</ul>
```

routing_model/index.html
```diff
@@ -319,14 +320,34 @@ <h2>Deciding when to accept an incoming relocation</h2>
...
+       and the given message_src of the current RPC match the CandidateInfo new_public_id.
```
pierrechevalier83: "match" -> "matches"

routing_model/index.html
```diff
@@ -319,14 +320,34 @@ <h2>Deciding when to accept an incoming relocation</h2>
...
+    </p>
+  </div>
+  <button class="collapsible">is_valid_waited_connection</button>
+  <div class="content">
+    <p>Return true if the given message_src of the current RPC is a one of our node and is in WaitingConnectionInfo state.
```
pierrechevalier83: "is a one of our node" -> "is one of our nodes"

routing_model/index.html
```diff
@@ -398,19 +419,21 @@ <h2>Relocated node connection</h2>
 
   <h2>Resource proof from a destination's point of view</h2>
   <div class=description>
-  <p>In the previous diagram, we ensured an incoming candidate would only be processed for resource_proof if we are the best fit we know and we are not currently processing another node.<br/>
+  <p>In the previous diagram, we ensured an incoming candidate is connected and in our peer_list with state WaitingProofing, and elders are ready to communicate with it.<br/>
```
pierrechevalier83: A bit hard to read.
Maybe this is easier:
"In the previous diagram, we ensured an incoming candidate would only reach the "WaitingProofing" state once it was connected to our section and able to communicate with its elders. At this stage, the node would be a member of our peer_list."

routing_model/index.html
```diff
@@ -197,14 +199,33 @@ <h2>Node as a member of a section</h2>
...
-    style StartCheckAndProcessElderChange fill:#f9f,stroke:#333,stroke-width:4px
+    ConcurentStartSrc --> StartRelocateSrc
+    style StartRelocateSrc fill:#f9f,stroke:#333,stroke-width:4px
+
+
+    ConcurentStartElder --> ConcurentStartDst
+    ConcurentStartDst{"Concurrent<br/>start dst"}
+
+    ConcurentStartDst --> StartTopLevelDst
+    style StartTopLevelDst fill:#f9f,stroke:#333,stroke-width:4px
```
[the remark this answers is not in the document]
jeanphilippeD: Yes, and so are some of the RPC names, I think I'll follow up with name clean up at some point soon, we should discuss what they should be.

routing_model/index.html
jeanphilippeD: This behavior is not in this PR unless I messed up something?

routing_model/index.html
jeanphilippeD: Yes, I'm not super happy with these names either. Fixed for now.

routing_model/index.html
jeanphilippeD: Replaced with:

and the interval should cover a range that will not be split if the section splits.

routing_model/index.html
pierrechevalier83: Oops. I was not looking at the exact correct graph...

routing_model/index.html
```diff
@@ -591,95 +611,109 @@ <h2>Deciding that a member of our section should be relocated away</h2>
...
+  <br/>
+  Also of note: While we are preventing OnlineState node from relocating when we are already relocating node, we may be actively relocating more nodes away from our section.<br/>
+  However, we will only do one at a time per CheckRelocateTimeOut event.
   </p>
+  <button class="collapsible">get_best_relocating_node_and_target</button>
+  <div class="content">
+    <p>Returns the best node to relocate and the target address to send it to<br/>
+    There may be mutliple nodes relocating, for example because of a merge. Take the best one (oldest), and choose a target address.<br/>
+    The target address is one managed by one of our neighbour. This could be random, or the current old_public_id with a single bit of the prefix flipped.<br/>
+    This would would help ensure that source and destination remain neighbours, even if the source split.<br/>
```
pierrechevalier83: This makes me wonder: what happens if we're relocating a node to a sibling, but then we merge and they get relocated to our own section? (Probably just works, but let's make sure)

routing_model/index.html
```diff
@@ -591,95 +611,109 @@ <h2>Deciding that a member of our section should be relocated away</h2>
...
+    Using a target address instead of a section allow to ensure we deliver the message even if the destination splits or merge.
+    </p>
+  </div>
+  <button class="collapsible">RefuseCandidate/<br/>RelocateResponse</button>
+  <div class="content">
+    <p>Exactly one of these RPCs will be sent to us from the destination section as a response to our section's ExpectCandidate RPC (see TryRelocating for more context).<br/>
```
pierrechevalier83: `s/TryRelocating/StartTopLevelDst` after the current reshuffling.

routing_model/index.html
```diff
@@ -289,47 +293,108 @@ <h2>Deciding when to accept an incoming relocation</h2>
...
+      CleanCandidates["for candidate in both<br/>-waiting_node_connecting()<br/>-CANDIDATES:<br/><br/>purge_node_info(candidate)"]
+      CleanCandidates --> ReStartCheckRelocatedNodeConnectionTimeout
+      ReStartCheckRelocatedNodeConnectionTimeout["CANDIDATES=waiting_node_connecting()<br/><br/>schedule(<br/>CheckRelocatedNodeConnectionTimeout)"]
+      ReStartCheckRelocatedNodeConnectionTimeout --> LoopEnd
 
 
-      AcceptAsCandidate["AcceptAsCandidate<br/>(shared state)"]
-      style AcceptAsCandidate fill:#f9f,stroke:#333,stroke-width:4px
-      AcceptAsCandidate --> SetCandidateNo
+      ParsecConsensus -- "ParsecCandidateInfo<br/>and<br/>is_valid_waited_info(<br/>CandidateInfo)" --> SetCandidateInfo
```
pierrechevalier83: I think this would be clearer with a "check?" box in between `ParsecConsensus` and the 3 variations of `CandidateInfo`. Currently, it looks visually like there were 4 different RPCs to handle...

routing_model/index.html
```diff
@@ -591,95 +611,109 @@ <h2>Deciding that a member of our section should be relocated away</h2>
...
-      ResetRelocating["RELOCATING=None<br/>(shared state)"]
-      ResetRelocating --> LoopEnd
+      SetRelocatingNodeState["set_node_state(get_node_to_relocate(), RelocatingState)"]
+      SetRelocatingNodeState --> LoopEnd
   </div>
 
   <h2>Relocating a member of our section away from it</h2>
   <div class=description>
-  <p>At this stage, we have decided that one of our section members should be relocated. The process is quite simple: we send an ExpectCandidate RPC to the destination section. That section will eventually either pass the node to a section with a shorter prefix, send us a RefuseCandidate RPC or eventually send us a RelocateResponse RPC. If the destination picked another section with a shorter prefix, that section will send us one of these RPCs (or suggest another section with a shorter prefix, but this recursion will end at some point).<br/>
+  <p>At this stage, we have decided that one of our section members should be relocated.<br/>
```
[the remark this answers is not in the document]
jeanphilippeD: I'll try to update.
What I was trying to say is this flow only cater to node we have decided to relocate, i.e it will do nothing unless a node is in RelocatingState.

routing_model/index.html
jeanphilippeD: Yes, there is nothing preventing self relocation.
I'll add a note.

routing_model/index.html
jeanphilippeD: Should I not remove the (see...) since this use to be the top level comment talking about its sub function, but now it is all in the stand alone routine.

PR #1592, opened 2019-04-08
Send Merge RPC without extra consensus round

jeanphilippeD:
We are already in a shared state when entering ProcessMerge as we come
from consensus on a timer event.
All nodes agree, so we can send the RPC imediately.

### DIFF
--- a/routing_model/index.html
+++ b/routing_model/index.html
@@ -746,7 +746,7 @@
   </div>
   <div class="mermaid">
       graph TB
-      ProcessElderChange["ProcessElderChange<br/>(Take elder changes)"]
+      ProcessElderChange["ProcessElderChange<br/>(Take elder changes)<br/>(shared state)"]
       style ProcessElderChange fill:#f9f,stroke:#333,stroke-width:4px
 
       EndRoutine["End of ProcessElderChange<br/>(shared state)"]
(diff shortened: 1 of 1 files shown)


routing_model/index.html
```diff
@@ -746,7 +746,7 @@ <h2>Process Adult/Elder promotion and demotion needed from last check</h2>
   </div>
   <div class="mermaid">
       graph TB
-      ProcessElderChange["ProcessElderChange<br/>(Take elder changes)"]
+      ProcessElderChange["ProcessElderChange<br/>(Take elder changes)<br/>(shared state)"]
```
pierrechevalier83: I'm not too convinced about these "shared state" annotations... What do they mean exactly? I would prefer to explicitely mention which state is shared (or leave it to the more detailed explanations)

routing_model/index.html
```diff
@@ -832,14 +825,14 @@ <h2>Handling merges</h2>
...
+      ProcessMerge["ProcessMerge<br/>(shared state)"]
       EndRoutine["End of ProcessMerge<br/>(shared state)"]
       style ProcessMerge fill:#f9f,stroke:#333,stroke-width:4px
       style EndRoutine fill:#f9f,stroke:#333,stroke-width:4px
 
-      ProcessMerge -->  VoteOurMerge
-      VoteOurMerge["vote_for(<br/>ParsecOurMerge)"]
-      VoteOurMerge --> LoopStart
+      ProcessMerge -->  SendMergeRpc
+      SendMergeRpc["OUR_MERGE=true<br/>send_rpc(Merge)"]
```
pierrechevalier83:
```
OUR_MERGE=true
```
I think this is unused now so could be removed
```
send_rpc(Merge)
```
Now it is done unconditionally, I think it would make sense to move it to `StartCheckAndProcessElderMergeChange`, that way it's the only place that needs to handle Merge RPCs

routing_model/index.html
```diff
@@ -832,14 +825,14 @@ <h2>Handling merges</h2>
   </div>
   <div class="mermaid">
       graph TB
-      ProcessMerge["ProcessMerge<br/>"]
+      ProcessMerge["ProcessMerge<br/>(shared state)"]
```
pierrechevalier83: Same comment about shared state, but happy to discuss it.

routing_model/index.html
jeanphilippeD: This refers to the chain.rs (i.e all information that we computed from consensused information).
We cannot update this if we are not triggered by a consensus event.
We probably want an explanation somewhere.

routing_model/index.html
jeanphilippeD: I do not think so, this is an implementation detail of ProcessMerge, as this change shows, it did not need to happen there.

jeanphilippeD: Also, it is a different Merge RPC in the sense that we are sending that one, but receiving the other.
However the main point is that when we really process the merge RPC is when we check for consensus, so I'm not sure that it would achieve the goal of 'only place that needs to handle Merge RPCs'

routing_model/index.html
pierrechevalier83: Ah! I see! OK, we can make it clear by documenting it. Maybe even come up with a better name for it such as: `in order` or so (we can do this in a different PR, though)

routing_model/index.html
pierrechevalier83: As discussed, we have a minor disagreement on this one, so we agreed to go with your suggestion.

PR #1593, opened 2019-04-08
Add mock_parsec feature

madadam:
So mock crust / mock crypto can be enabled independently from mock parsec.

Also rename `mock_parsec::init` to `mock_parsec::init_mock` for more explicitness.

### DIFF
--- a/scripts/tests
+++ b/scripts/tests
@@ -2,5 +2,7 @@
 
 set -x -e
 export RUSTFLAGS="-C opt-level=2 -C codegen-units=8"
+DIR=`dirname "$0"`
 
-cargo test --release --features=mock_parsec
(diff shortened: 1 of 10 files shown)


scripts/tests
```diff
@@ -2,5 +2,7 @@
 
 set -x -e
 export RUSTFLAGS="-C opt-level=2 -C codegen-units=8"
+DIR=`dirname "$0"`
 
-cargo test --release --features=mock_parsec
+cargo test --verbose --release --features=mock_parsec -- --skip aggressive_churn
+$DIR/travis_wait -l 1800 cargo test --verbose --release --features=mock_parsec aggressive_churn
```
pierrechevalier83: missing newline (as Viv pointed out :smile:)

PR #1594, opened 2019-04-09
Routing model handle replay,loss,trust and split

jeanphilippeD:
Rendered:
https://raw.githack.com/maidsafe/routing/34a0bb7adbcebafb25c1d923365f6c65bda87677/routing_model/index.html

(first 4 commits: https://raw.githack.com/maidsafe/routing/1b760bc491526a193a94022207450666b9fadd65/routing_model/index.html)

(first 2 commits: https://raw.githack.com/maidsafe/routing/81f5b6043660f59caefb0c19b32e05ca8d0abdd6/routing_model/index.html)

Note:
Initially remove re-send ExpectCandidate to section with shorter_prefix.
It complicates flow by involving unknown number of section in the process.
Re-add this functionality between the connection being established and the resource proof started as an immediate relocate with the same age.

Handles RPC Replay for ExpectCandidate/RelocateResponse:
 - Destination only trust a CandidateInfo if it has proof that the source decided on that relocation
 - Source only generate proof for a single relocation per node it relocates (The first one it consensus).

Handle Loss:
- Because we can now have multiple ExpectCandidate/RelocateResponse: Resend ExpectCandidate if it takes too long (We don't think it will come back), until we get a valid RelocateResponse.

Handle Trust:
- This only address the trust a node has in the new section it joins. In Alpha2, it already has the public Id of the elder it needs to connect too, so that was not an issue. But to handle merge/split, the node will not know the elder it needs to connect to. The Same public Id are given to the node as the initial SectionInfo, and the current set of elder gives a proof that it is a descendant section to the initial SectionInfo.

Handle Split:
- Handle similarly to merge and elder change: Vote for 2 new section info and wait for them before moving on. 

Different types of relocation are handled differently AgeIncrease, Hop and BackOnline.

### DIFF
--- a/routing_model/index.html
+++ b/routing_model/index.html
@@ -262,11 +262,10 @@
     Note that we could also see consensus on ParsecExpectCandidate before we ourselves voted for it in PARSEC, as long as enough members of our section did.
     </p>
   </div>
-  </div>
   <button class="collapsible">waiting_proofing_or_hop</button>
   <div class="content">
     <p>We want to accept at most one incoming relocation at a time into our section.<br/>
-    The waiting_proofing_or_hop function returns the list of nodes that we have yet to resource proof or relocated through a new hop, (States from WaitingCandidateInfo until it reaches OnlineState  or RelocatedState).<br/>
+    The waiting_proofing_or_hop function returns the list of nodes that we have yet to resource proof or relocate through a new hop, (States from WaitingCandidateInfo until it reaches OnlineState  or RelocatedState).<br/>
     When the output of this function is not empty and we reach consensus on ParsecExpectCandidate, we send a RefuseCandidate RPC to the would-be-incoming-node so they can try another section or try again later.
     </p>
   </div>
@@ -386,7 +385,7 @@
       ParsecConsensus -- "ParsecCandidateConnected<br/>and<br/>is_valid_waited_connection(<br/>message_src)" --> Balanced
       Balanced(("Check"))
       Balanced -- "shorter_prefix_section(<br/>).is_some()" --> RelocateToShorterPrefix
-      RelocateToShorterPrefix["set_node_state(<br/>candidate,<br/>RelocatingState(Hop)"]
+      RelocateToShorterPrefix["set_node_state(<br/>candidate,<br/>RelocatingState(Hop))"]
       RelocateToShorterPrefix --> SetConnected
 
       Balanced -- "Otherwise" --> SetWaitingProof
@@ -680,12 +679,12 @@
     Returns the best node to relocate and the target address to send it to<br/>
     There may be mutliple nodes relocating, for example because of a merge. Take the best one (oldest), and choose a target address.<br/>
     The target address is one managed by one of our neighbours. This could be random, or the current old_public_id with a single bit of the prefix flipped.<br/>
-    This would would help ensure that source and destination remain neighbours, even if the source splits.<br/>
+    This would help ensure that source and destination remain neighbours, even if the source splits.<br/>
     Using a target address instead of a section ensures we deliver the message even if the destination splits or merges.<br/>
     <br/>
     There are 3 states the node may be: RelocatingState(AgeIncrease), RelocatingState(Hop), and RelocatingState(BackOnline).<br/>
-    Node will be selected in this order: Relocating(AgeIncrease), RelocatingState(Hop) and then RelocatingState(BackOnline).<br/>
-    This ensures that we prioritizing good node relocation.
+    Nodes will be selected in this order: RelocatingState(AgeIncrease), RelocatingState(Hop) and then RelocatingState(BackOnline).<br/>
+    This ensures that we prioritize good node relocation.
     <br/>
     Note: It may be possible for a node to relocate to its sibling, and complete relocation after a merge occured.
     </p>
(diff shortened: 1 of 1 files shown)


routing_model/index.html
```diff
@@ -790,6 +850,15 @@ <h2>Process for Adult/Elder promotion and demotion including merge</h2>
...
+    <p>This function indicates that we need splitting.<br>
+    The details for the trigger are still in slight flux, but here are some possibilities:
+    <ul>
+      <li> The number of Online adults plus elders is above our min section size of 10 elder plus 90 Adults + buffer
+    </ul>
+    </p>
+  </div>
   <div class="mermaid">
       graph TB
       CheckAndProcessElderChange["StartCheckAndProcessElderMergeChange:<br/>No exit - Needs killed"]
```
pierrechevalier83: This is one more name that's getting really out of touch (maybe for another commit, though).

routing_model/index.html
```diff
@@ -342,6 +336,19 @@ <h2>Relocated node connection</h2>
...
     </p>
   </div>
+  <button class="collapsible">shorter_prefix_section</button>
+  <div class="content">
+    <p>If we know of a section that has a shorter prefix than ours, we prefer for them to receive this incoming node rather than ourselves as it will help keep the Network's sections tree balanced.<br/>
+    This shorter_prefix_section is a function that will return None if we are the shortest of any section we know, Some if there is a better candidate.<br/>
+    If it is Some, we will relay the RPC to them instead of accepting the rellocation to our own section.
+    </p>
+  </div>
+  <button class="collapsible">RelocatingState(Hop)</button>
```
pierrechevalier83: I don't think this name is consistent with other ones. Maybe the state could be: `RelocatingToShorterSection`

routing_model/index.html
```diff
@@ -1137,7 +1282,13 @@ <h2>Successfully relocate a node from source to destination section</h2>
     participant Node as Relocating Node
     participant Dst as Destination Section
 
-    Src->>+Dst: Routing RPC: ExpectCandidate
+
+    loop FindDestination
+      Src->>+Dst: Routing RPC: ExpectCandidate
+      opt Refuse
```
pierrechevalier83: The title of this section should loose the "successfully"

routing_model/index.html
```diff
@@ -1014,8 +1092,69 @@ <h2>Handling merges</h2>
...
+      MarkNewElderAdults["update_elder_status(new_section_info)"]
+      MarkNewElderAdults--> EndRoutine
+
+      WaitComplete--"!WAITED_VOTES<br/>.is_empty()<br/>(Wait not complete)"--> LoopEnd
+      LoopEnd --> LoopStart
+  </div>
+
   <h2>Handling members of our section going Online or Offline</h2>
   <div class=description>
+  <button class="collapsible">RelocatingState(BackOnline)</button>
```
pierrechevalier83: Again, I had rather not start nesting states.

routing_model/index.html
```diff
@@ -653,7 +653,8 @@ <h2>Relocating a member of our section away from it</h2>
   </p>
   <button class="collapsible">get_best_relocating_node_and_target</button>
   <div class="content">
-    <p>Returns the best node to relocate and the target address to send it to<br/>
+    <p>Takes ALREADY_RELOCATING for the nodes it will ignore.<br/>
```
pierrechevalier83: As it is stated, I think it's a bit unclear. If I understand the general idea right, if ALREADY_RELOCATING is not empty, this function will always return None to reduce the number of occasions where more than one nodes are relocating away from our section.

This is not how that sentence reads for me.

routing_model/index.html
jeanphilippeD: Yes, working on this clean up of names PR.

routing_model/index.html
jeanphilippeD: This is a state that contain some info, like WaitingCandidateInfo.
I do not want to split this into 3 states as it makes maintenance that much harder.

routing_model/index.html
jeanphilippeD: Why, if the node is rejected, the flow is different?

routing_model/index.html
jeanphilippeD: No, if ALREADY_RELOCATING is not empty, that function will not return a node that is in ALREADY_RELOCATING, it will return the next node to relocate that is not in ALREADY_RELOCATING.

routing_model/index.html
```diff
@@ -1014,8 +1092,69 @@ <h2>Handling merges</h2>
...
+ </p>
+  <button class="collapsible">WAITED_VOTES</button>
+  <div class="content">
+    <p>Wait for all these votes to reach consensus before reflecting the status change in our chain.<br>
+      Both section need to be consensused before we move on so we do not leave one behind with not enough voters.
+    </p>
+  </div>
+  <button class="collapsible">complete_split</button>
+  <div class="content">
+    <p>With the NewSectionInfo in hands, completing the split process consists on joining the correct newly formed section and leaving the old one behind.
```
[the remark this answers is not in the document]
jeanphilippeD: As discussed, leaving for now as mirror merge_complete

routing_model/index.html
pierrechevalier83: Because the diagram now deals with the successful and the unsuccessful situations. We can deal with it in a later PR, though

PR #1595, opened 2019-04-10
Fleming flow tests: add CheckOnlineOffline

octol:
Implementing tests for the CheckOnlineOffline flow in
https://raw.githack.com/maidsafe/routing/bd2d7d32cb1fcffc57e652df3fa5ff0f92f430fe/routing_model/index.html

### DIFF
--- a/routing_model/src/actions.rs
+++ b/routing_model/src/actions.rs
@@ -106,6 +106,13 @@ impl InnerAction {
         self.our_nodes.push(NodeChange::Online(node.node));
     }
 
+    fn set_offline_state(&mut self, name: &Name) {
+        let node = &mut self.our_current_nodes.get_mut(name).unwrap();
+        node.is_offline = true;
+        // Note: for test validation only
+        self.our_nodes.push(NodeChange::Offline(node.node));
+    }
+
     fn set_elder_state(&mut self, name: &Name, value: bool) {
         let node = &mut self.our_current_nodes.get_mut(name).unwrap();
 
@@ -166,6 +173,14 @@ impl Action {
             .set_online_state(&Name(candidate.0.name));
     }
 
+    pub fn set_node_offline_state(&self, node: Node) {
+        self.0.borrow_mut().set_offline_state(&Name(node.0.name));
+    }
+
+    pub fn set_node_back_online_state(&self, node: Node) {
+        self.0.borrow_mut().set_relocating_state(&Name(node.0.name));
+    }
+
     pub fn remove_node(&self, candidate: Candidate) {
         self.0.borrow_mut().remove_node(Node(candidate.0));
     }
@@ -183,10 +198,12 @@ impl Action {
                 .values()
                 .cloned()
                 .sorted_by(|left, right| {
-                    left.is_relocating
-                        .cmp(&right.is_relocating)
-                        .then(left.node.0.age.cmp(&right.node.0.age).reverse())
-                        .then(left.node.0.name.cmp(&right.node.0.name))
+                    left.is_offline.cmp(&right.is_offline).then(
+                        left.is_relocating
+                            .cmp(&right.is_relocating)
+                            .then(left.node.0.age.cmp(&right.node.0.age).reverse())
+                            .then(left.node.0.name.cmp(&right.node.0.name)),
+                    )
                 })
                 .collect_vec();
             let elder_size = std::cmp::min(3, sorted_values.len());
--- a/routing_model/src/flows_dst.rs
+++ b/routing_model/src/flows_dst.rs
@@ -11,7 +11,7 @@
 
 use crate::state::{AcceptAsCandidateState, CheckAndProcessElderChangeState, State};
 use crate::utilities::{
-    Candidate, ChangeElder, Event, LocalEvent, ParsecVote, Proof, Rpc, Section,
+    Candidate, ChangeElder, Event, LocalEvent, Node, ParsecVote, Proof, Rpc, Section,
 };
 
 #[derive(Debug, PartialEq, Default, Clone)]
@@ -343,6 +343,60 @@ impl CheckAndProcessElderChange {
 
     fn start_check_elder_timeout(&self) -> Self {
         self.0.action.schedule_event(LocalEvent::TimeoutCheckElder);
+        self.clone()
+    }
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct CheckOnlineOffline(pub State);
+
+impl CheckOnlineOffline {
+    pub fn try_next(&self, event: Event) -> Option<State> {
+        match event {
+            Event::ParsecConsensus(vote) => self.try_consensus(&vote),
+            Event::LocalEvent(local_event) => self.try_local_event(local_event),
+            // Delegate to other event loops
+            _ => None,
+        }
+        .map(|state: CheckOnlineOffline| state.0)
+    }
+
+    fn try_consensus(&self, vote: &ParsecVote) -> Option<Self> {
+        match vote {
+            ParsecVote::Offline(node) => Some(self.make_node_offline(*node)),
+            ParsecVote::BackOnline(node) => Some(self.make_node_back_online(*node)),
+            // Delegate to other event loops
+            _ => None,
+        }
+    }
+
+    fn try_local_event(&self, local_event: LocalEvent) -> Option<Self> {
+        match local_event {
+            LocalEvent::NodeDetectedOffline(node) => Some(self.vote_parsec_offline(node)),
+            LocalEvent::NodeDetectedBackOnline(node) => Some(self.vote_parsec_back_online(node)),
+            // Delegate to other event loops
+            _ => None,
+        }
+    }
+
+    fn vote_parsec_offline(&self, node: Node) -> Self {
+        self.0.action.vote_parsec(ParsecVote::Offline(node));
+        self.clone()
+    }
+
+    fn vote_parsec_back_online(&self, node: Node) -> Self {
+        self.0.action.vote_parsec(ParsecVote::BackOnline(node));
+        self.clone()
+    }
+
+    fn make_node_offline(&self, node: Node) -> Self {
+        self.0.action.set_node_offline_state(node);
+        self.clone()
+    }
+
+    /// A member of a section that was lost connection to became offline, but is now online again
+    fn make_node_back_online(&self, node: Node) -> Self {
+        self.0.action.set_node_back_online_state(node);
         self.clone()
     }
 }
--- a/routing_model/src/scenario_tests.rs
+++ b/routing_model/src/scenario_tests.rs
@@ -825,6 +825,76 @@ mod dst_tests {
             &AssertState::default(),
         );
     }
+
+    #[test]
+    fn test_local_events_offline_online_again_for_different_nodes() {
+        run_test(
+            "Get local event node detected offline online again different nodes",
+            &intial_state_old_elders(),
+            &[
+                LocalEvent::NodeDetectedOffline(NODE_ELDER_130).to_event(),
+                LocalEvent::NodeDetectedBackOnline(NODE_ELDER_131).to_event(),
+            ],
+            &AssertState {
+                action_our_votes: vec![
+                    ParsecVote::Offline(NODE_ELDER_130),
+                    ParsecVote::BackOnline(NODE_ELDER_131),
+                ],
+                ..AssertState::default()
+            },
+        );
+    }
+
+    #[test]
+    fn test_parsec_offline() {
+        run_test(
+            "Get parsec consensus offline",
+            &intial_state_old_elders(),
+            &[ParsecVote::Offline(NODE_ELDER_130).to_event()],
+            &AssertState {
+                action_our_nodes: vec![NodeChange::Offline(NODE_ELDER_130)],
+                ..AssertState::default()
+            },
+        );
+    }
+
+    #[test]
+    fn test_parsec_offline_then_check_elder() {
+        let initial_state = arrange_initial_state(
+            &intial_state_old_elders(),
+            &[ParsecVote::Offline(NODE_ELDER_130).to_event()],
+        );
+        run_test(
+            "Get parsec consensus offline then check elder",
+            &initial_state,
+            &[ParsecVote::CheckElder.to_event()],
+            &AssertState {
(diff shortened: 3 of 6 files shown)


routing_model/src/scenario_tests.rs
```diff
@@ -17,6 +17,7 @@ use crate::utilities::{
     NodeState, ParsecVote, Proof, ProofRequest, ProofSource, Rpc, Section, SectionInfo,
 };
 
+#[cfg(test)]
```
jeanphilippeD: should it be in libs.rs since all this file should only compile for tests?

octol: Yes that makes sense :)

routing_model/src/actions.rs
```diff
@@ -106,6 +106,11 @@ impl InnerAction {
         self.our_nodes.push(NodeChange::Online(node.node));
     }
 
+    fn set_offline_state(&mut self, name: &Name) {
+        let node = &mut self.our_current_nodes.get_mut(name).unwrap();
+        self.our_nodes.push(NodeChange::Offline(node.node));
```
jeanphilippeD: See `node.is_resource_proofing = false;` in set_online_state.

This is what is actually changing the state, we probably want something like that for offline.
(A new bool may be the simplest for now, but we will likely need to combine some of these into an enum as they are mutually exclusive).

jeanphilippeD: self.our_nodes.push(NodeChange::Offline(node.node)); is used only for validating what happened in our tests.

routing_model/src/flows_dst.rs
```diff
@@ -346,3 +346,58 @@ impl CheckAndProcessElderChange {
         self.clone()
     }
 }
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct CheckOnlineOffline(pub State);
```
jeanphilippeD: make sense with CheckAndProcessElderChange. I guess they could be moved in their own file in the future (not particularly a dst thing

routing_model/src/flows_dst.rs
```diff
@@ -346,3 +346,58 @@ impl CheckAndProcessElderChange {
...
+
+impl CheckOnlineOffline {
```
jeanphilippeD: Maybe indicate the commit the reference diagram was at in your PR
I assume: https://raw.githack.com/maidsafe/routing/bd2d7d32cb1fcffc57e652df3fa5ff0f92f430fe/routing_model/index.html

routing_model/src/flows_dst.rs
```diff
@@ -346,3 +346,58 @@ impl CheckAndProcessElderChange {
...
+            // Delegate to other event loops
+            _ => None,
+        }
+        .map(|state: CheckOnlineOffline| state.0)
+    }
+
+    fn try_consensus(&self, vote: &ParsecVote) -> Option<Self> {
+        match vote {
+            ParsecVote::Offline(node) => Some(self.make_node_offline(*node)),
+            ParsecVote::BackOnline(..) => None, // TODO
```
jeanphilippeD: Maybe worth completing in this PR.

routing_model/src/flows_dst.rs
```diff
@@ -346,3 +346,58 @@ impl CheckAndProcessElderChange {
...
+    }
+
+    fn vote_parsec_back_online(&self, node: Node) -> Self {
+        self.0.action.vote_parsec(ParsecVote::BackOnline(node));
+        self.clone()
+    }
+
+    fn make_node_offline(&self, node: Node) -> Self {
+        self.0.action.set_node_offline_state(node);
+        // TODO: exit_event_loop?
```
jeanphilippeD: nope, event loop never exit.

routing_model/src/scenario_tests.rs
```diff
@@ -825,6 +825,32 @@ mod dst_tests {
...
+        run_test(
+            "Get local event node detected online again",
+            &intial_state_old_elders(),
+            &[LocalEvent::NodeDetectedBackOnline(NODE_ELDER_130).to_event()],
+            &AssertState {
+                action_our_votes: vec![ParsecVote::BackOnline(NODE_ELDER_130)],
+                ..AssertState::default()
+            },
+        );
+    }
```
jeanphilippeD: Missing a test for the parsec consensus for offline.
Probably also want one for consensus offline, and then checkElder happens (Maybe with the check elder tests).

routing_model/src/state.rs
```diff
@@ -17,6 +17,12 @@ use crate::flows_node::*;
 use crate::flows_src::*;
 use crate::utilities::*;
 
+#[derive(Debug, PartialEq, Default, Clone)]
+pub struct CheckOnlineOfflineState {
```
jeanphilippeD: This routine need no local state, so it seem you can not have it at all.

routing_model/src/state.rs
```diff
@@ -96,6 +103,10 @@ impl State {
             return Some(next);
         }
 
+        if let Some(next) = self.as_check_online_offline().try_next(event) {
```
jeanphilippeD: Because it is Wait 5, it should be by the check elder.

routing_model/src/utilities.rs
```diff
@@ -250,6 +255,9 @@ impl ParsecVote {
             | ParsecVote::NewSectionInfo(_)
             | ParsecVote::RelocationTrigger
             | ParsecVote::CheckElder => None,
+
+            ParsecVote::Offline(_)
```
jeanphilippeD: group with other?

routing_model/src/actions.rs
```diff
@@ -177,6 +179,12 @@ impl Action {
             .set_offline_state(&Name(node.0.name));
     }
 
+    pub fn set_node_back_online_state(&self, node: Node) {
+        self.0
+            .borrow_mut()
+            .set_online_state(&Name(node.0.name));
```
jeanphilippeD: I guess this should also call the set_relocating to match the diagram I linked.

routing_model/src/flows_dst.rs
```diff
@@ -347,6 +347,8 @@ impl CheckAndProcessElderChange {
     }
 }
 
+/// Implementing the flow diagram according to the state at
+/// https://raw.githack.com/maidsafe/routing/bd2d7d32cb1fcffc57e652df3fa5ff0f92f430fe/routing_model/index.html
```
jeanphilippeD: I think this is probably better at the PR level rather than in the code.
Just my preference though, but also it should apply to the full flow rather than the one bit.

routing_model/src/scenario_tests.rs
```diff
@@ -178,6 +179,7 @@ fn run_test(test_name: &str, start_state: &State, events: &[Event], expected_sta
         },
         final_state.failure,
     );
+    dbg!(&final_state);
```
jeanphilippeD: I guess you want to remove that :-)

routing_model/src/actions.rs
octol: Ah yes good catch!

routing_model/src/scenario_tests.rs
```diff
@@ -39,6 +39,7 @@ const ADD_PROOFING_NODE_1: NodeChange =
     NodeChange::AddResourceProofing(Node(Attributes { name: 1, age: 10 }));
 const SET_ONLINE_NODE_1: NodeChange = NodeChange::Online(Node(Attributes { name: 1, age: 10 }));
 const REMOVE_NODE_1: NodeChange = NodeChange::Remove(Node(Attributes { name: 1, age: 10 }));
+const SET_OFFLINE_NODE_130: NodeChange = NodeChange::Offline(Node(Attributes { name: 130, age: 30 }));
```
jeanphilippeD: This is more because it is used a lot, and all these are about NODE_1.
You're probably better off using `NodeChange::Offline(NODE_ELDER_130.clone())` directly in place.

jeanphilippeD: clone is probably not necessary in my suggestion.

octol: Yep it wasn't necessary

routing_model/src/scenario_tests.rs
```diff
@@ -825,6 +825,110 @@ mod dst_tests {
...
+            &[LocalEvent::NodeDetectedOffline(NODE_ELDER_130).to_event()],
+            &AssertState {
+                action_our_votes: vec![ParsecVote::Offline(NODE_ELDER_130)],
+                ..AssertState::default()
+            },
+        );
+    }
+
+    #[test]
+    fn test_local_event_online_again() {
```
octol: Is this test non-sensical, or potentially exposing missing checks in the flows? Getting a BackOnline local event shouldn't happen here right?

jeanphilippeD: It really depend whether we trust our event provider. If it was an RPC, I'd say we need to handle it gracefully. This one is a bit controversial, regardless, I think this test make sense, as we would not want to filter it at this stage, if we wanted to do something, that would be when it comes out of Parsec.

The reason is that when this event occur we are not in shared state, so we could not trust that all nodes take the same action.

routing_model/src/scenario_tests.rs
```diff
@@ -825,6 +825,110 @@ mod dst_tests {
...
+            ],
+        );
+        run_test(
+            "Get parsec consensus offline then check elder",
+            &initial_state,
+            &[ParsecVote::CheckElder.to_event()],
+            &AssertState {
+                action_our_events: vec![LocalEvent::TimeoutCheckElder],
+                ..AssertState::default()
+            },
```
jeanphilippeD: That test seem to be failing, something in CheckElder does not take into account the offline state.
I think this is part of this PR's job.

routing_model/src/scenario_tests.rs
```diff
@@ -825,6 +825,110 @@ mod dst_tests {
...
+    }
+
+    #[test]
+    fn test_parsec_offline_and_local_event_back_online_then_parsec_oneline() {
+        let initial_state = arrange_initial_state(
+            &intial_state_old_elders(),
+            &[
+                LocalEvent::NodeDetectedOffline(NODE_ELDER_130).to_event(),
+                ParsecVote::Offline(NODE_ELDER_130).to_event(),
+                LocalEvent::NodeDetectedBackOnline(NODE_ELDER_130).to_event(),
```
jeanphilippeD: The local events are not needed here.
Both because they only enqueue votes, but also because that specific node may not have seen that event at all, but it will still process it correctly. (Same for the other tests).

routing_model/src/scenario_tests.rs
```diff
@@ -825,6 +825,110 @@ mod dst_tests {
             &AssertState::default(),
         );
     }
+
+    #[test]
+    fn test_local_event_offline() {
```
jeanphilippeD: I would probably have a single test processing both NodeDetectedOffline and NodeDetectedOnline on different nodes, combining together test_local_event_offline & test_local_event_online_again since they test very similar things.

routing_model/src/scenario_tests.rs
jeanphilippeD: Can make the test name shorter as a result.

routing_model/src/scenario_tests.rs
octol: Interesting, ok I'll have a look!

routing_model/src/scenario_tests.rs
octol: Good point!

routing_model/src/scenario_tests.rs
jeanphilippeD: To clarify, what I meant is I expect the assert state to be different and have votes for new elders.

jeanphilippeD: It may be this test would be better grouped with the other check elder tests. Maybe.

routing_model/src/actions.rs
```diff
@@ -197,6 +197,7 @@ impl Action {
             let mut sorted_values = our_current_nodes
                 .values()
                 .cloned()
+                .filter(|node_state| !node_state.is_offline)
```
jeanphilippeD: I think we probably more want to adjust the sorting the same way it is done for relocating a few lines bellow.

routing_model/src/scenario_tests.rs
```diff
@@ -893,10 +890,7 @@ mod dst_tests {
     fn test_parsec_offline_then_local_event_back_online() {
```
jeanphilippeD: I am not sure this test is necessary, is there any difference with what is tested by test_local_events_offline_online_again_for_different_nodes?
The one bellow that seem useful though.

routing_model/src/scenario_tests.rs
```diff
@@ -913,11 +907,7 @@ mod dst_tests {
     fn test_parsec_offline_and_local_event_back_online_then_parsec_oneline() {
```
jeanphilippeD: rename?

routing_model/src/scenario_tests.rs
```diff
@@ -825,6 +825,100 @@ mod dst_tests {
...
+    }
+
+    #[test]
+    fn test_local_event_offline_and_parsec_offline_then_check_elder() {
+        let initial_state = arrange_initial_state(
+            &intial_state_old_elders(),
+            &[ParsecVote::Offline(NODE_ELDER_130).to_event()],
+        );
+        let new_votes = vec![
+            ParsecVote::AddElderNode(YOUNG_ADULT_205),
```
jeanphilippeD: I think this will be different once you fixed the filter, because there should be a remove elder here

routing_model/src/scenario_tests.rs
octol: In the second case you're detecting back online a node that was consensus-ed away, maybe that doesn't matter? To be on the safe side I think we should keep both

jeanphilippeD: There are many thing that could be tested, so we have to make choices. I would not think keeping an extra test is necessarily the safe option since it adds maintainability cost. (The main good thing about TDD is that it guides you to only add test you need. (I'm not really a practitioner). In this case what code would you have to write to make this test pass that did not exist for the previous one?)

The local event should always be voted in Parsec, so trying that sepcific circumstance does not seem to add much, but I'll let you decide :-)

octol: Fair point, ok I'll remove!

PR #1599, opened 2019-04-11
Switch from mocking to faking work unit to decide relocation

jeanphilippeD:
This will support migration to the new flow diagram that include
work unit increment.

### DIFF
--- a/routing_model/src/actions.rs
+++ b/routing_model/src/actions.rs
@@ -29,7 +29,6 @@ pub struct InnerAction {
     pub our_nodes: Vec<NodeChange>,
 
     pub shortest_prefix: Option<Section>,
-    pub node_to_relocate: Option<Node>,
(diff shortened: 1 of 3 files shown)


routing_model/src/actions.rs
```diff
@@ -71,6 +69,14 @@ impl InnerAction {
         self.extend_current_nodes(&node_states)
     }
 
+    pub fn with_enough_work_to_relocate(mut self, nodes: &[Node]) -> Self {
+        self.our_current_nodes
+            .values_mut()
+            .filter(|state| nodes.contains(&state.node))
+            .for_each(|state| state.work_units_done = state.node.0.age);
```
octol: We're doing work == age for now?

jeanphilippeD: No, but for a fake, it is a good enough approximation..

PR #1600, opened 2019-04-11
Fleming design model: flow tests for merges

octol:
Add merges to CheckAndProcessElderChange flow tests

Render of graphs used: https://raw.githack.com/maidsafe/routing/bd2d7d32cb1fcffc57e652df3fa5ff0f92f430fe/routing_model/index.html

The sub-process ProcessMerge is missing and would be the next logical step

### DIFF
--- a/routing_model/src/flows_dst.rs
+++ b/routing_model/src/flows_dst.rs
@@ -271,6 +271,7 @@ impl CheckAndProcessElderChange {
     pub fn try_next(&self, event: Event) -> Option<MemberState> {
         match event {
             Event::ParsecConsensus(vote) => self.try_consensus(&vote),
+            Event::Rpc(rpc) => self.try_rpc(rpc),
             Event::LocalEvent(LocalEvent::TimeoutCheckElder) => {
                 Some(self.vote_parsec_check_elder())
             }
@@ -282,6 +283,13 @@ impl CheckAndProcessElderChange {
     fn try_consensus(&self, vote: &ParsecVote) -> Option<Self> {
         match vote {
             ParsecVote::CheckElder => Some(self.check_elder()),
+            _ => None,
+        }
+    }
+
+    fn try_rpc(&self, rpc: Rpc) -> Option<Self> {
+        match rpc {
+            Rpc::Merge => Some(self.vote_parsec_neighbour_merge()),
             _ => None,
         }
     }
@@ -307,6 +315,11 @@ impl CheckAndProcessElderChange {
 
     fn vote_parsec_check_elder(&self) -> Self {
         self.0.action.vote_parsec(ParsecVote::CheckElder);
+        self.clone()
+    }
+
+    fn vote_parsec_neighbour_merge(&self) -> Self {
+        self.0.action.vote_parsec(ParsecVote::NeighbourMerge);
         self.clone()
     }
 
--- a/routing_model/src/flows_dst.rs
+++ b/routing_model/src/flows_dst.rs
@@ -8,7 +8,9 @@
 
 use crate::{
     state::{AcceptAsCandidateState, MemberState, ProcessElderChangeState},
-    utilities::{Candidate, ChangeElder, Event, LocalEvent, Node, ParsecVote, Proof, Rpc, Section},
+    utilities::{
+        Candidate, ChangeElder, Event, LocalEvent, MergeInfo, Node, ParsecVote, Proof, Rpc, Section,
+    },
 };
 use unwrap::unwrap;
 
@@ -282,7 +284,8 @@ impl CheckAndProcessElderChange {
 
     fn try_consensus(&self, vote: &ParsecVote) -> Option<Self> {
         match vote {
-            ParsecVote::CheckElder => Some(self.check_elder()),
+            ParsecVote::NeighbourMerge(merge_info) => Some(self.store_merge_infos(*merge_info)),
+            ParsecVote::CheckElder => Some(self.check_merge()),
             _ => None,
         }
     }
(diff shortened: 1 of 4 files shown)


routing_model/src/flows_dst.rs
```diff
@@ -278,7 +280,15 @@ impl CheckAndProcessElderChange {
 
     fn try_consensus(&self, vote: &ParsecVote) -> Option<Self> {
         if ParsecVote::CheckElder == *vote {
```
jeanphilippeD: I think you'd be better off extracting the ProcessElderChange sub routine first.
You can see an example of sub routine in the source or destination flows, but feel free to implement the sub-routine differently if it make sense (these other sub routine will be removed to match the new graph, but the elder/merge/split will remain).

jeanphilippeD: This would allow you to refactor and all the test should still pass.

routing_model/src/flows_dst.rs
```diff
@@ -278,7 +280,15 @@ impl CheckAndProcessElderChange {
...
-            return Some(self.check_elder());
+            // TODO: something seems strange here
+            if self.has_merge_infos() {
+                // TODO delegate to ProcessMerge flow
+                return None;
+            } else {
+                return Some(self.check_elder());
+            }
+        } else if let ParsecVote::NeighbourMerge(info) = *vote {
```
jeanphilippeD: you'll want to use a match here.

routing_model/src/utilities.rs
```diff
@@ -74,6 +74,9 @@ pub struct SectionInfo(pub Section, pub i32 /*contain full membership */);
 #[derive(Debug, Clone, Copy, Default, PartialEq, PartialOrd, Ord, Eq)]
 pub struct GenesisPfxInfo(pub SectionInfo);
 
+#[derive(Debug, Clone, Copy, Default, PartialEq, PartialOrd, Ord, Eq)]
+pub struct NeighbourMergeInfo;
```
jeanphilippeD: Maybe call it MergeInfo, because it comes from the Merge RPC

routing_model/src/flows_dst.rs
```diff
@@ -343,6 +360,23 @@ impl CheckAndProcessElderChange {
...
+    fn vote_parsec_neighbour_merge(&self) -> Self {
+        self.0
+            .action
+            .vote_parsec(ParsecVote::NeighbourMerge(NeighbourMergeInfo));
+        self.clone()
+    }
+
+    fn store_merge_infos(&self, info: NeighbourMergeInfo) -> Self {
+        let mut state = self.clone();
+        state.mut_routine_state().neighbour_merge_info = Some(info);
```
jeanphilippeD: No, this is a function in the graph.
What this mean is that it needs to be implemented by the Action component.
Only local variable that appear in the graph as SCREAMING_CAMEL_CASE should be stored locally, or any local state you find yourself needing.

routing_model/src/scenario_tests.rs
```diff
@@ -175,6 +176,7 @@ fn run_test(test_name: &str, start_state: &State, events: &[Event], expected_sta
         },
         final_state.failure,
     );
+    dbg!(&final_state);
```
jeanphilippeD: Why did you need this? did the assert_eq not provide you the needed info?
If so, we should consider making sure it does.

routing_model/src/scenario_tests.rs
```diff
@@ -564,6 +566,65 @@ mod dst_tests {
         );
     }
 
+    #[test]
+    fn test_parsec_expect_candidate_and_online_then_neighbour_merge() {
+        let initial_state = arrange_initial_state(
+            &initial_state_old_elders(),
+            &[
+                ParsecVote::ExpectCandidate(CANDIDATE_1).to_event(),
+                ParsecVote::Online(CANDIDATE_1).to_event(),
```
jeanphilippeD: Why Online? It should only be used for resource proofing, I don't see it would do anything.

routing_model/src/scenario_tests.rs
```diff
@@ -564,6 +566,65 @@ mod dst_tests {
...
+    // JON
+    #[test]
+    fn test_parsec_neighbour_merge() {
+        run_test(
+            "Get Parsec NeighbourMergeInfo",
+            &initial_state_old_elders(),
+            &[ParsecVote::NeighbourMerge(NeighbourMergeInfo).to_event()],
+            &AssertState {
+                check_and_process_elder_change_routine: CheckAndProcessElderChangeState {
+                    neighbour_merge_info: Some(NeighbourMergeInfo),
```
jeanphilippeD: Because it is a store thing, for testing, you could also enqueu it in the our_events vector in InnerAction.
You'll also actually want to store it so you can get it back out when needed.

routing_model/src/state.rs
```diff
@@ -18,6 +18,7 @@ use crate::utilities::*;
 pub struct CheckAndProcessElderChangeState {
     pub wait_votes: Vec<ParsecVote>,
     pub change_elder: Option<ChangeElder>,
+    pub neighbour_merge_info: Option<NeighbourMergeInfo>,
```
jeanphilippeD: As mentioned, should not be there.
You see wait_votes as a couter part WAIT_VOTES.
and change_elder is helpful to implement the update to elder, but is not something that appear in the diagram at all.

routing_model/src/scenario_tests.rs
octol: Well it's the same info, but dbg gives it with linebreaks and indentation. I struggle to parse it when it's just one long long line :(

routing_model/src/scenario_tests.rs
octol: Yes I think this was actually just a placeholder

routing_model/src/utilities.rs
octol: Done

routing_model/src/flows_dst.rs
octol: Ah yes good point!

routing_model/src/scenario_tests.rs
jeanphilippeD: https://crates.io/crates/pretty-assertions ?

routing_model/src/flows_dst.rs
```diff
@@ -281,11 +284,41 @@ impl CheckAndProcessElderChange {
...
         }
     }
 
+    fn store_merge_infos(&self, merge_info: MergeInfo) -> Self {
+        self.0.action.store_merge_infos(merge_info);
+        self.clone()
+    }
+
+    fn merge_needed(&self) -> bool {
+        // TODO: trigger for merge needed still in flux
```
jeanphilippeD: For this PR, we would want to have this return true and false.

You can choose a criteria that is easy to implement and somewhat sensible, it does not have to relate to what the real merge needed require: I would say for now, just mock it (i.e have a field in InnerAction: merge_needed:bool

pierrechevalier83:
We have moved the routing model to its own repo instead of a branch in routing.
This PR was moved to https://github.com/maidsafe/routing_model/pull/1

PR #1601, opened 2019-04-12
Fleming design model: extract ProcessElderChange in flow tests

octol:
Extract out ProcessElderChange

### DIFF
--- a/routing_model/src/flows_dst.rs
+++ b/routing_model/src/flows_dst.rs
@@ -418,7 +418,7 @@ impl ProcessElderChange {
     fn mark_elder_change(&self) -> Self {
         let mut state = self.clone();
 
-        let change_elder = state.mut_routine_state().change_elder.clone(); // TODO: take it?
+        let change_elder = state.mut_routine_state().change_elder.clone();
         state.0.action.mark_elder_change(change_elder);
 
         state
--- a/routing_model/src/flows_dst.rs
+++ b/routing_model/src/flows_dst.rs
@@ -331,12 +331,12 @@ impl ProcessElderChange {
         self.0
             .with_check_and_process_elder_change_sub_routine_process_elder_change(Some(
                 ProcessElderChangeState {
-                    change_elder,
+                    change_elder: change_elder.clone(),
                     wait_votes: Vec::new(),
                 },
             ))
             .as_process_elder_change()
-            .vote_for_elder_change()
+            .vote_for_elder_change(change_elder)
     }
 
     fn exit_event_loop(&self) -> Self {
@@ -372,10 +372,8 @@ impl ProcessElderChange {
         }
     }
 
-    // WIP: in progress
-    fn vote_for_elder_change(&self) -> Self {
+    fn vote_for_elder_change(&self, change_elder: ChangeElder) -> Self {
         let mut state = self.clone();
-        let change_elder = self.change_elder();
 
         let votes = state.0.action.get_elder_change_votes(&change_elder);
         state.mut_routine_state().change_elder = change_elder;
@@ -411,17 +409,10 @@ impl ProcessElderChange {
         }
     }
 
-    fn change_elder(&self) -> ChangeElder {
-        self.routine_state().change_elder.clone()
-    }
-
     fn mark_elder_change(&self) -> Self {
-        let mut state = self.clone();
-
-        let change_elder = state.mut_routine_state().change_elder.clone();
-        state.0.action.mark_elder_change(change_elder);
-
-        state
+        let change_elder = self.routine_state().change_elder.clone();
+        self.0.action.mark_elder_change(change_elder);
+        self.clone()
     }
 
     fn start_check_elder_timeout(&self) -> Self {
--- a/routing_model/src/flows_dst.rs
+++ b/routing_model/src/flows_dst.rs
@@ -6,7 +6,9 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::state::{AcceptAsCandidateState, CheckAndProcessElderChangeState, MemberState};
+use crate::state::{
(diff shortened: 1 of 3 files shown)


routing_model/src/state.rs
```diff
@@ -165,6 +175,19 @@ impl State {
             ..self.clone()
         }
     }
+
+    pub fn with_check_and_process_elder_change_sub_routine_process_elder_change(
```
jeanphilippeD: yes, that is the kind of boiler plate I was talking about.

routing_model/src/state.rs
```diff
@@ -14,10 +14,16 @@ use crate::flows_node::*;
...
+#[derive(Debug, PartialEq, Clone)]
+pub struct ProcessElderChangeState {
+    pub change_elder: ChangeElder,
+}
+
 #[derive(Debug, PartialEq, Default, Clone)]
 pub struct CheckAndProcessElderChangeState {
     pub wait_votes: Vec<ParsecVote>,
     pub change_elder: Option<ChangeElder>,
+    pub sub_routine_process_elder_change: Option<ProcessElderChangeState>,
```
jeanphilippeD: The alternative to replace the boiler plate sub routine would be:

```
#[derive(Debug, PartialEq, Clone)]
pub struct ProcessElderChangeState {
    is_active: bool,
    pub change_elder: ChangeElder,
}

#[derive(Debug, PartialEq, Default, Clone)]
pub struct CheckAndProcessElderChangeState {
    pub wait_votes: Vec<ParsecVote>,
    pub sub_routine_process_elder_change: ProcessElderChangeState,
}
```

```
impl ProcessElderChange {
    pub fn start_event_loop(&self, change_elder: ChangeElder) -> Self {
        let state = self.clone();
        state.mut_routine_state().is_active = true;
        state.mut_routine_state().change_elder Some(change_elder);
        state.vote_for_elder_change()
    }

    fn routine_state(&self) -> &ProcessElderChangeState {
        &self.0.check_and_process_elder_change_routine.sub_routine_process_elder_change
    }

    fn mut_routine_state(&mut self) -> &mut ProcessElderChangeState {
        &mut self.0.check_and_process_elder_change_routine.sub_routine_process_elder_change
    }

```

octol: so wait_votes, I see you kept it in `CheckAndProcessElderChangeState` yet it only appears in `ProcessElderChange`. Any reason for this?

jeanphilippeD: got it wrong. fixed in the comment.

routing_model/src/flows_dst.rs
```diff
@@ -306,16 +295,74 @@ impl CheckAndProcessElderChange {
...
+        if !self.routine_state().wait_votes.contains(&vote) {
+            return None;
+        }
+
+        let mut state = self.clone();
+        let wait_votes = &mut state.mut_routine_state().wait_votes;
+        wait_votes.retain(|wait_vote| wait_vote != vote);
+
+        if wait_votes.is_empty() {
+            Some(state.mark_elder_change().start_check_elder_timeout())
```
jeanphilippeD: start_check_elder_timeout should not be there

routing_model/src/state.rs
```diff
@@ -73,6 +78,10 @@ impl State {
             return Some(next);
         }
 
+        if let Some(next) = self.as_process_elder_change().try_next(event) {
```
jeanphilippeD: This is not correct, should only run if active 
see
```
        if src.sub_routine_try_relocating.is_some() {
            if let Some(next) = self.as_try_relocating().try_next(event) {
                return Some(next);
            }
        }
```

octol: Good catch, this fixed most (except 2) test failures

routing_model/src/scenario_tests.rs
```diff
@@ -1005,6 +1020,16 @@ mod src_tests {
...
+                check_and_process_elder_change_routine: CheckAndProcessElderChangeState {
+                    sub_routine_process_elder_change: Some(ProcessElderChangeState {
+                        change_elder: ChangeElder {
+                            changes: vec![(YOUNG_ADULT_205, true), (NODE_ELDER_130, false)],
+                            new_section: SECTION_INFO_1,
+                        },
+                        wait_votes: Vec::new(),
+                    }),
+                    ..Default::default()
+                },
```
octol: @jeanphilippeD I added these blocks, they seem correct no? But I'm not 100% sure

jeanphilippeD: That does not seem right, if the state was not there before not clear why it is here now

octol: I think the difference is from 
old code:
`let change_elder = state.mut_routine_state().change_elder.take().unwrap();`
new code:
`let change_elder = state.mut_routine_state().change_elder.clone();`

jeanphilippeD: ```Some(state.mark_elder_change().start_check_elder_timeout())```
This is your problem, we do not exit the sub routine there, but we should have.
```wait_votes: Vec::new()``` means that we completed the tasks, so that sub routine should not exist anymore.

Comment (octol) [reply routing_model/src/scenario_tests.rs]: yep I'm adding the exit_event_loop logic now

File: routing_model/src/state.rs (hunk 2 above)
Comment (octol) [reply routing_model/src/state.rs]: But yeah, I think we should simplify to something like this at some point

File: routing_model/src/flows_dst.rs [hunk 6]
```diff
@@ -325,20 +388,42 @@ impl CheckAndProcessElderChange {
...
+
+    fn change_elder(&self) -> ChangeElder {
+        self.routine_state().change_elder.clone()
+    }
+
     fn mark_elder_change(&self) -> Self {
         let mut state = self.clone();
 
-        let change_elder = state.mut_routine_state().change_elder.take().unwrap();
+        let change_elder = state.mut_routine_state().change_elder.clone();
```
jeanphilippeD: What is it not using change_elder().
Probably better to not have the change_elder() function at all.

PR #1602, opened 2019-04-12
Use set_node_state like in more recent diagrams

jeanphilippeD:
Clean up NodeState to be closer to recent diagrams.
Do change relocating: There is still only one kind in this implementation.

### DIFF
--- a/routing_model/src/utilities.rs
+++ b/routing_model/src/utilities.rs
@@ -62,7 +62,7 @@ impl NodeChange {
 
     fn relocating(&self) -> bool {
         match &self {
-            NodeChange::State(_, State::RelocatingAgeIncrease) => true,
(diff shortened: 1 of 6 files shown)


routing_model/src/utilities.rs
```diff
@@ -18,52 +18,99 @@ pub struct Attributes {
...
+#[derive(Debug, Clone, PartialEq)]
 pub struct NodeState {
     pub node: Node,
     pub work_units_done: i32,
     pub is_elder: bool,
-    pub is_relocating: bool,
-    pub need_relocate: bool,
-    pub is_resource_proofing: bool,
-    // When a node that was previous online lost connection
-    pub is_offline: bool,
```
octol: Nice getting rid of all these bools :)

PR #1603, opened 2019-04-12
perf/mock_parsec: Optimize Network::send, don't deserialise.

afck:
This deserialisation took ~7% of CPU time in `merge_four_balanced_sections_into_one`.


afck:
Right, `bytes[..8]` panics if it doesn't have 8 bytes. Fixed now.

pierrechevalier83:
Superseded by PR 1606 which rebases it and fixes (well, ignores) a clippy warning.

PR #1604, opened 2019-04-12
Routing Model: Migrate source flow

jeanphilippeD:
Migrate the source flow to follow mostly the up to date diagram.
There are some separate concern not addressed like using different Relocating state for adults and back online nodes.

The multiple commit each address a part of the conversion with all test passing.
However, It can be reviewed as the complete change.

Also add the pretty_assertions crate to solve the assertion readability issues.


routing_model/src/scenario_tests.rs
```diff
@@ -906,13 +904,17 @@ mod dst_tests {
     }
 }
 
+//////////////////
+/// Src
+//////////////////
+
 mod src_tests {
```
octol: I was thinking later we could perhaps split these `mod *_tests` to their own files

jeanphilippeD: Could be, but I'm not 100% sure, each test is really testing the full system (i.e making sure other parts are not interfering, or make sure other parts are doing their job (Replace elders, relocate again...).

octol: Yes. Thought they test if from different perspectives. Actually my main motivation is for me it's much easier and faster when I can jump to the relevant file rather than scrolling back and forth. But then again maybe my workflow sucks ;)

octol: My thinking was that when we add these header style comments to separate code groups, it often suggests that there is a natural split at play

PR #1606, opened 2019-04-15
perf/mock_parsec: Optimize Network::send, don't deserialise.

pierrechevalier83:
This deserialisation took ~7% of CPU time in merge_four_balanced_sections_into_one.

### DIFF
--- a/src/mock_crust/support.rs
+++ b/src/mock_crust/support.rs
@@ -766,6 +766,7 @@ where
 }
 
 #[test]
+#[allow(clippy::let_unit_value)]
 fn test_is_parsec_req_resp() {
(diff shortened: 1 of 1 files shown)


pierrechevalier83:
I can't approve because it's now my own PR, but I've reviewed Andreas' commit and I'm fine with it as an optimization, @Viv-Rajkumar. (Of course, I find the second commit pristine! A diamond in a sea of zircons, if I may say so :rofl:)

Viv-Rajkumar:
> Of course, I find the second commit pristine! A diamond in a sea of zircons, if I may say so 🤣

so thats your pov for any clippy lints I take it? just suppress them. sounds good to me :)

PR #1608, opened 2019-04-15
 Flow tests: simplify ProcessElderChange subprocess

octol:
Implemented the suggested simplification from https://github.com/maidsafe/routing/pull/1601#discussion_r274847539

### DIFF
--- a/routing_model/src/flows_dst.rs
+++ b/routing_model/src/flows_dst.rs
@@ -328,20 +328,18 @@ pub struct ProcessElderChange(pub MemberState);
 
 impl ProcessElderChange {
     pub fn start_event_loop(&self, change_elder: ChangeElder) -> Self {
-        self.0
-            .with_check_and_process_elder_change_sub_routine_process_elder_change(Some(
-                ProcessElderChangeState {
-                    change_elder: change_elder.clone(),
-                    wait_votes: Vec::new(),
-                },
-            ))
-            .as_process_elder_change()
-            .vote_for_elder_change(change_elder)
+        let mut state = self.clone();
+        state.mut_routine_state().is_active = true;
+        state.mut_routine_state().change_elder = Some(change_elder.clone());
+        state.vote_for_elder_change(change_elder)
     }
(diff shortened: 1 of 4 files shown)


routing_model/src/flows_dst.rs
```diff
@@ -387,30 +385,21 @@ impl ProcessElderChange {
...
             .sub_routine_process_elder_change
-        {
-            Some(state) => state,
-            _ => panic!("Expect ProcessElderChange {:?}", &clone),
-        }
     }
 
     fn mark_elder_change(&self) -> Self {
-        let change_elder = self.routine_state().change_elder.clone();
+        let change_elder = self.routine_state().change_elder.clone().unwrap();
```
jeanphilippeD: take?

octol: You mean changing it to?
```
fn mark_elder_change(&self) -> Self {
    let mut state = self.clone();
    let change_elder = state.mut_routine_state().change_elder.take().unwrap();
    state.0.action.mark_elder_change(change_elder);
    state
}      
```

jeanphilippeD: Yes. If you think it make sense.
Your version also make sense to me.

jeanphilippeD: actually forget it. Yours seem simpler.

octol: I'll use your second version with `take` :)

octol: They are slightly different though? Even though it might not be visible from the rest of the code. Feels more correct to leave change_elder as empty

jeanphilippeD: since you clean it up on exit, it seems ok to leave it. But you decide.

octol: Let's give the second reviewer the casting vote ;)

Fraser999: I vote for `take()` :)

PR #1609, opened 2019-04-15
Minor updates

Fraser999:


routing_model/Cargo.toml
```diff
@@ -1,11 +1,15 @@
 [package]
+authors = ["MaidSafe Developers <dev@maidsafe.net>"]
+description = "A secured storage DHT"
```
jeanphilippeD: is that right?

routing_model/Cargo.toml
```diff
@@ -1,11 +1,15 @@
...
 version = "0.1.0"
-authors = ["Jean-Philippe DUFRAIGNE <jean.dufraigne@maidsafe.net>"]
 edition = "2018"
 
-[dependencies]
-lazy_static = "~1.0.1"
-unwrap = "~1.2.1"
-itertools = "~0.8.0"
+[dev-dependencies]
+lazy_static = "1.0.1"
```
jeanphilippeD: i thought we were supposed to have the `~`

routing_model/src/actions.rs
```diff
@@ -79,20 +81,25 @@ impl InnerAction {
     }
 
     pub fn with_section_members(mut self, section: SectionInfo, nodes: &[Node]) -> Self {
-        self.section_members.insert(section, nodes.to_vec());
+        assert!(self
```
jeanphilippeD: I'm not super happy to see that whole statement in an assert.
I'd rather have:
```
let inserted = self.section_members.insert(section, nodes.to_vec());
assert!(inserted.is_none());
```

Or if it is just to shut up the warning:
```let _ = ...```

I'm not sure this assert bring a sensible restriction to this function.

File: routing_model/src/actions.rs [hunk 4]
```diff
@@ -79,20 +81,25 @@ impl InnerAction {
...
+            .is_none());
         self
     }
 
     fn add_node(&mut self, node_state: NodeState) {
         self.our_nodes
             .push(NodeChange::AddWithState(node_state.node, node_state.state));
-        self.our_current_nodes
-            .insert(node_state.node.name(), node_state);
+        assert!(self
```
jeanphilippeD: same

routing_model/src/actions.rs
```diff
@@ -79,20 +81,25 @@ impl InnerAction {
...
+            .our_current_nodes
+            .insert(node_state.node.name(), node_state)
+            .is_none());
     }
 
     fn remove_node(&mut self, node: Node) {
         self.our_nodes.push(NodeChange::Remove(node));
-        self.our_current_nodes.remove(&Name(node.0.name));
+        unwrap!(self.our_current_nodes.remove(&Name(node.0.name)));
```
jeanphilippeD: same

routing_model/src/flows_dst.rs
```diff
@@ -263,10 +261,6 @@ pub struct CheckAndProcessElderChange(pub MemberState);
 
 // CheckAndProcessElderChange Sub Routine
 impl CheckAndProcessElderChange {
-    fn start_event_loop(&self) -> Self {
```
jeanphilippeD: it may not be used/tested, but it reflect the diagram, the proper fix would be to add a test for it.

routing_model/src/flows_src.rs
```diff
@@ -6,17 +6,16 @@
...
+    state::{MemberState, StartRelocateSrcState},
+    utilities::{Candidate, Event, LocalEvent, ParsecVote, RelocatedInfo, Rpc, SectionInfo},
+};
+use unwrap::unwrap;
 
 #[derive(Debug, PartialEq, Default, Clone)]
 pub struct TopLevelSrc(pub MemberState);
 
 impl TopLevelSrc {
-    fn start_event_loop(&self) -> Self {
```
jeanphilippeD: Same

routing_model/src/flows_src.rs
```diff
@@ -89,10 +88,6 @@ pub struct StartRelocateSrc(pub MemberState);
 
 // StartRelocateSrc Sub Routine
 impl StartRelocateSrc {
-    fn start_event_loop(&self) -> Self {
```
jeanphilippeD: Same

routing_model/src/flows_src.rs
```diff
@@ -147,10 +142,11 @@ impl StartRelocateSrc {
             .get_best_relocating_node_and_target(&state.routine_state().already_relocating)
         {
             state.0.action.send_rpc(Rpc::ExpectCandidate(candidate));
-            state
+            assert!(state
```
jeanphilippeD: I think I'd rather the assert be a separate statement.

routing_model/src/flows_src.rs
```diff
@@ -184,10 +180,10 @@ impl StartRelocateSrc {
 
     fn allow_resend(&self, candidate: Candidate) -> Self {
         let mut state = self.clone();
-        state
+        unwrap!(state
```
jeanphilippeD: why unwrap in some cases and assert in others?

routing_model/src/flows_src.rs
```diff
@@ -212,7 +208,6 @@ impl StartRelocateSrc {
     //
     // Routine state
     //
-
```
jeanphilippeD: Not sure about removing that space.
That block apply to all the functions bellow, not just the one?

routing_model/Cargo.toml
Fraser999: Nope - I wrote a description and have obviously obliterated it with the Routing one :D  I'll fix this.

routing_model/Cargo.toml
Fraser999: Yeah, I thought so too, but Routing seems to be doing its own thing; different from our other crates.

jeanphilippeD: I thought "we" said that routing needing fixing at some point.

routing_model/src/actions.rs
Fraser999: I'm still pretty new to the code in here, but it didn't seem reasonable to me to allow that insertion to replace an existing element.  If that _is_ expected, I'd prefer some comment as to the rationale (I can add such a comment if you think that is to be expected).

I think the clippy warning is useful, in that it's at least making us consider whether we really mean to silently drop existing entries.

I'll update to create a temporary `inserted` as suggested for now.

routing_model/src/flows_dst.rs
Fraser999: Ok, I'll revert this (and the other two instances), but I'll add a TODO comment to remove the allow lint they'll need once we implement tests for them.

routing_model/src/flows_src.rs
Fraser999: `unwrap!` is for values that have `Some` or `Ok`.  We have `unwrap_err!` for `Err`, but there's no `unwrap!` equivalent for `None`, so I used `assert!` for those.

routing_model/src/flows_src.rs
Fraser999: I'll add it back and the other two instances below.

routing_model/flow.md
```diff
@@ -11,7 +11,7 @@ Notes:
 Todo:
```
jeanphilippeD: This file should be removed, it is completly out of date, only the index.html is up to date.

routing_model/Cargo.toml
jeanphilippeD: did you meant to keep that?

Fraser999: I was waiting for an answer in Slack.

routing_model/src/flows_src.rs
jeanphilippeD: They do seem to have a semantic difference, unwrap suggest we use the value, assert suggest we are testing an invariant/ pre/post-condition.

routing_model/Cargo.toml
Fraser999: Yup - I'll put the tildes back in.  Routing will get a similar update sometime too.

pierrechevalier83:
> Looks great.
> Mostly some style things so feel free to disregard these if they don't make sense in our code base.
> There are 3 unused functions that I'd rather keep, either with a test, or an allow unused for now as they match the diagram.

Can you keep track of this commit JP instead? That way, we can re-introduce them when they're used and we don't have to live with dead code in the interim.

pierrechevalier83:
Change has been approved by both reviewers. Merging.

PR #1612, opened 2019-04-16
Prevent invalid peer disconnects, toml update and unack_msg update

Viv-Rajkumar:
- Do not disconnect from routing peers during a pfx change.
- Update Cargo.toml to use ~ for deps.
- Add a custom Debug impl for `NetworkEvent`
- Adds src_section to unack_msg to remove the previous PfxSection patch
- enable messages_during_churn test
- clear peer_mgr candidate post pfx change.


pierrechevalier83:
I would try and make the commit message a bit more helpful to parsing the intent of the code change.
Suggestion:
```
fix: Do not disconnect from peers during pfx change

* This prevents false positive matches of joining nodes.
* Also remove unnecessary checks for peer states that are mutually exclusive with the Routing state.

* chore: Update Cargo.toml to use ~ for deps.
```

tests/mock_crust/churn.rs
```diff
@@ -155,11 +155,11 @@ fn random_churn<R: Rng>(
     let len = nodes.len();
     let section_count = count_sections(nodes);
     if section_count > 1 && !rng.gen_weighted_bool(section_count as u32) {
-        // Only drop less than GRP_SIZE quorum to prevent collapsing any grp in the network.
-        let max_drop = cmp::max(
-            (nodes[0].chain().min_sec_size() * QUORUM_NUMERATOR / QUORUM_DENOMINATOR) - 1,
-            1,
-        );
+        // Use min_sec_size than section size to prevent collapsing any groups.
```
pierrechevalier83: maybe "rather than" or "instead of" rather than "than" :smile:

src/states/node.rs
```diff
@@ -991,6 +991,10 @@ impl Node {
         self.gen_pfx_info = Some(gen_pfx_info);
         let _ = self.init_parsec()?; // We don't reset the chain on prefix change.
 
+        // Clear any pending candidate
+        // TODO: maybe also disconnect if currently connected.
```
pierrechevalier83: Maybe could tackle now rather than keeping it as a TODO?

Viv-Rajkumar: It'd possibly be redundant tho based on crust changes if we arent tracking non-routing peers via peer_mgr PeerState anyways. This was more a point to consider it if we for some reason didnt do the crust integration like so

src/states/node.rs
```diff
@@ -3342,36 +3342,33 @@ impl Bootstrapped for Node {
             }
             return Ok(());
         }
-        if !self.add_to_pending_acks(&routing_msg, route, expires_at) {
+
+        use crate::routing_table::Authority::*;
+        let sending_sec = if route == 0 {
+            match routing_msg.src {
+                ClientManager(_) | NaeManager(_) | NodeManager(_) | ManagedNode(_) | Section(_)
+                | PrefixSection(_)
```
afck: I think the idea with `PrefixSection` was that the node collecting the signatures could be in our sibling section. If we put in a different source section, our signature won't match. Even with the "N/3→N/3" change, I think that might still be relevant?

afck: (As discussed, this is probably not relevant anymore. Possibly error if source is `PrefixSection`.)

PR #1613, opened 2019-04-18
Refactor Node state: extract ProvingNode state

madadam:
The bulk of this PR is the extraction of new node state called `ProvingNode` which corresponds to node undergoing resource proof. 

The PR also includes these additional changes:
- extraction of some common bits from `Client`, `JoiningNode` and `ProvingNode` into the `Unapproved` trait.
- addition of `time` module which provides `Instant` and `Duration` and selects the real or fake clock impls depending on the presence of the `mock` feature.
- some time constants changed from `u64` to `Duration` for better type safety and convenience (enabled by `Duration::from_secs` etc. being `const fn`s now).


src/states/proving_node.rs
```diff
@@ -0,0 +1,408 @@
...
+    xor_name::XorName,
+    CrustEvent, PubConnectionInfo, Service,
+};
+use maidsafe_utilities::serialisation;
+use std::{
+    collections::BTreeSet,
+    fmt::{self, Display, Formatter},
+};
+
+pub struct ProvingNode {
```
jeanphilippeD: Adam, could you point me how you created this file (proving_node.rs), where did you pick the pieces from (since it is not removed in this commit)

src/states/common/unapproved.rs
```diff
@@ -0,0 +1,51 @@
...
+    ) -> Result<Option<RoutingMessage>, RoutingError> {
+        hop_msg.verify(pub_id.signing_public_key())?;
+
+        let signed_msg = hop_msg.content;
+        signed_msg.check_integrity(self.min_section_size())?;
+
+        let routing_msg = signed_msg.into_routing_message();
+        let in_authority = self.in_authority(&routing_msg.dst);
+        if in_authority {
+            self.send_ack(&routing_msg, 0);
```
jeanphilippeD: This was not in src/states/client.rs.
Only in src/states/joining_node.rs, could you explain the change in behaviour

src/states/proving_node.rs
madadam: The pieces are mostly extracted from `node.rs`, but they are removed from it only in later commits. It was easier for me to work that way but I can see how it can make the reviewing harder. Sorry about that :flushed:

jeanphilippeD: No problem, just needed the right source to compare.

src/states/common/unapproved.rs
madadam: There is no change, the difference is handled with the `send_ack` param now.

jeanphilippeD: Sorry I do not understand what you mean, can you be more specific.

src/states/common/unrelocated.rs
```diff
@@ -0,0 +1,64 @@
...
+use crate::{
+    error::{Result, RoutingError},
+    id::PublicId,
+    messages::{RoutingMessage, SignedMessage},
+    routing_table::Authority,
+    time::Instant,
+};
+use std::collections::BTreeSet;
+
+pub trait Unrelocated: Bootstrapped {
```
jeanphilippeD: It looks fine, but what is the benefit of using a trait over a free function.
Is there a problem with passing the PublicId?

src/states/common/unapproved.rs
madadam: Sorry, sure: if you look at the code before this PR, then you can see that the `handle_hop_message` method in [Client](https://github.com/maidsafe/routing/blob/master/src/states/client.rs#L207-L241) and in [JoiningNode](https://github.com/maidsafe/routing/blob/master/src/states/joining_node.rs#L219-L255) are almost identical. The only difference is that `JoiningNode` does `send_ack` but `Client` doesn't. So I extracted that logic into `filter_hop_message` where the `send_ack` difference is handled using additional argument.

src/states/common/unrelocated.rs
madadam: It's mostly about making the borrow checker happy. I'd prefer to use either free functions or composition but it would require more extensive refactoring. Maybe something for the future?

src/states/common/unapproved.rs
jeanphilippeD: I was looking that the commit https://github.com/maidsafe/routing/pull/1613/commits/340c8328a7b07e001cc5a5b2171319ed03b02790 which did the extraction but not have it, I guess you fixed it in a latter commit then.

src/states/proving_node.rs
```diff
@@ -0,0 +1,649 @@
...
+            .handle_timeout(token, log_ident, outbox)
+        {
+            transition
+        } else {
+            self.resend_unacknowledged_timed_out_msgs(token);
+            Transition::Stay
+        }
+    }
+
+    fn send_connection_info_request(
```
jeanphilippeD: This is mostly a copy paste of the one still in src/states/node.rs.
We probably should share the bulk of the implementation.

Is there other like these, I'll try to point them out.

src/states/node.rs
```diff
@@ -1285,15 +1192,16 @@ impl Node {
         use crate::messages::MessageContent::*;
         use crate::Authority::{Client, ManagedNode, PrefixSection, Section};
 
-        if !self.chain.is_member() && !self.is_first_node {
+        if !self.chain.is_member() {
```
jeanphilippeD: Why did this change? Is it because of https://github.com/maidsafe/routing/pull/1613/commits/55136aa72e9e82dc298b5021fda7170ed7846b68 ?

Could you say a bit more about this change in the PR description, I'm not sure I see all the implication, because is_first_node is still a member, so some thing still needs it.

src/states/node.rs
```diff
@@ -1302,18 +1210,6 @@ impl Node {
                     self.routing_msg_backlog.push(routing_msg);
                     return Ok(());
                 }
-                ConnectionInfoRequest { .. } => {
```
jeanphilippeD: Why did this go away, it does not seem to be a ProvingNode thing?

src/states/node.rs
madadam: > Is it because of 55136aa

Precisely. First node now has its chain initialised from the beginning, so `chain.is_member()` does always return `true` for it.

src/states/proving_node.rs
madadam: You are right. I'll try to extract the common parts and put them into `Relocated`.

src/resource_prover.rs
```diff
@@ -6,41 +6,40 @@
...
 /// satisfy the proof of resource challenge within this time, no `CandidateApproval` is sent.
-pub const RESOURCE_PROOF_DURATION_SECS: u64 = 300;
+pub const RESOURCE_PROOF_DURATION: Duration = Duration::from_secs(300);
 /// Maximum time a new node will wait to receive `NodeApproval` after receiving a
 /// `RelocateResponse`. This covers the built-in delay of the process and also allows time for the
 /// message to accumulate and be sent via four different routes.
-const APPROVAL_TIMEOUT_SECS: u64 =
-    RESOURCE_PROOF_DURATION_SECS + ACCUMULATION_TIMEOUT_SECS + (4 * ACK_TIMEOUT_SECS);
+const APPROVAL_TIMEOUT: Duration = Duration::from_secs(
+    RESOURCE_PROOF_DURATION.as_secs()
```
jeanphilippeD: Do these not work directly on the time? instead of back and forth conversion to secs?

madadam: The `+` operator isn't `const` function, so can't be used in `const` definitions yet ([Rust playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a1ef3b3529ddb9298785e7b333b77389))

src/states/client.rs
```diff
@@ -200,42 +194,21 @@ impl Client {
...
-        {
-            FilteringResult::KnownMessage | FilteringResult::KnownMessageAndRoute => {
-                return Err(RoutingError::FilterCheckFailed);
-            }
-            FilteringResult::NewMessage => (),
-        }
-
-        if !in_authority {
-            return Ok(Transition::Stay);
+        if let Some(routing_msg) = self.filter_hop_message(hop_msg, pub_id, false)? {
```
jeanphilippeD: I know what that false is for but it is not very nice.
Could it be a constant implemented by the trait https://doc.rust-lang.org/edition-guide/rust-2018/trait-system/associated-constants.html ? Would it help?

src/states/common/relocated.rs
```diff
@@ -0,0 +1,357 @@
...
+    // Filter, then convert the message to a `Hop` and serialise.
+    // Send this byte string.
+    fn send_signed_message_to_peer(
+        &mut self,
+        signed_msg: SignedMessage,
+        target: &PublicId,
+        route: u8,
+        sent_to: BTreeSet<XorName>,
+    ) -> Result<(), RoutingError> {
+        if self.filter_outgoing_routing_msg(signed_msg.routing_message(), target, route) {
```
jeanphilippeD: Why move the filter? I mean it make sense to me to do filter first, but it also seem to change things like before, it would have called disconnect_peer but it won't now.

src/states/node.rs
madadam: OK, so the difference regarding this between the fleming branch and this PR is only in how we handle the case when we are already approved, but not yet established and we receive `ConnectionInfoRequest` whose `src.name` matches out `joining_prefix`. In fleming, we would handle the message, but after this PR we only backlog it. This may or might not be incorrect. I guess the question is whether non-established node needs to handle connection requests now, or whether it is OK for it to wait until it becomes established. In case we need to handle them, another question is what to use in place of `joining_prefix`, as `Node` no longer has that member. I guess `chain.our_prefix()` ?

src/states/client.rs
madadam: :+1: Sounds like a good idea!

src/states/node.rs
jeanphilippeD: I do not know if it is correct or not, but if we are unsure, maybe better to keep it as it was?
Also, since this is a refactoring PR, we should probably explicitly document any change of behaviour in the PR description.

madadam: You are probably right. I'll keep the original behaviour.

pierrechevalier83:
Just leaving a general comment for future PRs:

I love refactoring commits, but a combination of this branch's history being a bit approximate and github being clunky with git makes this PR pretty hard to review.

I think there is a really easy way to improve on that in the future, though:
For commits that are straightforward refactorings, affect most of the codebase and are easily isolated in their own commits, rebase them to the bottom of your branch and make it its own PR.
This way, all the changes that are a bit harder to keep cleanly separated can be in their own PR and viewing all the files at the same time becomes an option.
This flow would allow to be a bit relaxed about making perfectly well isolated commits where it's more tricky, while remaining easier to review at a glance.

As a reviewer, my workflow for this one is the following:
* read all the commits; give good attention to the simple ones that are pervasive and take note of them as good to ignore later
* review all files at once and mentally ignore all the pervasive changes that I've already reviewed
Just a little bit inconvenient :wink:

src/states/joining_node.rs
```diff
@@ -437,6 +377,19 @@ impl Bootstrapped for JoiningNode {
     }
 }
 
+impl Unapproved for JoiningNode {
+    const SEND_ACK: bool = true;
+
+    fn get_proxy_public_id(&self, proxy_name: &XorName) -> Result<&PublicId> {
```
jeanphilippeD: Can you move this and the duplicate implementation into pub trait Unapproved, and only expose a function like:
(not sure about the name).

```
fn get_unchecked_proxy_public_id(&self) -> &PublicId {
    &self.proxy_pub_id
}
```

src/states/node.rs
```diff
@@ -1516,98 +1407,29 @@ impl Node {
...
 
-    fn init_parsec(&mut self) -> Result<bool, RoutingError> {
-        let genesis_info = if let Some(ref genesis_info) = self.gen_pfx_info {
-            genesis_info.clone()
-        } else {
-            return Ok(false);
-        };
-
-        if self
+    fn init_parsec(&mut self) -> bool {
```
jeanphilippeD: why return a bool since we do not use it?

jeanphilippeD: Also can simplify the implementation of init_parsec as a result.

src/states/node.rs
```diff
@@ -3424,8 +2760,120 @@ impl Bootstrapped for Node {
...
+        if self.chain.is_member() && self.chain.our_info().members().contains(&pub_id) {
+            self.vote_for_event(NetworkEvent::Offline(pub_id));
+        }
+    }
+
+    fn is_peer_valid(&self, pub_id: &PublicId) -> bool {
+        !self.chain.is_member() || self.chain.is_peer_valid(pub_id)
+    }
+
+    fn add_to_routing_table(&mut self, pub_id: &PublicId, outbox: &mut EventBox) {
```
jeanphilippeD: This one also seem mostly duplicated, could we move it to the trait?

src/states/proving_node.rs
```diff
@@ -0,0 +1,601 @@
...
+            if let Err(error) = self.send_connection_info_request(*pub_id, src, dst, &mut outbox) {
+                debug!(
+                    "{} - Failed to send connection info request to {:?}: {:?}",
+                    self, pub_id, error
+                );
+            }
+        }
+    }
+
+    pub fn handle_action(&mut self, action: Action, outbox: &mut EventBox) -> Transition {
```
jeanphilippeD: With the current pattern these seem to be multiple somewhat copy pasted implementation.

src/states/proving_node.rs
```diff
@@ -0,0 +1,601 @@
...
+            }
+            Action::Terminate => {
+                return Transition::Terminate;
+            }
+        }
+
+        Transition::Stay
+    }
+
+    pub fn handle_crust_event(
```
jeanphilippeD: Same for this, a lot of copy paste. Not sure we can do anything about it.

src/states/node.rs
```diff
@@ -3424,8 +2760,120 @@ impl Bootstrapped for Node {
...
+            full_id,
+            &gen_pfx_info.first_info.members(),
+            &gen_pfx_info.latest_info.members(),
+            ConsensusMode::Single,
+        )
+    }
+}
+
+#[cfg(feature = "mock_parsec")]
+fn create_parsec(full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec<NetworkEvent, FullId> {
```
jeanphilippeD: Avoiding the duplication?

```
fn create_parsec(full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec<NetworkEvent, FullId> {
    if gen_pfx_info
        .first_info
        .members()
        .contains(full_id.public_id())
    {
        Parsec::from_genesis(
            #[cfg(feature = "mock_parsec")]
            *gen_pfx_info.first_info.hash(),
            section_hash,
            full_id,
            &gen_pfx_info.first_info.members(),
            ConsensusMode::Single,
        )
    } else {
        Parsec::from_existing(
            #[cfg(feature = "mock_parsec")]
            *gen_pfx_info.first_info.hash(),
            full_id,
            &gen_pfx_info.first_info.members(),
            &gen_pfx_info.latest_info.members(),
            ConsensusMode::Single,
        )
    }
}
```

src/states/common/relocated.rs
madadam: I think I wanted to avoid having to `clone` the routing_msg, but you are right, it changes the logic. Will fix.

src/states/joining_node.rs
madadam: This is also used in `ProvingNode` where it wouldn't work, because there is no `self.proxy_pub_id`. I can try to extract the duplicate bits into a free function though.

jeanphilippeD: I tend to think it is pretty much always worth it, for maintenance to avoid duplicate code.
But I'll let you decide.

src/states/proving_node.rs
madadam: There are differences though. Don't see an obvious way to deduplicate it without making the code convoluted. Might be worth to think about more radical reorganization of the code to avoid this.

src/states/proving_node.rs
madadam: Same as above: https://github.com/maidsafe/routing/pull/1613#discussion_r276741015

jeanphilippeD: Yes, it was more something that showed up in this diff that is likely to cause issue in the future, but we probably cannot really do anything at this point.

src/states/client.rs
```diff
@@ -200,42 +194,21 @@ impl Client {
...
-        {
-            FilteringResult::KnownMessage | FilteringResult::KnownMessageAndRoute => {
-                return Err(RoutingError::FilterCheckFailed);
-            }
-            FilteringResult::NewMessage => (),
-        }
-
-        if !in_authority {
-            return Ok(Transition::Stay);
+        if let Some(routing_msg) = self.filter_hop_message(hop_msg, pub_id)? {
```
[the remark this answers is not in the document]
pierrechevalier83: Nevermind: it's the same comment JP did later, and I can see that the Unapproved trait has SEND_ACK set to false which means it's behaving the same as it used to.

src/state_machine.rs
```diff
@@ -400,6 +410,13 @@ impl StateMachine {
                 };
                 self.state = new_state;
             }
+            IntoNode { gen_pfx_info } => {
+                let new_state = match mem::replace(&mut self.state, State::Terminated) {
```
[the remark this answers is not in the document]
madadam: I don't think there is a way to do it without the temporary state and without involving `unsafe`, but if you think of one, I'm all ears :D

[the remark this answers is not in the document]
jeanphilippeD: It also follow the existing pattern.

src/states/node.rs
```diff
@@ -490,36 +443,6 @@ impl Node {
...
-        self.process_connection(pub_id, outbox);
-    }
-
-    fn handle_connect_failure(&mut self, pub_id: PublicId, outbox: &mut EventBox) {
-        if let Some(&PeerState::CrustConnecting) = self.peer_mgr.get_peer(&pub_id).map(Peer::state)
-        {
-            debug!("{} Failed to connect to peer {:?}.", self, pub_id);
-        }
-        let _ = self.dropped_peer(&pub_id, outbox, true);
-        if self.chain.is_member() && self.chain.our_info().members().contains(&pub_id) {
```
pierrechevalier83: Lines 518-520 didn't make it to proving_node.rs handle_connect_failure.
I would like to confirm it's intentional.

madadam: `ProvingNode` doesn't have neither `Parsec` nor `Chain`, so those lines make no sense for it.

madadam: Another way to look at it is that anything that is behind `if self.chain.is_member() { ... }` doesn't belong to `ProvingNode`, because node in that state isn't a full section member yet.

src/states/node.rs
```diff
@@ -1516,98 +1417,19 @@ impl Node {
...
-             network!",
-            self
-        );
-
-        self.gen_pfx_info = Some(gen_info.clone());
-        if self.init_parsec()? {
-            self.init_chain();
-        }
-
-        self.poke_timer_token = Some(self.timer.schedule(Duration::from_secs(POKE_TIMEOUT_SECS)));
```
pierrechevalier83: For my info, what does this poke_timer_token represent? (I see you kept it in `Node`, but not in `ResourceProving` state)

src/states/node.rs
pierrechevalier83: Looks intentional: we can't vote Offline for someone who's still `ResourceProving`. Also, I see this remains the same in this state (way later in this file)

src/states/proving_node.rs
```diff
@@ -0,0 +1,595 @@
...
+                );
+            }
+        }
+    }
+
+    pub fn handle_action(&mut self, action: Action, outbox: &mut EventBox) -> Transition {
+        match action {
+            Action::ClientSendRequest { result_tx, .. }
+            | Action::NodeSendMessage { result_tx, .. } => {
+                // TODO (adam): should we backlog `NodeSendMessage`s and handle them once we are
```
pierrechevalier83: This is still under consideration, so I agree with the current implementation. (Would keep as is, with the TODO for further consideration. Just giving my opinion to answer this question)

src/states/node.rs
pierrechevalier83: Yes, sorry: I realised this later (Had posted prematurely by hitting the wrong button :p)

src/states/node.rs
```diff
@@ -1285,32 +1192,21 @@ impl Node {
...
                 ConnectionInfoRequest { .. } => {
-                    if !self.joining_prefix.matches(&routing_msg.src.name()) {
-                        // Doesn't allow other node connect to us before node approval
-                        trace!(
-                            "{} Not approved yet. Delaying message handling: {:?}",
-                            self,
-                            routing_msg
-                        );
-                        self.routing_msg_backlog.push(routing_msg);
+                    if !self.chain.our_prefix().matches(&routing_msg.src.name()) {
```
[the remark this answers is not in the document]
madadam: `ProvingNode` doesn't have `Chain` yet, so the prefix of the section it is becoming member of is stored in `joining_prefix`. When it is approved and becomes `Node`, it gets `Chain` and so can use `Chain::our_prefix` instead.

src/states/node.rs
madadam: "Poke" is a message send from a newly joined node to a random member of the section to nudge them to send the first parsec gossip. A node upon receiving the poke will immediately respond by sending parsec request back. It isn't in `ProvingNode` because there is no `Parsec` yet.

src/states/proving_node.rs
madadam: We can also handle this in the upper layers (vault): either prevent sending them until the node is a full member, or backlog them there.

src/states/node.rs
```diff
@@ -2034,297 +1853,6 @@ impl Node {
...
-                }
-                return;
-            }
-            Ok(()) => (),
-        }
-
-        if self.notified_nodes.insert(*pub_id) {
-            info!("{} Added {} to routing table.", self, pub_id);
-            outbox.send_event(Event::NodeAdded(*pub_id.name()));
-            self.print_rt_size();
```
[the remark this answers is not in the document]
madadam: Before this PR, it seems the `NodeAdded` event was raised regardless of whether the node is approved or not. But the corresponding `NodeLost` event is raised only for members: https://github.com/maidsafe/routing/blob/fleming/src/states/node.rs#L3144 which seems strange. I wonder whether there is a deep reason for this asymmetry or whether it's just an omission. And what the correct behaviour should be. So yeah, good question :)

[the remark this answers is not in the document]
pierrechevalier83: That is an interesting question, although I was speaking specifically about `print_rt_size` here. I don't see why this only happens in the Node state.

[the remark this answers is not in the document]
madadam: I see. Looks `print_rt_size` depends on `Chain` too: https://github.com/maidsafe/routing/blob/fleming/src/states/node.rs#L294-L313

src/state_machine.rs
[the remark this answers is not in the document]
afck: I'd love to have a proper solution to this problem! Maybe it would be slightly cleaner to isolate the place where we assign the temporary value, to separate it from the business logic itself (that's what I did [elsewhere](https://github.com/poanetwork/hbbft/blob/a64f62506d1c3e98ad805c61f2c825439d2025d3/src/subset/proposal_state.rs#L162)); so we'd have an
```rust
fn do_apply_transition(state: State, crust_tx: …, …, transition: …, outbox: …) -> State;
```
and in `apply_transition` itself we'd only do the swap:
```rust
let state = mem::replace(&mut self.state, State::Terminated);
let new_state = do_apply_transition(state, &mut self crust_tx, …, transition, outbox);
self.state = new_state;
```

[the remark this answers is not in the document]
madadam: Adding a function like that should be doable. Not sure we can do much better than that though.

PR #1615, opened 2019-04-24
Miscellaneous refactoring

madadam:


src/states/proving_node.rs
```diff
@@ -0,0 +1,555 @@
...
+            self,
+            our_section.len()
+        );
+
+        let src = Authority::Client {
+            client_id: *self.full_id.public_id(),
+            proxy_node_name: *proxy_pub_id.name(),
+        };
+        // There will be no events raised as a result of these calls, so safe to just use a
+        // throwaway `EventBox` here.
```
afck: And I guess we can't remove the arguments because they're trait methods?
Could the trait have an associated type instead, that could be set to `EventBox` or `()`?

src/state_machine.rs
```diff
@@ -6,35 +6,46 @@
...
+        match $self {
+            State::Bootstrapping($state) => $expr,
+            State::Client($state) => $expr,
+            State::RelocatingNode($state) => $expr,
+            State::ProvingNode($state) => $expr,
+            State::Node($state) => $expr,
+            State::Terminated => $term_expr,
+        }
+    };
+}
```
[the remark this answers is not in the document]
madadam: I don't think this would work. The states have each their own type and the current state is known only at runtime, but the closure type needs to be known at compile time.

src/states/proving_node.rs
madadam: Good idea. I'll give it a try.

src/states/common/relocated.rs
```diff
@@ -0,0 +1,452 @@
...
+    fn add_to_notified_nodes(&mut self, pub_id: PublicId) -> bool;
+    fn add_to_routing_table_success(&mut self, pub_id: &PublicId);
+    fn add_to_routing_table_failure(&mut self, pub_id: &PublicId);
+
+    fn handle_connection_info_prepared(
+        &mut self,
+        result_token: u32,
+        result: Result<PrivConnectionInfo, CrustError>,
+    ) -> Transition {
+        let our_connection_info = match result {
```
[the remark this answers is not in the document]
madadam: I don't think I can `return Transition::Stay` from inside the closure.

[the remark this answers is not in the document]
afck: Ah, right, this _returns_! :+1:

src/state_machine.rs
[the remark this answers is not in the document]
afck: Ah, you're right, of course! So `F` would have to be `Fn(Box<dyn Base>) -> T` or something, but that probably causes other problems…

src/states/proving_node.rs
madadam: Unfortunately, this turned out to not be possible. But I added a `outbox` param to the function, so it is now handled normally like anywhere else, removing this weird special case.

src/states/mod.rs
```diff
@@ -9,14 +9,37 @@
...
+//        │   ┌────────────────┐ ┌─────────────┐
+//        │   │ RelocatingNode │ │ ProvingNode │
+//        │   └────────────────┘ └─────────────┘
+//        │                        │
+//        │                        │
+//        ▼                        ▼
+// ┌────────┐                    ┌──────┐
+// │ Client │                    │ Node │
+// └────────┘                    └──────┘
+//
```
octol: Did you create this manually? Looks fantastic! :)

madadam: It uses the Unicode box drawing symbols instead of just ASCII, but yes, it's all manual.

pierrechevalier83: I :heart: Unicode box drawing :smile: https://github.com/pierrechevalier83/matrix_display/blob/master/src/style.rs

PR #1616, opened 2019-04-24
Refactor Node state: extract EstablishingNode state

madadam:
This is a refactoring PR with almost no functional changes. Exceptions are some minor changes for the sake of simplicity, clarity or performance. They are listed below:

1. When `EstablishingNode` receives `NodeApproval`, it backlogs it and once it becomes `Node` it logs and ignores it. Previously it would immediately log and ignore it.
2. In `EstablishingNode`, `handle_hop_message` and `send_routing_message_via_route` now work the same way as in `ProvingNode` as opposed to the way they work in `Node`. This makes it more explicit what the responsibilities of each of the states are.

### DIFF
--- a/src/mock_parsec/mod.rs
+++ b/src/mock_parsec/mod.rs
@@ -184,10 +184,21 @@ where
         unimplemented!()
     }
 
+    // TODO: rename this to `has_unpolled_observations`
     pub fn has_unconsensused_observations(&self) -> bool {
-        self.observations
-            .iter()
-            .any(|(_, obs)| obs.state != ConsensusState::Polled)
+        state::with(self.section_hash, |state| {
+            state
+                .observations
+                .iter()
+                .any(|(holder, observation_state)| {
+                    !observation_state.consensused()
+                        || self
+                            .observations
+                            .get(&*holder)
+                            .map(|info| info.state != ConsensusState::Polled)
+                            .unwrap_or(true)
+                })
+        })
     }
 
     pub fn our_unpolled_observations(&self) -> impl Iterator<Item = &Observation<T, S::PublicId>> {
--- a/src/mock_parsec/mod.rs
+++ b/src/mock_parsec/mod.rs
@@ -186,19 +186,12 @@ where
 
     // TODO: rename this to `has_unpolled_observations`
     pub fn has_unconsensused_observations(&self) -> bool {
-        state::with(self.section_hash, |state| {
+        state::with::<T, S::PublicId, _, _>(self.section_hash, |state| {
             state
                 .observations
                 .iter()
-                .any(|(holder, observation_state)| {
-                    !observation_state.consensused()
-                        || self
-                            .observations
-                            .get(&*holder)
-                            .map(|info| info.state != ConsensusState::Polled)
-                            .unwrap_or(true)
-                })
-        })
+                .any(|(_, observation_state)| !observation_state.consensused())
+        }) || self.our_unpolled_observations().next().is_some()
     }
 
     pub fn our_unpolled_observations(&self) -> impl Iterator<Item = &Observation<T, S::PublicId>> {
--- a/src/state_machine.rs
+++ b/src/state_machine.rs
@@ -13,7 +13,7 @@ use crate::{
     outbox::EventBox,
     routing_table::Prefix,
     states::common::Base,
-    states::{Bootstrapping, Client, EstablishingNode, Node, ProvingNode, RelocatingNode},
+    states::{BootstrappingPeer, Client, EstablishingNode, Node, ProvingNode, RelocatingNode},
     timer::Timer,
     types::RoutingActionSender,
     xor_name::XorName,
@@ -37,7 +37,7 @@ use std::{
 macro_rules! state_dispatch {
     ($self:expr, $state:pat => $expr:expr, Terminated => $term_expr:expr) => {
         match $self {
-            State::Bootstrapping($state) => $expr,
+            State::BootstrappingPeer($state) => $expr,
             State::Client($state) => $expr,
             State::RelocatingNode($state) => $expr,
             State::ProvingNode($state) => $expr,
@@ -64,7 +64,7 @@ pub struct StateMachine {
 // FIXME - See https://maidsafe.atlassian.net/browse/MAID-2026 for info on removing this exclusion.
 #[allow(clippy::large_enum_variant)]
 pub enum State {
-    Bootstrapping(Bootstrapping),
+    BootstrappingPeer(BootstrappingPeer),
     Client(Client),
     RelocatingNode(RelocatingNode),
     ProvingNode(ProvingNode),
@@ -127,7 +127,7 @@ impl State {
         match *self {
             State::EstablishingNode(ref state) => Some(state.chain()),
             State::Node(ref state) => Some(state.chain()),
-            State::Bootstrapping(_)
+            State::BootstrappingPeer(_)
             | State::Client(_)
             | State::RelocatingNode(_)
             | State::ProvingNode(_)
@@ -200,7 +200,7 @@ impl State {
 
     pub fn get_timed_out_tokens(&mut self) -> Vec<u64> {
         match *self {
-            State::Bootstrapping(_) | State::Terminated => vec![],
+            State::BootstrappingPeer(_) | State::Terminated => vec![],
             State::Client(ref mut state) => state.get_timed_out_tokens(),
             State::RelocatingNode(ref mut state) => state.get_timed_out_tokens(),
             State::ProvingNode(ref mut state) => state.get_timed_out_tokens(),
@@ -220,7 +220,7 @@ impl State {
     pub fn has_unpolled_observations(&self, filter_opaque: bool) -> bool {
         match *self {
             State::Terminated
-            | State::Bootstrapping(_)
+            | State::BootstrappingPeer(_)
             | State::Client(_)
             | State::RelocatingNode(_)
             | State::ProvingNode(_) => false,
@@ -239,7 +239,7 @@ impl State {
 
     pub fn in_authority(&self, auth: &Authority<XorName>) -> bool {
         match *self {
-            State::Terminated | State::Bootstrapping(_) => false,
+            State::Terminated | State::BootstrappingPeer(_) => false,
             State::Client(ref state) => state.in_authority(auth),
             State::RelocatingNode(ref state) => state.in_authority(auth),
             State::ProvingNode(ref state) => state.in_authority(auth),
@@ -250,7 +250,7 @@ impl State {
 
     pub fn has_unacked_msg(&self) -> bool {
         match *self {
-            State::Terminated | State::Bootstrapping(_) => false,
+            State::Terminated | State::BootstrappingPeer(_) => false,
             State::Client(ref state) => state.ack_mgr().has_unacked_msg(),
             State::RelocatingNode(ref state) => state.ack_mgr().has_unacked_msg(),
             State::ProvingNode(ref state) => state.ack_mgr().has_unacked_msg(),
@@ -409,7 +409,7 @@ impl StateMachine {
         match transition {
             Stay => (),
             IntoBootstrapped { proxy_public_id } => self.state.replace_with(|state| match state {
-                State::Bootstrapping(src) => src.into_target_state(proxy_public_id, outbox),
+                State::BootstrappingPeer(src) => src.into_target_state(proxy_public_id, outbox),
                 _ => unreachable!(),
             }),
             IntoBootstrapping {
--- a/src/state_machine.rs
+++ b/src/state_machine.rs
@@ -193,49 +193,58 @@ impl State {
 
     pub fn get_timed_out_tokens(&mut self) -> Vec<u64> {
         match *self {
-            State::Node(ref mut state) => state.get_timed_out_tokens(),
+            State::Bootstrapping(_) | State::Terminated => vec![],
             State::Client(ref mut state) => state.get_timed_out_tokens(),
             State::RelocatingNode(ref mut state) => state.get_timed_out_tokens(),
-            _ => vec![],
+            State::ProvingNode(ref mut state) => state.get_timed_out_tokens(),
+            State::Node(ref mut state) => state.get_timed_out_tokens(),
         }
     }
 
     pub fn get_clients_usage(&self) -> Option<BTreeMap<IpAddr, u64>> {
-        match *self {
-            State::Node(ref state) => Some(state.get_clients_usage()),
-            _ => None,
+        if let State::Node(ref state) = *self {
+            Some(state.get_clients_usage())
+        } else {
+            None
         }
     }
 
     pub fn has_unconsensused_observations(&self, filter_opaque: bool) -> bool {
         match *self {
+            State::Terminated
+            | State::Bootstrapping(_)
+            | State::Client(_)
+            | State::RelocatingNode(_)
+            | State::ProvingNode(_) => false,
             State::Node(ref state) => state.has_unconsensused_observations(filter_opaque),
-            _ => false,
         }
     }
 
     pub fn is_routing_peer(&self, pub_id: &PublicId) -> bool {
-        match *self {
-            State::Node(ref state) => state.is_routing_peer(pub_id),
-            _ => false,
+        if let State::Node(ref state) = *self {
+            state.is_routing_peer(pub_id)
+        } else {
+            false
         }
     }
 
     pub fn in_authority(&self, auth: &Authority<XorName>) -> bool {
         match *self {
-            State::Node(ref state) => state.in_authority(auth),
+            State::Terminated | State::Bootstrapping(_) => false,
             State::Client(ref state) => state.in_authority(auth),
             State::RelocatingNode(ref state) => state.in_authority(auth),
-            _ => false,
+            State::ProvingNode(ref state) => state.in_authority(auth),
+            State::Node(ref state) => state.in_authority(auth),
         }
     }
 
     pub fn has_unacked_msg(&self) -> bool {
         match *self {
-            State::Node(ref state) => state.ack_mgr().has_unacked_msg(),
+            State::Terminated | State::Bootstrapping(_) => false,
             State::Client(ref state) => state.ack_mgr().has_unacked_msg(),
             State::RelocatingNode(ref state) => state.ack_mgr().has_unacked_msg(),
-            _ => false,
+            State::ProvingNode(ref state) => state.ack_mgr().has_unacked_msg(),
+            State::Node(ref state) => state.ack_mgr().has_unacked_msg(),
         }
     }
 }
--- a/src/state_machine.rs
+++ b/src/state_machine.rs
@@ -8,12 +8,12 @@
 
 use crate::{
     action::Action,
-    chain::GenesisPfxInfo,
+    chain::{GenesisPfxInfo, SectionInfo},
     id::{FullId, PublicId},
     outbox::EventBox,
     routing_table::Prefix,
     states::common::Base,
-    states::{Bootstrapping, Client, Node, ProvingNode, RelocatingNode},
+    states::{Bootstrapping, Client, EstablishingNode, Node, ProvingNode, RelocatingNode},
     timer::Timer,
     types::RoutingActionSender,
     xor_name::XorName,
@@ -41,6 +41,7 @@ macro_rules! state_dispatch {
             State::Client($state) => $expr,
             State::RelocatingNode($state) => $expr,
             State::ProvingNode($state) => $expr,
+            State::EstablishingNode($state) => $expr,
             State::Node($state) => $expr,
             State::Terminated => $term_expr,
         }
@@ -67,6 +68,7 @@ pub enum State {
     Client(Client),
     RelocatingNode(RelocatingNode),
     ProvingNode(ProvingNode),
+    EstablishingNode(EstablishingNode),
     Node(Node),
     Terminated,
 }
@@ -123,8 +125,13 @@ impl State {
     #[cfg(feature = "mock_base")]
     fn chain(&self) -> Option<&Chain> {
         match *self {
+            State::EstablishingNode(ref state) => Some(state.chain()),
             State::Node(ref state) => Some(state.chain()),
-            _ => None,
+            State::Bootstrapping(_)
+            | State::Client(_)
+            | State::RelocatingNode(_)
+            | State::ProvingNode(_)
+            | State::Terminated => None,
         }
     }
 
@@ -197,6 +204,7 @@ impl State {
             State::Client(ref mut state) => state.get_timed_out_tokens(),
             State::RelocatingNode(ref mut state) => state.get_timed_out_tokens(),
             State::ProvingNode(ref mut state) => state.get_timed_out_tokens(),
+            State::EstablishingNode(ref mut state) => state.get_timed_out_tokens(),
             State::Node(ref mut state) => state.get_timed_out_tokens(),
         }
     }
@@ -216,6 +224,9 @@ impl State {
             | State::Client(_)
             | State::RelocatingNode(_)
             | State::ProvingNode(_) => false,
+            State::EstablishingNode(ref state) => {
+                state.has_unconsensused_observations(filter_opaque)
+            }
             State::Node(ref state) => state.has_unconsensused_observations(filter_opaque),
         }
     }
@@ -234,6 +245,7 @@ impl State {
             State::Client(ref state) => state.in_authority(auth),
             State::RelocatingNode(ref state) => state.in_authority(auth),
             State::ProvingNode(ref state) => state.in_authority(auth),
+            State::EstablishingNode(ref state) => state.in_authority(auth),
             State::Node(ref state) => state.in_authority(auth),
         }
     }
@@ -244,6 +256,7 @@ impl State {
             State::Client(ref state) => state.ack_mgr().has_unacked_msg(),
             State::RelocatingNode(ref state) => state.ack_mgr().has_unacked_msg(),
             State::ProvingNode(ref state) => state.ack_mgr().has_unacked_msg(),
+            State::EstablishingNode(ref state) => state.ack_mgr().has_unacked_msg(),
             State::Node(ref state) => state.ack_mgr().has_unacked_msg(),
         }
     }
@@ -263,9 +276,14 @@ pub enum Transition {
         new_id: FullId,
         our_section: (Prefix<XorName>, BTreeSet<PublicId>),
     },
-    // `ProvingNode` state transitioning to `Node`.
-    IntoNode {
+    // `ProvingNode` state transitioning to `EstablishingNode`.
+    IntoEstablishingNode {
         gen_pfx_info: GenesisPfxInfo,
+    },
+    // `EstablishingNode` state transition to `Node`.
+    IntoNode {
+        sec_info: SectionInfo,
+        old_pfx: Prefix<XorName>,
     },
     Terminate,
 }
@@ -416,8 +434,14 @@ impl StateMachine {
                     _ => unreachable!(),
                 })
             }
-            IntoNode { gen_pfx_info } => self.state.replace_with(|state| match state {
-                State::ProvingNode(src) => src.into_node(gen_pfx_info),
+            IntoEstablishingNode { gen_pfx_info } => {
+                self.state.replace_with(|state| match state {
+                    State::ProvingNode(src) => src.into_establishing_node(gen_pfx_info),
+                    _ => unreachable!(),
+                });
+            }
+            IntoNode { sec_info, old_pfx } => self.state.replace_with(|state| match state {
+                State::EstablishingNode(src) => src.into_node(sec_info, old_pfx, outbox),
                 _ => unreachable!(),
             }),
             Terminate => self.terminate(),
--- a/src/state_machine.rs
+++ b/src/state_machine.rs
@@ -217,17 +217,15 @@ impl State {
         }
     }
 
-    pub fn has_unconsensused_observations(&self, filter_opaque: bool) -> bool {
+    pub fn has_unpolled_observations(&self, filter_opaque: bool) -> bool {
         match *self {
             State::Terminated
             | State::Bootstrapping(_)
             | State::Client(_)
             | State::RelocatingNode(_)
             | State::ProvingNode(_) => false,
-            State::EstablishingNode(ref state) => {
-                state.has_unconsensused_observations(filter_opaque)
-            }
-            State::Node(ref state) => state.has_unconsensused_observations(filter_opaque),
+            State::EstablishingNode(ref state) => state.has_unpolled_observations(filter_opaque),
+            State::Node(ref state) => state.has_unpolled_observations(filter_opaque),
         }
     }
 
--- a/src/state_machine.rs
+++ b/src/state_machine.rs
@@ -432,12 +432,10 @@ impl StateMachine {
                     _ => unreachable!(),
                 })
             }
-            IntoEstablishingNode { gen_pfx_info } => {
-                self.state.replace_with(|state| match state {
-                    State::ProvingNode(src) => src.into_establishing_node(gen_pfx_info),
-                    _ => unreachable!(),
-                });
-            }
+            IntoEstablishingNode { gen_pfx_info } => self.state.replace_with(|state| match state {
+                State::ProvingNode(src) => src.into_establishing_node(gen_pfx_info, outbox),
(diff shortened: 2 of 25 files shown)


src/states/establishing_node.rs
```diff
@@ -0,0 +1,463 @@
...
+        let chain = Chain::new(min_section_size, public_id, gen_pfx_info.clone());
+
+        let node = Self {
+            ack_mgr,
+            cache,
+            chain,
+            crust_service,
+            full_id,
+            gen_pfx_info,
+            msg_backlog,
```
jeanphilippeD: `routing_msg_backlog: vec![]`

This was initialized with empty vector, while the msg_queue field took the initial content.
Now, we initialize `routing_msg_backlog` with what would have gone to `msg_queue`.

I'm not clear why this is correct?
Do we have test for this, what is that about?

src/states/establishing_node.rs
```diff
@@ -0,0 +1,463 @@
...
+        node
+    }
+
+    pub fn into_node(
+        self,
+        sec_info: SectionInfo,
+        old_pfx: Prefix<XorName>,
+        outbox: &mut EventBox,
+    ) -> State {
+        let node = Node::from_establishing_node(
```
jeanphilippeD: This is not ideal since it rely on type being different to ensure things are ok. And I'm not sure we were doing the right thing with message_backlog comment above.
I'm wondering if we should have a struct here that we pass in from_establishing_node, would also avoid #[allow(clippy::too_many_arguments)]. (Not in this commit either follow up or preceding).

src/states/establishing_node.rs
madadam: `msg_backlog` is for messages that are to be processed later, when we become `Node`. `msg_queue` is for messages that are to be processed now (-ish). Only `ProvingNode` and `EstablishingNode` need `msg_backlog` but don't need `msg_queue` (they either backlog an incoming message, or handle it immediately). `Node`, on the other hand, doesn't need `msg_backlog` (it handles all messages), but needs `msg_queue` (it needs to delay message handling. I think to prevent too deep/infinite recursion. Not 100% sure, but it was there before this PR and I didn't want to change it).

jeanphilippeD: I'm ok with what was there before (even if i don't necessarily understand all of it).

My question was specifically about what I saw as a change of behaviour: before Node started with empty msg_backlog, now the state that replace it EstablishingNode use what Node would take to populate msg_queue, and use it to populate msg_backlog.

src/states/establishing_node.rs
madadam: We can add a struct to avoid the `too_many_arguments` lint, but then we would be just adding more code to satisfy the linter while not really making things cleaner. Seems pointless to me. On the other hand, we could change `Node::from_establishing_node` to take `EstablishingNode` directly, but then we would have to change all fields in `EstablishingNode` to `pub(super)` or something. Might be worth it, not sure.

jeanphilippeD: The benefit i see is with a struct, the field names have to be specified.
I do not think we want to the API with what fields we need in the EstablishingNode iteself.
```
function(
      value1,
      value2,
);

Struct {
     value1,
     value3: value2
};
```

src/states/establishing_node.rs
madadam: Turns out that if `ProvingNode` doesn't handle a message, `EstablishingNode` doesn't handle it either. So it's OK to just transfer `msg_backlog` from `ProvingNode` to `EstablishingNode` without giving `EstablishingNode` chance to handle them first.

Maybe this is something to be documented somewhere? Or alternatively, should we loop through `msg_backlog` when transitioning to `EstablishingNode` to make the code more future proof?

src/states/establishing_node.rs
madadam: It's not public API though.

src/states/node.rs
```diff
@@ -654,34 +687,6 @@ impl Node {
...
-                    if !self.chain.our_prefix().matches(&routing_msg.src.name()) {
-                        self.add_routing_message_to_backlog(routing_msg);
-                        return Ok(());
-                    }
-                }
-                Relocate { .. }
-                | ConnectionInfoResponse { .. }
-                | RelocateResponse { .. }
-                | Ack(..)
-                | NodeApproval { .. } => {
```
jeanphilippeD: I think there is a change of behaviour for Establishing:
NodeApproval was handled as normal, thus generate an error, but now it is pushed to the backlog, I think.

```
    fn handle_routing_message(
        &mut self,
        msg: RoutingMessage,
        outbox: &mut EventBox,
        use crate::{messages::MessageContent::*, routing_table::Authority::*};

        let src_name = msg.src.name();

        match msg {
            RoutingMessage {
                content:
                    ConnectionInfoRequest {
                        encrypted_conn_info,
                        pub_id,
                        msg_id,
                    },
                src: ManagedNode(_),
                dst: ManagedNode(_),
            } => {
                if self.our_prefix().matches(&src_name) {
                    self.handle_connection_info_request(
                        encrypted_conn_info,
                        pub_id,
                        msg_id,
                        msg.src,
                        msg.dst,
                        outbox,
                    )
                } else {
                    self.add_message_to_backlog(RoutingMessage {
                        content: ConnectionInfoRequest {
                            encrypted_conn_info,
                            pub_id,
                            msg_id,
                        },
                        ..msg
                    });
                    Ok(())
                }
            }
            RoutingMessage {
                content:
                    ConnectionInfoResponse {
                        encrypted_conn_info,
                        pub_id,
                        msg_id,
                    },
                src: ManagedNode(src_name),
                dst: Client { .. },
            } => self.handle_connection_info_response(
                encrypted_conn_info,
                pub_id,
                msg_id,
                src_name,
                msg.dst,
            ),
            RoutingMessage {
                content: Ack(ack, _),
                ..
            } => {
                self.handle_ack_response(ack);
                Ok(())
            }
            _ => {
                self.add_message_to_backlog(msg);
                Ok(())
            }
        }
    }
```

src/states/establishing_node.rs
jeanphilippeD: I'd say if we probably want to keep the pass through behaviour in this PR:
I'd add the msg_queue field to EstablishingNode that is only in intialized in from_... and used in into_...

Then it is more clearly as it was before.

src/states/establishing_node.rs
jeanphilippeD: `pub fn from_proving_node(`.
In my mind it it the public API of Node with respect to the other mod, even if it is not use outside the crate.

src/states/node.rs
madadam: `NodeApproval` is a message sent to `ProvingNode` to inform it that it passed the resource proof. It doesn't make sense to send it to `EstablishingNode` or `Node`, so it's ignored there (with a log message). I don't think it matters much whether we log the error in `EstablishingNode` already, or just backlog it and log it once we are `Node`. It doesn't affect anything and it makes the code simpler. So even though strictly speaking it is a behaviour change, it's a very minor one and I think it's justified by making the code sightly simpler.

jeanphilippeD: Ideally Refactoring commit does not change any behaviour.
With such a big change, it is difficult to know what was intended change in behaviour and unintended ones (difficult enough to detect if there are any change). As a result review need to flag all of them and assume all are unintended.

Best to document all these kind of changes so we know it is intentional, and explain why they are sensible, and keep them to a minimum / separate commit if sensible.

src/states/establishing_node.rs
```diff
@@ -0,0 +1,463 @@
...
+    fn dispatch_routing_message(
+        &mut self,
+        msg: RoutingMessage,
+        outbox: &mut EventBox,
+    ) -> Result<Transition, RoutingError> {
+        self.handle_routing_message(msg, outbox)
+            .map(|()| Transition::Stay)
+    }
+
+    fn handle_parsec_request(
```
jeanphilippeD: Maybe as a follow up commit remove duplication with version in node, and code in parsec.rs.

Ideally the following would not be duplicated.

```rust
parsec.handle_request(&pub_id, request)
            .map(|response| response.into_message(msg_version))
            .map_err(|err| {
                debug!("{} - Error handling parsec request: {:?}", log_ident, err);
                err
            })
            .ok();
```

Same for parsec.handle_response.

Ideally the code calling handle_request followed by parsec_poll_all would also not be duplicated.

jeanphilippeD: I wonder if we would not be better off keeping the member as a ParsecMap, then this issue would not exists, and have a `take_sole_parsec` for the into_...

jeanphilippeD: as it happen, with NodeAgeing, we will probably stay in establishing through multiple Parsec, so having the map may be good?

jeanphilippeD: (If we go back to a map, better to do it in that initial commit,

src/states/establishing_node.rs
```diff
@@ -0,0 +1,463 @@
...
+                self.handle_parsec_response(version, par_response, pub_id, outbox)
+            }
+            _ => {
+                debug!("{} Unhandled direct message: {:?}", self, msg);
+                Ok(Transition::Stay)
+            }
+        }
+    }
+
+    fn handle_hop_message(
```
jeanphilippeD: Looking at node.rs handle_hop_message it is not clear to me how we have that implementation:

```rust
    fn handle_hop_message(
        &mut self,
        hop_msg: HopMessage,
        pub_id: PublicId,
        _: &mut EventBox,
        hop_msg.verify(pub_id.signing_public_key())?;
        let mut client_ip = None;
        let mut hop_name_result = match self.peer_mgr.get_peer(&pub_id).map(Peer::state) {
            Some(&PeerState::Bootstrapper { .. }) => {
                warn!(
                    "{} Hop message received from bootstrapper {:?}, disconnecting.",
                    self, pub_id
                );
                Err(RoutingError::InvalidStateForOperation)
            }
            Some(&PeerState::Client { ip, .. }) => {
                client_ip = Some(ip);
                Ok(*self.name())
            }
            Some(&PeerState::JoiningNode) => Ok(*self.name()),
            Some(&PeerState::Candidate) | Some(&PeerState::Proxy) | Some(&PeerState::Routing) => {
                Ok(*pub_id.name())
            }
            Some(&PeerState::ConnectionInfoPreparing { .. })
            | Some(&PeerState::ConnectionInfoReady(_))
            | Some(&PeerState::CrustConnecting)
            | Some(&PeerState::Connected)
            | None => {
                if self.dropped_clients.contains_key(&pub_id) {
                    debug!(
                        "{} Ignoring {:?} from recently-disconnected client {:?}.",
                        self, hop_msg, pub_id
                    );
                    return Ok(Transition::Stay);
                } else {
                    Ok(*self.name())
                    // FIXME - confirm we can return with an error here by running soak tests
                    // debug!("{} Invalid sender {} of {:?}", self, pub_id, hop_msg);
                    // return Err(RoutingError::InvalidSource);
                }
            }
        };

        if let Some(ip) = client_ip {
            match self.check_valid_client_message(&ip, hop_msg.content.routing_message()) {
                Ok(added_bytes) => {
                    self.proxy_load_amount += added_bytes;
                    self.peer_mgr.add_client_traffic(&pub_id, added_bytes);
                }
                Err(e) => hop_name_result = Err(e),
            }
        }

        match hop_name_result {
            Ok(hop_name) => {
                let HopMessage {
                    content,
                    route,
                    sent_to,
                    ..
                } = hop_msg;
                    .map(|()| Transition::Stay)
            }
            Err(RoutingError::ExceedsRateLimit(hash)) => {
                trace!(
                    "{} Temporarily can't proxy messages from client {:?} (rate-limit hit).",
                    self,
                    pub_id
                );
                self.send_direct_message(
                    pub_id,
                    DirectMessage::ProxyRateLimitExceeded {
                        ack: Ack::compute(hop_msg.content.routing_message())?,
                    },
                );
                Err(RoutingError::ExceedsRateLimit(hash))
            }
            Err(error) => {
                self.ban_and_disconnect_peer(&pub_id);
                Err(error)
            }
        }
    }

```

src/states/establishing_node.rs
```diff
@@ -0,0 +1,463 @@
...
+
+    fn routing_msg_filter(&mut self) -> &mut RoutingMessageFilter {
+        &mut self.routing_msg_filter
+    }
+
+    fn timer(&mut self) -> &mut Timer {
+        &mut self.timer
+    }
+
+    fn send_routing_message_via_route(
```
jeanphilippeD: Similarly, looking at node.rs, not clear how we come to this implmenation.

```rust
    // Constructs a signed message, finds the node responsible for accumulation, and either sends
    // this node a signature or tries to accumulate signatures for this message (on success, the
    // accumulator handles or forwards the message).
    fn send_routing_message_via_route(
        &mut self,
        routing_msg: RoutingMessage,
        route: u8,
        expires_at: Option<Instant>,
        if !self.in_authority(&routing_msg.src) {
            if route == 0 {
                log_or_panic!(
                    LogLevel::Error,
                    "{} Not part of the source authority. Not sending message {:?}.",
                    self,
                    routing_msg
                );
            }
            return Ok(());
        }
        if !self.add_to_pending_acks(&routing_msg, route, expires_at) {
            debug!(
                "{} already received an ack for {:?} - so not resending it.",
                self, routing_msg
            );
            return Ok(());
        }
        use crate::routing_table::Authority::*;
        let sending_sec = match routing_msg.src {
            ClientManager(_) | NaeManager(_) | NodeManager(_) | ManagedNode(_) | Section(_)
                if self.chain.is_member() =>
            {
                Some(self.chain.our_info().clone())
            }
            PrefixSection(ref pfx) if self.chain.is_member() => {
                let src_section = match self.chain.our_info_for_prefix(pfx) {
                    Some(a) => a.clone(),
                    None => {
                        // Can no longer represent sending Pfx.
                        return Ok(());
                    }
                };
                Some(src_section)
            }
            Client { .. } => None,
            _ => {
                // Cannot send routing msgs as a Node until established.
                return Ok(());
            }
        };

        if route > 0 {
            trace!(
                "{} Resending Msg: {:?} via route: {} and src_section: {:?}",
                self,
                routing_msg,
                route,
                sending_sec
            );
        }

        let signed_msg = SignedMessage::new(routing_msg, &self.full_id, sending_sec)?;

        match self.get_signature_target(&signed_msg.routing_message().src, route) {
            None => Ok(()),
            Some(our_name) if our_name == *self.name() => {
                let min_section_size = self.min_section_size();
                if let Some((mut msg, route)) =
                    self.sig_accumulator
                        .add_message(signed_msg, min_section_size, route)
                {
                    if self.in_authority(&msg.routing_message().dst) {
                        self.handle_signed_message(msg, route, our_name, &BTreeSet::new())?;
                    } else {
                        self.send_signed_message(&mut msg, route, &our_name, &BTreeSet::new())?;
                    }
                }
                Ok(())
            }
            Some(target_name) => {
                if let Some(&pub_id) = self.peer_mgr.get_pub_id(&target_name) {
                    let direct_msg = signed_msg
                        .routing_message()
                        .to_signature(self.full_id.signing_private_key())?;
                    self.send_direct_message(pub_id, direct_msg);
                    Ok(())
                } else {
                    Err(RoutingError::RoutingTable(RoutingTableError::NoSuchPeer))
                }
            }
        }
    }

```

src/states/establishing_node.rs
```diff
@@ -0,0 +1,463 @@
...
+    fn handle_offline_event(
+        &mut self,
+        pub_id: PublicId,
+        _: &mut EventBox,
+    ) -> Result<(), RoutingError> {
+        let _ = self.chain.remove_member(pub_id)?;
+        Ok(())
+    }
+
+    fn handle_section_info_event(
```
jeanphilippeD: from node.rs
We use to do quite a few things before, it is not clear to me why they were not necessary, or where they moved.

```rust
    fn handle_section_info_event(
        &mut self,
        sec_info: SectionInfo,
        old_pfx: Prefix<XorName>,
        outbox: &mut EventBox,
        if !self.peer_mgr.is_established() && self.chain.is_member() {
            // We have just found a `SectionInfo` block that contains us. Now we can supply our
            // votes for all latest neighbour infos that have accumulated so far.
            let ni_events = self
                .chain
                .neighbour_infos()
                .map(|ni| ni.clone().into_network_event())
                .collect_vec();

            ni_events.into_iter().for_each(|ni_event| {
                self.vote_for_event(ni_event);
            });

            self.node_established(outbox);
        }

        if sec_info.prefix().is_extension_of(&old_pfx) {
            outbox.send_event(Event::SectionSplit(*sec_info.prefix()));
            self.send_neighbour_infos();
        } else if old_pfx.is_extension_of(sec_info.prefix()) {
            outbox.send_event(Event::SectionMerge(*sec_info.prefix()));
        }

        let our_name = *self.full_id.public_id().name();
        let self_sec_update = sec_info.prefix().matches(&our_name);

        if self.chain.is_member() {
            self.update_peer_states(outbox);

            if self_sec_update {
                self.send_neighbour_infos();
            } else {
                // Vote for neighbour update if we haven't done so already.
                // vote_for_event is expected to only generate a new vote if required.
                self.vote_for_event(sec_info.into_network_event());
            }
        }

        let _ = self.merge_if_necessary();

        Ok(Transition::Stay)
    }
```

src/states/node.rs
```diff
@@ -284,19 +284,57 @@ impl Node {
...
+        debug!("{} - State changed to Node.", self);
+        trace!(
+            "{} - Node Established. Prefixes: {:?}",
+            self,
+            self.chain.prefixes()
+        );
+
+        // We have just become established. Now we can supply our votes for all latest neighbour
+        // infos that have accumulated so far.
+        let neighbour_info_events = self
```
jeanphilippeD: Ha, this is where that moved.

src/states/node.rs
```diff
@@ -888,35 +883,6 @@ impl Node {
             .init(self.full_id.clone(), &self.gen_pfx_info, &log_ident)
     }
 
-    // Completes a Node startup process and allows a node to behave as a `ManagedNode`.
-    // A given node is considered "established" when it exists in `chain.our_info().members()`
-    // first node: this method is skipped entirely as it behaves as a Node from startup.
-    fn node_established(&mut self, outbox: &mut EventBox) {
-        self.peer_mgr.set_established();
-        outbox.send_event(Event::Connected);
```
jeanphilippeD: so this happens in init now.

src/states/node.rs
```diff
@@ -888,35 +883,6 @@ impl Node {
...
-        self.update_peer_states(outbox);
-
-        // Allow other peers to bootstrap via us.
-        if let Err(err) = self.crust_service.set_accept_bootstrap(true) {
-            warn!("{} Unable to accept bootstrap connections. {:?}", self, err);
-        }
-        self.crust_service.set_service_discovery_listen(true);
-
-        let backlog = mem::replace(&mut self.routing_msg_backlog, vec![]);
-        backlog
```
jeanphilippeD: I guess this was made unecessary because we pass in msg_backlog to initialize  msg_queue.
Something to be careful about if keeping 2 queues for Establishing.

src/states/node.rs
```diff
@@ -1141,19 +1107,6 @@ impl Node {
...
         }
 
-        if !self.chain.is_member() {
-            debug!(
-                "{} Client {:?} rejected: We are not an established node yet.",
-                self, pub_id
-            );
-            self.send_direct_message(
-                pub_id,
-                DirectMessage::BootstrapResponse(Err(BootstrapResponseError::NotApproved)),
```
jeanphilippeD: I am not clear where this has moved or what made it unecessary in EstablishingNode?

src/states/node.rs
```diff
@@ -1651,18 +1575,6 @@ impl Node {
             }
         }
 
-        // Until we're established do not relay any messages.
-        let src = signed_msg.routing_message().src;
```
jeanphilippeD: Also not 100% clear how this now work.

src/states/establishing_node.rs
madadam: But I fail to see how that would be functionally different from what is in the PR now. It would only make things more complex and confusing.

What I would consider doing is something like this: In `EstablishingNode::from_proving_node`, I would set `self.msg_backlog` to empty `Vec` initially, and then do:

```rust
for msg in msg_backlog {
    self.dispatch_routing_msg(msg, outbox);
}
```
That way, any message in `msg_backlog` that `ProvingNode` put there and that `EstablishingNode` actually knows how to handle, would be handled. The rest would be moved over to `self.msg_backlog`. Currently there are no such messages that `EstablishingNode` handles but `ProvingNode` does not, but it would make the code more robust in terms of future changes. I'm not convinced it's needed though.

src/states/establishing_node.rs
madadam: OK. This might be worth doing, but this pattern is used in other states too, so best to do it everywhere to maintain consistency.

src/states/establishing_node.rs
jeanphilippeD: > It would only make things more complex and confusing.

jeanphilippeD: > It would only make things more complex and confusing.

Yes, I think I understand things better now.

`for msg in msg_backlog {` yes, i think it may be sensible. That seem to be more like the behaviour we had before this PR.

Also maybe change the name of the parameter to msg_queue.

src/states/establishing_node.rs
madadam: Same issue as https://github.com/maidsafe/routing/pull/1616#discussion_r278593427

src/states/establishing_node.rs
jeanphilippeD: The reason I was suggesting this, is primarily because that it is a much simpler diff if we use ParsecMap everywhere (since then there is no need for duplicate code or de-duplication, the code is there). The added benefit is it also seem to be closer to what we will want.

madadam: OK, I think you might be right. I'll give it a try.

src/states/node.rs
madadam: You are right, this got accidentally lost. I'll put it back.

src/states/node.rs
madadam: For non-established states, routing message sending goes via proxy and is handled by `NotEstablished::send_routing_message_via_proxy`. So the check that the sender is ourselves should have gone there. I'll add it.

src/states/node.rs
```diff
@@ -168,38 +171,26 @@ impl Node {
...
+            source.full_id,
+            source.gen_pfx_info,
             false,
-            msg_queue,
-            notified_nodes,
-            parsec_map,
-            peer_mgr,
-            routing_msg_filter,
-            timer,
+            source.msg_backlog.into_iter().collect(),
```
jeanphilippeD: Kind of lost the benefit here because there is a new function.

So we would need to pass the source to new if possible, otherwise that is the flexibility we lost that we had with all the parameter, that would still be there with pretty much no additional noise with a separate struct (i.e the parameter are just moved to the struct, same amount of code).

I quite like your approach, because it remove the need for the argument to be described again.

However, it create some coupling and force you to make these member public.
I'm thinking it could be cleaner with a separate struct as it would not have these quirks.

src/states/establishing_node.rs
madadam: This isn't resolved yet.

src/states/establishing_node.rs
madadam: This isn't resolved yet.

src/states/node.rs
jeanphilippeD: > Best to document all these kind of changes so we know it is intentional, and explain why they are sensible, and keep them to a minimum / separate commit if sensible.

jeanphilippeD: > Best to document all these kind of changes so we know it is intentional, and explain why they are sensible, and keep them to a minimum / separate commit if sensible.

Maybe add it to the PR description, which probably could be fleshed out a bit.

src/states/establishing_node.rs
jeanphilippeD: Could you comment on that?

jeanphilippeD: never mind, saw the comment bellow where it moved.

src/states/node.rs
jeanphilippeD: This is a follow up on resolution for https://github.com/maidsafe/routing/pull/1616#pullrequestreview-230608692.

Could you comment on that.

madadam: I'm OK to change the code to use separate structs if you think it's worth it. But I'm also OK with keeping it as is, because I don't think the additional coupling is too bad in practice and it makes the code simpler. Your call.

jeanphilippeD: I think it is worth it, it will be cleaner and mirror simply the interface that existed before.
But mostly I'm wanting to fix this `new` function that this comment is on as it push around the problem we were trying to solve (i.e `new` need to take the struct, so names are tied together).

src/states/bootstrapping.rs
```diff
@@ -49,19 +49,18 @@ pub enum TargetState {
 
 // State of Client or Node while bootstrapping.
 pub struct Bootstrapping {
```
pierrechevalier83: This commit (simplify state conversions) is officially my favourite commit of the day so far :smile:

src/states/establishing_node.rs
```diff
@@ -39,52 +38,69 @@ use std::{
...
+}
+
 pub struct EstablishingNode {
-    pub(super) ack_mgr: AckManager,
-    pub(super) cache: Box<Cache>,
-    pub(super) chain: Chain,
-    pub(super) crust_service: Service,
-    pub(super) full_id: FullId,
-    pub(super) gen_pfx_info: GenesisPfxInfo,
+    ack_mgr: AckManager,
```
pierrechevalier83: Wouldn't it make sense to just use composition to avoid repeating all these fields?

src/states/bootstrapping.rs
```diff
@@ -49,15 +54,15 @@ pub enum TargetState {
...
-    pub(super) action_sender: RoutingActionSender,
-    pub(super) cache: Box<Cache>,
-    pub(super) crust_service: Service,
-    pub(super) full_id: FullId,
-    pub(super) min_section_size: usize,
-    pub(super) timer: Timer,
+    action_sender: RoutingActionSender,
```
pierrechevalier83: Wouldn't it make sense to just use composition to avoid repeating all these fields?

src/states/proving_node.rs
```diff
@@ -39,62 +39,71 @@ use std::{
...
+    pub full_id: FullId,
+    pub min_section_size: usize,
+    pub old_full_id: FullId,
+    pub our_section: (Prefix<XorName>, BTreeSet<PublicId>),
+    pub proxy_pub_id: PublicId,
+    pub timer: Timer,
+}
+
 pub struct ProvingNode {
-    pub(super) ack_mgr: AckManager,
```
pierrechevalier83: Wouldn't it make sense to just use composition to avoid repeating all these fields?

src/states/relocating_node.rs
```diff
@@ -39,6 +39,16 @@ use std::{
...
+    pub cache: Box<Cache>,
+    pub crust_service: Service,
+    pub full_id: FullId,
+    pub min_section_size: usize,
+    pub proxy_pub_id: PublicId,
+    pub timer: Timer,
+}
+
 pub struct RelocatingNode {
     action_sender: RoutingActionSender,
```
pierrechevalier83: Wouldn't it make sense to just use composition to avoid repeating all these fields?

src/states/node.rs
```diff
@@ -77,6 +74,21 @@ const CLIENT_BAN_DURATION: Duration = Duration::from_secs(2 * 60 * 60);
...
+    pub msg_queue: VecDeque<RoutingMessage>,
+    pub notified_nodes: BTreeSet<PublicId>,
+    pub parsec_map: ParsecMap,
+    pub peer_mgr: PeerManager,
+    pub routing_msg_filter: RoutingMessageFilter,
+    pub timer: Timer,
+}
+
 pub struct Node {
     ack_mgr: AckManager,
```
pierrechevalier83: Wouldn't it make sense to just use composition to avoid repeating all these fields?

src/states/client.rs
```diff
@@ -37,6 +34,15 @@ use std::{
...
+    pub full_id: FullId,
+    pub min_section_size: usize,
+    pub msg_expiry_dur: Duration,
+    pub proxy_pub_id: PublicId,
+    pub timer: Timer,
+}
+
 /// A node connecting a user to the network, as opposed to a routing / data storage node.
 ///
 /// Each client has a _proxy_: a node through which all requests are routed.
```
pierrechevalier83: Wouldn't it make sense to just use composition to avoid repeating all these fields?

src/state_machine.rs
```diff
@@ -193,49 +193,58 @@ impl State {
 
     pub fn get_timed_out_tokens(&mut self) -> Vec<u64> {
         match *self {
-            State::Node(ref mut state) => state.get_timed_out_tokens(),
+            State::Bootstrapping(_) | State::Terminated => vec![],
             State::Client(ref mut state) => state.get_timed_out_tokens(),
```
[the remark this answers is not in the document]
madadam: No because all the `state`s are actually different types. ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6249b0de048643e437f73096f10909c0))

src/state_machine.rs
```diff
@@ -193,49 +193,58 @@ impl State {
...
+        } else {
+            false
         }
     }
 
     pub fn in_authority(&self, auth: &Authority<XorName>) -> bool {
         match *self {
-            State::Node(ref state) => state.in_authority(auth),
+            State::Terminated | State::Bootstrapping(_) => false,
             State::Client(ref state) => state.in_authority(auth),
```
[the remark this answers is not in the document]
madadam: Same answer :)

src/states/mod.rs
```diff
@@ -38,8 +40,30 @@ pub use self::{
...
+//        │                        │
+//        │                        │
 //        ▼                        ▼
 // ┌────────┐                    ┌──────┐
 // │ Client │                    │ Node │
 // └────────┘                    └──────┘
 //
+//
+// # Common traits
+//                           Bootstrapping
```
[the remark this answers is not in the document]
madadam: `Bootstrapping` can become `Client` or `Node`, but the other states are all going to become only `Node`. I think it makes some sense to have this distinction. On the other hand I don't feel too strongly about it and having shorter names would be nice too. Your call.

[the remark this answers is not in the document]
madadam: Hm, so I wonder: I can remove the `Node` suffix, but that would still make the `Client` and `Node` names inconsistent with the rest, as the rest are verbs in the present continuous tense (`Bootstrapping`, `Relocating`, `Proving`, ...), but `Client` and `Node` are nouns. I can maybe go the opposite way and rename `Node` and `Client` to `EstablishedNode` and `EstablishedClient`, but that would still leave `Bootstrapping` out as it can be both node and client. Not sure. Naming stuff is hard. Maybe best to leave this as is for now and maybe return to it in the future?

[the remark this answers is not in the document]
Viv-Rajkumar: you could switch `Bootstrapping` -> `BootstrappingPeer`. we normally go with `Peer` when common to Node & Client but not like it was enforced.

src/states/establishing_node.rs
madadam: After spending some time on this I'm now about 90% convinced this is actually correct. Although it's a logic change I believe the way this was handled before was actually somehow buggy (although in a harmless way) because until a node is established, it isn't supposed to handle these things. That's why I'd prefer to keep things as they are now. Worst case, this will pop up in the soak testing at which point I'll change it back. I'm going to at least mention this in the PR description.

src/states/establishing_node.rs
madadam: See [here](https://github.com/maidsafe/routing/pull/1616#discussion_r281622247).

src/mock_parsec/mod.rs
```diff
@@ -184,10 +184,21 @@ where
         unimplemented!()
     }
 
+    // TODO: rename this to `has_unpolled_observations`
```
pierrechevalier83: This isn't done yet because you would want to update Parsec to one that exposes the new API in the same PR?
I would add this to the comment.

madadam: Yes, I'd like to make the change in Parsec first, to avoid the situation where routing becomes uncompilable with `--features=mock_base` (that is, with real parsec).

madadam: Yes, I'd like to make the change in Parsec first, to avoid the situation where routing becomes uncompilable with `--features=mock_base` (that is, with real parsec).

PR #1617, opened 2019-04-26
mock serialise: 4x speed up aggressive churn

jeanphilippeD:
Not serialize when sending messages.
Not serialising full content when verifying HopMessage

mock_crust::churn::aggressive_churn
seed: Some([0,1,2,4]):
Before: 100 seconds  (features=mock_parsec)
After:   23 seconds  (features=mock_serialise_send)

Note:
- The big remaining serialisation item is SignedMessage::check_integrity, we may remove it as part of N/3 message passing. Removing it aggressive_churn is 13.5 seconds.
- After that, what remains is in mock parsec, and the Message clone used to mock serialization.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ serde_json = "~1.0.8"
 [features]
 mock = ["lru_time_cache/fake_clock", "safe_crypto/mock", "parsec/mock", "parsec/malice-detection"]
 mock_parsec = ["mock"]
+mock_serialise = ["mock"]
 
 [[example]]
 bench = false
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -51,6 +51,7 @@ pub const CLIENT_GET_PRIORITY: u8 = 3;
 /// Wrapper of all messages.
 ///
 /// This is the only type allowed to be sent / received on the network.
+#[cfg_attr(feature = "mock_serialise", derive(Clone))]
 #[derive(Debug, Serialize, Deserialize)]
 // FIXME - See https://maidsafe.atlassian.net/browse/MAID-2026 for info on removing this exclusion.
 #[allow(clippy::large_enum_variant)]
@@ -73,6 +74,7 @@ impl Message {
 /// Messages sent via a direct connection.
 ///
 /// Allows routing to directly send specific messages between nodes.
+#[cfg_attr(feature = "mock_serialise", derive(Clone))]
 #[derive(Serialize, Deserialize)]
 // FIXME - See https://maidsafe.atlassian.net/browse/MAID-2026 for info on removing this exclusion.
 #[allow(clippy::large_enum_variant)]
@@ -153,6 +155,7 @@ impl DirectMessage {
 /// To relay a `SignedMessage` via another node, the `SignedMessage` is wrapped in a `HopMessage`.
 /// The `signature` is from the node that sends this directly to a node in its routing table. To
 /// prevent Man-in-the-middle attacks, the `content` is signed by the original sender.
+#[cfg_attr(feature = "mock_serialise", derive(Clone))]
 #[derive(Serialize, Deserialize)]
 pub struct HopMessage {
     /// Wrapped signed message.
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -178,7 +178,7 @@ impl HopMessage {
         sent_to: BTreeSet<XorName>,
         signing_key: &SecretSignKey,
     ) -> Result<HopMessage> {
-        let bytes_to_sign = serialise(&content)?;
+        let bytes_to_sign = serialise(HopMessage::content_to_serialise(&content))?;
         Ok(HopMessage {
             content: content,
             route: route,
@@ -192,12 +192,22 @@ impl HopMessage {
     /// This does not imply that the message came from a known node. That requires a check against
     /// the routing table to identify the name associated with the `verification_key`.
     pub fn verify(&self, verification_key: &PublicSignKey) -> Result<()> {
-        let signed_bytes = serialise(&self.content)?;
+        let signed_bytes = serialise(HopMessage::content_to_serialise(&self.content))?;
         if verification_key.verify_detached(&self.signature, &signed_bytes) {
             Ok(())
         } else {
             Err(RoutingError::FailedSignature)
         }
+    }
+
+    #[cfg(not(feature = "mock_serialise"))]
+    fn content_to_serialise(content: &SignedMessage) -> &SignedMessage {
+        content
+    }
+
+    #[cfg(feature = "mock_serialise")]
+    fn content_to_serialise(_content: &SignedMessage) -> &[u8; 18] {
+        b"HopMessage.content"
     }
 }
 
--- a/src/states/bootstrapping.rs
+++ b/src/states/bootstrapping.rs
@@ -17,9 +17,11 @@ use crate::messages::{DirectMessage, Message};
 use crate::outbox::EventBox;
 use crate::routing_table::{Authority, Prefix};
 use crate::state_machine::{State, Transition};
+use crate::states::common::from_crust_bytes;
 use crate::timer::Timer;
 use crate::types::RoutingActionSender;
 use crate::xor_name::XorName;
+use crate::CrustBytes;
 use crate::{CrustEvent, Service};
 use maidsafe_utilities::serialisation;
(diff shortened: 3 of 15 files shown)


src/messages/mod.rs
```diff
@@ -189,13 +189,23 @@ impl HopMessage {
...
             Err(RoutingError::FailedSignature)
         }
     }
+
+    fn serialise_content(_content: &SignedMessage) -> Result<Vec<u8>> {
+        #[cfg(not(feature = "mock_serialise_send"))]
+        let result = serialise(_content)?;
+
+        #[cfg(feature = "mock_serialise_send")]
+        let result = b"HopMessage.content".to_vec();
```
afck: Even with mock serialization and mock cryptography, wouldn't we still want to distinguish between signatures of different messages? (But maybe not; the huge speedup may be worth it.)

src/states/bootstrapping.rs
```diff
@@ -264,14 +265,18 @@ impl Bootstrapping {
...
     }
 
-    fn handle_new_message(&mut self, pub_id: PublicId, bytes: Vec<u8>) -> Result<Transition> {
-        match serialisation::deserialise(&bytes) {
+    fn handle_new_message(
+        &mut self,
+        pub_id: PublicId,
+        bytes: CrustSendReceiveData,
+    ) -> Result<Transition> {
+        match self.from_curst_send_receive_data(bytes) {
```
afck: Typo: `from_crust_send_receive_data`
(Also: We could apply `?` here instead of matching `Err`.)

src/states/bootstrapping.rs
```diff
@@ -370,7 +375,7 @@ impl Display for Bootstrapping {
     }
 }
 
-#[cfg(all(test, feature = "mock"))]
+#[cfg(all(test, feature = "mock", not(feature = "mock_serialise_send")))]
```
pierrechevalier83: Can you explain why not test the bootstrapping when the mock_serialise_send feature is passed?

src/states/common/base.rs
```diff
@@ -62,15 +65,73 @@ pub trait Base: Display {
...
+        if let Err(err) = self.crust_service().send(dst_id, data, priority) {
             info!("{} Connection to {} failed: {:?}", self, dst_id, err);
             // TODO: Handle lost peer, but avoid a cascade of sending messages and handling more
             //       lost peers: https://maidsafe.atlassian.net/browse/MAID-1924
             // self.crust_service().disconnect(*pub_id);
             // return self.handle_lost_peer(*pub_id).map(|_| Err(err.into()));
         }
     }
+
+    fn to_curst_send_receive_data(
```
pierrechevalier83: s/curst/crust

src/states/common/base.rs
```diff
@@ -62,15 +65,73 @@ pub trait Base: Display {
...
+        &self,
+        message: Message,
+    ) -> Result<CrustSendReceiveData, (serialisation::SerialisationError, Message)> {
+        to_curst_send_receive_data(message)
+    }
+
+    fn from_curst_send_receive_data(
```
pierrechevalier83: s/curst/crust/

src/states/common/base.rs
```diff
@@ -62,15 +65,73 @@ pub trait Base: Display {
...
+        )?;
+        let message = Message::Hop(hop_msg);
+        Ok(self
+            .to_curst_send_receive_data(message)
+            .map_err(|(err, _)| err)?)
+    }
+}
+
+#[cfg(not(feature = "mock_serialise_send"))]
+fn to_curst_send_receive_data(
```
pierrechevalier83: s/curst/crust

src/states/common/base.rs
```diff
@@ -62,15 +65,73 @@ pub trait Base: Display {
...
+    message: Message,
+) -> Result<CrustSendReceiveData, (serialisation::SerialisationError, Message)> {
+    let result = serialisation::serialise(&message);
+    result.map_err(|err| (err, message))
+}
+
+#[cfg(not(feature = "mock_serialise_send"))]
+fn from_curst_send_receive_data(data: CrustSendReceiveData) -> Result<Message, RoutingError> {
```
pierrechevalier83: s/curst/crust

src/states/common/base.rs
```diff
@@ -62,15 +65,73 @@ pub trait Base: Display {
...
+    serialisation::deserialise(&data).map_err(RoutingError::SerialisationError)
+}
+
+#[cfg(feature = "mock_serialise_send")]
+fn to_curst_send_receive_data(
```
pierrechevalier83: s/curst/crust

src/messages/mod.rs
jeanphilippeD: If there is a bug like that, I assume test without mock_serialisation_send would catch it super quick (probably in CI). I think it serve a different purpose, so I do not believe this will be an issue.

src/states/common/base.rs
```diff
@@ -62,15 +65,73 @@ pub trait Base: Display {
...
+    message: Message,
+) -> Result<CrustSendReceiveData, (serialisation::SerialisationError, Message)> {
+    Ok(message)
+}
+
+#[cfg(feature = "mock_serialise_send")]
+fn from_curst_send_receive_data(data: CrustSendReceiveData) -> Result<Message, RoutingError> {
```
pierrechevalier83: s/curst/crust (Don't mean to be noisy, but I thought it may help to have all instances pointed out for checking after you've renamed them all)

src/states/bootstrapping.rs
jeanphilippeD: It is using deserializing, and modifying types that are changed by that feature.
It is possible to make this test work, it did not seem so important, but I can fix it before we decide to merge (wanted early feedback while Viv is not on holiday).

Cargo.toml
```diff
@@ -40,6 +40,7 @@ serde_json = "~1.0.8"
 [features]
 mock = ["lru_time_cache/fake_clock", "safe_crypto/mock", "parsec/mock", "parsec/malice-detection"]
 mock_parsec = ["mock"]
+mock_serialise_send = ["mock_parsec"]
```
Viv-Rajkumar: we prolly also want to update https://github.com/maidsafe/routing/blob/fleming/scripts/tests#L7-L8 with this PR

jeanphilippeD: We can, I was not sure if we wanted to change the CI.

Option 1: keep it as is (not so fast not so slow, test serialization)
Option 2: test both (slowest option: need to build twice and run twice).
Option 3: only test with no serialization: then missing serialization coverage.

That is why I left it as is as I thought it may be only used for soak testing initially.

Viv-Rajkumar: hmm its a fair point, sure we can stick to option-1 I guess for now until CI time becomes an issue in this repo or due to this repo for other shared runners. Ideally yeh would be good for this to be in opt-3 and go with opt-2 for something like nightly/weekly CI

jeanphilippeD:
> Conceptually, looks good to me.
> I'm not sure I'm a fan of the names: `CrustSendReceiveData`, `from_crust_send_receive_data` and `to_crust_send_receive_data`...
> Maybe something like `SerializedData`, `deserialize` and `serialize` would read better everywhere IMO. Not every function has to care that btw, in `mock_serialise_send` (should this be `mock_serialization` ?) we actually do something different.

> I agree with Pierre's comment, with the exception that I'd suggest maybe CrustBytes instead of SerializedData as it's slightly shorter, and deserialise/serialise with UK spelling for consistency.

I'd rather avoid serialise/deserialise because it seem an implementation detail since we abstract it to add a mock, but this abstraction is reasonable..
I'm ok with `CrustBytes`, so what about to_crust_bytes()/from_crust_bytes().
Or 'CrustData' and to_crust_data()/from_crust_data().

I'm also coming around to `mock_serialisation` or `mock_serialise`, even if we only mock some part of our serialisation

Cargo.toml
```diff
@@ -40,6 +40,7 @@ serde_json = "~1.0.8"
 [features]
 mock = ["lru_time_cache/fake_clock", "safe_crypto/mock", "parsec/mock", "parsec/malice-detection"]
```
pierrechevalier83: I don't think this is what we had in mind... Let's grab a quick chat.

pierrechevalier83: After quick chat, further change will make it into a different PR. Good to go :)

PR #1618, opened 2019-04-29
chore/routing: change 'mock' to mock all, and rename to `mock_base`.

jeanphilippeD:
Rename `mock` to `mock_base`.
Re-add `mock` feature to mean mock all: mock_parsec,mock_serialise.
For fastest soak testing use features=mock.

Test:
time RUSTFLAGS=-g cargo test --release --features=mock
 => 1 min

time RUSTFLAGS=-g cargo test --release --features=mock_serialise --
--skip aggres --skip messages_during_churn
 => 22 min

mock_crust::churn::aggressive_churn seed: Some([0,1,2,4])
 => Killed after 2 hours as it was still running.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,9 +38,10 @@ libc = "~0.2.29"
 serde_json = "~1.0.8"
 
 [features]
-mock = ["lru_time_cache/fake_clock", "safe_crypto/mock", "parsec/mock", "parsec/malice-detection"]
-mock_parsec = ["mock"]
-mock_serialise = ["mock"]
+mock_base = ["lru_time_cache/fake_clock", "safe_crypto/mock", "parsec/mock", "parsec/malice-detection"]
+mock_parsec = ["mock_base"]
+mock_serialise = ["mock_base"]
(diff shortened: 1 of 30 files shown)


Cargo.toml
```diff
@@ -38,9 +38,10 @@ libc = "~0.2.29"
 serde_json = "~1.0.8"
 
 [features]
-mock = ["lru_time_cache/fake_clock", "safe_crypto/mock", "parsec/mock", "parsec/malice-detection"]
-mock_parsec = ["mock"]
-mock_serialise = ["mock"]
+mock_base = ["lru_time_cache/fake_clock", "safe_crypto/mock", "parsec/mock", "parsec/malice-detection"]
```
madadam: Why is malice-detection enabled in mock, but not otherwise?

Viv-Rajkumar: I think that predates this PR but we just havent set that to be a default feature yet, but should be fine to set now I guess or expect parsec to have that enabled by disabled and feature gate to disable once stable.

Merging this for now though and can take this hopefully to a different PR

PR #1619, opened 2019-04-29
chore/routing: update lock file in CI.

jeanphilippeD:
Update travis and appveyor to run cargo update before running clippy
or tests so depedencies are updated when they change without a
cargo.toml change.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,6 +24,7 @@ configuration:
   - Release
 
 build_script:
+  - cargo update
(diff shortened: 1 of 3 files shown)


Viv-Rajkumar:
Closing for now as this wasnt the source of the issue. rust_sodium-sys 0.10.4 isnt getting resolved due to conflicting deps for lazy_static. That potentially needs sorted from safe_crypto(just needs published)->parsec->this crate.

PR #1621, opened 2019-05-01
make churn tests more aggressive and minor refactoring

Viv-Rajkumar:
- refactor: use is_opaque from parsec
- make messages_during_churn match new node acceptance threshold match aggressive_churn
- randomise proxy and inserted indexes for new nodes to highlight any sequential poll issues with mock_base.

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -167,13 +167,9 @@ impl ParsecMap {
         };
 
         if filter_opaque {
-            parsec.our_unpolled_observations().any(|obs| {
-                if let Observation::OpaquePayload(_) = obs {
-                    true
-                } else {
-                    false
-                }
-            })
+            parsec
+                .our_unpolled_observations()
+                .any(parsec::Observation::is_opaque)
         } else {
             parsec.has_unconsensused_observations()
         }
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -169,7 +169,7 @@ impl ParsecMap {
         if filter_opaque {
             parsec
                 .our_unpolled_observations()
-                .any(parsec::Observation::is_opaque)
+                .any(Observation::is_opaque)
         } else {
             parsec.has_unconsensused_observations()
         }
--- a/tests/mock_crust/churn.rs
+++ b/tests/mock_crust/churn.rs
@@ -107,10 +107,10 @@ fn add_nodes<R: Rng>(
         }
     }
 
-    added_nodes.drain(..).for_each(|added_node| {
+    for added_node in added_nodes {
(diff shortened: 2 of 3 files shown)


tests/mock_crust/churn.rs
```diff
@@ -90,24 +85,61 @@ fn add_nodes_and_poll<R: Rng>(
...
+            if skip_some_prefixes && !rng.gen_weighted_bool(prefixes.len() as u32) {
                 continue;
             }
-            nodes.push(node);
+            added_nodes.push(node);
         }
     }
 
-    poll_and_resend(&mut nodes, &mut []);
+    added_nodes.drain(..).for_each(|added_node| {
```
madadam: Although this does what it's supposed to do, it feels convoluted. Why not just:
```rust
for added_node in added_nodes {
    ...
}
```
?

tests/mock_crust/churn.rs
```diff
@@ -90,24 +85,61 @@ fn add_nodes_and_poll<R: Rng>(
...
+        .iter()
+        .enumerate()
+        .filter_map(|(index, node)| {
+            if !node.inner.is_node() {
+                Some(index)
+            } else {
+                None
+            }
+        })
+        .collect::<BTreeSet<_>>()
```
madadam: No need to spell the type to `collect` to as the compiler can infer it here.

tests/mock_crust/churn.rs
```diff
@@ -90,24 +85,61 @@ fn add_nodes_and_poll<R: Rng>(
...
+}
 
+/// Checks if the given indices have been accepted to the network.
+/// allow_add_failure: Allows nodes to fail getting accepted.
+fn check_added_indices<R: Rng>(
```
madadam: The name of this function is misleading: it does much more than what the name says. It:
1. Checks all new nodes were successfully added
2. Removes failed nodes
3. Clears relocation overrides
4. Polls the nodes
5. Shuffles the nodes

So either it should be given more suitable name or maybe better be split into several functions each with clear responsibility.

src/parsec.rs
```diff
@@ -167,13 +167,9 @@ impl ParsecMap {
...
-            parsec.our_unpolled_observations().any(|obs| {
-                if let Observation::OpaquePayload(_) = obs {
-                    true
-                } else {
-                    false
-                }
-            })
+            parsec
+                .our_unpolled_observations()
+                .any(parsec::Observation::is_opaque)
```
madadam: No need to use the qualified name. `Observation` is in scope here.

PR #1622, opened 2019-05-01
Replace AcceptAsCandidate RPC

jeanphilippeD:
Add:
-ExpectCandidate struct from MessageContent::ExpectCandidate.
-NetworkEvent::ExpectCandidate to consensus the RPC and the
 RPC destination used as the source of the response.

Remove:
-MessageContent::AcceptAsCandidate replacing its function with consensus
 through Parsec of NetworkEvent::ExpectCandidate.
-Candidate::Expecting that was used when waiting AcceptAsCandidate.
-CANDIDATE_ACCEPT_TIMEOUT_SECS

Replace:
-expect_candidate with has_candidate: Do not raise an error in handling
 the parsec event because we already have a candidate.

Update handling ExpectCandidate:
-handle_expect_candidate only vote for the received RPC.
-new handle_expect_candidate_event take on combined responsabilities
 of handle_expect_candidate and handle_expect_candidate_event.
-only processed events if member of chain or established to be allowed
 to vote.

### DIFF
--- a/src/chain/network_event.rs
+++ b/src/chain/network_event.rs
@@ -10,10 +10,23 @@ use super::{ProofSet, ProvingSection, SectionInfo};
 use crate::id::PublicId;
 use crate::parsec;
 use crate::sha3::Digest256;
+use crate::types::MessageId;
 use crate::{Authority, RoutingError, XorName};
 use hex_fmt::HexFmt;
 use maidsafe_utilities::serialisation::serialise;
 use std::fmt::{self, Debug, Formatter};
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
+pub struct ExpectCandidatePayload {
+    /// The joining node's current public ID.
+    pub old_public_id: PublicId,
+    /// The joining node's current authority.
+    pub old_client_auth: Authority<XorName>,
+    /// The message's unique identifier.
+    pub message_id: MessageId,
+    // The routing_msg.dst
+    pub dst_name: XorName,
+}
 
 /// Routing Network events
 // TODO: Box `SectionInfo`?
@@ -25,6 +38,10 @@ pub enum NetworkEvent {
     OurMerge,
     NeighbourMerge(Digest256),
     SectionInfo(SectionInfo),
+
+    /// Voted for received ExpectCandidate RPC.
+    ExpectCandidate(ExpectCandidatePayload),
+
     /// A list of proofs for a neighbour section, starting from the current section.
     ProvingSections(Vec<ProvingSection>, SectionInfo),
 }
@@ -76,6 +93,7 @@ impl Debug for NetworkEvent {
             NetworkEvent::SectionInfo(ref sec_info) => {
                 write!(formatter, "SectionInfo({:?})", sec_info)
             }
+            NetworkEvent::ExpectCandidate(ref vote) => write!(formatter, "SectionInfo({:?})", vote),
             NetworkEvent::ProvingSections(_, ref sec_info) => {
                 write!(formatter, "ProvingSections(_, {:?})", sec_info)
             }
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -535,8 +535,8 @@ impl RoutingMessage {
 /// Once in `JoiningNode` state, A sends a `Relocate` request to the `NaeManager` section authority
 /// X of A's current name. X computes a target destination Y to which A should relocate and sends
 /// that section's `NaeManager`s an `ExpectCandidate` containing A's current public ID. Each member
-/// of Y caches A's public ID, and sends `AcceptAsCandidate` to self section. Once Y receives
-/// `AcceptAsCandidate`, sends a `RelocateResponse` back to A, which includes an address space range
+/// of Y votes for `ExpectCandidate`. Once Y accumulates votes for `ExpectCandidate`, send a
+/// `RelocateResponse` back to A, which includes an address space range
 /// into which A should relocate and also the public IDs of the members of Y. A then disconnects
 /// from the network and reconnects with a new ID which falls within the specified address range.
 /// After connecting to the members of Y, it begins the resource proof process. Upon successful
@@ -639,19 +639,6 @@ pub enum MessageContent {
         cacheable: bool,
         /// The `part_index`-th part of the serialised user message.
         payload: Vec<u8>,
-    },
-    /// Confirm with section that the candidate is about to resource prove.
-    ///
-    /// Sent from the `NaeManager` to the `NaeManager`.
-    AcceptAsCandidate {
-        /// The joining node's current public ID.
-        old_public_id: PublicId,
-        /// The joining node's current authority.
-        old_client_auth: Authority<XorName>,
-        /// The interval into which the joining node should join.
-        target_interval: (XorName, XorName),
-        /// The message's unique identifier.
-        message_id: MessageId,
     },
     /// Approves the joining node as a routing node.
     ///
@@ -807,16 +794,6 @@ impl Debug for MessageContent {
                 hash[0],
                 hash[1],
                 hash[2]
-            ),
-            AcceptAsCandidate {
-                ref old_public_id,
-                ref old_client_auth,
-                ref target_interval,
-                ref message_id,
-            } => write!(
-                formatter,
-                "AcceptAsCandidate {{ {:?}, {:?}, {:?}, {:?} }}",
-                old_public_id, old_client_auth, target_interval, message_id
             ),
             NodeApproval(ref gen_info) => write!(formatter, "NodeApproval {{ {:?} }}", gen_info),
         }
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -31,15 +31,12 @@ const CONNECTING_PEER_TIMEOUT_SECS: u64 = 150;
 /// Time (in seconds) the node waits for a peer to either become valid once connected to it or to
 /// transition once bootstrapped to it.
 const CONNECTED_PEER_TIMEOUT_SECS: u64 = 120;
-/// Time (in seconds) after which a `VotedFor` candidate will be removed.
-const CANDIDATE_ACCEPT_TIMEOUT_SECS: u64 = 120;
 
 #[cfg(feature = "mock_base")]
 #[doc(hidden)]
 pub mod test_consts {
     pub const ACCUMULATION_TIMEOUT_SECS: u64 = super::ACCUMULATION_TIMEOUT.as_secs();
     pub const ACK_TIMEOUT_SECS: u64 = crate::ack_manager::ACK_TIMEOUT.as_secs();
-    pub const CANDIDATE_ACCEPT_TIMEOUT_SECS: u64 = super::CANDIDATE_ACCEPT_TIMEOUT_SECS;
     pub const RESOURCE_PROOF_DURATION_SECS: u64 = super::RESOURCE_PROOF_DURATION.as_secs();
     pub const CONNECTING_PEER_TIMEOUT_SECS: u64 = super::CONNECTING_PEER_TIMEOUT_SECS;
     pub const CONNECTED_PEER_TIMEOUT_SECS: u64 = super::CONNECTED_PEER_TIMEOUT_SECS;
@@ -261,10 +258,6 @@ impl Peer {
 #[derive(Debug, Eq, PartialEq)]
 enum Candidate {
     None,
-    Expecting {
-        timestamp: Instant,
-        old_pub_id: PublicId,
-    },
     AcceptedForResourceProof {
         res_proof_start: Instant,
         target_interval: (XorName, XorName),
@@ -283,9 +276,6 @@ impl Candidate {
     fn is_expired(&self) -> bool {
         match *self {
             Candidate::None => false,
-            Candidate::Expecting { ref timestamp, .. } => {
-                timestamp.elapsed() > Duration::from_secs(CANDIDATE_ACCEPT_TIMEOUT_SECS)
-            }
             Candidate::AcceptedForResourceProof {
                 res_proof_start, ..
             }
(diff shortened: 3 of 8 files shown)


src/states/node.rs
```diff
@@ -1877,98 +1852,117 @@ impl Node {
...
         &mut self,
-        old_pub_id: PublicId,
-        old_client_auth: Authority<XorName>,
-        relocation_dst: Authority<XorName>,
-        message_id: MessageId,
+        rpc: ExpectCandidate,
+        dst_name: XorName,
     ) -> Result<(), RoutingError> {
-        self.remove_expired_peers();
+        if !self.chain.is_member() || !self.peer_mgr.is_established() {
```
Viv-Rajkumar: why are we needing both checks than just `chain.is_member()` here?

src/states/node.rs
```diff
@@ -1877,98 +1852,117 @@ impl Node {
...
-        if min_len_prefix != *self.our_prefix() && !forbid_join_balancing {
-            let request_content = MessageContent::ExpectCandidate {
-                old_public_id: old_pub_id,
-                old_client_auth: old_client_auth,
-                message_id: message_id,
-            };
-            let src = relocation_dst;
-            let dst = Authority::Section(min_len_prefix.substituted_in(relocation_dst.name()));
-            return self.send_routing_message(src, dst, request_content);
+        if let Some(target_interval) = self.accept_candidate_with_interval(&vote) {
```
Viv-Rajkumar: considering `accept_candidate_with_interval` is a single use method and its just 10lines and this one is also abt 15lines, I'd personally would just inline it, but I'm ok if this layout helps you guys more.

src/states/node.rs
```diff
@@ -1877,98 +1852,117 @@ impl Node {
...
+    }
+
+    // Handles an accumulated `ExpectCandidate` event.
+    // Context: a node is joining our section. Sends the node our section. If the
+    // network is unbalanced, sends `ExpectCandidate` on to a section with a shorter prefix.
+    fn handle_expect_candidate_event(
+        &mut self,
+        vote: ExpectCandidateVote,
+    ) -> Result<(), RoutingError> {
+        if !self.chain.is_member() || !self.peer_mgr.is_established() {
```
Viv-Rajkumar: same as prev comment

src/chain/network_event.rs
```diff
@@ -10,11 +10,37 @@ use super::{ProofSet, ProvingSection, SectionInfo};
...
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
+pub struct ExpectCandidateVote {
+    // The routing_msg.content
+    pub rpc: ExpectCandidate,
+    // The routing_msg.dst
+    pub dst_name: XorName,
+}
+
+impl ExpectCandidateVote {
```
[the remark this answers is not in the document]
jeanphilippeD: It makes passing it around a lot easier though, and I do pass it around.

src/chain/network_event.rs
```diff
@@ -10,11 +10,37 @@ use super::{ProofSet, ProvingSection, SectionInfo};
...
 use crate::parsec;
 use crate::sha3::Digest256;
+use crate::types::MessageId;
 use crate::{Authority, RoutingError, XorName};
 use hex_fmt::HexFmt;
 use maidsafe_utilities::serialisation::serialise;
 use std::fmt::{self, Debug, Formatter};
 
+#[derive(Debug, Ord, PartialOrd, Eq, PartialEq, Clone, Hash, Serialize, Deserialize)]
+pub struct ExpectCandidate {
```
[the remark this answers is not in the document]
jeanphilippeD: Can you be more specific?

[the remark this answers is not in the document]
jeanphilippeD: Can you be more specific?
The reason I put it there is we were importing NetworkEvent in node.rs

src/states/node.rs
jeanphilippeD: On my first attempt, tests were still failing, but trying it again, it seem to be enough.
Will fix.

src/states/node.rs
jeanphilippeD: This is more in line with single level of abstraction principle, it does an awful lot in this 15 lines.
This allow to highlight all that is important in this function.

src/chain/network_event.rs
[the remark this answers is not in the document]
Viv-Rajkumar: my point was more abt its usage from messages/mod but I guess its redundant considering thats what we're doing for other types like GenPfxinfo and the likes too. so yeh I'd be fine with this as is.

src/chain/network_event.rs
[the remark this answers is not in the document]
Viv-Rajkumar: yeh its fair play if we pass it around / you guys prefer it, but the name still is kind of weird considering its NetworkEvent we vote for in parsec from routing and this is a type within that enum and also called `..Vote`.

In this case the impl is also sorta weird to me cos we only use `into_network_event` I think once and if we constructed it directly, could have saved this entire impl definition. I'm prolly just biased to more primitive types that way instead of wrapper code.

jeanphilippeD:
> just confirming if it was intentional in this PR.

No, it was not intentional. However, as you say, it is ok, though not unlikely to not accumulate if too many nodes in the new prefix did not receive the RPC.
I think it make sense to not add special handling in this PR.

src/states/node.rs
Viv-Rajkumar: yeh but making code traversal annoying juggling between multiple methods. RLS helps with that mostly but sure I'd go with what majority are comfortable with.

Viv-Rajkumar:
> > just confirming if it was intentional in this PR.
> 
> No, it was not intentional. However, as you say, it is ok, though not unlikely to not accumulate if too many nodes in the new prefix did not receive the RPC.
> I think it make sense to not add special handling in this PR.

hmm I wouldn't see this as special case handling normally. If anything what we have with https://github.com/maidsafe/routing/blob/fleming/src/states/node.rs#L934 is not great thus allowing this to have not get picked up compile time. If we matched there explicitly, introducing a new `NetworkEvent` would have forced us to make the choice explicitly here.

src/messages/mod.rs
```diff
@@ -535,8 +535,8 @@ impl RoutingMessage {
 /// Once in `JoiningNode` state, A sends a `Relocate` request to the `NaeManager` section authority
 /// X of A's current name. X computes a target destination Y to which A should relocate and sends
 /// that section's `NaeManager`s an `ExpectCandidate` containing A's current public ID. Each member
-/// of Y caches A's public ID, and sends `AcceptAsCandidate` to self section. Once Y receives
-/// `AcceptAsCandidate`, sends a `RelocateResponse` back to A, which includes an address space range
+/// of Y vote for `ExpectCandidate`. Once Y accumulate vote for `ExpectCandidate`, sends a
```
pierrechevalier83: typos:
* s/vote/votes
* s/accumulate/accumulates
* s/vote/votes
* s/sends/send

src/peer_manager.rs
```diff
@@ -353,23 +343,14 @@ impl PeerManager {
...
-            return Err(RoutingError::AlreadyHandlingJoinRequest);
-        }
-        self.candidate = Candidate::Expecting {
-            timestamp: Instant::now(),
-            old_pub_id: old_pub_id,
-        };
-        Ok(())
+    /// Return true if already has a candidate
+    pub fn has_candidate(&self) -> bool {
+        self.candidate != Candidate::None
```
pierrechevalier83: I was going to recommend using `is_some` before I realised this `None` was a `Candidate` variant.

Side-note: should we maybe use an Option of Candidate in many of these places rather than mixing concepts in Candidate? (I'm not expecting you to do it here since it's pre-existing code... Just throwing a thought out there)

src/states/node.rs
```diff
@@ -1736,21 +1727,42 @@ impl Node {
...
+    }
 
-        if old_pub_id == *self.full_id.public_id() {
-            return Ok(()); // This is a delayed message belonging to our own relocate request.
+    // Handles an accumulated `ExpectCandidate` event.
+    // Context: a node is joining our section. Sends the node our section. If the
+    // network is unbalanced, sends `ExpectCandidate` on to a section with a shorter prefix.
+    fn handle_expect_candidate_event(
+        &mut self,
+        vote: ExpectCandidatePayload,
```
pierrechevalier83: Why not borrow `ExpectCandidatePayload` instead of moving it here?

src/states/node.rs
```diff
@@ -510,6 +510,9 @@ impl Node {
                 NetworkEvent::SectionInfo(ref sec_info) => {
                     self.handle_section_info_event(sec_info, our_pfx, outbox)?;
                 }
+                NetworkEvent::ExpectCandidate(ref vote) => {
+                    self.handle_expect_candidate_event(vote.clone())?
```
pierrechevalier83: Do you need to clone here or could you borrow in `handle_expect_candidate_event`?

src/states/node.rs
```diff
@@ -771,18 +774,28 @@ impl Node {
...
         }
         let our_pfx = *self.chain.our_prefix();
-        // filter cached events to SectionInfo where we benefit from additional signatures for
-        // neighbours for sec-msg-relay. Online/Offline events we only need to re-vote events which
-        // havent yet accumulated in old prefix and that are relevant to our new prefix.
+
         cached_events
             .iter()
             .filter(|event| match **event {
+                // Only re-vote not yet accumulated events and still relevant to our new prefix.
```
pierrechevalier83: This comment (L781 and piece of functionality L783) are duplicated, so should probably go to a function.
Actually, a good name for that function can probably remove the need for the comment :wink:

src/states/node.rs
```diff
@@ -1736,21 +1727,42 @@ impl Node {
...
+            dst_name,
+            message_id,
+        }));
+        Ok(())
+    }
 
-        if old_pub_id == *self.full_id.public_id() {
-            return Ok(()); // This is a delayed message belonging to our own relocate request.
+    // Handles an accumulated `ExpectCandidate` event.
+    // Context: a node is joining our section. Sends the node our section. If the
```
pierrechevalier83: typos:
* s/Sends/send
* s/sends/send

src/states/node.rs
```diff
@@ -1939,25 +1973,28 @@ impl Node {
                 iter::once(*self.full_id().public_id()).collect::<BTreeSet<PublicId>>(),
             )
         };
+
+        let response_src = Authority::Section(vote.dst_name);
         let response_content = MessageContent::RelocateResponse {
             target_interval: target_interval,
             section: own_section,
-            message_id: rpc.message_id,
+            message_id: vote.rpc.message_id,
```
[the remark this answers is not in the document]
jeanphilippeD: no longer a problem

src/states/node.rs
```diff
@@ -1894,41 +1894,75 @@ impl Node {
...
-            return self.send_routing_message(relocation_dst, dst, request_content);
+        // Nothing to do with this event.
+        Ok(())
+    }
+
+    // Return Prefix of section with shorter prefix to resend the `ExpectCandidate` to.
+    // Return None if we are the best section.
+    fn need_resend_expect_candidate_to_prefix(&self) -> Option<Prefix<XorName>> {
+        if cfg!(feature = "mock_base") && self.next_relocation_interval.is_some() {
+            // Break things if doing section balancing.
```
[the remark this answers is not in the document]
jeanphilippeD: I'll reword, though I did not expect "break things" to be confusing this way.

src/peer_manager.rs
jeanphilippeD: I am not sure what is the best pattern, both express the same exactly the same thing. We should have this discussion at some point.

src/states/node.rs
jeanphilippeD: We then move things out. Removed the unnecessary clone at the call site.
Though all the members are copy... But that should be enough to address this comment I think.

src/states/node.rs
jeanphilippeD: fixed, can move by removing ref.

src/states/node.rs
jeanphilippeD: Since it would be a free function taking 4 parameters, I'd rather not.
I'm happy to add a closure though.

jeanphilippeD: Though if you have a name in mind for this, because I'm not sure I see an obvious one that is not like the sentence in the comment.

src/states/node.rs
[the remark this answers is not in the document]
jeanphilippeD: Going for:
```
// Forbid section balancing: It Breaks things in this case.
```

src/states/node.rs
```diff
@@ -1894,41 +1894,75 @@ impl Node {
...
+        Some(min_len_prefix)
+    }
+
+    // Forward ExpectCandidate to section with given prefix.
+    fn forward_expect_candidate_to_prefix_section(
+        &mut self,
+        vote: ExpectCandidateVote,
+        prefix: Prefix<XorName>,
+    ) -> Result<(), RoutingError> {
+        let request_content = MessageContent::ExpectCandidate(vote.rpc);
```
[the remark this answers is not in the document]
jeanphilippeD: Most of the other self.send_routing_message have (src, dst, content).
I'm thinking maybe make more sense to call it just content like all the others?
I'll also fix send_relocate_response to follow the same convention of having src, dst, content.

What do you think?

src/states/node.rs
jeanphilippeD: Not convinced will get a win there.

src/states/node.rs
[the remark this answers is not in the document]
pierrechevalier83: Yup. Code reads better now :smile:

src/peer_manager.rs
pierrechevalier83: Sure. We can chat about it. My argument for using `Option` is that you get the nicities like `is_some`, `map`, `and_then`,... for free without having to manually re-code them everywhere like a barbarian :wink:

pierrechevalier83: Also, in terms of SRP, if you're an `Option`, you're an `Option`. If you're a `Candidate`, you're a `Candidate`... It's not nearly as bad as mixing memory management with other functionality, but it's got similarities with that anti-pattern...

jeanphilippeD: Moved that conversation to rust-help

src/states/node.rs
pierrechevalier83: Your argument for a lambda is convincing: What about:
```
// Only revote not yet accumulated events that are still relevant to our new prefix.
let should_revote = | prefix: Prefix | { our_pfx.matches(prefix) && !completed_events.contains(event) };
...
    NetworkEvent::Online(pub_id, _) | NetworkEvent::Offline(pub_id) => {
        should_revote(pub_id.name())
    }
    NetworkEvent::ExpectCandidate(ExpectCandidatePayload { ref dst_name, .. }) => {
        should_revote(dst_name)
    }
```

jeanphilippeD: Is that within the filter closure, because otherwise you are also needing event.
That is why I'm, not convinced.

Also the name is somewhat problematic, because other variant decide whether to re-vote differently.

PR #1625, opened 2019-05-07
fix soak test related to message loss

jeanphilippeD:


tests/mock_crust/merge.rs
```diff
@@ -126,7 +142,7 @@ fn concurrent_merge() {
     // Poll the nodes, check the invariant and ensure the network has merged to `0` and `1`.
```
Viv-Rajkumar: needs a comment update

tests/mock_crust/merge.rs
```diff
@@ -94,14 +95,29 @@ fn concurrent_merge() {
...
-    let mut nodes = create_connected_nodes_until_split(&network, vec![2, 2, 2, 2], false);
+    let mut nodes = create_connected_nodes_until_split(&network, vec![2, 2, 2, 3, 3], false);
     verify_invariant_for_all_nodes(&mut nodes);
     rng.shuffle(&mut nodes);
 
-    // Choose two sections to drop nodes from, one of `00`/`01` and the other one of `10`/`11`.
-    let prefix_0_to_drop_from = Prefix::new(1, XorName([0; XOR_NAME_LEN])).pushed(rng.gen());
-    let prefix_1_to_drop_from = Prefix::new(1, XorName([255; XOR_NAME_LEN])).pushed(rng.gen());
-    let prefixes_to_drop_from = vec![prefix_0_to_drop_from, prefix_1_to_drop_from];
+    // Choose two random sections to drop nodes from.
```
Viv-Rajkumar: can we explain this a bit more to make sure we don't just revert to a "simple random" selection and ensure a concurrent merge.

jeanphilippeD: I started with picking 2 random sections and it fail for the reason mentioned in the new comment, so there is no risk of us reverting to a simple random here.

Added:
```
    // Choose two random sections to drop nodes from:
    // Not sibling sections as we would only merge 2 section instead of 4.
    // Not one without sibling, this section would merge with 2 other sections.
    //
    // This setup also allows testing concurrent merges where one of the merging
    // section is not a neighbour of either of the other two merging sections.
```

Viv-Rajkumar: what I meant as simple random was things like randomly picking two sections and thus not having a concurrent merge. The extra detail does make that clear as not expected here. so good with that now.

PR #1627, opened 2019-05-10
test candidate fail resource proof

jeanphilippeD:
Refactor in multiple commits to ease access to node state in test avoiding much boiler plate code.
Add new test verifying behaviour when a candidate does not go through with resource proof.
Extend poll_and_resend with pool_and_resend_until to stop progression when condition reached.

### DIFF
--- a/src/node.rs
+++ b/src/node.rs
@@ -512,17 +512,17 @@ impl Node {
     /// Returns the first `count` names of the nodes in the routing table which are closest
     /// to the given one.
     pub fn close_group(&self, name: XorName, count: usize) -> Option<Vec<XorName>> {
-        self.machine.close_group(name, count)
+        self.machine.current().close_group(name, count)
     }
 
     /// Returns the `PublicId` of this node.
     pub fn id(&self) -> Result<PublicId, RoutingError> {
-        self.machine.id().ok_or(RoutingError::Terminated)
+        self.machine.current().id().ok_or(RoutingError::Terminated)
     }
 
     /// Returns the minimum section size this vault is using.
     pub fn min_section_size(&self) -> usize {
-        self.machine.min_section_size()
+        self.machine.current().min_section_size()
     }
 
     fn send_action(
@@ -573,33 +573,65 @@ impl EventStepper for Node {
 impl Node {
     /// Returns the chain for this node.
     pub fn chain(&self) -> Option<&Chain> {
-        self.machine.chain()
+        self.machine.current().chain()
+    }
+
+    /// Returns this node state.
+    pub fn node_state(&self) -> Option<&crate::states::Node> {
+        match *self.machine.current() {
+            State::Node(ref state) => Some(state),
+            _ => None,
+        }
+    }
+
+    /// Returns this node mut state.
+    pub fn node_state_mut(&mut self) -> Option<&mut crate::states::Node> {
+        match *self.machine.current_mut() {
+            State::Node(ref mut state) => Some(state),
+            _ => None,
+        }
+    }
+
+    /// Returns this node state unwraped: assume state is Node.
+    pub fn node_state_unchecked(&self) -> &crate::states::Node {
+        unwrap!(self.node_state(), "Should be State::Node")
     }
 
     /// Returns the list of banned clients' IPs held by this node.
     pub fn get_banned_client_ips(&self) -> BTreeSet<IpAddr> {
-        self.machine.current().get_banned_client_ips()
+        self.node_state_unchecked().get_banned_client_ips()
     }
 
     /// Returns whether the current state is `Node`.
     pub fn is_node(&self) -> bool {
-        if let State::Node(..) = *self.machine.current() {
-            true
-        } else {
-            false
+        self.node_state().is_some()
+    }
+
+    /// Returns whether the current state is `ProvingNode`.
+    pub fn proving_node_state(&self) -> Option<&crate::states::ProvingNode> {
+        match *self.machine.current() {
+            State::ProvingNode(ref state) => Some(state),
+            _ => None,
         }
+    }
+
+    /// Returns whether the current state is `ProvingNode`.
+    pub fn is_proving_node(&self) -> bool {
+        self.proving_node_state().is_some()
     }
 
     /// Sets a name to be used when the next node relocation request is received by this node.
     pub fn set_next_relocation_dst(&mut self, dst: Option<XorName>) {
-        self.machine.current_mut().set_next_relocation_dst(dst)
+        let _ = self
+            .node_state_mut()
+            .map(|state| state.set_next_relocation_dst(dst));
     }
 
     /// Sets an interval to be used when a node is required to generate a new name.
     pub fn set_next_relocation_interval(&mut self, interval: Option<(XorName, XorName)>) {
-        self.machine
-            .current_mut()
-            .set_next_relocation_interval(interval)
+        let _ = self
+            .node_state_mut()
+            .map(|state| state.set_next_relocation_interval(interval));
     }
 
     /// Indicates if there are any pending observations in the parsec object
@@ -611,7 +643,9 @@ impl Node {
 
     /// Indicates if a given `PublicId` is in the peer manager as a routing peer
     pub fn is_routing_peer(&self, pub_id: &PublicId) -> bool {
-        self.machine.current().is_routing_peer(pub_id)
+        self.node_state()
+            .map(|state| state.is_routing_peer(pub_id))
+            .unwrap_or(false)
     }
 
     /// Checks whether there is un-acked messages.
--- a/src/node.rs
+++ b/src/node.rs
@@ -34,8 +34,6 @@ use safe_crypto::PublicSignKey;
 use std::collections::{BTreeMap, BTreeSet};
 #[cfg(feature = "mock_base")]
 use std::fmt::{self, Display, Formatter};
-#[cfg(feature = "mock_base")]
-use std::net::IpAddr;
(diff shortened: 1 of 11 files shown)


tests/mock_crust/utils.rs
```diff
@@ -332,9 +332,24 @@ impl Cache for TestCache {
...
+/// need_stop: can be used for an early return from poll_all
+pub fn poll_all_until(
+    nodes: &mut [TestNode],
+    clients: &mut [TestClient],
+    need_stop: &Fn(&[TestNode]) -> bool,
+) -> bool {
     assert!(!nodes.is_empty());
     let mut result = false;
     for _ in 0..MAX_POLL_CALLS {
+        if need_stop(nodes) {
```
Viv-Rajkumar: I do prefer `should_stop` compared to `need_stop` btw.

tests/mock_crust/mod.rs
```diff
@@ -54,6 +75,51 @@ fn disconnect_on_rebootstrap() {
...
+    poll_and_resend(&mut nodes, &mut []);
+
+    assert_eq!(
+        nodes_in_authority(&nodes, &proving_node.name()),
+        nodes_with_candidate(&nodes),
+        "All members of destination section accepted node as candidate"
+    );
+
+    // Continue after candidate time out:
+    FakeClock::advance_time(60 * 60 * 1000);
```
Viv-Rajkumar: ah I just put this in my diff as a tmp placeholder. not that it should affect anything but if wanting to be precise, we could advance time specifically using the RES_PROOF consts to forward time and check accordingly.

tests/mock_crust/mod.rs
```diff
@@ -54,6 +75,51 @@ fn disconnect_on_rebootstrap() {
...
+    let mut nodes = create_connected_nodes_until_split(&network, vec![1, 1], false);
+    let bootstrap_config = BootstrapConfig::with_contacts(&[nodes[0].handle.endpoint()]);
+    nodes.insert(
+        0,
+        TestNode::builder(&network)
+            .bootstrap_config(bootstrap_config)
+            .create(),
+    );
+
+    // Initiate connection until the node receive RelocateResponse:
```
Viv-Rajkumar: well if being precise, its more until the node switches its state to a `ProvingNode` as it does potentially restart and get a new identity and maybe other things too than just on receipt of the RelRsp rpc

tests/mock_crust/mod.rs
```diff
@@ -54,6 +75,51 @@ fn disconnect_on_rebootstrap() {
     expect_next_event!(unwrap!(nodes.last_mut()), Event::Terminated);
 }
 
+#[test]
+fn candidate_timeout_resource_proof() {
+    let network = Network::new(MIN_SECTION_SIZE, None);
+    let mut nodes = create_connected_nodes_until_split(&network, vec![1, 1], false);
```
Viv-Rajkumar: why do we need multiple sections here? sure I can reason with that in saying "it does check all nodes have lost a candidate cos only one section should have accepted the candidate and then purged it", but that should maybe be a seperate test to confirm a joining node is only accepted to one section. 

I'm just a bit partial about keeping smaller tests specific to what they're testing cos otherwise almost all tests would benefit from a bigger network setup ofc. not a biggie tho anyways if we wanted to leave as is.

src/states/node.rs
```diff
@@ -2169,6 +2162,10 @@ impl Node {
             .get_peer(pub_id)
             .map_or(false, Peer::is_routing)
     }
+
+    pub fn has_resource_proof_candidate(&self) -> bool {
```
Viv-Rajkumar: I'd have matched the inner method name here to just `has_candidate` and if needed explicit difference eventually then name them throught accordingly.

tests/mock_crust/mod.rs
```diff
@@ -54,6 +75,51 @@ fn disconnect_on_rebootstrap() {
...
+        nodes[0].inner.is_proving_node()
+    });
+    let proving_node = nodes.remove(0);
+
+    assert!(
+        proving_node.inner.is_proving_node(),
+        "Accepted as candidate"
+    );
+
+    // Continue without the joining node until all node consensused events:
```
Viv-Rajkumar: > node consensused events:
maybe switch to `nodes idle` poll_and_resend isnt just abt consensus events, also includes ack_timeouts and stuff ofc.

tests/mock_crust/mod.rs
jeanphilippeD: I was a bit confused initially by some of the logs and wondered whether the proxy node disconnection was because the node changed identity or whether we timed out the candidate.

As it is the test take 0.03 second on average to run on its own. So it seem reasonable to have a more representative flow.

src/states/node.rs
jeanphilippeD: Will rename has_candidate.

tests/mock_crust/mod.rs
jeanphilippeD: I think I'll leave it as is.
Because node could (will?) have different starting instant, it seem difficult to try to be precise here.

src/node.rs
```diff
@@ -573,33 +573,65 @@ impl EventStepper for Node {
...
+    /// Returns this node state.
+    pub fn node_state(&self) -> Option<&crate::states::Node> {
+        match *self.machine.current() {
+            State::Node(ref state) => Some(state),
+            _ => None,
+        }
+    }
+
+    /// Returns this node mut state.
+    pub fn mut_node_state(&mut self) -> Option<&mut crate::states::Node> {
```
madadam: This should be called `node_state_mut` to follow the convention that is common in rust and its ecosystem.

tests/mock_crust/mod.rs
```diff
@@ -104,13 +97,16 @@ fn candidate_timeout_resource_proof() {
...
-        nodes_in_authority(&nodes, &proving_node.name()),
+        nodes.iter().map(TestNode::name).collect_vec(),
         nodes_with_candidate(&nodes),
         "All members of destination section accepted node as candidate"
     );
 
     // Continue after candidate time out:
-    FakeClock::advance_time(60 * 60 * 1000);
+    FakeClock::advance_time(
+        1000 * (routing::test_consts::RESOURCE_PROOF_DURATION_SECS
```
madadam: Minor: I'd bring `test_consts` into scope to not have to prefix it with `routing::` all the time.

src/node.rs
```diff
@@ -577,35 +569,62 @@ impl EventStepper for Node {
...
-    /// Returns the list of banned clients' IPs held by this node.
-    pub fn get_banned_client_ips(&self) -> BTreeSet<IpAddr> {
-        self.machine.current().get_banned_client_ips()
+    /// Returns the chain for this node.
+    pub fn chain(&self) -> Option<&Chain> {
+        self.machine.current().chain()
+    }
+
+    /// Returns this node state.
+    pub fn node_state(&self) -> Option<&crate::states::Node> {
```
madadam: I'm going to put this comment here, but it involves other changes in this PR as well:
I'm not sure I see the benefit of exposing the internal state like this and removing the forwarding functions. It seems to go contrary to what's generally considered good practices: separation of concerns and low coupling.

src/peer_manager.rs
```diff
@@ -342,7 +342,7 @@ impl PeerManager {
     }
 
     /// Return true if already has a candidate
-    pub fn has_candidate(&self) -> bool {
+    pub fn has_resource_proof_candidate(&self) -> bool {
```
madadam: I'm not sure about this rename. I think "candidate" here means someone who is aspiring to join a section and "resource proof" is what they have to perform to do that. `resource_proof_candidate` sounds like "resource proof" itself is the thing they are aspiring to.

src/node.rs
jeanphilippeD: I agree with this in general, but I do think testing have its own requirements that do not always fit well with all the other code good practices.

I'm not sure the low coupling benefit is really there, when we are exposing internals through multiple layer of indirection.
Right now this is more making it extra convoluted to things when needed, and resulted in 3 different way of accessing that internals.

src/peer_manager.rs
jeanphilippeD: Well they are aspiring to "resource proof", initially. If they pass resource proof, the section will take them in. I rather keep a less generic name than has_candidate, but I'm open to other more specific names.

src/node.rs
madadam: I'd prefer to have only one way to do things and that way to be the forwarding methods to expose only what's really necessary. That said, I don't consider this to be that important in the end and with your approach at least there is less code which is always nice. I'm OK with it.

src/peer_manager.rs
madadam: The way I see it is that we can have multiple candidates aspiring to become section members but only one of those is selected for the resource proof at a time. So all except the one currently undergoing the proof can be considered "resource proof candidates", because they are waiting for the resource proof. The other one is not a candidate for resource proof anymore because they are already undergoing it. So maybe this should be `member_candidate`?

Anyway, I thiks this is a very minor semantic issue and probably not worth spending that much time on, so I'll leave this up to you.

jeanphilippeD: I kind of see where you are going, but I'm not sure `member_candidate` is really clear.
Because the Candidate enum is all about resource proof, I'll keep that attached to the name for now.
This code will change a lot in the close future as things need to eventually move in the chain and follow the designed flow, so it is something we can and should address when things have cleaned up a bit.

madadam: OK, fair enough. Let's keep it.

PR #1628, opened 2019-05-13
consistent failed candidate when timing out

jeanphilippeD:
We want to ensure eventually that the candidate is a state shared by
all peers in the section. To do that all modification must be the
result of reaching consensus on an event.

We use a new consensused event PurgeCandidate to decide together
whether a candidate failed or succeeded.

### DIFF
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -275,12 +275,15 @@ enum Candidate {
         challenge: Option<ResourceProofChallenge>,
         passed_our_challenge: bool,
     },
+    AcceptedWaitingSectionInfo {
+        new_pub_id: PublicId,
+    },
 }
 
 impl Candidate {
     fn is_expired(&self) -> bool {
         match *self {
-            Candidate::None => false,
+            Candidate::None | Candidate::AcceptedWaitingSectionInfo { .. } => false,
             Candidate::AcceptedForResourceProof {
                 res_proof_start, ..
             }
@@ -430,7 +433,11 @@ impl PeerManager {
                 );
                 return Err(RoutingError::UnknownCandidate);
             }
-            _ => return Err(RoutingError::UnknownCandidate),
+            Candidate::None
+            | Candidate::AcceptedForResourceProof { .. }
+            | Candidate::AcceptedWaitingSectionInfo { .. } => {
+                return Err(RoutingError::UnknownCandidate)
+            }
         };
 
         if !self.is_connected(new_pub_id) {
@@ -459,6 +466,11 @@ impl PeerManager {
             } if pub_id == *new_pub_id => (),
             _ => return Err(RoutingError::UnknownCandidate),
         }
+
+        // Do not accept candidate until we complete our current one.
+        self.candidate = Candidate::AcceptedWaitingSectionInfo {
+            new_pub_id: *new_pub_id,
+        };
 
         if let Some(peer) = self.peers.get_mut(new_pub_id) {
             let is_connected = peer.is_connected();
@@ -588,6 +600,11 @@ impl PeerManager {
                 }
                 trace!("{}and is not yet approved by our section.", log_prefix);
             }
+            Candidate::AcceptedWaitingSectionInfo { new_pub_id } => trace!(
+                "{}{} candidate waits for its SectionInfo.",
+                log_prefix,
+                new_pub_id
+            ),
         }
     }
 
@@ -648,7 +665,7 @@ impl PeerManager {
     pub fn remove_expired_peers(&mut self) -> Vec<PublicId> {
         let remove_candidate = if self.candidate.is_expired() {
             match self.candidate {
-                Candidate::None => None,
+                Candidate::None | Candidate::AcceptedWaitingSectionInfo { .. } => None,
                 Candidate::AcceptedForResourceProof { ref old_pub_id, .. } => Some(*old_pub_id),
                 Candidate::ResourceProof { ref new_pub_id, .. } => Some(*new_pub_id),
             }
@@ -972,14 +989,23 @@ impl PeerManager {
     }
 
     // Forget about the current candidate
-    pub fn remove_candidate(&mut self) {
+    pub fn reset_candidate(&mut self) {
         self.candidate = Candidate::None;
+    }
+
+    // Forget about the current candidate
+    pub fn reset_candidate_if_member(&mut self, members: &BTreeSet<PublicId>) {
+        if let Candidate::AcceptedWaitingSectionInfo { ref new_pub_id } = self.candidate {
+            if members.contains(new_pub_id) {
+                self.candidate = Candidate::None;
+            }
+        }
     }
 
     /// Removes the given peer. Returns whether the peer was actually present.
     pub fn remove_peer(&mut self, pub_id: &PublicId) -> bool {
         let remove_candidate = match self.candidate {
-            Candidate::None => false,
+            Candidate::None | Candidate::AcceptedWaitingSectionInfo { .. } => false,
             Candidate::AcceptedForResourceProof { ref old_pub_id, .. } => {
                 // only consider candidate cleanup via old_id if candidate is also expired.
                 // else candidate may simply be restarting.
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -258,42 +258,94 @@ impl Peer {
     }
 }
 
+/// A candidate (if any) may be in different stages of the resource proof process.
+/// As they are accepted for resource proof, a timer will start.
+/// On expiry of this timer, we will vote for `PurgeCandidate` and set expired_once, but the
+/// resource proof will continue until we reach consensus on either `PurgeCandidate` or `Online`.
+/// On termination of our part of the resource proof (When our challenge is completed, we will
+/// vote the candidate `Online`.
+/// Regardless of our own opinion (be it a vote for `Online`, a vote for PurgeCandidate or a
+/// vote for both) we will wait for consensus to be reached on either of these and take the
+/// first such event to reach consensus as the source of truth.
+/// Finally, if `Online` is consensused first, before allowing a new candidate, we wait for its
+/// SectionInfo to be consensused, so this new member will process `ExpectCandidate` the same way
+/// the other members will.
 // FIXME - See https://maidsafe.atlassian.net/browse/MAID-2026 for info on removing this exclusion.
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug, Eq, PartialEq)]
 enum Candidate {
+    /// No-one is currently in the resource proof process.
+    /// We can take on a new candidate on consensus of an `ExpectCandidate` event.
     None,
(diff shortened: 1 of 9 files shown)


src/peer_manager.rs
```diff
@@ -971,23 +1036,30 @@ impl PeerManager {
...
+    pub fn reset_candidate_with_old_public_id(&mut self, old_pub_id: &PublicId) {
+        if Some(old_pub_id) == self.candidate.old_pub_id() {
             self.candidate = Candidate::None;
         }
+    }
 
-        self.peers.remove(pub_id).is_some() || remove_candidate
+    /// Removes the given peer. Returns whether the peer was actually present.
+    pub fn remove_peer(&mut self, pub_id: &PublicId) -> bool {
+        self.peers.remove(pub_id).is_some()
```
madadam: What happens now when we call `remove_peer` with a `pub_id` of the candidate? Aren't we going to end up in inconsistent state, because `self.candidate` will be still set, but the corresponding entry in `self.peers` will be gone? Maybe we shouldn't remove the entry from `self.peers` in this case?

src/states/node.rs
```diff
@@ -433,14 +433,17 @@ impl Node {
...
-                NetworkEvent::ExpectCandidate(ExpectCandidatePayload { ref dst_name, .. }) => {
-                    our_pfx.matches(dst_name) && !completed_events.contains(event)
-                }
+                // Drop candidates that have not completed:
+                // Called peer_manager.remove_candidate reset the candidate so it can be shared by
+                // all nodes: Because new node may not have voted for it, Forget the votes in
+                // flight as well.
+                NetworkEvent::Online(_, _)
+                | NetworkEvent::ExpectCandidate(_)
+                | NetworkEvent::PurgeCandidate(_) => false,
```
madadam: I'm confused about this. Why is it OK to ignore these events now?

src/states/node.rs
```diff
@@ -2458,3 +2476,336 @@ fn create_first_section_info(public_id: PublicId) -> Result<SectionInfo, Routing
         err
     })
 }
+
+#[cfg(all(test, feature = "mock_parsec"))]
+mod tests {
```
madadam: I would move all these tests to a separate file (say `src/states/node_tests.rs` or `src/states/node/tests.ts`). This one is already too long.

src/peer_manager.rs
```diff
@@ -971,23 +1036,30 @@ impl PeerManager {
...
-                // only consider candidate cleanup via old_id if candidate is also expired.
-                // else candidate may simply be restarting.
-                old_pub_id == pub_id && self.candidate.is_expired()
+    /// Forget about the current candidate.
+    pub fn reset_candidate(&mut self) {
+        self.candidate = Candidate::None;
+    }
+
+    /// Forget about the current candidate if it is a member of the given section.
+    pub fn reset_candidate_member_of(&mut self, members: &BTreeSet<PublicId>) {
```
madadam: This name sounds weird. What about `reset_candidate_if_member_of`?

src/states/node.rs
```diff
@@ -2458,3 +2476,336 @@ fn create_first_section_info(public_id: PublicId) -> Result<SectionInfo, Routing
...
+        pub machine: StateMachine,
+        pub full_id: FullId,
+        pub other_full_ids: Vec<FullId>,
+        pub other_parsec_map: Vec<ParsecMap>,
+        pub ev_buffer: EventBuf,
+        pub section_info: SectionInfo,
+    }
+
+    impl NoteUnderTest {
+        fn make_node() -> NoteUnderTest {
```
madadam: Any reason this method isn't just called `new`?

madadam: You can avoid repeating the type name in the return type by using `Self` instead.

src/states/node.rs
```diff
@@ -2458,3 +2476,336 @@ fn create_first_section_info(public_id: PublicId) -> Result<SectionInfo, Routing
...
+            let full_id = full_ids[0].clone();
+            let machine = make_state_machine(&full_id, &gen_pfx_info, &mut ev_buffer);
+
+            let other_full_ids = full_ids[1..4].iter().cloned().collect_vec();
+            let other_parsec_map = other_full_ids
+                .iter()
+                .map(|full_id| ParsecMap::new(full_id.clone(), &gen_pfx_info))
+                .collect_vec();
+
+            NoteUnderTest {
```
madadam: Again, you can just use `Self { ... }` to avoid repeating the type name.

src/states/node.rs
```diff
@@ -2458,3 +2476,336 @@ fn create_first_section_info(public_id: PublicId) -> Result<SectionInfo, Routing
...
+                full_id,
+                other_full_ids,
+                other_parsec_map,
+                ev_buffer,
+                section_info,
+            }
+        }
+
+        fn node_state(&self) -> &Node {
+            match *self.machine.current() {
```
madadam: You added `StateMachine::node_state`. Why not use it here too?

src/states/node.rs
```diff
@@ -2458,3 +2476,336 @@ fn create_first_section_info(public_id: PublicId) -> Result<SectionInfo, Routing
...
+    use utils::LogIdent;
+
+    const ACCUMULATE_VOTE_COUNT: usize = 3;
+    const NOT_ACCUMULATE_ALONE_VOTE_COUNT: usize = 2;
+
+    struct NoteUnderTest {
+        pub machine: StateMachine,
+        pub full_id: FullId,
+        pub other_full_ids: Vec<FullId>,
+        pub other_parsec_map: Vec<ParsecMap>,
```
madadam: I'd maybe merge these two fields into one: `other_nodes: Vec<(FullId, ParsecMap)>` (or better use a struct instead of tuple) to make sure they never get out of sync.

src/states/node.rs
```diff
@@ -2458,3 +2476,336 @@ fn create_first_section_info(public_id: PublicId) -> Result<SectionInfo, Routing
...
+    use crate::state_machine::{State, StateMachine};
+    use crate::xor_name::XOR_NAME_LEN;
+    use utils::LogIdent;
+
+    const ACCUMULATE_VOTE_COUNT: usize = 3;
+    const NOT_ACCUMULATE_ALONE_VOTE_COUNT: usize = 2;
+
+    struct NoteUnderTest {
+        pub machine: StateMachine,
+        pub full_id: FullId,
```
madadam: I'd consider adding a `full_id` method to `StateMachine` (test-only) so you don't have to duplicate it here (and possibly risking them getting out of sync).

src/states/node.rs
```diff
@@ -2458,3 +2476,336 @@ fn create_first_section_info(public_id: PublicId) -> Result<SectionInfo, Routing
...
+        pub other_full_ids: Vec<FullId>,
+        pub other_parsec_map: Vec<ParsecMap>,
+        pub ev_buffer: EventBuf,
+        pub section_info: SectionInfo,
+    }
+
+    impl NoteUnderTest {
+        fn make_node() -> NoteUnderTest {
+            let full_ids = (0..4).map(|_| FullId::new()).collect_vec();
```
madadam: Where does the number `4` come from? Maybe extract it to a named constant?

src/states/node.rs
Viv-Rajkumar: more so just not carrying candidate info into a new prefix than ignoring. currently candidate info isnt part of the shared state, so you could have a pfx change with members having different candidates, to prevent that, we're resetting candidate info on pfx change and thus these in flight events that also influence the old candidate state from the old prefix are filtered out to not be replayed in the new prefix to prevent it from affecting a new candidate in the next prefix. thus giving all nodes entering that prefix the same blank state if that helps.

madadam: OK. But isn't the old candidate going to be forgotten now? Shouldn't we at least inform them somehow that they should repeat the join attempt with the new prefix?

src/peer_manager.rs
Viv-Rajkumar: > Aren't we going to end up in inconsistent state, because self.candidate will be still set, but the corresponding entry in self.peers will be gone?
that should be an acceptable scenario in general as in the section has a candidate that maybe some members cant connect to and thus dont have in peer_mgr. so the candidate info should be uncoupled to the candidates state in peer_mgr itself in terms of connected or not. So while I'm not sure if we call remove_peer for a valid candidate, even if we did we should expect it to only result in an eventual purge candidate for that peer later and them not being accepted into the section.

src/states/node.rs
jeanphilippeD: They can't, we are not waiting for them anymore. At this point they are gone from the network and have to rejoin when they time out on their side. (This is not something that should happen in the final design).

src/states/node.rs
jeanphilippeD: 1- Are you ok for it to be a follow up PR?
2- Which version exactly do we want? I would want it to have the same level of access as the current test have.

src/states/node.rs
madadam: If we send them a message to restart the candidate process it would speed things up as they wouldn't have to wait for the timeout. Probably not important enough to deserve complicating the code though. I'm OK with it.

src/states/node.rs
jeanphilippeD: I thought I did, but I must have misplaced it :-)

src/states/node.rs
Viv-Rajkumar: yeh the timeout from joining node side being the key ^^ cos even without a pfx change, we dont inform a joining node they failed res-proof/join process. thats just left to their timeout of not getting a NodeApproval ultimately to switch to establishing state and that handles these too for now while eventually as jp mentioned the candidate info itself with node-ageing wouldnt be ignored like now.

src/states/node.rs
madadam: Hm good point about the access. I think `tests` needs to remain a submodule of `node`, so that means `src/states/node/tests.rs`. Don't mind if it's done in a different PR.

src/states/node.rs
jeanphilippeD: I think that would make the use of them less practical/readable, I'm not sure it is a good trade-off for test code.

src/states/node.rs
jeanphilippeD: The risk is super low, since it is test code that would likely fail if it happened. Not sure it is worth adding noise to the StateMachine yet.

src/states/node.rs
jeanphilippeD: Will do, this is essentially the minimal member count that allow a consensus without one member participation.

src/states/node.rs
madadam: OK, your call.

src/states/node.rs
jeanphilippeD: I wanted to be more specific, but I did not succeed, will try to come up with a good name or use new.

src/states/node.rs
madadam: I think it's a good practice to reduce the duplication of information. Probably not a big issue here yet though.

PR #1629, opened 2019-05-14
fix/states: remove candidate when PurgeCandidate event comes first

jeanphilippeD:
Add new tests to cover when PurgeCandidate comes before Online event.
Fix the code to ensure the test pass:
 - Add new field to NetworkEvent::Online with the old_public_id.
 - Use that field to decide whether we should accept the new member
   or vote it offline consistently imediately.

Both auth_client and old_public info are needed because the current
node did not receive the CandidateInfo. Keep them together in a
struct that will be serialized as related_info.

Note:
-This is not ideal because it introduce a new quirk: when candidate is
 rejected, it is added then removed. This did not happen before.
 The benefit is all node behave consistently, and this will be better
 addressed soon when introducing Opaque Online/Offline NetworkEvent.
-Now PeerManager::candidate has an expectation to only change
 from None to not None when reaching consensus on ExpectCandidate and
 from not None to None when reaching Consensus on PrugeCandidate or the
 SectionInfo containing the candidate.
 => We must no longer fall back to reseting it to None on error.

Also improve error message for soak test failure.
Also fix issue when receiving CandidateInfo signed incorrectly.

### DIFF
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -409,6 +409,16 @@ impl PeerManager {
         self.candidate != Candidate::None
     }
 
+    /// Return true if received CandidateInfo
+    #[cfg(all(test, feature = "mock_parsec"))]
+    pub fn has_candidate_info(&self) -> bool {
+        if let Candidate::ResourceProof { .. } = self.candidate {
+            true
+        } else {
+            false
+        }
+    }
+
     /// Our section decided that the candidate should be selected next.
     /// replace any ongoing (i.e. unapproved) candidate with the new state
     /// `AcceptedForResourceProof` for the given candidate.
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -7,6 +7,7 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::{
+    chain::OnlinePayload,
     crust::CrustUser,
     error::RoutingError,
     id::PublicId,
@@ -25,7 +26,7 @@ use rand;
 use resource_proof::ResourceProof;
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap, VecDeque},
-    error, fmt, mem,
+    error, fmt,
     net::IpAddr,
 };
 
@@ -416,6 +417,13 @@ impl PeerManager {
         old_pub_id: PublicId,
         target_interval: (XorName, XorName),
     ) {
+        if self.has_resource_proof_candidate() {
+            log_or_panic!(
+                LogLevel::Error,
+                "accept_as_candidate when processing one already"
+            );
+        }
+
         self.candidate = Candidate::AcceptedForResourceProof {
             res_proof_start: Instant::now(),
             expired_once: false,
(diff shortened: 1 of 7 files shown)


src/peer_manager.rs
```diff
@@ -467,14 +467,15 @@ impl PeerManager {
     pub fn verified_candidate_info(
         &self,
         log_ident: &LogIdent,
-    ) -> Result<(PublicId, Authority<XorName>), RoutingError> {
-        let (new_pub_id, new_client_auth) = match self.candidate {
+    ) -> Result<(PublicId, Authority<XorName>, PublicId), RoutingError> {
```
madadam: Should update the doc comment of this method to reflect the changed return type. I'd also consider turning that tuple into a struct with named fields to make things clearer.

src/states/node.rs
```diff
@@ -1418,13 +1418,14 @@ impl Node {
...
 
     fn send_candidate_approval(&mut self) {
-        let (new_id, client_auth) = match self.peer_mgr.verified_candidate_info(&self.log_ident()) {
-            Err(_) => {
-                trace!("{} No candidate for which to send CandidateApproval.", self);
-                return;
-            }
-            Ok(result) => result,
-        };
+        let (new_id, client_auth, old_public_id) =
```
madadam: Either rename `new_id` to `new_public_id` or `old_public_id` to `old_id`, to keep the names consistent.

src/peer_manager.rs
madadam: Or maybe return `(PublicId, OnlinePayload)` ?

src/peer_manager.rs
```diff
@@ -505,21 +514,43 @@ impl PeerManager {
...
+            // Unkwown candidate, Candidate was purged before: refuse it.
+            false
+        }
     }
 
-    /// Handles accumulated candidate approval. Marks the candidate as `Approved` and returns if the
-    /// candidate is connected or `Err` if the peer is not the candidate or we're missing its info.
+    /// Handles approved accumulated candidate. Returns if the candidate is connected or
+    /// `Err` if the peer is not the candidate or we're missing its info.
     pub fn handle_candidate_approval(
```
madadam: So this function now seems to just return whether the peer is connected. I'd consider renaming it to `is_peer_connected` or something like that.

src/states/node.rs
```diff
@@ -2640,8 +2683,156 @@ mod tests {
...
+
+        fn offline_payload(&self) -> PublicId {
+            *self.candidate_info.new_full_id.public_id()
+        }
+
+        fn purge_payload(&self) -> PublicId {
+            *self.candidate_info.old_full_id.public_id()
+        }
+
+        fn expect_candidate_rpc(&self) -> RoutingMessage {
```
madadam: A nitpick, but we don't use the term "RPC" anywhere in routing codebase - instead, we use "message". I'd stick to that term to not unnecessarilly introduce new terminology.

tests/mock_crust/utils.rs
```diff
@@ -680,14 +680,31 @@ pub fn verify_invariant_for_all_nodes(nodes: &mut [TestNode]) {
     let min_section_size = nodes[0].handle.0.borrow().network.min_section_size();
     verify_chain_invariant(nodes.iter().map(TestNode::chain), min_section_size);
 
+    let mut all_missing_peers = BTreeSet::<PublicId>::new();
```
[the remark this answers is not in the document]
jeanphilippeD: Cannot infer type. I think it is because I use extend.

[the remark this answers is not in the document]
madadam: Weird, it looks like it should be able to. Nevermind then.

tests/mock_crust/utils.rs
```diff
@@ -680,14 +680,31 @@ pub fn verify_invariant_for_all_nodes(nodes: &mut [TestNode]) {
...
+            .iter()
+            .filter(|pub_id| !node.inner.is_node_peer(pub_id))
+            .cloned()
+            .collect_vec();
+        if !missing_peers.is_empty() {
+            error!(
+                "verify_invariant_for_all_nodes: node {}: missing: {:?}",
+                our_id, &missing_peers
+            );
+            all_missing_peers.extend(missing_peers.into_iter());
```
madadam: The call to `into_iter` is redundant here.

PR #1631, opened 2019-05-20
use opaque Offline/Online event

jeanphilippeD:
Avoid temporarily adding candidate that fail resource proof.
Open door for Node Ageing Adults that do not become Elder. 

Some additional test uncover more issues that will not be addressed here:
https://github.com/maidsafe/routing/pull/1636

### DIFF
--- a/src/chain/network_event.rs
+++ b/src/chain/network_event.rs
@@ -41,13 +41,21 @@ pub struct OnlinePayload {
 #[allow(clippy::large_enum_variant)]
 #[derive(Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub enum NetworkEvent {
+    /// Add new elder once we agreed to add a candidate
+    AddElder(PublicId, Authority<XorName>),
+    /// Remove elder once we agreed to remove the peer
+    RemoveElder(PublicId),
+
+    /// Voted for candidate that pass resource proof
     Online(PublicId, OnlinePayload),
+    /// Voted for candidate we no longer consider online.
     Offline(PublicId),
+
     OurMerge,
     NeighbourMerge(Digest256),
     SectionInfo(SectionInfo),
 
-    /// Voted for received ExpectCandidate RPC.
+    /// Voted for received ExpectCandidate Message.
     ExpectCandidate(ExpectCandidatePayload),
 
     // Voted for timeout expired for this candidate old_public_id.
@@ -77,11 +85,11 @@ impl NetworkEvent {
     /// Convert `NetworkEvent` into a Parsec Observation
     pub fn into_obs(self) -> Result<parsec::Observation<NetworkEvent, PublicId>, RoutingError> {
         Ok(match self {
-            NetworkEvent::Online(id, auth) => parsec::Observation::Add {
+            NetworkEvent::AddElder(id, auth) => parsec::Observation::Add {
                 peer_id: id,
                 related_info: serialise(&auth)?,
             },
-            NetworkEvent::Offline(id) => parsec::Observation::Remove {
+            NetworkEvent::RemoveElder(id) => parsec::Observation::Remove {
                 peer_id: id,
                 related_info: Default::default(),
             },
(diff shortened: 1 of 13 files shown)


src/peer_manager.rs
```diff
@@ -118,7 +118,7 @@ pub enum PeerState {
     /// Connected peer is a joining node and waiting for approval of routing.
     Candidate,
     /// We are connected to the peer who is a full node.
-    Node,
+    Node(bool),
```
madadam: Better add a comment explaining what the `bool` is. Or better, use a named field: `Node { was_joining: bool }` (just a suggestion. Feel free to use any name you think is appropriate).

src/peer_manager.rs
```diff
@@ -254,6 +254,7 @@ impl Peer {
     fn is_joining_node(&self) -> bool {
         match self.state {
             PeerState::JoiningNode => true,
+            PeerState::Node(val) => val,
```
madadam: This feels wrong somehow. A peer with `PeerState::Node` is a full node, not a joining node. The fact that this function can return `true` for a full node is surprising and confusing. I'd suggest a different approach: revert this function as it was and instead add a new function: `was_joining_node` or `is_joined_node` or whatever, which returns true only if `self.state == PeerState::Node(true)`.

src/peer_manager.rs
```diff
@@ -1073,7 +1073,17 @@ impl PeerManager {
 
     /// Removes the given peer. Returns whether the peer was actually present.
     pub fn remove_peer(&mut self, pub_id: &PublicId) -> bool {
-        self.peers.remove(pub_id).is_some()
+        if let Some(mut peer) = self.peers.remove(pub_id) {
+            if peer.is_joining_node() && peer.is_node() {
```
madadam: ...then use the function here: `if peer.was_joining_node { ...`  (or whatever you decide to call it).

src/states/node/mod.rs
```diff
@@ -1729,7 +1729,10 @@ impl Node {
         outbox: &mut EventBox,
         try_reconnect: bool,
     ) -> bool {
-        if self.peer_mgr.remove_peer(&pub_id) {
+        // Calling remove twice to remove a potential JoiningNode we have as a Node than purely
+        // demoting it
+        // TODO: Avoid calling this twice.
+        if self.peer_mgr.remove_peer(&pub_id) || self.peer_mgr.remove_peer(&pub_id) {
```
madadam: This is not ideal (but the TODO suggest you are aware of it). Maybe it would be better to have two functions: `demote_peer` and `remove_peer` ? Or add a param to `remove_peer` which indicates whether to remove or just demote (could be just `bool`, but an `enum` would be better). Another alternative would be to only change the return type of `remove_peer` to indicate whether the peer was removed or only demoted. That would make this `if` condition a bit messier, but at least it would be more explicit about what's going on.

src/chain/network_event.rs
```diff
@@ -41,13 +41,21 @@ pub struct OnlinePayload {
 #[allow(clippy::large_enum_variant)]
 #[derive(Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub enum NetworkEvent {
+    /// Add new elder once we agreed to add a candidate
+    AddElder(PublicId, Authority<XorName>),
+    /// Remove elder once we agreed to remove the peer
+    RemoveElder(PublicId),
+
+    /// Voted for candidate that pass resource proof
     Online(PublicId, OnlinePayload),
```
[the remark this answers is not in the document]
jeanphilippeD: Could be, this was not in the impl spec though.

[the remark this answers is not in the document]
jeanphilippeD: A better name may be AddCandidate. But I'll leave that rename out of this PR as I'm not 100% clear it would be better yet.

src/peer_manager.rs
jeanphilippeD: Most use of this need bot functions, to work together. I've renamed to is_or_was_joining_node.

src/chain/network_event.rs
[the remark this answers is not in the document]
madadam: OK. Let's leave if out of this PR then.

src/states/node/mod.rs
jeanphilippeD: Added remove_peer_no_joining_checks.

PR #1632, opened 2019-05-20
mock-parsec: reevaluate previously unconsensused votes

madadam:
So that observations which initially didn't have enough votes for consensus, but later due to some peers being removed have enough, are properly consensused.

### DIFF
--- a/src/mock_parsec/mod.rs
+++ b/src/mock_parsec/mod.rs
@@ -20,7 +20,7 @@ pub use self::{
 };
 pub use parsec::{NetworkEvent, Proof, PublicId, SecretId};
 
-use self::observation::{ObservationHolder, ObservationState};
+use self::{observation::ObservationHolder, state::SectionState};
 use crate::sha3::Digest256;
 use std::{
     collections::{BTreeMap, BTreeSet},
@@ -96,29 +96,16 @@ where
 
     pub fn vote_for(&mut self, observation: Observation<T, S::PublicId>) -> Result<(), Error> {
         state::with(self.section_hash, |state| {
-            let entry = state.observations.entry(ObservationHolder::new(
-                observation,
-                self.our_id.public_id(),
-                self.consensus_mode,
-            ));
-            let holder = entry.key().clone();
-
-            if let Some(block) = entry.or_insert_with(ObservationState::new).vote(
-                &self.our_id,
-                &self.peer_list,
-                self.consensus_mode,
-                &holder,
-            ) {
-                state.blocks.push((block, holder.clone()))
-            }
-
+            let holder =
+                ObservationHolder::new(observation, self.our_id.public_id(), self.consensus_mode);
+            state.vote(&self.our_id, holder.clone());
             self.observations
                 .entry(holder)
                 .or_insert_with(ObservationInfo::new)
                 .our = true;
+            self.compute_consensus(state)
         });
 
-        self.update_blocks();
         Ok(())
     }
 
@@ -145,7 +132,7 @@ where
         _src: &S::PublicId,
         _req: Request<T, S::PublicId>,
     ) -> Result<Response<T, S::PublicId>, Error> {
-        self.update_blocks();
+        state::with(self.section_hash, |state| self.compute_consensus(state));
         Ok(Response::new())
     }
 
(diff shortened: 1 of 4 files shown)


src/mock_parsec/mod.rs
```diff
@@ -200,31 +184,37 @@ where
...
+
                 self.first_unconsensused += 1;
                 self.observations
                     .entry(holder.clone())
                     .or_insert_with(ObservationInfo::new)
                     .set_consensused();
             }
-        })
+
+            if !recompute {
```
jeanphilippeD: I guess it could be written as a while?

src/mock_parsec/mod.rs
```diff
@@ -200,31 +184,37 @@ where
...
-        state::with(self.section_hash, |state| {
-            while let Some((block, holder)) = state.blocks.get(self.first_unconsensused) {
-                self.handle_consensus(block.payload());
+    fn compute_consensus(&mut self, state: &mut SectionState<T, S::PublicId>) {
+        loop {
+            state.compute_consensus(&self.peer_list, self.consensus_mode);
+
+            let mut recompute = false;
+
+            while let Some((block, holder)) = state.get_block(self.first_unconsensused) {
```
jeanphilippeD: Could probably be left in its own function `update_blocks`? to avoid nested loops.

src/mock_parsec/tests.rs
```diff
@@ -204,6 +204,77 @@ fn consensus_mode_single() {
...
+                &genesis_group,
+                ConsensusMode::Single,
+            )
+        })
+        .collect();
+
+    // Drain the `Genesis` blocks.
+    gossip_all(&mut nodes);
+    for node in &mut nodes {
+        for _ in poll_all(node) {}
```
jeanphilippeD: why the `for _ in` instead of only `poll_all(node)`?

src/mock_parsec/tests.rs
```diff
@@ -204,6 +204,77 @@ fn consensus_mode_single() {
...
+#[test]
+fn reevaluate_previously_insufficient_votes() {
+    init_mock();
+
+    // Create 5 nodes (A, B, C, D, E). 3 of them vote to remove D, which is not enough to reach
+    // consensus. Then all vote to remove E which gets consensused and E is removed. The previous
+    // votes to remove D are now enough to reach consensus and so D should be removed too.
+
+    let genesis_group: BTreeSet<_> = (0..5).map(|n| PeerId(n)).collect();
+    let mut nodes: Vec<_> = genesis_group
```
jeanphilippeD: Maybe worth making that a well named helper function that take a number and a ConsensusMode and return nodes.
And use it for the other 2 tests as well?

src/mock_parsec/tests.rs
```diff
@@ -204,6 +204,77 @@ fn consensus_mode_single() {
...
+    // Drain the `Genesis` blocks.
+    gossip_all(&mut nodes);
+    for node in &mut nodes {
+        for _ in poll_all(node) {}
+    }
+
+    // 3 votes to Remove(D) - should not get consensus yet.
+    for node in nodes.iter_mut().take(3) {
+        unwrap!(node.vote_for(Observation::Remove {
+            peer_id: PeerId(3),
```
jeanphilippeD: Maybe give it a name let D =  PeerId(3) at the top. (or maybe Dave)?

src/mock_parsec/tests.rs
```diff
@@ -204,6 +204,77 @@ fn consensus_mode_single() {
...
+    gossip_all(&mut nodes);
+
+    for node in &mut nodes {
+        assert_eq!(poll_all(node).next(), None, "{:?}", node.our_pub_id());
+    }
+
+    // Remove E
+    for node in &mut nodes {
+        unwrap!(node.vote_for(Observation::Remove {
+            peer_id: PeerId(4),
```
jeanphilippeD: Same

src/mock_parsec/tests.rs
```diff
@@ -204,6 +204,77 @@ fn consensus_mode_single() {
...
+
+    gossip_all(&mut nodes);
+
+    for node in &mut nodes {
+        assert_eq!(poll_all(node).next(), None, "{:?}", node.our_pub_id());
+    }
+
+    // Remove E
+    for node in &mut nodes {
+        unwrap!(node.vote_for(Observation::Remove {
```
jeanphilippeD: Could possibly have a helper used here and bellow.
```
vote_for_remove(&mut nodes, dave);
```

src/mock_parsec/tests.rs
```diff
@@ -204,6 +204,77 @@ fn consensus_mode_single() {
...
+            peer_id: PeerId(4),
+            related_info: vec![]
+        }))
+    }
+
+    gossip_all(&mut nodes);
+
+    for node in &mut nodes {
+        let block = unwrap!(node.poll());
+        match block.payload() {
```
jeanphilippeD: Could possibly have a helper used here and bellow.
```
assert_peer_removed(&mut nodes, dave);
nodes.remove(dave_index);
```

src/mock_parsec/tests.rs
madadam: `poll_all` returns an iterator. I needed to consume it.

PR #1634, opened 2019-05-22
chore: remove unnecessary clone

d1vyank:
Addresses a minor TODO to remove an unnecessary clone() operation.

### DIFF
--- a/src/states/node/mod.rs
+++ b/src/states/node/mod.rs
@@ -549,17 +549,36 @@ impl Node {
                 return Err(RoutingError::InvalidProvingSection);
             }
         } else {
-            // Remove any untrusted trailing section infos.
-            while match signed_msg.previous_hop() {
-                None => true,
-                Some(hop) => !self.is_trusted(hop)?,
-            } {
-                // We don't know the last hop! Try the one before that.
-                if !signed_msg.pop_previous_hop() {
-                    debug!("{} Untrusted message: {:?}", self, signed_msg);
-                    return Err(RoutingError::NotEnoughSignatures);
+            let mut num_trusted_hops = signed_msg.proving_sections().len();
+            let mut trusted_hop_found = false;
+            for (index, hop) in signed_msg.proving_sections()
+                .iter()
+                .map(|ps| &ps.sec_info)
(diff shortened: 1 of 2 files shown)


src/states/node/mod.rs
```diff
@@ -550,15 +550,13 @@ impl Node {
...
-            // TODO: remove wasted clone. Only useful when msg isnt trusted for log msg.
-            let msg_clone = signed_msg.clone();
             while match signed_msg.previous_hop() {
                 None => true,
                 Some(hop) => !self.is_trusted(hop)?,
             } {
                 // We don't know the last hop! Try the one before that.
                 if !signed_msg.pop_previous_hop() {
-                    debug!("{} Untrusted message: {:?}", self, msg_clone);
+                    debug!("{} Untrusted message: {:?}", self, signed_msg);
```
Viv-Rajkumar: hmm I think the TODO isn't maybe too clear. by just using signed_msg here instead of msg_clone, we'd only be logging the message without any proving_sections in it I think as the `pop_previous_hop` calls through the while loop would be stripping them out. So the point of the cloned object before starting the while loop was to keep a copy of the full signed_msg(with proving_sections) for the log here which ofc can be done better by potentially not popping out invalid ones if we can check is_trusted across the list directly and then print this message and then do the pop out...

d1vyank: Thank you for the feedback! I missed the mutation on `signed_msg` between the clone() and log operations. I'm taking your suggestion to run the `is_trusted` check over the list and logging the message before popping the untrusted hops.
However, while doing so, I noticed that the existing implementation does not error out if the `source_section` field in the message fails the `is_trusted` check, the code only requires that this field is set. Is the source section always trusted?

Viv-Rajkumar: no problem, although this does add a bit of complexity to a portion of code thats going to get deprecated with the secure msg delivery RFCs coming up. So for now I'd probably just vote to close this PR in favor of this TODO getting deprecated soon enough.

d1vyank: That sounds fair, closing.

PR #1635, opened 2019-05-22
chore: optimize our section info cleanup

d1vyank:
Addresses a TODO in the code to remove elements from a sorted list of sections more efficiently.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -371,9 +371,11 @@ impl Chain {
         // Check the lowest version of our info that any neighbour has and remove everything less
         // than it
         let our_oldest_ver = self.their_knowledge.values().min().map_or(0, |&v| v);
-        // TODO since the list is sorted according to the version, choose something more efficient
-        self.our_infos
-            .retain(|&(ref si, _)| *si.version() >= our_oldest_ver);
+        let oldest_ver_index = self
+            .our_infos
+            .binary_search_by_key(&our_oldest_ver, |&(ref si, _)| *si.version())
+            .unwrap_or(0);
+        let _ = self.our_infos.drain(0..oldest_ver_index);
 
         self.check_and_clean_neighbour_infos(None);
         self.state = ChainState::Normal;
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -371,11 +371,19 @@ impl Chain {
         // Check the lowest version of our info that any neighbour has and remove everything less
         // than it
         let our_oldest_ver = self.their_knowledge.values().min().map_or(0, |&v| v);
-        let oldest_ver_index = self
+        let version_retention_threshold_index = self
             .our_infos
-            .binary_search_by_key(&our_oldest_ver, |&(ref si, _)| *si.version())
-            .unwrap_or(0);
-        let _ = self.our_infos.drain(0..oldest_ver_index);
+            .binary_search_by_key(&our_oldest_ver, |(si, _)| *si.version())
+            .unwrap_or_else(|_index| {
+                if our_oldest_ver > 0 {
+                    log_or_panic!(
+                        LogLevel::Warn,
+                        "Oldest version indicated by neighbours not found in our infos"
+                    );
+                }
+                usize::min_value()
+            });
+        let _ = self.our_infos.drain(0..version_retention_threshold_index);
 
         self.check_and_clean_neighbour_infos(None);
         self.state = ChainState::Normal;


src/chain/chain.rs
```diff
@@ -371,9 +371,11 @@ impl Chain {
         // Check the lowest version of our info that any neighbour has and remove everything less
         // than it
         let our_oldest_ver = self.their_knowledge.values().min().map_or(0, |&v| v);
-        // TODO since the list is sorted according to the version, choose something more efficient
-        self.our_infos
-            .retain(|&(ref si, _)| *si.version() >= our_oldest_ver);
+        let oldest_ver_index = self
+            .our_infos
+            .binary_search_by_key(&our_oldest_ver, |&(ref si, _)| *si.version())
+            .unwrap_or(0);
```
madadam: Better would be to use `unwrap_or_else(|index| index)`, to cover the cases where no entry in `our_infos` has its `version` equal to `our_oldest_ver`. Although such case might be impossible in practice, I'd still prefer bit more defensive approach.

afck: Isn't `unwrap_or(0)` (if in doubt, drop nothing) safer than `unwrap_or_else(|index| index)`? Maybe we should even `log_or_panic!` if `Err(i)` with `i > 0` is returned?

d1vyank: Agree that it's safer to use 0 and drop nothing. I've added a `log_or_panic!` for this case.

d1vyank: Update: While debugging a test failure I noticed that, in the tests, the `their_knowledge` map held by `chain` is always empty. I believe this is because `NeighbourInfo` messages are currently not handled: https://github.com/maidsafe/routing/blob/096e7c88d2136d4011ca3e5350e90eb7345b4957/src/states/elder/mod.rs#L711
Is this something that needs to be addressed? I'm happy to help with this, if so.

src/chain/chain.rs
```diff
@@ -371,9 +371,19 @@ impl Chain {
         // Check the lowest version of our info that any neighbour has and remove everything less
         // than it
         let our_oldest_ver = self.their_knowledge.values().min().map_or(0, |&v| v);
-        // TODO since the list is sorted according to the version, choose something more efficient
-        self.our_infos
-            .retain(|&(ref si, _)| *si.version() >= our_oldest_ver);
+        let oldest_ver_index = self
```
maqi: First, the result from this calculation is not `the index of the oldest version` but shall be `the index of the first version older than our_oldest_ver` ? so the variable name is little bit mis-leading?

Second, in case we have a list of versions of [0, 1, 2, 3, 5, 8], and the our_oldest_ver to be 7.
The previous retain approach will give us a result of [8].
However the drain approach will give us [0, 1, 2, 3, 5, 8], because the result from binary_search is `Err(4)` which will then be replaced as `0` and cause no entry to be drained.
To make the behaviour consistent, maybe we need somthing like:
```
let oldest_ver_index = match self
    .our_infos
    .binary_search_by_key(&our_oldest_ver, |&(ref si, _)| *si.version())
{
    Ok(index) => index,
    Err(index) => cmp::min(index + 1, self.our_infos.len()),
};
```

If we don't want to do so, maybe worth to put a comment, mentioning the `non-equal` situation won't happen?

d1vyank: I'll change this variable name to `version_retention_threshold_index` to be more clear.

In another discussion above, we agreed that it would be safer to not drop anything in the non-equal situation you described and add a `log_or_panic` to warn us if this pathological case arises. @madadam, @afck, thoughts?

madadam: Yes, I agree with this change now.

src/chain/chain.rs
```diff
@@ -371,9 +371,19 @@ impl Chain {
...
+            .our_infos
+            .binary_search_by_key(&our_oldest_ver, |&(ref si, _)| *si.version())
```
madadam: Minor: you can simplify the lambda params to just `|(si, _)|` due to match ergonomics - the compiler can figure the `ref`s and `&`s on its own.

src/chain/chain.rs
```diff
@@ -371,9 +371,19 @@ impl Chain {
...
+            .unwrap_or_else(|_index| {
+                if our_oldest_ver > 0 {
+                    log_or_panic!(
+                        LogLevel::Warn,
+                        "Oldest version indicated by neighbours not found in our info"
```
madadam: Nitpick: "...not found in our info**s**".

src/chain/chain.rs
Viv-Rajkumar: its actually the `NeighbourConfirm` rpc which would be updating their_knowledge.

`NeighbourInfo` is just an empty(well with a digest to not get filtered) RoutingMessage rpc used to explicitly notify a neighbour of churn to update their neighbour_infos. Its handling(generic to any other rpc) occurs at the signed_msg layer to discover a neighbours new SectionInfo(the portion of code in your previous PR where the TODO was for the extra clone). 

As for NeighbourConfirm, while its handling code exists `handle_neighbour_confirm`, the rpc is currently not sent out when handling a new sec_info event intentionally as that would affect unack_msgs potentially not being able to get resent. With the msg_relay changes getting discussed/ongoing, once that gets into code, unacked_msgs shouldnt exist and this RPC should be ok to get enabled to be dispatched and have their_knowledge active.

src/chain/chain.rs
d1vyank: Done

src/chain/chain.rs
d1vyank: Done

src/chain/chain.rs
d1vyank: > First, the result from this calculation is not the index of the oldest version but shall be the index of the first version older than our_oldest_ver ? so the variable name is little bit mis-leading?

Done

PR #1638, opened 2019-05-23
define CANDIDATE_EXPIRED_TIMEOUT const

maqi:
This PR contains work convering the cleanup point of defining CANDIDATE_EXPIRED_TIMEOUT const

### DIFF
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -13,7 +13,6 @@ use crate::{
     id::PublicId,
     resource_prover::RESOURCE_PROOF_DURATION,
     routing_table::Authority,
-    signature_accumulator::ACCUMULATION_TIMEOUT,
     time::{Duration, Instant},
     types::MessageId,
     utils::LogIdent,
@@ -32,6 +31,10 @@ use std::{
 
 /// Time (in seconds) after which a joining node will get dropped from the map of joining nodes.
 const JOINING_NODE_TIMEOUT_SECS: u64 = 900;
+/// Duration after which a candidate is considered as expired.
+/// Using a larger timeout to allow Online to accumulate via gossip.
+const CANDIDATE_EXPIRED_TIMEOUT: Duration =
+    Duration::from_secs(RESOURCE_PROOF_DURATION.as_secs() + 90);
 /// Time (in seconds) after which the connection to a peer is considered failed.
 const CONNECTING_PEER_TIMEOUT_SECS: u64 = 150;
 /// Time (in seconds) the node waits for a peer to either become valid once connected to it or to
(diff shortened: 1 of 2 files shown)


src/peer_manager.rs
```diff
@@ -314,9 +317,9 @@ impl Candidate {
             | Candidate::ResourceProof {
                 res_proof_start, ..
             } => {
-                // TODO: need better fix. Using a larger timeout to allow Online to accumulate via gossip
-                // than the prev timeout for grp-msg accumulation.
-                res_proof_start.elapsed() > RESOURCE_PROOF_DURATION + ACCUMULATION_TIMEOUT * 3
+                // TODO: need better fix. Using a larger timeout to allow Online to accumulate via
```
jeanphilippeD: Need to remove TODO

src/peer_manager.rs
```diff
@@ -314,9 +317,9 @@ impl Candidate {
...
+                // gossip than the prev timeout for grp-msg accumulation.
+                res_proof_start.elapsed() > RESOURCE_PROOF_DURATION + CANDIDATE_EXPIRED_TIMEOUT
```
jeanphilippeD: I am not 100% sure about the name or the comment explaining the name.
With the name and comment, it feels like it should read `res_proof_start.elapsed() > CANDIDATE_EXPIRED_TIMEOUT`.

Maybe you could have CANDIDATE_EXPIRED_TIMEOUT = RESOURCE_PROOF_DURATION + Duration::from_secs(90);

src/peer_manager.rs
```diff
@@ -32,6 +32,8 @@ use std::{
 
 /// Time (in seconds) after which a joining node will get dropped from the map of joining nodes.
 const JOINING_NODE_TIMEOUT_SECS: u64 = 900;
+/// Duration (in seconds) after which a candidate is considered as expired.
+const CANDIDATE_EXPIRED_TIMEOUT: Duration = Duration::from_secs(90);
```
jeanphilippeD: Did you ran soak test for a while?

src/peer_manager.rs
```diff
@@ -32,6 +31,11 @@ use std::{
 
 /// Time (in seconds) after which a joining node will get dropped from the map of joining nodes.
 const JOINING_NODE_TIMEOUT_SECS: u64 = 900;
+/// Duration (in seconds) after which a candidate is considered as expired.
```
jeanphilippeD: (in seconds) => Not needed, it is not in seconds.

src/peer_manager.rs
```diff
@@ -32,6 +31,11 @@ use std::{
...
+/// Using a larger timeout to allow Online to accumulate via gossip than the prev timeout for
```
jeanphilippeD: I think you can remove the part about ` than the prev timeout for grp-msg accumulation.`

src/peer_manager.rs
jeanphilippeD: Not fixed?

src/peer_manager.rs
jeanphilippeD: Not fixed?

jeanphilippeD:
Merged as part of #1639

PR #1640, opened 2019-05-23
Cover proxy split with extra hop

jeanphilippeD:


src/states/node/mod.rs
```diff
@@ -1538,7 +1538,12 @@ impl Node {
         sent_to: BTreeSet<XorName>,
     ) -> Result<(), RoutingError> {
         if !self.crust_service().is_connected(target) {
-            trace!("{} Not connected to {:?}. Dropping peer.", self, target);
+            trace!(
+                "{} Not connected to {:?}. Dropping peer and message: {:?}",
+                self,
+                target,
+                &signed_msg
```
Viv-Rajkumar: wasnt just `signed_msg.routing_message()` enough here if we wanted msg details, or were you interested in things to do with proving_sections/... they normally tend to be the large part of messages for log spam.

tests/mock_crust/mod.rs
```diff
@@ -206,48 +207,155 @@ fn multi_split() {
...
+        rng.shuffle(&mut sections);
+        sections
+    };
+
+    simultaneous_joining_nodes(network, nodes, &sections, false);
+}
+
+#[test]
+fn simultaneous_joining_nodes_three_section_with_one_ready_to_split() {
+    let _ = maidsafe_utilities::log::init(false);
```
Viv-Rajkumar: should be gone I guess

src/states/node/mod.rs
jeanphilippeD: No I was only interested in the first part, will update.

tests/mock_crust/mod.rs
jeanphilippeD: Yes.

tests/mock_crust/mod.rs
```diff
@@ -203,48 +214,197 @@ fn multi_split() {
...
+        },
+    ];
+    simultaneous_joining_nodes(network, nodes, &nodes_to_add_setup);
+}
+
+#[test]
+fn simultaneous_joining_nodes_three_section_with_one_ready_to_split() {
+    // TODO: Use same section size once we have a reliable message relay that handle split.
+    // Allow for more route hops otherwise NodeApproaval get losts.
+    let min_section_size = MIN_SECTION_SIZE + 1;
```
maqi: typo: NodeApproaval -> NodeApproval

could you pls explain bit more regarding why there is lost of NodeApproval, and more route hop will avoid lost? thx

tests/mock_crust/utils.rs
```diff
@@ -621,22 +608,134 @@ pub fn add_connected_nodes_until_split(
...
+        | Event::TimerTicked
+        | Event::SectionSplit(..) => (),
+        event => panic!("Got unexpected event: {:?}", event),
+    });
+    clear_relocation_overrides(nodes);
+
+    trace!("Created testnet comprising {:?}", prefixes);
+}
+
+// Add connected nodes to the given prefixes, so adding one extra node in any of the
```
maqi: `till adding one extra...` ?

tests/mock_crust/utils.rs
```diff
@@ -621,22 +608,134 @@ pub fn add_connected_nodes_until_split(
...
+// returned sub-prefixes, will trigger a split in the parent prefix.
+pub fn add_connected_nodes_until_one_away_from_split(
+    network: &Network<PublicId>,
+    nodes: &mut Vec<TestNode>,
+    prefixes_to_nearly_split: &[Prefix<XorName>],
+    use_cache: bool,
```
maqi: maybe don't need this flag?  the function is only be used once, with the flag being false

tests/mock_crust/utils.rs
```diff
@@ -621,22 +608,134 @@ pub fn add_connected_nodes_until_split(
...
+    for node in nodes.iter_mut() {
         node.inner.set_next_relocation_dst(None);
         node.inner.set_next_relocation_interval(None);
     }
+}
 
-    trace!("Created testnet comprising {:?}", prefixes);
+// Returns how many nodes to add to each sub-prefixes to reach split with one extra node.
+// The second field of the return value contain the sub-prefixes to add the final node to split.
+fn prefixes_and_count_to_split_with_only_one_extra_node(
```
maqi: I think the returned value in `Vec<PrefixAndSize>` is the `expected target section size`, not `how many nodes to add` ?

tests/mock_crust/utils.rs
```diff
@@ -621,22 +608,134 @@ pub fn add_connected_nodes_until_split(
...
+    use_cache: bool,
+) -> Vec<Prefix<XorName>> {
+    let (prefixes_and_counts, prefixes_to_add_to_split) =
+        prefixes_and_count_to_split_with_only_one_extra_node(nodes, prefixes_to_nearly_split);
+
+    add_connected_nodes_until_sized(network, nodes, &prefixes_and_counts, use_cache);
+    prefixes_to_add_to_split
+}
+
+// This adds new nodes until we reach the requested size for each prefix. No split expected.
```
maqi: => `keeps adding new nodes till reach the requested size ...` ?

tests/mock_crust/mod.rs
```diff
@@ -203,48 +214,197 @@ fn multi_split() {
...
-        if prefix0.matches(&node.name()) {
-            node.inner.set_next_relocation_dst(Some(name0));
-        } else {
-            node.inner.set_next_relocation_dst(Some(name1));
-        }
-    }
+struct SimultaneousJoiningNode {
+    // relocation_dst to set for nodes in the section with src_section_prefix: Must match section.
+    dst_section_prefix: Prefix<XorName>,
+    // section prefix that will contain the new node to add: Must match section.
```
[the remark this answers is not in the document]
jeanphilippeD: It means that this must be the prefix of an existing section.

`new node` mean the new joining node. The node will start up with an identity that is within the src_section: It does not join it, it sends it a Relocate request. The src_section will then send an ExpectCandidate to the dst_section.

tests/mock_crust/mod.rs
```diff
@@ -203,48 +214,197 @@ fn multi_split() {
...
+    src_section_prefix: Prefix<XorName>,
+    // The relocation_interval to set for nodes in the section with dst_section_prefix:
+    // Must be within dst_section_prefix. If none let production code decide.
```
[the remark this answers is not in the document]
jeanphilippeD: We don't, this is a pre-condition for the test to work.
It is something that the person writing the test must do (it is not very difficult, and the test will likely fail if it is not done).

tests/mock_crust/mod.rs
```diff
@@ -203,48 +214,197 @@ fn multi_split() {
...
-            break;
-        }
+// Proceed with testing joining nodes at the same time with the given configuration.
+fn simultaneous_joining_nodes(
+    network: Network<PublicId>,
+    mut nodes: Nodes,
+    nodes_to_add_setup: &[SimultaneousJoiningNode],
+) {
+    //
+    // Arrange
```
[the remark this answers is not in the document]
jeanphilippeD: This is part of the Arrange/Act/Assert pattern (AAA Pattern).
The Arrange is the setup phase were things needed for the test, but not the thing we are really testing, are setup.

tests/mock_crust/mod.rs
```diff
@@ -203,48 +214,197 @@ fn multi_split() {
...
+    rng.shuffle(&mut nodes);
+
+    let mut nodes_to_add = Vec::new();
+    for setup in nodes_to_add_setup {
+        // Set the specified relocation destination on the nodes of the given prefixes
+        let relocation_dst = setup.dst_section_prefix.substituted_in(rng.gen());
+        nodes_with_prefix_mut(&mut nodes, &setup.src_section_prefix)
+            .for_each(|node| node.inner.set_next_relocation_dst(Some(relocation_dst)));
+
+        // Set the specified relocation interval on the nodes of the given prefixes
```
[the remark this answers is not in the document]
jeanphilippeD: I don't think it would be correct.
We are selecting the nodes `of a given prefix` and setting the specified `relocation interval` to each of them.

tests/mock_crust/mod.rs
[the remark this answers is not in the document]
maqi: > It means that this must be the prefix of an existing section.
Then probably no need to have the word of `Must match section`? the word of `Must match section` does clarifify anything to me, but causing more confustion. :(

tests/mock_crust/mod.rs
jeanphilippeD: I'll try to fix the comment to be a bit clearer.

I'll explain detail here, as I don't think it will be really relevant to the code.
The problem is during split it is possible to loose messages with our current implementation because some node will disconnect. Adding an extra route mean that there are more chances for the message to go through after the split (it is not a full fix but makes it unlikely enough that it will do until we switch to N/3 routing).

tests/mock_crust/utils.rs
jeanphilippeD: It is consistent with the other similar functions.
It seemed reasonable to keep a consistent interface.

tests/mock_crust/utils.rs
jeanphilippeD: Correct, will fix.

tests/mock_crust/mod.rs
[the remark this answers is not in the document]
jeanphilippeD: Ok, I'll reword slightly.

tests/mock_crust/utils.rs
jeanphilippeD: I'll reword

tests/mock_crust/mod.rs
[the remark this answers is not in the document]
maqi: hmm...   first I only see this function using this, and there is no Act part marked out?
Second, there is 
```
    // Arrange
    // Add new nodes and process until complete
```
below. 
So maybe as the same put one extra line explaining what is being arranged?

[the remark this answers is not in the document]
jeanphilippeD: Got that wrong, that second Arrange should be Act.
I'll also add a line explaining the setup part

PR #1643, opened 2019-05-27
clean up timer to vote node is approved

jeanphilippeD:
candidate_timer_token was used before we moved to Parsec to ensure
the RPC from different elders would be sent at the same time.

This is no longer needed with Parsec: Send it when resource proof is
complete.

### DIFF
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -459,7 +459,9 @@ impl PeerManager {
                 ref mut passed_our_challenge,
                 ref res_proof_start,
                 ..
-            } if new_pub_id == pub_id => (challenge, passed_our_challenge, res_proof_start),
+            } if *passed_our_challenge || new_pub_id == pub_id => {
+                (challenge, passed_our_challenge, res_proof_start)
+            }
             _ => return Err(RoutingError::UnknownCandidate),
         };
 
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -446,90 +446,48 @@ impl PeerManager {
     /// elapsed since the candidate was inserted.
     pub fn verify_candidate(
         &mut self,
-        new_pub_id: &PublicId,
+        new_public_id: &PublicId,
         part_index: usize,
         part_count: usize,
         proof_part: Vec<u8>,
         leading_zero_bytes: u64,
-    ) -> Result<Option<(usize, u8, Duration)>, RoutingError> {
-        let (challenge, passed_our_challenge, res_proof_start) = match self.candidate {
-            Candidate::ResourceProof {
-                new_pub_id: ref pub_id,
-                challenge: Some(ref mut challenge),
-                ref mut passed_our_challenge,
-                ref res_proof_start,
-                ..
-            } if *passed_our_challenge || new_pub_id == pub_id => {
-                (challenge, passed_our_challenge, res_proof_start)
-            }
-            _ => return Err(RoutingError::UnknownCandidate),
-        };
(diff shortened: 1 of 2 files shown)


src/peer_manager.rs
```diff
@@ -446,90 +446,50 @@ impl PeerManager {
     /// elapsed since the candidate was inserted.
     pub fn verify_candidate(
         &mut self,
-        new_pub_id: &PublicId,
```
maqi: we still have many other places using this `new_pub_id`, do they need to be replaced as well?

src/peer_manager.rs
```diff
@@ -446,90 +446,50 @@ impl PeerManager {
...
+                    challenge,
+                    passed_our_challenge,
+                    res_proof_start,
+                    OnlinePayload {
+                        new_public_id: *new_pub_id,
+                        client_auth: *new_client_auth,
+                        old_public_id: *old_pub_id,
+                    },
+                ),
+                _ => return Err(RoutingError::UnknownCandidate),
```
maqi: In case the `passed_our_challenge` is `true`, then the Error shall not be `UnknownCandidate` ?
or such case won't happen?

src/peer_manager.rs
jeanphilippeD: I replaced that one for convenience because it clashed with the other and required `new_pub_id: ref pub_id,`. I don't think we need to change the other ones soon.

src/peer_manager.rs
jeanphilippeD: Yes, this replace:

```rust
        if self.candidate_timer_token.is_none() {
            debug!(
                "{} Won't handle resource proof response from {:?} - not currently waiting.",
                self, pub_id
            );
            return;
        }
```

Once we completed the challenge, we stop processing for that candidate.
It would happen if the candidate had send more or duplicate ResourceProofResponse.
What we should do is ignore them and log something.

maqi: But previously we just return siliently.
Now, it will be returned with an error. And the caller will generate a log of:
```
                debug!(
                    "{} Failed to verify candidate {}: {:?}",
                    self, pub_id, error
                );
```
which I think is quite misleading?

jeanphilippeD: Previously, the caller would have done the filtering and generating a log, never calling this function.
Now we always call this function and an error then get a debug message similar to what happen when we receive an unexpected ResourceProofResponse, which seem reasonable.

(BTW, Most of this code will change again soon to implement the rest of the impl spec. Not a reason to make nonsensical changes though, I thought it was a sensible thing.)

maqi: I am just concerned that this in-accurate `Failed to verify candidate` logs (and will be many) could be noisy when we need to carry out the debugging work. Especially when got deployed to testnet. (an error msg that could be generated many times during the normal process will blow up the log file a lot and could be a burden to the debugging work.)

jeanphilippeD: We replace a debug log with another debug log, so it should not impact the size of the logs.
Also, it should basically not happen with our implementation right now because we do not send the same ResourceProofResponse twice.

It is not really in-accurate, we received a ResourceProof response and failed to verify that candidate then (we happen to have verified it before).

I can update the log message wording if you have a suggestion?

maqi: Well, as you say, if `verified it before` can be considered as `failed to verify`, and such case shall not happen (at least not often in normal) as we don't send ResourceProofResponse  twice, then I am ok with leave the code as it is now.
No need to reword the log message.

PR #1646, opened 2019-05-28
refactor/data: move ImmutableData to safe-nd

m-cat:
Closes https://github.com/maidsafe/safe_client_libs/issues/809


jeanphilippeD:
Do we want it in the fleming branch as well/instead?

m-cat:
> Do we want it in the fleming branch as well/instead?

From @ustulation :

> Yes you can do it for the fleming branch too. But i see it this way - it's highly unlikely it'll be ready by the time SCL is done, which os the only super critical thing right now. Anything else is just a scope creep and can be tackled later. The goal here is to get SCL running without disrupting stuff elsewhere and by 13th. In that light anyone from the routing team itself is welcome to port it to fleming branch else we will ofc come back to it later and do it.

pierrechevalier83:
I agree with @jeanphilippeD : We probably want this in `fleming` instead, which is kind of acting as our `master` branch. It may make the second commit unnecessary.

Otherwise, the change is looking good :heavy_plus_sign:

pierrechevalier83:
Closing for now. A new PR will be raised once the safe-nd API is ready, and it will be based on the fleming branch.

PR #1647, opened 2019-05-29
Extract shared state

madadam:
This PR adds a new struct - `SharedState` - which contains all the eventually-consistent data shared among all elders of a section. 

It also changes the way `our_infos` are stored to statically guarantee they are never empty, which simplifies some code and allows to remove one `unwrap` from production code.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -7,7 +7,7 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use super::{
-    shared_state::{SectionChange, SharedState},
+    shared_state::{PrefixChange, SharedState},
     GenesisPfxInfo, NeighbourSigs, NetworkEvent, Proof, ProofSet, ProvingSection, SectionInfo,
 };
 use crate::error::RoutingError;
@@ -220,7 +220,7 @@ impl Chain {
                 // and we'd want to perform the merge eventually with our current latest state.
                 let our_hash = *self.state.new_info.hash();
                 let _ = self.state.merging.insert(our_hash);
-                self.state.change = SectionChange::Merging;
+                self.state.change = PrefixChange::Merging;
             }
             NetworkEvent::NeighbourMerge(digest) => {
                 // TODO: Check that the section is known and not already merged.
@@ -235,7 +235,7 @@ impl Chain {
     /// If we need to split also returns an additional sibling `SectionInfo`.
     /// Should not be called while a pfx change is in progress.
     pub fn add_member(&mut self, pub_id: PublicId) -> Result<Vec<SectionInfo>, RoutingError> {
-        if self.state.change != SectionChange::None {
+        if self.state.change != PrefixChange::None {
             log_or_panic!(
                 LogLevel::Warn,
                 "Adding {:?} to chain during pfx change.",
@@ -256,7 +256,7 @@ impl Chain {
 
         if self.should_split(&members)? {
             let (our_info, other_info) = self.split_self(members.clone())?;
-            self.state.change = SectionChange::Splitting;
+            self.state.change = PrefixChange::Splitting;
             return Ok(vec![our_info, other_info]);
         }
 
@@ -272,7 +272,7 @@ impl Chain {
     /// Removes a member from our section, creating a new `our_info` in the process.
     /// Should not be called while a pfx change is in progress.
     pub fn remove_member(&mut self, pub_id: PublicId) -> Result<SectionInfo, RoutingError> {
-        if self.state.change != SectionChange::None {
+        if self.state.change != PrefixChange::None {
             log_or_panic!(
                 LogLevel::Warn,
                 "Removing {:?} from chain during pfx change.",
@@ -300,7 +300,7 @@ impl Chain {
         if self.state.new_info.members().len() < self.min_sec_size {
             // set to merge state to prevent extending chain any further.
             // We'd still not Vote for OurMerge until we've updated our_infos
-            self.state.change = SectionChange::Merging;
+            self.state.change = PrefixChange::Merging;
         }
 
         Ok(self.state.new_info.clone())
@@ -356,7 +356,7 @@ impl Chain {
         }
 
         self.check_and_clean_neighbour_infos(None);
-        self.state.change = SectionChange::None;
+        self.state.change = PrefixChange::None;
 
         let completed_events = mem::replace(&mut self.completed_events, Default::default());
         let chain_acc = mem::replace(&mut self.chain_accumulator, Default::default());
@@ -399,7 +399,7 @@ impl Chain {
     }
 
     /// Returns whether our section is in the process of changing (splitting or merging).
-    pub fn change(&self) -> SectionChange {
+    pub fn prefix_change(&self) -> PrefixChange {
         self.state.change
     }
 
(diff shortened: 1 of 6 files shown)


src/chain/shared_state.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub fn our_info_by_hash(&self, hash: &Digest256) -> Option<&SectionInfo> {
+        self.our_infos
+            .iter()
+            .find(|(sec_info, _)| sec_info.hash() == hash)
+            .map(|(sec_info, _)| sec_info)
+    }
+
+    /// Returns the section info matching our own name with the given version number.
+    pub fn our_info_by_version(&self, version: u64) -> Option<&SectionInfo> {
+        // TODO: Binary search? Reverse order? Benchmark which one is fastest in practice.
```
jeanphilippeD: Do we need that TODO, either it will show in profiling as an issue and we can fix it or it won't and we don't want to do anything. (Can be follow up commit because it is copied).

src/chain/shared_state.rs
```diff
@@ -0,0 +1,174 @@
...
+        self.our_infos
+            .iter()
+            .find(|(sec_info, _)| *sec_info.version() == version)
+            .map(|(sec_info, _)| sec_info)
+    }
+
+    /// Remove our infos that are older than the given version.
+    pub(super) fn clean_our_infos(&mut self, oldest_version: u64) {
+        // TODO since the list is sorted according to the version, choose something more efficient
```
jeanphilippeD: Will this clash with PR 1635. (We probably should rebase 1647 on top of it).
Otherwise I'd probably say the same.

src/chain/chain.rs
```diff
@@ -408,35 +381,29 @@ impl Chain {
...
     pub fn our_prefix(&self) -> &Prefix<XorName> {
-        self.opt_our_info()
-            .map_or(&DEFAULT_PREFIX, SectionInfo::prefix)
+        self.state.our_prefix()
     }
 
-    /// Returns our current chain state.
-    pub fn state(&self) -> &ChainState {
+    /// Returns the shared state of the section.
+    pub fn shared_state(&self) -> &SharedState {
```
jeanphilippeD: Not sure why we made the full SharedState accessible while before we only allow ChainState.

src/chain/chain.rs
```diff
@@ -712,7 +654,7 @@ impl Chain {
             | NetworkEvent::Offline(_)
             | NetworkEvent::ExpectCandidate(_)
             | NetworkEvent::PurgeCandidate(_) => {
-                self.state() == &ChainState::Normal && self.our_info().is_quorum(proofs)
```
jeanphilippeD: Not sure we needed ChainState -> SectionChange as part of this change.
Not 100% sure it helps clarity, in particular, the Section does change when adding or removing a member, but it will have SectionChange::None. (Also make the change bigger than strictly needed).

src/states/common/approved.rs
```diff
@@ -168,7 +168,7 @@ pub trait Approved: Relocated {
                 }
             }
 
-            match self.chain_poll_all(outbox)? {
+            match self.chain_poll(outbox)? {
```
jeanphilippeD: Why that change here?

src/chain/shared_state.rs
```diff
@@ -0,0 +1,174 @@
...
+        let pfx = self.our_prefix();
+        if pfx.is_empty() || self.change == SectionChange::Splitting {
+            return false;
+        }
+
+        if self.our_info().members().len() < min_section_size {
+            return true;
+        }
+
+        for si in neighbour_infos {
```
jeanphilippeD: Not 100% sure about moving from any algo to a for loop help clarify things.

src/chain/shared_state.rs
madadam: It was there before so didn't want to lose it in case we want to address it at some point (which is out of scope of this PR).

src/chain/shared_state.rs
madadam: I'll address it when #1635 is merged.

src/chain/chain.rs
madadam: I can remove it, but I anticipate we are going to need it eventually (at least in tests).

src/chain/chain.rs
madadam: The motivation for the rename was to avoid the clash/confusion between `SharedState` and `ChainState` (and the corresponding variable names). I find `SectionChange` a decent enough name as it more explicitly describes what's going on ("state" is too generic). I'm always open to better suggestion though.

Regarding adding/removing elders: `SectionChange::None` means there is no ongoing change in progress. Adding/removing elders are instantaneous changes which don't involve any ongoing process.

src/states/common/approved.rs
madadam: This was a leftover from the state refactoring PRs. It should have been renamed back then, but it slipped through. As it is a trivial change, I though sneaking it into this PR is OK.

src/chain/shared_state.rs
madadam: Matter of taste, but I think `for` is cleaner for simple loops. Anyway, it's questionable in this case, so I'll revert it to `any`.

src/chain/chain.rs
jeanphilippeD: It is not so immediate though, when a candidate is voted Online, then it is added to Parsec, and the new_info, then the section info is voted, and then it is consensused. But if that helped for now, I'm ok with the rename, can be fixed latter if it is becoming an issue.

src/chain/shared_state.rs
```diff
@@ -172,3 +167,61 @@ pub enum SectionChange {
...
+        self.head.iter().chain(iter::once(&self.tail))
+    }
+
+    pub fn last(&self) -> &T {
+        &self.tail
+    }
+
+    // Retains only elements matching the predicate. If no element matches, the last one is
+    // retained anyway, to maintain the non-empty invariant.
+    pub fn retain<F>(&mut self, mut f: F)
```
jeanphilippeD: I'm really not so sure about that NonEmptyList.
We now have this wide contract for retain that somehow keep the last element.

It seems strange to me as a level of abstraction.
It seems either we should have a more specific OurInfo object with private member(s) and the invariant held by the interface and providing roughly this interface without retain, but instead provide a `keep_version_above` that would not fully empty the vector. (You could then have a simpler function than `retain` because you know to keep the `tail` always). 

Or make the member of SharedState private and hold invariant there.

src/chain/shared_state.rs
jeanphilippeD: I was thinking ok, but since this code is moved, I'd rather fix that in this PR. Unless you think this TODO is valid. (I'd rather avoid getting Performance fix that are not needed).

jeanphilippeD: I was thinking ok, but since this code is moved, I'd rather fix that in this PR. Unless you think this TODO is valid. (I'd rather avoid getting Performance fix that are not needed).

@Viv-Rajkumar could you comment?

madadam: I'm against addressing the comment in this PR, as it can be hugely out of scope (as to do it properly it would require profiling and/or benchmarking which is not trivial). I'm OK with just deleting the comment. We should have a separate performance pass for routing at some point anyway.

jeanphilippeD: Yes, I did not mean to ask you to address the comment, only remove it.

src/chain/chain.rs
Viv-Rajkumar: I find the name `SectionChange` a bit vague too. maybe even `SectionStatus` and change None to Normal or `PrefixChange` where even None could make sense. the point with the ChainState::Normal was to indicate normal activity which ofc goes into a stalled state when we're changing prefixes in Split/Merge state. SectionChange::None in that convention sounds like we've stopped mutating this section. minor point anyways tho as long as the meaning behind it isnt lost.

PR #1648, opened 2019-05-29
split candidate information that should be shared state

jeanphilippeD:


src/chain/candidate.rs
```diff
@@ -0,0 +1,141 @@
...
+            };
+            true
+        } else {
+            // Unkwown candidate, Candidate was purged before: refuse it.
+            false
+        }
+    }
+
+    /// The public id of the candidate we are waiting to approve.
+    pub fn waiting_candidate_old_public_id(&self) -> Option<&PublicId> {
```
madadam: This method might be confusingly named. "waiting candidate" to me suggest it refers to the `ApprovedWaitingSectionInfo` variant, but that variants yields `None`. I think it would be renamed to `accepted_candidate_old_public_id` (or just `accepted_old_public_id`).

src/states/elder/mod.rs
```diff
@@ -2157,7 +2179,7 @@ impl Elder {
     }
 
     pub fn set_next_relocation_interval(&mut self, interval: Option<(XorName, XorName)>) {
```
madadam: Why not change this method to take `XorTargetInterval` too?

src/states/elder/mod.rs
```diff
@@ -1304,7 +1323,7 @@ impl Elder {
         let src = Authority::Section(vote.dst_name);
         let dst = vote.old_client_auth;
         let content = MessageContent::RelocateResponse {
-            target_interval: target_interval,
+            target_interval: (target_interval.0, target_interval.1),
```
madadam: Why not change the type of this field to `XorTargetInterval` too?

src/utils.rs
```diff
@@ -59,6 +59,9 @@ impl Display for LogIdent {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
+pub struct XorTargetInterval(pub XorName, pub XorName);
```
madadam: I think this name is unnecessarily too specific. What makes this type a "target" interval? I would consider naming it just `XorInterval`.

That said, I'd actually prefer not to define new type for this at all and use [`RangeInclusive`](https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html)`<XorName>` instead:
1. It would be obvious that the upper bound is inclusive (currently not obvious, as in rust ranges/intervals are more often upper-exclusive)
2. It gives you some convenience, like the [`contains`](https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html#method.contains) method, which can help avoid subtle bugs due to the inclusive/exclusive confusion.

src/chain/candidate.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        *self = Candidate::AcceptedForResourceProof {
+            old_public_id: old_public_id,
+            target_interval: target_interval,
+        };
+    }
+
+    /// Return the target interval if we are waiting for candidate info for that PublicId.
+    pub fn waiting_for_candidate_interval(
```
madadam: If you agree with https://github.com/maidsafe/routing/pull/1648/files#r288895197, then maybe also rename this to `accepted_target_interval` (or similar).

src/chain/candidate.rs
```diff
@@ -0,0 +1,141 @@
...
+        old_public_id: PublicId,
+    },
+    /// We consensused the candidate online. We are waiting for the SectionInfo to consensus
+    /// and this new node to start handling events before allowing a new candidate.
+    ApprovedWaitingSectionInfo { new_pub_id: PublicId },
+}
+
+impl Candidate {
+    /// Return true if already has a candidate
+    pub fn has_resource_proof_candidate(&self) -> bool {
```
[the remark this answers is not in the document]
jeanphilippeD: Because all of these public function are forwarded from chain, I'd rather keep the candidate in them. Last time we had similar discussion we decided that it was better to have the forwarding function and the function have the same name, so I apply the same logic (not 100% sure it is the best, or if we should apply the same logic).

src/chain/candidate.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub fn accept_as_candidate(
+        &mut self,
+        old_public_id: PublicId,
+        target_interval: XorTargetInterval,
+    ) {
+        if self.has_resource_proof_candidate() {
+            log_or_panic!(
+                LogLevel::Error,
+                "accept_as_candidate when processing one already"
+            );
```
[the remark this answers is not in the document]
jeanphilippeD: We broke a precondition, so everything is off.
But mostly, the implicit post-condition is the candidate was set and that would break that.

src/chain/candidate.rs
jeanphilippeD: sounds good.

src/chain/candidate.rs
jeanphilippeD: Will go for `accepted_candidate_target_interval`.

src/states/elder/mod.rs
jeanphilippeD: It was slowly getting everywhere, I did not think it was warranted at this point to update the message and all its handlers.
If you think it is, I can add an extra commit?

src/states/elder/mod.rs
jeanphilippeD: Same reason, really, I did not really want to change all the test code, but if you think it is needed, I add an extra commit (or same commit as other suggestion).

src/utils.rs
jeanphilippeD: I like the idea of using RangeInclusive.
I'm not 100% sure the type is too specific, it is used often enough that it seem reasonable to use a dedicated `new type`.

So i'm a bit unsure on the resolution for this.

src/chain/candidate.rs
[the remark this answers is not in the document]
madadam: - Repeating the name of the receiver in the method name is unidiomatic (see the std library: there is `Vec::len`, not `Vec::vec_len`, `File::open`, not `File::open_file`, `slice::sort_by_key` not `slice::sort_slice_by_key`, etc... 
- I don't think the fact that these function are only forwarded to matter. In fact, this type should not even be aware of how it's used since that violates separation of concerns. 
- Also this sets a weird precedent. Should we now follow this pattern in the future in similar situations? If so when? Always? Sometimes? etc...
- Ultimately it just makes the code verbose for no good reason. Verbosity is good if it helps clarity, but that is dubious in this case.

Anyway, up to you. If you insist on this, I won't block the merge over it.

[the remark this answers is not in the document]
jeanphilippeD: I'll fix it up, it make sense to me. Probably interpreted some previous review wrong.

src/chain/candidate.rs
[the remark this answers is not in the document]
madadam: Not sure I agree. I think when a precondition is broken, doing nothing seems safer than doing something. Not 100% sure though. My preferred approach would be hard `panic` here. But that's a discussion for another time I guess.

src/states/elder/mod.rs
madadam: I wouldn't say it's needed, but it would make the code cleaner. Currently there are two different representations of xor-name intervals for no good reason which adds complexity and confusion.

src/utils.rs
madadam: I'd definitely use `RangeInclusive` if only so you can use `interval.contains(name)` instead of `name >= interval.0 && name <= interval.1`. I'd leave it up to you whether to use `RangeInclusive<XorName>` directly or via a typedef.

jeanphilippeD:
About `Needs to be rebased and the candidate field moved to SharedState`

jeanphilippeD:
About `Needs to be rebased and the candidate field moved to SharedState`
Because neighour_info is not in there yet, I don;t think we want to move candidate there just yet.

madadam:
> About Needs to be rebased and the candidate field moved to SharedState
Because neighour_info is not in there yet, I don;t think we want to move candidate there just yet.

Does `candidate` depend on (or is depended on by) `neighbour_info`? If so, then it makes sense to not do it yet. Otherwise I'd rather do.

jeanphilippeD:
> > About Needs to be rebased and the candidate field moved to SharedState
> > Because neighour_info is not in there yet, I don;t think we want to move candidate there just yet.
> 
> Does `candidate` depend on (or is depended on by) `neighbour_info`? If so, then it makes sense to not do it yet. Otherwise I'd rather do.

We use `neighbour_info` to decide whether to forward the ExpectCandidate or accept it and modify our candidate. So yes, candidate state depends on `neighbour_info`.

src/chain/candidate.rs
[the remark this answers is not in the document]
jeanphilippeD: Here are the fixed name in candidate:
The forwarding function will still have candidate in the name but some will likely be updated too.

Does that seem ok?

```
    /// Return true if no candidate info
    pub fn is_none(&self) -> bool {

    /// Forget about the current candidate.
    pub fn reset(&mut self) {

    /// Forget about the current candidate if it is a member of the given section.
    pub fn reset_if_member_of(&mut self, members: &BTreeSet<PublicId>) {

    /// Our section decided that the candidate should be resource proofed next.
    /// Pre-condition: is_none.
    pub fn accept_for_resource_proof(
        &mut self,
        old_public_id: PublicId,
        target_interval: XorTargetInterval,
    ) {


    /// Try to accept as memeber.
    /// If the candidate was already purged or is unexpected, return false.
    /// Otherwise marks the candidate as `ApprovedWaitingSectionInfo`.
    pub fn try_accept_as_member(&mut self, online_payload: &OnlinePayload) -> bool {

    /// Return the target interval if we are resource proofing for that old PublicId.
    pub fn matching_target_interval(&self, old_public_id: &PublicId) -> Option<&XorTargetInterval> {


    /// The public id of the candidate we are resource proofing.
    pub fn old_public_id(&self) -> Option<&PublicId> {


    /// The target interval of the candidate we are resource proofing.
    fn target_interval(&self) -> Option<&XorTargetInterval> {

    /// Logs info about ongoing candidate state, if any.
    pub fn show_status(&self, log_ident: &LogIdent) {
```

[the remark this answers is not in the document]
madadam: Perfect!

src/utils.rs
jeanphilippeD: I may do that in a follow up commit in this PR.

tests/mock_crust/mod.rs
```diff
@@ -249,9 +251,11 @@ fn simultaneous_joining_nodes(
         // Set the specified relocation interval on the nodes of the given prefixes
         let relocation_interval = setup
             .dst_relocation_interval_prefix
-            .map(|prefix| (prefix.lower_bound(), prefix.upper_bound()));
-        nodes_with_prefix_mut(&mut nodes, &setup.dst_section_prefix)
-            .for_each(|node| node.inner.set_next_relocation_interval(relocation_interval));
+            .map(|prefix| XorTargetInterval(prefix.lower_bound(), prefix.upper_bound()));
+        nodes_with_prefix_mut(&mut nodes, &setup.dst_section_prefix).for_each(|node| {
+            node.inner
+                .set_next_relocation_interval(relocation_interval.clone())
```
madadam: Can we `#[derive(Copy)]` for `XorTargetInterval`?

tests/mock_crust/utils.rs
```diff
@@ -872,7 +873,10 @@ fn add_node_to_section<T: Rng>(
     nodes.iter_mut().for_each(|node| {
         node.inner.set_next_relocation_dst(Some(relocation_name));
         node.inner
-            .set_next_relocation_interval(Some((prefix.lower_bound(), prefix.upper_bound())));
+            .set_next_relocation_interval(Some(XorTargetInterval(
+                prefix.lower_bound(),
+                prefix.upper_bound(),
+            )));
```
madadam: Seems like it might be worth adding a `From<Prefix>` impl for XorTargetInterval for convenience.

tests/mock_crust/mod.rs
jeanphilippeD: We could, but would it be a good thing?

tests/mock_crust/utils.rs
jeanphilippeD: Only used a few times, and later commits in that PR add an inclusive_range() function to prefix that is more generally used.

tests/mock_crust/mod.rs
madadam: It would avoid this `clone` here, for one thing. But not sure. `RangeInclusive` is not `Copy` so maybe this shouldn't be either? Up to you.

jeanphilippeD: I'll leave it out for now then.

PR #1649, opened 2019-05-31
Resource proofing with lagging elders

maqi:
This PR contains the work of making a joining_node resend ConnectionInfoReq and CandidateInfo to an elder in case of not having connection to it or hasn't started resource proofing with it. Which allows resource proofing to be carried out with lagging elders.
A test is also create to confirm that the update will work as expected.

### DIFF
--- a/src/states/elder/mod.rs
+++ b/src/states/elder/mod.rs
@@ -129,6 +129,8 @@ pub struct Elder {
     gen_pfx_info: GenesisPfxInfo,
     gossip_timer_token: u64,
     chain: Chain,
+    #[cfg(feature = "mock_base")]
+    ignore_candidate_info_counter: u8,
 }
 
 impl Elder {
@@ -228,6 +230,8 @@ impl Elder {
             gen_pfx_info: details.gen_pfx_info,
             gossip_timer_token,
             chain: details.chain,
+            #[cfg(feature = "mock_base")]
+            ignore_candidate_info_counter: 0,
         }
     }
 
@@ -1057,6 +1061,13 @@ impl Elder {
         new_client_auth: &Authority<XorName>,
         outbox: &mut EventBox,
     ) {
+        #[cfg(feature = "mock_base")]
+        {
+            if self.ignore_candidate_info_counter > 0 {
+                self.ignore_candidate_info_counter -= 1;
+                return;
+            }
+        }
         debug!(
             "{} Handling CandidateInfo from {}->{}.",
             self, old_pub_id, new_pub_id
@@ -2192,6 +2203,10 @@ impl Elder {
 
     pub fn has_resource_proof_candidate(&self) -> bool {
         self.chain.has_resource_proof_candidate()
+    }
+
+    pub fn set_ignore_candidate_info_counter(&mut self, counter: u8) {
+        self.ignore_candidate_info_counter = counter;
     }
 }
 
--- a/src/states/proving_node.rs
+++ b/src/states/proving_node.rs
@@ -36,10 +36,14 @@ use crate::{
     Service,
 };
 use maidsafe_utilities::serialisation;
+use std::collections::btree_map::BTreeMap;
+use std::time::Duration;
 use std::{
     collections::BTreeSet,
     fmt::{self, Display, Formatter},
 };
+
+const RESEND_TIMEOUT: Duration = Duration::from_secs(20);
 
 pub struct ProvingNodeDetails {
     pub action_sender: RoutingActionSender,
@@ -71,6 +75,8 @@ pub struct ProvingNode {
     resource_prover: ResourceProver,
     routing_msg_filter: RoutingMessageFilter,
     timer: Timer,
+    resource_proofing_status: BTreeMap<PublicId, bool>,
+    resend_token: Option<u64>,
 }
 
 impl ProvingNode {
@@ -103,6 +109,8 @@ impl ProvingNode {
             joining_prefix: details.our_section.0,
             old_full_id: details.old_full_id,
             resource_prover,
+            resource_proofing_status: BTreeMap::new(),
+            resend_token: None,
         };
         node.init(details.our_section.1, &details.proxy_pub_id, outbox);
         node
@@ -143,6 +151,7 @@ impl ProvingNode {
                 );
             }
         }
+        self.resend_token = Some(self.timer.schedule(RESEND_TIMEOUT));
     }
 
     pub fn into_establishing_node(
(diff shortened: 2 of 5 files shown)


src/states/elder/mod.rs
```diff
@@ -1057,11 +1061,26 @@ impl Elder {
...
+                self.ignore_candidate_info_counter -= 1;
+                return;
+            }
+        }
         debug!(
             "{} Handling CandidateInfo from {}->{}.",
             self, old_pub_id, new_pub_id
         );
 
+        if self.peer_mgr.has_candidate_info() {
```
jeanphilippeD: I do not think we want that here.
It seem the bug is either in `self.peer_mgr.handle_candidate_info`, or the match processing the result.

When we call this we already know this is a valid candidate we are processing since `self.chain.matching_candidate_target_interval` did not lead us to return early, so we probably should not disconnect it then.

src/states/proving_node.rs
```diff
@@ -248,11 +329,14 @@ impl Base for ProvingNode {
...
 
     fn handle_connect_failure(&mut self, pub_id: PublicId, outbox: &mut EventBox) -> Transition {
+        debug!("{} received connection failure from {:?}", self, pub_id);
+        let _ = self.resource_proofing_status.remove(&pub_id);
         RelocatedNotEstablished::handle_connect_failure(self, pub_id, outbox)
     }
 
     fn handle_lost_peer(&mut self, pub_id: PublicId, outbox: &mut EventBox) -> Transition {
         debug!("{} Received LostPeer - {}", self, pub_id);
+        let _ = self.resource_proofing_status.remove(&pub_id);
```
jeanphilippeD: Why is that need to change?

src/states/proving_node.rs
```diff
@@ -293,6 +377,9 @@ impl Base for ProvingNode {
                     difficulty,
                     log_ident,
                 );
+                if let Some(status) = self.resource_proofing_status.get_mut(&pub_id) {
```
jeanphilippeD: Would we be better off using a BTreeSet?

tests/mock_crust/mod.rs
```diff
@@ -172,6 +172,23 @@ fn node_joins_in_front() {
     verify_invariant_for_all_nodes(&mut nodes);
 }
 
+#[test]
+fn joining_node_with_ignoring_candidate_info() {
+    let network = Network::new(MIN_SECTION_SIZE, None);
+    let mut nodes = create_connected_nodes(&network, MIN_SECTION_SIZE);
+    let bootstrap_config = BootstrapConfig::with_contacts(&[nodes[0].handle.endpoint()]);
+    for node in nodes.iter_mut() {
```
jeanphilippeD: I think we may only want to do that on some of the nodes.

tests/mock_crust/utils.rs
```diff
@@ -418,6 +418,36 @@ pub fn poll_and_resend_until(
     panic!("poll_and_resend has been called {} times.", MAX_POLL_CALLS);
 }
 
+/// A special advancing pattern to test the ability of resending `CandidateInfo`.
+pub fn poll_and_resend_until_reduced_advance(nodes: &mut [TestNode]) {
```
jeanphilippeD: As discussed in mumble, best to not add a duplicate function and make that extra advance clock work in normal poll_and_resend_until.

jeanphilippeD:
Lots of good things, left a few comments. Maybe good to add Jon as reviewer.

src/states/proving_node.rs
maqi: If it's not received, the disconnected peer will be left over in the `resource_proofing_status` map, which will cause the `connection_info_req` to be sent out forever and exceeding the polling max_limitation

src/states/proving_node.rs
maqi: we need the status value to decide whether to send out CandidateInfo or not, so I think a map is required?

jeanphilippeD: Yes, we are populating the map initially and iterating over it. make sense,=.

tests/mock_crust/utils.rs
```diff
@@ -381,15 +381,17 @@ pub fn poll_all_until(
...
 }
 
 /// Polls and processes all events, until there are no unacknowledged messages left.
 /// should_stop: can be used for an early return from poll_and_resend
+/// extra_advance: this is so far only used for the ignoring candidate_info test.
 pub fn poll_and_resend_until(
     nodes: &mut [TestNode],
     clients: &mut [TestClient],
     should_stop: &Fn(&[TestNode]) -> bool,
+    mut extra_advance: Option<u64>,
```
jeanphilippeD: Why do we need that field? which test fail if we do it unconditionally?

jeanphilippeD: Lets keep it as you did for now, we can reconsider how best to improve it later.

tests/mock_crust/mod.rs
```diff
@@ -177,15 +180,18 @@ fn joining_node_with_ignoring_candidate_info() {
     let network = Network::new(MIN_SECTION_SIZE, None);
     let mut nodes = create_connected_nodes(&network, MIN_SECTION_SIZE);
     let bootstrap_config = BootstrapConfig::with_contacts(&[nodes[0].handle.endpoint()]);
-    for node in nodes.iter_mut() {
-        node.inner.set_ignore_candidate_info_counter(1);
+    // Make half of the elders ingoring the candidate_info to simulate lagging.
+    // Without the resending, this will block the joining node to be approved.
+    for i in 0..((nodes.len() + 1) / 2) {
```
jeanphilippeD: for node in nodes.iter_mut().take(2) {
...
}

src/states/elder/mod.rs
jeanphilippeD: I'm confused a bit, so what was the fix for that?

src/states/proving_node.rs
jeanphilippeD: Why did that change with this PR? why did it not happen before as well?

src/states/elder/mod.rs
maqi: This itself is not trying to fix a bug. It's mainly just my over-caustious about the duplicated candidate_info.
I double check the code flow after that point and the case of duplication shall be addressed already.
So I just removed this check here.

src/states/proving_node.rs
maqi: The `resource_proofing_status` map is introduced by this PR only. 

Previously, a disconnected connection is will have only one `disconnected` msg to be sent to peer.
But now, with the new `resending` attempt, if the entry doesn't get cleaned up, we will send out ConnectionInfo request each time reached timeout.

jeanphilippeD: Thanks, yes I got confused there.
still, should it not be set to false instead?

> But now, with the new resending attempt, if the entry doesn't get cleaned up, we will send out ConnectionInfo request each time reached timeout.

Won't it eventually stop when reaching the trying to prove themselves timeout?

maqi: >Won't it eventually stop when reaching the trying to prove themselves timeout
No. As long as the node is in the proving_node state and having the map which contains the entry, the connection_info_req will be sent out each time a timeout is reached.

tests/mock_crust/utils.rs
maqi: >Why do we need that field
that's for the `ignoring_candidate_info` test.
>which test fail if we do it unconditionally
the new `joining_node_with_ignoring_candidate_info` test

src/states/proving_node.rs
jeanphilippeD: Yes, that is what it is suppose to do, but does it not move to Terminated state?

tests/mock_crust/mod.rs
maqi: hmm... `2` only works for the `min_network_size` which is 4. 
I thought we want a more general pattern?
how about: `nodes.iter_mut().take((nodes.len() + 1) / 2)` ?

jeanphilippeD: Ok with your suggestion, but why the +1?
Why not `take(nodes.len() / 2)`?

src/states/proving_node.rs
maqi: > does it not move to Terminated state

No. there is no timeout on proving_node itself.
There is only timeout on the elders side which they just reject and disconnect.
But because the proving_node itself keeps sending out msg, the `poll` loop won't break and exceeds the max

tests/mock_crust/mod.rs
maqi: for `odd` numbers, `5 / 2 = 2`, which is `less than half`
maybe it doesn't matter, as we require supermajority.
I will use `nodes.len() / 2` then.

tests/mock_crust/mod.rs
```diff
@@ -172,6 +172,26 @@ fn node_joins_in_front() {
...
+    for node in nodes.iter_mut().take(MIN_SECTION_SIZE / 2) {
+        node.inner.set_ignore_candidate_info_counter(1);
+    }
+    nodes.push(
+        TestNode::builder(&network)
+            .bootstrap_config(bootstrap_config.clone())
+            .create(),
+    );
+    let dummy = |_nodes: &[TestNode]| false;
+    poll_and_resend_until(&mut nodes, &mut [], &dummy);
```
jeanphilippeD: Can use poll_and_resend now :-)

src/states/elder/mod.rs
jeanphilippeD: I forgot and meant to say:
This is a good place for a unit tests. You add a new test that follow the pattern of the existing one to exercise this corner case and verify it does the right thing.

PR #1676, opened 2019-06-05
Mock quic-p2p

madadam:
This PR adds a mock version of the quick-p2p crate. It does not replaces crust with quick-p2p yet. That will be done in followup PR(s).

Additional change:

- move all the mock functionality (mock crust, mock parsec and mock quick-p2p) to the top-level `mock` module for improved code organization. 
- remove unused `mock_crypto` module
- rename `tests/mock_crust` to `tests/mock_network` in preparation for crust removal
- small simplification in mock crust intended to make the quick-p2p integration slightly easier

New dependencies:

- bytes - needed as it is used by quick-p2p instead of `Vec<u8>`.
- fxhash (optional) - used only in mock quick-p2p to guarantee consistent iteration order of hash sets/maps which is necessary for repeatable test results.


src/mock/quick_p2p/tests.rs
```diff
@@ -0,0 +1,542 @@
...
+    let network = Network::new(None);
+
+    let mut a = Agent::node(&network);
+    let b_addr = network.gen_addr();
+
+    let msg = Bytes::from_static(b"message from A");
+    a.send(b_addr, msg.clone());
+    network.poll();
+
+    // Note: the real quick-p2p will only emit `UnsentUserMessage` when a connection to the peer
```
jeanphilippeD: That seem a bit weird, so we send a message, but because there is nothing, we will never know it was not sent?

src/mock/quick_p2p/tests.rs
```diff
@@ -0,0 +1,542 @@
...
+    b.expect_new_message(&a.addr(), &msg);
+}
+
+#[test]
+fn send_multiple_messages_without_connecting_first() {
+    let network = Network::new(None);
+    let mut a = Agent::node(&network);
+    let b = Agent::node(&network);
+
+    let msg0 = Bytes::from_static(b"message 0 from A");
```
jeanphilippeD: collection?
```
let msg_a = [b"message 0 from A", b"message 1 from A", b"message 2 from A"];
msg_a.iter().for_each(|msg| a.send(b.addr(), msg.clone()))

...
msg_a.iter().for_each(|msg| b.expect_new_message(&a.addr(), &msg)))
```

src/mock/quick_p2p/tests.rs
```diff
@@ -0,0 +1,542 @@
...
+
+    let msg2 = Bytes::from_static(b"message 2 from A");
+    a.send(b.addr(), msg2.clone());
+
+    network.poll();
+
+    a.expect_connected_to_node(&b.addr());
+    b.expect_connected_to_node(&a.addr());
+
+    b.expect_new_message(&a.addr(), &msg0);
```
jeanphilippeD: Is that always true? No re-ordering?

src/mock/quick_p2p/tests.rs
madadam: Agree, but it's an issue with the real quic-p2p and I wanted to keep the API identical. Might be worth raising it there (if not already raised).

src/mock/quick_p2p/tests.rs
madadam: Within the same node, `ConnectedTo` is guaranteed to be raised before `NewMessage`. There is no guarantee on the order different nodes will receive the messages, but that doesn't matter anyway.

jeanphilippeD: I mean is the mock doing no re-ordering while the real code may do so?
Is that correct?
Is that not a problem?

madadam: There is randomness in the order messages are processed (see `Network::poll`) which is supposed to simulate the uncertain order in the real world. So for example, when we have something like this:
```rust
a.send(c, "msg0");
b.send(c, "msg1");
```
then there is no guarantee which message will be delivered first. This non-guarantee is followed in this mock code too. 

However, in the case of this test, the `ConnectedTo`  is guaranteed to be raised before any `NewMessage` from the same peer. This is the case in the real code too.

src/mock/quick_p2p/tests.rs
```diff
@@ -0,0 +1,542 @@
...
+use super::{Builder, Config, Event, Network, NodeInfo, OurType, Peer, QuicP2p};
+use bytes::Bytes;
+use std::{
+    iter,
+    net::SocketAddr,
+    sync::mpsc::{self, Receiver, TryRecvError},
+};
+
+// Assert that the expression matches the expected pattern.
+macro_rules! assert_match {
```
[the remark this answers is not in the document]
madadam: Could you elaborate on what trouble this could cause? I find this macro very useful and wish it was part of the standard library.

src/mock/quick_p2p/tests.rs
jeanphilippeD: That is a different example than the test case:

```
a.send(b, "msg0");
a.send(b, "msg1");
```
Can these messages be seen in a different order by b?
in the real world?
in the tests?

src/mock/quick_p2p/tests.rs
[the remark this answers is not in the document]
jeanphilippeD: I think it was causing confusion on what the code was actually verifying. (i.e was used with actual variables, but only doing pattern matching). What I know is that it was removed from Parsec test on purpose.

src/mock/quick_p2p/tests.rs
madadam: As, I see what you mean now. So yes, in that case they can be reordered in the real world, but currently not in mock. Do you think it can be a problem? Is it worth implementing the reordering in the mock too?

jeanphilippeD: I think it will, so our test can detect these issues with mock. However, I'm happy for it to be a follow up. Maybe leave a TODO. (Could be something someone else does).

src/mock/quick_p2p/tests.rs
jeanphilippeD: As discussed on slack, a TODO is fine with someone else picking up bring the API fix to the mock that will latter be fixed in quick-p2p lib.

src/mock/quick_p2p/tests.rs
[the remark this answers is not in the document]
jeanphilippeD: Fine to leave for someone else if need addressing.

PR #1677, opened 2019-06-05
Fix concurrent merge test fail by adding a ProvingSection cache

octol:
This is another attempt at fixing the problem described in 9fdadac3c296a672f1acf3705355af5855ae5e95 which triggers failures in mock_crust::merge::concurrent_merge.

When adding new sections either cache the ProvingSections in case a prefix change causes the accumulation to fail so we can vote for them again in the new prefix, or vote for them straight away if we just have gone through a prefix change where the other half of the section has not caught up yet.

### DIFF
--- a/src/states/adult.rs
+++ b/src/states/adult.rs
@@ -357,6 +357,14 @@ impl Approved for Adult {
         &mut self.chain
     }
 
+    fn set_pfx_successfully_polled(&mut self, _: bool) {
+        // Doesn't do anything
+    }
+
+    fn is_pfx_successfully_polled(&self) -> bool {
+        false
+    }
+
     fn handle_add_elder_event(
         &mut self,
         new_pub_id: PublicId,
--- a/src/states/elder/mod.rs
+++ b/src/states/elder/mod.rs
@@ -72,6 +72,7 @@ const CANDIDATE_STATUS_INTERVAL: Duration = Duration::from_secs(60);
 const CLIENT_BAN_DURATION: Duration = Duration::from_secs(2 * 60 * 60);
 /// Duration for which clients' IDs we disconnected from are retained.
 const DROPPED_CLIENT_TIMEOUT: Duration = Duration::from_secs(2 * 60 * 60);
+const PROVING_SECTION_CACHE_SIZE: usize = 100;
 
 pub struct ElderDetails {
     pub ack_mgr: AckManager,
@@ -131,6 +132,10 @@ pub struct Elder {
     chain: Chain,
     #[cfg(feature = "mock_base")]
     ignore_candidate_info_counter: u8,
+    /// Cached proving sections used to in case a prefix change causes a section info not to
+    /// accumulate.
+    proving_section_cache: LruCache<(Vec<ProvingSection>, SectionInfo), ()>,
+    pfx_is_successfully_polled: bool,
 }
 
 impl Elder {
@@ -232,6 +237,8 @@ impl Elder {
             chain: details.chain,
             #[cfg(feature = "mock_base")]
             ignore_candidate_info_counter: 0,
+            proving_section_cache: LruCache::with_capacity(PROVING_SECTION_CACHE_SIZE),
+            pfx_is_successfully_polled: false,
         }
     }
 
@@ -468,6 +475,10 @@ impl Elder {
                 self.vote_for_event(event.clone());
             });
 
+        // This also clears the cache, something that *might* be a problem during a multi stage
+        // merge.
+        self.vote_for_neighbours_in_proving_section_cache();
+
         Ok(())
     }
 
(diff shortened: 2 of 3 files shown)


src/states/elder/mod.rs
```diff
@@ -129,6 +130,10 @@ pub struct Elder {
     gen_pfx_info: GenesisPfxInfo,
     gossip_timer_token: u64,
     chain: Chain,
+    /// Cached proving sections used to in case a prefix change causes a section info not to
+    /// accumulate.
+    set_of_proving_sections: VecDeque<(Vec<ProvingSection>, SectionInfo)>,
```
Viv-Rajkumar: why not use an LruCache object for this one?

src/states/elder/mod.rs
```diff
@@ -464,6 +471,18 @@ impl Elder {
                 self.vote_for_event(event.clone());
             });
 
+        let ps = self
+            .set_of_proving_sections
+            .iter()
+            .filter(|(_, si)| our_pfx.is_neighbour(si.prefix()))
```
Viv-Rajkumar: hmm is this correct cos this is only going to be considering the leaf sec_info, not any sec_infos from within the ProvingSection itself? maybe take a look at `handle_proving_sections_event`. That should show how its not just the leaf sec_info we aim to discover about but any within the ProvingSection too.

Viv-Rajkumar: we should also check if the version in this sec_info is relevant as in if we cached v-5 and neighbour_infos has v-6 for the same pfx, they're still our neighbour according to this filter but no point voting for this excess PS here

src/states/elder/mod.rs
```diff
@@ -464,6 +471,18 @@ impl Elder {
...
+            .cloned()
+            .collect_vec();
+        ps.into_iter().for_each(|(ps, si)| {
+            self.vote_for_event(NetworkEvent::ProvingSections(ps, si));
+        });
+        self.set_of_proving_sections
+            .retain(|(_, si)| !our_pfx.is_neighbour(si.prefix()));
```
Viv-Rajkumar: similar to filter case before

src/states/elder/mod.rs
```diff
@@ -570,12 +589,16 @@ impl Elder {
...
                     //       slows down the tests. Find out why, maybe enable it in more cases.
-                    if self.add_new_sections(signed_msg.section_infos())
-                        && (!self.in_authority(&signed_msg.routing_message().dst)
-                            || signed_msg.routing_message().dst.is_single())
-                    {
+                    if self.add_new_sections(signed_msg.section_infos()) {
                         let ps = signed_msg.proving_sections().clone();
-                        self.vote_for_event(NetworkEvent::ProvingSections(ps, si.clone()));
+                        if !self.in_authority(&signed_msg.routing_message().dst)
+                            || signed_msg.routing_message().dst.is_single()
```
Viv-Rajkumar: guess this part of the or condition shouldnt be needed with this patch now?

src/states/elder/mod.rs
```diff
@@ -616,6 +639,25 @@ impl Elder {
         Ok(())
     }
 
+    fn add_to_proving_section_cache(
+        &mut self,
+        proving_secs: Vec<ProvingSection>,
+        sec_info: SectionInfo,
+    ) {
+        if self.chain.is_new_neighbour(&sec_info) {
+            self.set_of_proving_sections
```
Viv-Rajkumar: similar to filter, not sure if just checking leaf sec_info as valid neighbour is enough if the src_section is a couple hops away but proving_section contains new info about our valid neighbour.

src/states/elder/mod.rs
```diff
@@ -616,6 +639,25 @@ impl Elder {
...
+                .push_back((proving_secs, sec_info));
+        }
+        while self.set_of_proving_sections.len() > PROVING_SECTION_CACHE_SIZE {
```
Viv-Rajkumar: guess if lru cache, this manual stuff may not be needed and also becomes a time bound queue

src/states/adult.rs
```diff
@@ -357,6 +357,14 @@ impl Approved for Adult {
         &mut self.chain
     }
 
+    fn set_pfx_successfully_polled(&mut self, _: bool) {
+        // Doesn't do anything
+    }
+
+    fn is_pfx_successfully_polled(&self) -> bool {
+        false
```
Viv-Rajkumar: will have to think about this. right now an adult is someone staying in this state until they're established, so they certainly do go past the Genesis gossip stage, so this returning blanket false seems weird. Although this cache logic and how much we need adults to also handle this I need to think about too.

src/states/elder/mod.rs
```diff
@@ -616,6 +639,25 @@ impl Elder {
...
+            let _ = self.set_of_proving_sections.pop_front();
+        }
+    }
+
+    fn remove_from_proving_section_cache(&mut self, sec_info: &SectionInfo) {
```
Viv-Rajkumar: this also depends on if we only choose to cache if leaf sec_info is interesting or any msg going through us.

src/states/elder/mod.rs
jeanphilippeD: Do we want it to be timebound though?

Viv-Rajkumar: you can choose for either with lru if I remember right or both. I think it does have a max_capacity ctor too

src/states/elder/mod.rs
octol: > hmm is this correct cos this is only going to be considering the leaf sec_info, not any sec_infos from within the ProvingSection itself? maybe take a look at handle_proving_sections_event. That should show how its not just the leaf sec_info we aim to discover about but any within the ProvingSection too.

After a discussion we decided that the cache should primarily function on the leaf

octol: > hmm is this correct cos this is only going to be considering the leaf sec_info, not any sec_infos from within the ProvingSection itself? maybe take a look at handle_proving_sections_event. That should show how its not just the leaf sec_info we aim to discover about but any within the ProvingSection too.

octol:
Added a slight simplification by using the SectionInfo as key

PR #1678, opened 2019-06-06
test: assert quorum in request tests

d1vyank:
Addresses a TODO to assert that 'quorum' number of put and get requests are received.

### DIFF
--- a/tests/mock_network/requests.rs
+++ b/tests/mock_network/requests.rs
@@ -12,11 +12,13 @@ use super::{
 use routing::mock_crust::Network;
 use routing::{
     Authority, ClientError, Event, EventStream, ImmutableData, MessageId, Request, Response,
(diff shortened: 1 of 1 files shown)


tests/mock_network/requests.rs
```diff
@@ -56,13 +58,14 @@ fn successful_put_request() {
         }
     }
 
-    // TODO: Assert a quorum here.
+    assert!(request_received_count >= quorum);
     assert!(2 * request_received_count > min_section_size);
```
jeanphilippeD: I think we want to replace the old assertion with the new one as they check the same value

d1vyank: Removed the original assertion, thanks!

jeanphilippeD:
Could you rebase ontop of the latest fleming commits so we can merge it?

PR #1688, opened 2019-06-07
test/quic_p2p: pop random message

maqi:
This PR contains the change to make mock quic_p2p poping message randomly.

### DIFF
--- a/src/mock/quick_p2p/network.rs
+++ b/src/mock/quick_p2p/network.rs
@@ -15,7 +15,7 @@ use maidsafe_utilities::SeededRng;
 use rand::Rng;
 use std::{
     cell::RefCell,
-    cmp::Ordering,
+    cmp::{self, Ordering},
     collections::{hash_map::Entry, VecDeque},
     net::{Ipv4Addr, SocketAddr, SocketAddrV4},
     rc::{Rc, Weak},
@@ -161,8 +161,6 @@ impl Inner {
     }
 
     fn pop_random_packet(&mut self) -> Option<(Connection, Packet)> {
-        // TODO: randomly reorder subsequent `Packet::Message`s to simulate out-of-order delivery.
-
         let connections: Vec<_> = self
             .connections
             .iter()
@@ -183,7 +181,7 @@ impl Inner {
     fn pop_packet(&mut self, connection: Connection) -> Option<Packet> {
         match self.connections.entry(connection) {
             Entry::Occupied(mut entry) => {
-                let packet = entry.get_mut().pop();
+                let packet = entry.get_mut().pop_random_msg(&mut self.rng);
                 if entry.get().is_empty() {
                     let _ = entry.remove_entry();
                 }
@@ -218,12 +216,26 @@ impl Queue {
         self.0.push_back(packet)
     }
 
-    fn pop(&mut self) -> Option<Packet> {
-        self.0.pop_front()
-    }
-
     fn is_empty(&self) -> bool {
         self.0.is_empty()
+    }
+
+    // This function will pop random msg from the queue.
+    fn pop_random_msg(&mut self, rng: &mut SeededRng) -> Option<Packet> {
+        let first_non_msg_packet = self
+            .0
+            .iter()
+            .position(|packet| {
+                if let Packet::Message(_) = packet {
+                    false
+                } else {
+                    true
+                }
+            })
+            .unwrap_or(0);
+
+        let selected = rng.gen_range(0, cmp::max(first_non_msg_packet, 1));
+        self.0.remove(selected)
     }
 }
 
--- a/src/mock/quick_p2p/tests.rs
+++ b/src/mock/quick_p2p/tests.rs
@@ -9,6 +9,7 @@
 use super::{Builder, Config, Event, Network, NodeInfo, OurType, Peer, QuicP2p};
 use bytes::Bytes;
 use std::{
+    collections::BTreeSet,
     iter,
     net::SocketAddr,
     sync::mpsc::{self, Receiver, TryRecvError},
@@ -309,13 +310,8 @@ fn send_multiple_messages_without_connecting_first() {
     a.expect_connected_to_node(&b.addr());
     b.expect_connected_to_node(&a.addr());
 
-    // TODO: We shouldn't rely on the messages being delivered in the same order they were sent.
-    //       We should also change the implementation to introduce random reordering of the
-    //       messages to more faithfully simulate real quick-p2p which doesn't guarantee the order
-    //       either.
-    for msg in &msgs {
-        b.expect_new_message(&a.addr(), msg);
-    }
+    let received_messages = b.received_messages(&a.addr());
+    expected_messages_received(msgs.to_vec(), received_messages);
 }
 
 #[test]
@@ -527,6 +523,20 @@ impl Agent {
     fn expect_none(&self) {
         assert_eq!(self.rx.try_recv(), Err(TryRecvError::Empty));
     }
+
+    fn received_messages(&self, src_addr: &SocketAddr) -> BTreeSet<Bytes> {
+        let mut received_messages = BTreeSet::new();
+        while let Ok(Event::NewMessage { peer_addr, msg }) = self.rx.try_recv() {
+            assert_eq!(peer_addr, *src_addr);
+            let _ = received_messages.insert(msg);
+        }
+        received_messages
+    }
+}
+
+fn expected_messages_received(sent: Vec<Bytes>, received: BTreeSet<Bytes>) {
+    let expected: BTreeSet<Bytes> = sent.iter().cloned().collect();
+    assert_eq!(expected, received);
 }
 
 fn establish_connection(network: &Network, a: &mut Agent, b: &mut Agent) {


src/mock/quick_p2p/network.rs
```diff
@@ -218,13 +216,27 @@ impl Queue {
...
-    fn pop(&mut self) -> Option<Packet> {
-        self.0.pop_front()
-    }
-
     fn is_empty(&self) -> bool {
         self.0.is_empty()
     }
+
+    // This function will pop random msg from the queue.
+    fn pop(&mut self, rng: &mut SeededRng) -> Option<Packet> {
```
pierrechevalier83: I would definitely rename it because to me `pop` suggests either `pop_front` or `pop_back`.
Maybe `pop_random_msg` ?

src/mock/quick_p2p/network.rs
```diff
@@ -218,13 +216,27 @@ impl Queue {
...
+        let mut first_non_random_msg = 0;
+        for packet in self.0.iter() {
```
pierrechevalier83: Maybe this for loop can be replaced with
```
let first_non_random_msg = self.0.iter().position(|packet| packet == Packet::Disconnect).unwrap_or(0);
```
(or some variation if `Packet` isn't `PartialEq`)

src/mock/quick_p2p/tests.rs
```diff
@@ -309,13 +310,8 @@ fn send_multiple_messages_without_connecting_first() {
...
 
-    // TODO: We shouldn't rely on the messages being delivered in the same order they were sent.
-    //       We should also change the implementation to introduce random reordering of the
-    //       messages to more faithfully simulate real quick-p2p which doesn't guarantee the order
-    //       either.
-    for msg in &msgs {
-        b.expect_new_message(&a.addr(), msg);
-    }
+    let received_messages = b.received_messages(&a.addr());
+    expected_messags_received(msgs.to_vec(), received_messages);
```
pierrechevalier83: typo: "messags" either use "msgs" or "messages" (I prefer the latter)

src/mock/quick_p2p/tests.rs
```diff
@@ -527,6 +523,22 @@ impl Agent {
...
+        let mut received_messages = BTreeSet::new();
+        while let Ok(Event::NewMessage { peer_addr, msg }) = self.rx.try_recv() {
+            assert_eq!(peer_addr, *src_addr);
+            let _ = received_messages.insert(msg);
+        }
+        received_messages
+    }
+}
+
+fn expected_messags_received(expected: Vec<Bytes>, mut received: BTreeSet<Bytes>) {
```
pierrechevalier83: Why take 2 different types for things that are conceptually similar?
This function should probably take two arguments of the same type, `assert!` they're equal and reset received to `ThatCollection::new()`.

In particular, I see a `to_vec()` at line 314 which makes me think the `Vec` isn't really a natural representation for this `expected` argument.

src/mock/quick_p2p/network.rs
maqi: As I understand, the `pop` function shall still be remained ?

src/mock/quick_p2p/tests.rs
maqi: first, we don't expect these two are equal, as the received order could be different.
second, I made these two different purposely to diff `expected` and `received`

src/mock/quick_p2p/network.rs
```diff
@@ -218,13 +216,31 @@ impl Queue {
...
+        let first_non_random_msg = self
+            .0
+            .iter()
+            .position(|packet| {
+                if let Packet::Disconnect = packet {
+                    true
+                } else {
+                    false
+                }
+            })
```
madadam: This is incorrect. It should be something like:

```rust
if let Packet::Message = packet {
    false
} else {
    true
}
```
that is, it should give you the first **non**-message `Packet`. The way it is currently it can still reorder `Message` with other types (say `ConnectSuccess`) which can cause issues (and would not match the real quic-p2p behaviour).

src/mock/quick_p2p/network.rs
```diff
@@ -218,13 +216,31 @@ impl Queue {
...
-        self.0.pop_front()
-    }
-
     fn is_empty(&self) -> bool {
         self.0.is_empty()
     }
+
+    // This function will pop random msg from the queue.
+    fn pop_random_msg(&mut self, rng: &mut SeededRng) -> Option<Packet> {
+        let first_non_random_msg = self
```
madadam: This variable is weirdly named. I'd call it something like `first_non_msg_packet` (that is: the first packet that is not `Message`).

src/mock/quick_p2p/network.rs
```diff
@@ -218,13 +216,31 @@ impl Queue {
...
+                    false
+                }
+            })
+            .unwrap_or(0);
+
+        let selected = if first_non_random_msg == 0 {
+            0
+        } else {
+            rng.gen_range(0, first_non_random_msg)
+        };
```
madadam: This can be more succinctly expressed as `let selected = rng.gen_range(0, cmp::max(first_non_random_msg, 1))`

src/mock/quick_p2p/tests.rs
```diff
@@ -527,6 +523,22 @@ impl Agent {
...
+        let mut received_messages = BTreeSet::new();
+        while let Ok(Event::NewMessage { peer_addr, msg }) = self.rx.try_recv() {
+            assert_eq!(peer_addr, *src_addr);
+            let _ = received_messages.insert(msg);
+        }
+        received_messages
+    }
+}
+
+fn expected_messages_received(expected: Vec<Bytes>, mut received: BTreeSet<Bytes>) {
```
madadam: Why no take `expected` by `BTreeSet` as well. Then you would be able to do just `assert_eq!(expected, received)`.

src/mock/quick_p2p/network.rs
maqi: Ah, ok.
So, only `Message` can be re-ordered, other Packet shall remain the same.
I will update it.

src/mock/quick_p2p/tests.rs
maqi: The `expected` was a Vec in the order of sending out messages.
We need to convert it into BTreeSet somewhere anyway.
I made it here as it can also diff the expected and received explicitly by type.

madadam: There are two issue with this (both relatively minor though):
1. The parameter `expected: Vec<Bytes>` suggest that the order is also what is expected. If the parameter was `expected: BTreeSet<Bytes>` then it would be more clear that we only care about the elements being the same, but not their order.
2. The implementation of this function could be changed to just `assert_eq!(expected, received)` which is not only shorter, but it would produce more useful error messages in case of failure.

maqi: but we still need to convert that send_out msgs (which is a `vec` into `BTreeSet`) somewhere, and with more tests, such convertion may happen multiple times.
How about:
```
fn expected_messages_received(sent: Vec<Bytes>, mut received: BTreeSet<Bytes>) {
    let expected: BTreeSet<Bytes> = sent.iter().cloned().collect();
    assert_eq!(expected, received);
}
```
so basicaly the convertion got capsulated in this function?

madadam: That works. You don't need the `mut` though.

src/mock/quick_p2p/network.rs
pierrechevalier83: If you're asking me whether that's what I'm suggesting, it isn't. I'm just saying that I would rename it. If you think `pop` should remain for a different reason, can you explain it?

src/mock/quick_p2p/tests.rs
pierrechevalier83: uhm. If you want for the types to be different for type-safety, I would suggest to be explicit and create a `struct Expected (BTreeSet<Bytes>)` and a `struct Actual(BtreeSet<Bytes>)`. Probably overkill, but better than changing one to a Vec somewhat arbitrarily imo.

maqi: Adam had the same concern and we had a discussion at https://github.com/maidsafe/routing/pull/1688#discussion_r291955790
And reached an agreement.

src/mock/quick_p2p/network.rs
maqi: I thought we may need a `sequenced pop` for some test.
However it turned not, and I have changed it.

src/mock/quick_p2p/tests.rs
pierrechevalier83: Uhm. I'm not sure it's exactly the same concern. To be clear, I think the implementation now is better than what was there before. I am simply still unsure about passing a `Vec` for the `sent` instead of a `BTreeSet`. If I'm not mistaken, we're converting from `BTreeSet` to `vec` and then vice-versa.

maqi: > we're converting from BTreeSet to vec and then vice-versa

No we don't. The vection is constructed directly in the testing code:
```
    let msgs = [
        Bytes::from_static(b"message 0 from A"),
        Bytes::from_static(b"message 1 from A"),
        Bytes::from_static(b"message 2 from A"),
    ];
```

pierrechevalier83: Uhm. OK. I reread the thread. I had missed something from the msg before last. I may have been confused about the type at the call site. I'm OK as is.

PR #1689, opened 2019-06-10
Remove all verification logic from hops

### LINKED ISSUE #1661: Remove all verification logic from hops

pierrechevalier83:
With secure message relay, we used to verify each hop message to avoid
transmitting spammy messages to the rest of the network.
This pattern is going away with the introduction of Reliable Message
Delivery where we ensure that any message passed to us will be
transmitted.
The logic to ensure message validity will only happen at destination
once Secure Message Delivery is introduced.

Fixes #1661

### DIFF
--- a/src/states/elder/mod.rs
+++ b/src/states/elder/mod.rs
@@ -546,7 +546,7 @@ impl Elder {
         Ok(())
     }
 
-    // Verify the message, then, if it is for us, handle the enclosed routing message and swarm it
+    // If the message is for us, verify it then, handle the enclosed routing message and swarm it
     // to the rest of our section when destination is targeting multiple; if not, forward it.
     fn handle_signed_message(
         &mut self,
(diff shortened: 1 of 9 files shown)


src/states/elder/mod.rs
```diff
@@ -550,19 +550,6 @@ impl Elder {
...
-                None => true,
-                Some(hop) => !self.is_trusted(hop)?,
-            } {
-                // We don't know the last hop! Try the one before that.
-                if !signed_msg.pop_previous_hop() {
-                    debug!("{} Untrusted message: {:?}", self, msg_clone);
-                    return Err(RoutingError::NotEnoughSignatures);
-                }
-            }
             // Now that we validated the sections, inform our peers about any new ones.
```
octol: Is this comment now out of date?

pierrechevalier83:
> LGTM :) Although either you or me will have some conflicts to resolve after the other one's PR gets merged.

Probably you, given the size of both PRs but should be pretty trivial as this is mostly pure removal :)

src/states/elder/mod.rs
pierrechevalier83: Uhm. Good catch.

PR #1692, opened 2019-06-13
Replace crust with quic-p2p

madadam:
This PR replaces crust with quic-p2p as the low-level networking layer. This necessitated some changes that are described here:

## PeerMap

Unlike crust, quic-p2p does not carry public ids of the peers, only their connection infos. To associate the two a new structure called `PeerMap` is introduced. This structure can track a connection in several states:
1. No connection established and public id unknown (no entry in `PeerMap`)
2. Network-layer connection established, but public id unknown
3. Connection info and public id known, but the connection not established yet
4. Connection established and public id known

It is only in the fourth state that we consider a peer as connected. The rest of routing doesn't have to worry about states 1 to 3.
Note: `PeerMap` does not replace `PeerManager` yet, so both still exists. Such change might be done as a separate task in the future. `PeerManager` did get simplified a bit though.

## DirectMessage

Because quic-p2p does not carry public ids, we need to embed them in `DirectMessage`, otherwise the recipient wouldn't be able to tell what the public id of the sender is. For this purpose a new type - `SignedDirectMessage` is introduced which contains the `DirectMessage`, the `PublicId` of the sender and the signature (which signs both content and the public_id).

## HopMessage

When handling `HopMessage`, we lookup the public id by the socket address. If not found (which should rarely happen), we log error and ignore the message.

## Bootstrap blacklist

quic-p2p does not support bootstrap blacklist. This can cause issues in some (relatively rare) cases. For example, if a node we want to bootstrap against does not send us `BootstrapRespose` or sends a negative one, we can't tell quic-p2p to skip that node when rebootstrapping, which might result in us unsuccessfully trying to bootstrap against the same node over and over. There is a couple of tests which currently fail because of this. Those tests were `#[ignored]` for now.

## Connection logic

The connection and disconnection logic got simplified. This is how it works now:

### Bootstrapping

1. A new node that wants to join the network initiates the bootstrap by calling `QuicP2p::bootstrap`.
2. This established network-level bootstrap connection with some existing node
3. The existing node receives `ConnectedTo` quic-p2p event which contains the connection details of the joining node
4. The new node sends `BootstrapRequest` direct message to the existing node
5. The existing node receives the `BootstrapRequest` and learn the public id of the new node. At this point the connection from their point of view is established and it insert an entry into `PeerManager` with either `PeerState::Client` or `PeerState::JoiningNode`, depending on the connection type.
6. The existng node sends `BootstrapResponse` back to the new node
7. The new node receives the `BootstrapResponse` and learns the public id of the existing node which makes the connection established from their side too. The bootstrap is finished and the node transitions to the next state.
8. If the new node does not receive `BootstrapResponse` in time or if it receives negative response, it rebootstraps (see also the current issue with bootstrap blacklist). 

### Connecting

1. A freshly bootstrapped node wants to connect to the other nodes in the section, or an existing member want to connect to some other member after a section change. Let's call the node that initiates the connection A and the other node B.
2. A sends a `ConnectRequest` **routing** message to B (that is, it's destination authority is `ManagedNode(B)`). This message contains the public id and encrypted connection info of A. It also inserts the public id of B into the peer manager with `PeerState::Connecting`.
3. B receives the `ConnectRequest`, and decrypts the connection info. It then stores the public id and the connection info into `PeerMap` (not connected yet), stores the public id in `PeerManager` as `PeerState::Connecting` and sends a `ConnectResponse` **direct** message back to A
4. Sending and receiving the direct message results in the connection being established. Both A and B change the state of the other node in `PeerManager` to `PeerState::Connected`.


src/state_machine.rs
```diff
@@ -285,99 +276,47 @@ impl StateMachine {
...
-            MaidSafeEventCategory::Crust,
-            category_tx.clone(),
+        let (network_tx, network_rx) = mpmc::unbounded();
+        let (action_tx, action_rx) = mpmc::unbounded();
+
+        let network_service = unwrap!(
+            NetworkBuilder::new(network_tx)
+                .with_config(network_config)
+                .build(),
+            "Unsable to start network service"
```
pierrechevalier83: Typo: s/Unsable/Unable/

Cargo.toml
```diff
@@ -14,7 +14,8 @@ edition = "2018"
 bytes = "~0.4.12"
 config_file_handler = "~0.11.0"
 crossbeam-channel = "~0.3.8"
-crust = { git = "https://github.com/maidsafe/crust", rev = "028de4ea8" }
+# crust = { git = "https://github.com/maidsafe/crust", rev = "028de4ea8" }
+crust = { path = "../crust" }
```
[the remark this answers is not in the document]
madadam: Yep, the crust dependency will go away in the final PR.

src/mock/quic_p2p/network.rs
```diff
@@ -193,8 +230,29 @@ impl Inner {
...
+        if let Some(node) = self.find_node(addr0) {
+            node.borrow_mut().disconnect(*addr1)
+        }
+    }
 }
 
+// The 4-byte tags of `Message::Direct` and `DirectMessage::ParsecRequest`.
+// A serialised Parsec request message starts with these bytes.
+#[cfg(not(feature = "mock_serialise"))]
+static PARSEC_REQ_MSG_TAGS: &[u8] = &[0, 0, 0, 0, 9, 0, 0, 0];
```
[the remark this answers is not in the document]
madadam: It's duplicated currently, but when crust (and mock_crust) is gone it won't be.

src/mock/quic_p2p/tests.rs
```diff
@@ -554,3 +639,34 @@ fn assert_connected_to_node(event: Event, addr: &SocketAddr) {
     );
     assert_eq!(actual_peer_addr, *addr);
 }
+
+// Generate unique message.
+fn gen_message() -> NetworkBytes {
+    use std::sync::atomic::{AtomicUsize, Ordering};
+
+    static COUNTER: AtomicUsize = AtomicUsize::new(0);
+    let num = COUNTER.fetch_add(1, Ordering::Relaxed);
```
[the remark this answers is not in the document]
madadam: `fetch_add` returns the previous value, so we are starting at 0 here.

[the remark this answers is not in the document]
pierrechevalier83: Ah OK! Thanks for the clarification :smile:

src/client.rs
```diff
@@ -549,20 +543,23 @@ impl Client {
     #[allow(clippy::new_ret_no_self)]
     pub fn new(
         keys: Option<FullId>,
-        bootstrap_config: Option<BootstrapConfig>,
+        network_config: Option<NetworkConfig>,
         config: Config,
         msg_expiry_dur: Duration,
     ) -> Result<Client, RoutingError> {
+        // TODO: if `network_config` is `None`, use `NetworkConfig::read_or_construct_default`.
```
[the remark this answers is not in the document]
madadam: Forgot to remove the TODO. It does no longer apply.

src/messages/direct.rs
```diff
@@ -34,18 +34,16 @@ pub enum DirectMessage {
...
     /// bootstrapping is successful; if `false` the sender is not available as a bootstrap node.
     BootstrapResponse(Result<(), BootstrapResponseError>),
+    /// Sent from members of a section to a joining node in response to `ConnectRequest` (which is
+    /// a routing message)
+    ConnectResponse,
     /// Sent from a node which is still joining the network to another node, to allow the latter to
     /// add the former to its routing table.
     CandidateInfo {
         /// `PublicId` from before relocation.
         old_public_id: PublicId,
```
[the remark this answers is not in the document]
madadam: The new public id is now kept in `SignedDirectMessage` which is why it was removed here. But this one is still the old public id, so I'd say the name should stay so the semantics is clear.

src/messages/direct.rs
```diff
@@ -34,18 +34,16 @@ pub enum DirectMessage {
...
-        /// `PublicId` from after relocation.
-        new_public_id: PublicId,
-        /// Signature of concatenated `PublicId`s using the pre-relocation key.
+        /// Signature of both old and new `PublicId`s using the pre-relocation key.
         signature_using_old: Signature,
```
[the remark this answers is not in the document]
madadam: Same argument as https://github.com/maidsafe/routing/pull/1692/files#r293792843

src/messages/direct.rs
```diff
@@ -0,0 +1,234 @@
...
+        )
+    }
+}
+
+#[cfg(not(feature = "mock_serialise"))]
+mod implementation {
+    use super::*;
+    use maidsafe_utilities::serialisation;
+
+    pub fn sign(src_full_id: &FullId, content: &DirectMessage) -> Result<Signature, RoutingError> {
```
pierrechevalier83: I think I would rename this to `sign_public_id_and_content` because the signature as is could be taken as: "sign this content with the private key from this FullId". Being more explicit would make the call site unambiguous.

src/messages/direct.rs
```diff
@@ -0,0 +1,234 @@
...
+        let mut serialised = serialisation::serialise(src_full_id.public_id())?;
+        serialised.extend_from_slice(&serialisation::serialise(content)?);
+
+        let signature = src_full_id.signing_private_key().sign_detached(&serialised);
+        Ok(signature)
+    }
+
+    pub fn verify(
```
pierrechevalier83: if you rename `sign`, this should probably be verbosified too for consistency :smile:

src/states/elder/mod.rs
```diff
@@ -1112,8 +1059,11 @@ impl Elder {
             return;
         }
 
-        let (difficulty, target_size) = if self.disable_resource_proof
-            || self.network_service.is_peer_hard_coded(new_pub_id)
+        // TODO (quic-p2p): quic-p2p doesn't currently have an API to check if a contact is
```
pierrechevalier83: Given the code I reviewed in quic-p2p, I'm assuming this TODO will be addressed in this PR (?)

src/peer_manager.rs
```diff
@@ -227,6 +142,14 @@ impl Peer {
         self.timestamp.elapsed() >= Duration::from_secs(timeout)
     }
 
+    /// Returns whether the peer is in the `Connecting` state.
+    pub fn is_connecting(&self) -> bool {
+        match self.state {
```
pierrechevalier83: For these kind of situations, I'm in favour of deriving `PartialEq` for `PeerState` and simplifying this code except if there is a good reason not to do so.

src/states/elder/mod.rs
madadam: Yes, as soon as https://github.com/maidsafe/quic-p2p/pull/53 is merged I'll push a commit to address this.

src/states/common/base.rs
```diff
@@ -245,20 +232,28 @@ pub trait Base: Display {
...
     fn handle_new_deserialised_message(
         &mut self,
-        pub_id: PublicId,
+        src_addr: SocketAddr,
         message: Message,
         outbox: &mut EventBox,
     ) -> Result<Transition, RoutingError> {
         match message {
-            Message::Hop(msg) => self.handle_hop_message(msg, pub_id, outbox),
+            Message::Hop(msg) => self.handle_hop_message(msg, outbox),
```
fizyk20: Wouldn't it be better to try and get the public ID from the PeerMap?

Or maybe we should make the `HopMessage` a variant of `DirectMessage`? This way we would get the ID of the previous hop from the `SignedDirectMessage`... and hop messages _are_ passed directly between nodes, anyway.

The workaround with using `sent_to` is kind of okay, but we're removing `sent_to` as part of RMD, so it will cause problems again...

src/states/common/base.rs
```diff
@@ -270,29 +265,50 @@ pub trait Base: Display {
...
+
+        self.send_message_over_network(peer, message);
+    }
 
+    fn send_message_over_network(&mut self, dst: ConnectionInfo, message: Message) {
         match to_network_bytes(message) {
-            Ok(bytes) => {
-                self.send_or_drop(dst_id, bytes, priority);
-            }
+            Ok(bytes) => self.network_service_mut().send(dst, bytes),
```
fizyk20: Doesn't `send` return any kind of result? If not, it might cause trouble for RMD - the plan was to depend on whether sending a message was successful quite a bit (and resend to someone else if it wasn't).

src/states/elder/mod.rs
```diff
@@ -540,8 +520,10 @@ impl Elder {
             self.sig_accumulator
                 .add_proof(min_section_size, digest, proof)
         {
-            let hop = *self.name(); // we accumulated the message, so now we act as the last hop
-            self.handle_signed_message(signed_msg, route, hop, &BTreeSet::new())?;
+            // we accumulated the message, so now we act as the last hop
+            let mut sent_to = BTreeSet::new();
+            let _ = sent_to.insert(*self.name());
+            self.handle_signed_message(signed_msg, route, &sent_to)?;
```
fizyk20: Like in the other comment - `sent_to` is on the chopping block, so it would be nice to find another way of handling this.

src/states/elder/mod.rs
```diff
@@ -1006,49 +943,32 @@ impl Elder {
...
+            self,
+            pub_id,
+            peer_type
+        );
 
-        if (peer_kind == CrustUser::Client || !self.is_first_node)
-            && self.chain.len() < self.min_section_size() - 1
-        {
+        // Check min section size.
+        if !self.is_first_node && self.chain.len() < self.min_section_size() - 1 {
```
fizyk20: Won't this make the first node accept clients even before the first section grows enough to be fully functional?

src/messages/direct.rs
madadam: Not sure I get you completely, but signing the public id is important to know the sender is authentic. Singing the content is something I decided to add, because it was easy to do and I think it provides additional layer of security we didn't have before this PR (previously `DirectMessage` weren't signed). For example, if only the public id were signed, not the content, then an advresary can in theory capture a legit message, take the public_id and the signature, put their own content and pass it on and the recipient would have no way of knowing the message was tampered with.

src/states/common/base.rs
madadam: > Wouldn't it be better to try and get the public ID from the PeerMap?

We might not necessarily have it. We only have the public id if we received at least one `DirectMessage` from them previously, or if we received `ConnectRequest`. Not sure that's guaranteed to happen. Maybe it is. This might need more thinking through...

src/states/common/base.rs
madadam: It doesn't, but if the message fails to be sent, we might receive `UnsentUserMessage` event from quic-p2p later. In this PR we are ignoring such event but there is a `TODO` to address it later (and we have a [github issue](https://github.com/maidsafe/routing/issues/1684) for it too).

src/states/elder/mod.rs
madadam: Understood. This might require a discussion.

src/states/elder/mod.rs
madadam: Yes, but that's what the original (before this PR) behaviour was. 

Maybe the first node should only accept nodes, not clients though (until reaching min section size)? I think before this PR, we weren't able to tell whether the bootstrapping peer is a client or node until the relocation, so we accepted both. But now we know it*, so maybe we can restrict this to nodes only. What do you think?

*) With crust, both client and nodes started as `CrustUser::Cient` and nodes then switched to `CrustUser::Node` after relocation. Now we identify as client or node right from the start which I'm not 100% sure it's correct.

src/states/elder/mod.rs
madadam: Here is an idea: we can exclude the sender using their connection info instead of pub_id.

fizyk20: We need to filter at the point of choosing the targets, though, so that the number of targets stays correct. If we can map the IDs to connection infos at this point, though, can't we just translate the connection info into the pub_id to be filtered?

src/states/elder/mod.rs
fizyk20: >Yes, but that's what the original (before this PR) behaviour was.
Hmm, but before this PR, this `if` would cause even the first node to reject clients if the section was too small, wouldn't it? That would be even weirder with what you are saying, because it would mean that before this PR everyone would be rejected when the section was small. What am I missing? Is it that this happens at some later stage of bootstrapping?

madadam: Oops, I think I got confused and am spreading misinformation now :grimacing: . Ignore the previous comment from me. You were right. I'm going to modify the code.

src/states/elder/mod.rs
madadam: So reverse lookup the pub_id by the connection info? Sounds doable. Let me try that.

src/messages/direct.rs
fizyk20: My point is that just signing the content would probably be enough (so we could name the function `sign_content` or something) - the public ID is just a piece of data that enables validation of the signature.

Basically, if the signature of the content is valid when we use the given public ID, it already means that someone with the corresponding private key signed it - no need to sign the ID, too.

madadam: I see. Yeah, that makes sense. I'll update the code.

src/chain/chain.rs
```diff
@@ -1056,14 +1056,20 @@ impl Chain {
...
+        } else {
+            section.iter().collect_vec()
+        };
+
         if names.is_empty() {
             return Err(Error::CannotRoute);
         }
 
-        Ok(*Self::get_routeth_name(names, &target, route))
+        Ok(*Chain::get_routeth_name(names, &target, route))
```
pierrechevalier83: It's very minor, but I don't think this bit needs undone :smile:

src/peer_manager.rs
```diff
@@ -175,6 +175,15 @@ impl Peer {
         }
     }
 
+    // If the peer is a client, return its IP, otherwise returns `None`.
```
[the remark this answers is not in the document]
madadam: It's not exposed. It's actually not even `pub`, so doc comments would make no sense. Actually the other methods in that file shouldn't be doc-commented either.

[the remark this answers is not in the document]
eminence: Doc comments can make sense on non-pub methods/fields if you consider that some IDEs will use the doc comments in various parts of the UI.  For example, VS Code and IntelliJ with both do this.

PR #1693, opened 2019-06-18
Simplify connection handling

### LINKED ISSUE #1683: Connection logic is simpler in quic-p2p
=> Some state can be removed from peer manager
=> More fleshing out needed

madadam:
This is a followup to https://github.com/maidsafe/routing/pull/1692.

Closes #1683

### DIFF
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -138,11 +138,6 @@ impl Peer {
         self.timestamp.elapsed() >= Duration::from_secs(timeout)
     }
 
-    /// Returns whether the peer is in the `Connecting` state.
-    pub fn is_connecting(&self) -> bool {
-        self.state == PeerState::Connecting
-    }
-
     /// Returns whether the peer is a full node.
     pub fn is_node(&self) -> bool {
         match self.state {
@@ -475,11 +470,6 @@ impl PeerManager {
     /// Returns an iterator over all connected peers.
     pub fn connected_peers(&self) -> impl Iterator<Item = &Peer> {
         self.peers.values().filter(|peer| peer.is_connected())
-    }
-
-    /// Returns if the given peer is in the `Connecting` state.
-    pub fn is_connecting(&self, pub_id: &PublicId) -> bool {
-        self.peers.get(pub_id).map_or(false, Peer::is_connecting)
     }
 
     /// Returns if the given peer is our proxy node.
--- a/src/peer_manager.rs
+++ b/src/peer_manager.rs
@@ -20,7 +20,10 @@ use itertools::Itertools;
 use log::LogLevel;
 use resource_proof::ResourceProof;
 use std::{
-    collections::{BTreeMap, BTreeSet, VecDeque},
+    collections::{
+        btree_map::{BTreeMap, Entry},
+        BTreeSet, VecDeque,
+    },
     net::IpAddr,
 };
 
@@ -84,26 +87,16 @@ impl<'a> From<&'a ConnectionInfo> for PeerState {
 /// Represents peer we are connected or attempting connection to.
 #[derive(Debug)]
 pub struct Peer {
-    pub_id: PublicId,
     state: PeerState,
     timestamp: Instant,
 }
 
 impl Peer {
-    pub fn new(pub_id: PublicId, state: PeerState) -> Self {
+    pub fn new(state: PeerState) -> Self {
         Self {
-            pub_id,
             state,
             timestamp: Instant::now(),
         }
-    }
-
-    pub fn pub_id(&self) -> &PublicId {
-        &self.pub_id
-    }
-
-    pub fn name(&self) -> &XorName {
-        self.pub_id.name()
     }
 
     pub fn state(&self) -> &PeerState {
@@ -468,8 +461,8 @@ impl PeerManager {
     }
 
     /// Returns an iterator over all connected peers.
-    pub fn connected_peers(&self) -> impl Iterator<Item = &Peer> {
-        self.peers.values().filter(|peer| peer.is_connected())
+    pub fn connected_peers(&self) -> impl Iterator<Item = (&PublicId, &Peer)> {
+        self.peers.iter().filter(|(_, peer)| peer.is_connected())
     }
 
     /// Returns if the given peer is our proxy node.
@@ -487,9 +480,9 @@ impl PeerManager {
     /// Returns the proxy node's name if we have a proxy.
     pub fn get_proxy_name(&self) -> Option<&XorName> {
         self.peers
-            .values()
-            .find(|peer| peer.state == PeerState::Proxy)
-            .map(Peer::name)
+            .iter()
+            .find(|(_, peer)| peer.state == PeerState::Proxy)
+            .map(|(pub_id, _)| pub_id.name())
     }
 
     /// Return old public id of expired candidate only once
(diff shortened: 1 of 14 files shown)


src/peer_map.rs
```diff
@@ -222,112 +142,52 @@ mod tests {
...
-        let node_info = node_info("198.51.100.0:5555");
-        let pub_id = *FullId::new().public_id();
+        assert!(peer_map.get_connection_info(&pub_id).is_none());
+        assert!(peer_map.get_public_id(&node_info.peer_addr).is_none());
 
+        peer_map.handle_connect_request(pub_id, node_info.clone());
         assert_eq!(
-            peer_map.handle_connected_to(ConnectionInfo::Node {
+            peer_map.get_connection_info(&pub_id),
+            Some(&ConnectionInfo::Node {
```
jeanphilippeD: Would it be more readable if we had
`let conn_info = conn_info("198.51.100.0:5555");` under `let node_info...` so we can use it in this assert?

src/states/adult.rs
```diff
@@ -285,10 +276,9 @@ impl Base for Adult {
         }
 
         if let Some(routing_msg) = self.filter_hop_message(msg)? {
-            self.dispatch_routing_message(routing_msg, outbox)
-        } else {
-            Ok(Transition::Stay)
-        }
+            self.dispatch_routing_message(routing_msg, outbox)?;
+        } 
+        Ok(Transition::Stay)
```
jeanphilippeD: Is it better form?

src/peer_manager.rs
```diff
@@ -503,9 +496,10 @@ impl PeerManager {
     /// Returns the proxy node's name if we have a proxy.
     pub fn get_proxy_name(&self) -> Option<&XorName> {
         self.peers
-            .values()
-            .find(|peer| peer.state == PeerState::Proxy)
-            .map(Peer::name)
+            .iter()
+            .filter(|(_, peer)| peer.state == PeerState::Proxy)
+            .map(|(pub_id, _)| pub_id.name())
```
jeanphilippeD: Why replace find with filter + next, find seem more appropriate?

src/peer_manager.rs
```diff
@@ -652,17 +636,18 @@ impl PeerManager {
     /// Removes the given peer. Returns whether the peer was actually present.
     /// If the peer was joining before, it is demoted back to JoiningNode and false is returned.
     pub fn remove_peer(&mut self, pub_id: &PublicId) -> bool {
-        if let Some(mut peer) = self.peers.remove(pub_id) {
-            if peer.is_or_was_joining_node() && peer.is_node() {
-                peer.state = PeerState::JoiningNode;
-                let _ = self.peers.insert(peer.pub_id, peer);
-                return false;
+        match self.peers.entry(*pub_id) {
```
jeanphilippeD: Why the refactor here? not clear the code looks clearer.

src/peer_map.rs
```diff
@@ -222,112 +142,52 @@ mod tests {
...
-
-        assert_eq!(
-            peer_map.handle_connected_to(conn_info.clone()),
-            Err(ConnectionError::Incomplete)
-        );
+        peer_map.handle_connected_to(conn_info.clone());
+        assert!(peer_map.get_connection_info(&pub_id).is_none());
+        assert!(peer_map.get_public_id(&conn_info.peer_addr()).is_none());
 
+        peer_map.handle_direct_message(pub_id, conn_info.peer_addr());
```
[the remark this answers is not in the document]
madadam: No. Previously the function returned a value indicating when the connection has been established. The test was there to make sure that value is returned only once. In this PR that logic is gone (not needed anymore), so the test is not necessary.

src/peer_manager.rs
madadam: I prefer to use the [entry API](https://doc.rust-lang.org/std/collections/btree_map/enum.Entry.html) in cases where the alternative is to use double lookup (`contains_key` followed by `insert` or `remove` followed by `insert`, etc...). It is more efficient and I think also cleaner.

src/states/adult.rs
pierrechevalier83: `dispatch_routing_message` now returns `OK(())` (adult.rs line 151), so it's necessary

src/peer_map.rs
```diff
@@ -40,172 +35,100 @@ impl PeerMap {
...
-                    let _ = self.routing_to_network.insert(pub_id, conn_info.clone());
-                    Ok(conn_info)
-                }
-                State::Routing(_) => Self::invalid_state(),
-                State::Complete(_) => Err(ConnectionError::AlreadyConnected),
-            }
-        } else {
-            Self::invalid_state()
+    pub fn handle_direct_message(&mut self, pub_id: PublicId, socket_addr: SocketAddr) {
+        if let Some(peer_type) = self.pending.remove(&socket_addr) {
```
pierrechevalier83: This entire commits looks very good to me: lots of simplifications etc., clear abstractions. I'm just unclear on one thing in the new workflow: why does a Peer need to be in the `pending` map (i.e be pending connection) for us to handle a direct message from them? Shouldn't it also work if we find them in our `reverse` map, i.e: if we're already connected to them?
I may be confusing myself, but hopefully this explains what's unclear to me and you can help me understand.

src/states/elder/mod.rs
```diff
@@ -1681,7 +1676,7 @@ impl Elder {
         if self
             .peer_mgr
             .connected_peers()
-            .filter(|p| p.is_node())
```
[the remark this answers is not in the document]
madadam: Problem is `connected_peers` returns only `impl Iterator<Item = (&PublicId, &Peer)>`, so there is no `values` unfortunately.

src/peer_map.rs
madadam: When someone whom we haven't established a connection yet sends us `DirectMessage`, the flow goes like this:
1. First quic-p2p emits `Event::ConnectedTo` which contains their connection info. We put it into `pending`.
2. Then quic-p2p emits `Event::NewMessage` which contains the socket address of the sender and the message itself (which if it is `DirectMessage` contains the public id of the sender). So we lookup the previously stored connection info in `pending` by the socket address and then store the connection in both `forward` and `reverse`.

It is necessary to do it like this, because `ConnectedTo` contains only the connection info and `NewMessage` contains only socket address (not the full connection info).

So when we receive a `DirectMessage` from someone for the first time, the above flow happens. When we receive any subsequent messages, there is no entry in `pending` for them anymore (because we removed it before), so there is nothing more to do. The full connection information is already stored in `forward` and `reverse`.

Now I'm thinking that this method (and the others too) might be poorly named as `PeerMap` doesn't really know or care about `DirectMessage`s. It only cares about the association of connection infos to public ids. I'll try to come up with better names.

jeanphilippeD: Would be good to put something along that information in the PR description.

madadam: I can put it there, problem is this PR doesn't introduce this flow. It was already introduced in the previous PR (where it was actually explained in the PR description, although in less details). So putting it here might be a bit misleading? I think I'll rather expand the comment for `PeerMap`.

jeanphilippeD: It probably would not be misleading, but likely difficult to find. A code comment is a good idea (thought if you cannot put all the information there, PR description is still probably good.)

madadam: Misleading in that it would suggest the change was introduced in this PR when in reality it wasn't. Probably not a major problem. Still, I'm putting it as a comment so it's easier to find.

src/states/elder/mod.rs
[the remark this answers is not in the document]
pierrechevalier83: Ah OK. Missed that. Looks good then.

src/peer_map.rs
pierrechevalier83: @madadam, I agree with the naming: really it's the name of the function that confused me here now you've explained it :)

Issue #1660, opened 2019-05-31
Replace signature accumulation in Delivery Group with PARSEC consensus (Probably no longer wanted)


nadiaburborough:
!!! This may not be something we want to do as we may want to avoid using Parsec for `Get`.

=> Note: as the outcome of this, we won’t enforce correct behaviour from other nodes, but we do want for correct nodes to always behave correctly (i.e: only send a message with section authority when it is definitive that a quorum of the section wants to send this message)

Bart+JP: not necessarily as easy as we thought so not necessarily a good idea to do.
Was suppose to simplify code. Avoid timers.

jeanphilippeD:
This is no longer wanted as we want to be able to do Secure Message Relay without using Parsec, For example for the Get Mutable data flow.

jeanphilippeD:
Closed as per https://github.com/maidsafe/routing/issues/1660#issuecomment-525771143

PR #1697, opened 2019-06-25
WIP feat/routing: replace signature_accumulator with parsec consensus

maqi:


Lindsey17:
PR resolves #1660

src/resource_prover.rs
```diff
@@ -30,6 +29,10 @@ use std::{
 };
 use unwrap::unwrap;
 
+/// Time (in seconds) within which a message and a quorum of signatures need to arrive to
+/// accumulate.
+pub const ACCUMULATION_TIMEOUT: Duration = Duration::from_secs(30);
```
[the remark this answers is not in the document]
maqi: hmm... I am afraid there is some testing configuration depends on this.
I will have a try to remove it, if it doesn't fail any test.

src/state_machine.rs
```diff
@@ -182,7 +182,10 @@ impl State {
             | State::Client(_)
             | State::RelocatingNode(_)
             | State::ProvingNode(_)
-            | State::Terminated => None,
+            | State::Terminated => {
+                println!("State {:?} doesn't have a chain", self);
```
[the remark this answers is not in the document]
maqi: this is pushed by mistake, I will remove it.

src/states/adult.rs
```diff
@@ -360,6 +361,11 @@ impl Approved for Adult {
         false
     }
 
+    fn handle_accumulated_message(&mut self, _digest: Digest256) -> Result<(), RoutingError> {
+        // TODO: need to ensure whether there is situation an Adult will call this function.
```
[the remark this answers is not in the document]
maqi: we then need to pass the `cache` as well  :(

src/states/elder/mod.rs
```diff
@@ -2045,6 +1942,17 @@ impl Approved for Elder {
         self.pfx_is_successfully_polled
     }
 
+    fn handle_accumulated_message(&mut self, digest: Digest256) -> Result<(), RoutingError> {
+        let routing_message =
+            if let Some(routing_message) = self.signed_messages_accumulator.get(&digest) {
```
pierrechevalier83: This should probably be `.remove`.
* It would avoid cloning
* Currently, this is going to be an ever growing cache which is probably not what we want.

src/states/elder/mod.rs
```diff
@@ -231,6 +228,7 @@ impl Elder {
             ignore_candidate_info_counter: 0,
             proving_section_cache: LruCache::with_capacity(PROVING_SECTION_CACHE_SIZE),
             pfx_is_successfully_polled: false,
+            signed_messages_accumulator: BTreeMap::new(),
```
pierrechevalier83: Not sure about this name. Why not: `signed_messages`? Calling it "accumulator" makes it sound like it was more than just a simple cache.

src/states/elder/mod.rs
```diff
@@ -2045,6 +1942,17 @@ impl Approved for Elder {
...
     }
 
+    fn handle_accumulated_message(&mut self, digest: Digest256) -> Result<(), RoutingError> {
+        let routing_message =
+            if let Some(routing_message) = self.signed_messages_accumulator.get(&digest) {
+                routing_message.clone()
+            } else {
+                return Ok(());
+            };
+        self.msg_queue.push_back(routing_message);
```
pierrechevalier83: Are we sure that pushing a msg to the msg_queue is what we want? I'm struggling to convince myself of the correctness because it's quite different from the previous code, so I'm not mapping it properly in my head...

maqi: It also took me some time to catch up with the changed behaviour.  :(
At least according to the test, this pushing to the msg_queue allowed nodes joining network properly for certain steps.

src/states/common/approved.rs
```diff
@@ -52,6 +53,9 @@ pub trait Approved: Relocated {
     /// Handles an accumulated `Offline` event.
     fn handle_offline_event(&mut self, pub_id: PublicId) -> Result<(), RoutingError>;
 
+    /// Handles an accumulated message.
+    fn handle_accumulated_message(&mut self, digest: Digest256) -> Result<(), RoutingError>;
```
pierrechevalier83: For consistency, this should probably be named: `handle_message_event`

src/states/adult.rs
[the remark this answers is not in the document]
pierrechevalier83: This doesn't really answer my question. Basically, in the big picture; what is the role of adults in message delivery?
Do we expect they never accumulate a message? (and should panic)
Do we expect they should accumulate messages in some cases (for instance if they get demoted from elders and only accumulate the message after demotion) and ignoring the message in this case should be OK for the network? (In which case, a clear comment describing this situation and convincing that the network invariants are upheld and then return `OK()` would do)
Do we need to do something? (in which case we should do it)

jeanphilippeD:
As per Qi's comment https://github.com/maidsafe/routing/pull/1697#pullrequestreview-262949933 
Closing this PR until we want to use it again.

PR #1698, opened 2019-06-25
Convert some previously ignored integration tests to unit tests

### LINKED ISSUE #1694: Decide what to do about the #[ignore]d tests due to bootstrap blacklist missing in quic-p2p
quic-p2p doesn't support bootstrap blacklist yet, so some tests that relied on it were `#[ignore]`d. Decide what to do about them: either implement bootstrap blacklist in quic-p2p and then remove the `#[ignore]` attribute, or modify the tests to not require bootstrap blacklist (perhaps by converting them to unit tests - similar to the tests in `src/states/elder/tests.rs`) or remove them.

madadam:
Closes #1694


src/peer_manager.rs
```diff
@@ -511,27 +511,13 @@ impl PeerManager {
...
-
-        // Allow only one client per IP
-        self.peers
-            .values()
-            .filter_map(Peer::client_ip)
-            .filter(|other_ip| *other_ip == client_ip)
-            .take(2)
-            .count()
-            >= 2
+    /// Checks whether we can accept more clients.
```
pierrechevalier83: I would add a line with "Allow one client per IP". Personally, it helped me understand the implementation.

src/states/elder/tests.rs
```diff
@@ -52,6 +54,30 @@ impl CandidateInfo {
     }
 }
 
+struct ClientInfo {
+    full_id: FullId,
+    addr: SocketAddr,
+}
+
+impl ClientInfo {
+    fn with_addr(addr: &str) -> Self {
```
maqi: normally just `new` for the constructor?
`with_addr` indicating there is another constructor `without addr` ?

src/states/elder/tests.rs
```diff
@@ -64,6 +90,10 @@ struct ElderUnderTest {
 
 impl ElderUnderTest {
     fn new() -> Self {
+        Self::with_min_section_size(DEFAULT_MIN_SECTION_SIZE)
```
maqi: hmm... so there is other places calling this constructor with different size?

src/states/elder/tests.rs
```diff
@@ -413,7 +469,7 @@ fn new_elder_state(
     let parsec_map = ParsecMap::new(full_id.clone(), gen_pfx_info);
     let chain = Chain::new(min_section_size, public_id, gen_pfx_info.clone());
     let peer_map = PeerMap::new();
-    let peer_mgr = PeerManager::new(public_id, true);
+    let peer_mgr = PeerManager::new(public_id, false);
```
maqi: why the flag needs to be changed from `true` to `false`?

tests/mock_network/requests.rs
```diff
@@ -227,77 +225,3 @@ fn failed_get_request() {
...
     assert_eq!(response_received_count, 1);
 }
-
-#[test]
-// TODO (quic-p2p): this test relies on a behaviour of mock-crust which drops all in-flight messages
-// on disconnect. Verify whether this behaviour should be transitioned over to quic-p2p. If yes,
-// then do so and then re-enable this test. If not, modify this test accordignly (or remove it if
-// not relevant anymore).
-#[ignore]
-fn disconnect_on_get_request() {
```
maqi: the title of the PR indicates this ignored test shall be moved to somewhere else.
however it seems not appear among the above new created tests?

src/states/elder/tests.rs
pierrechevalier83: I think `new` is idiomatic for a constructor that takes no argument. If you take an argument, especially a `&str` where the type isn't particularly meaningful, `with_addr` is a better name as it tells the user what argument they should pass with just the signature. The name of the variable is local and not part of the signature.

tests/mock_network/requests.rs
pierrechevalier83: If you look at it commit by commit, some are moved and some are removed. When tests are removed, Adam explained why in the appropriate commit messages.

src/states/elder/tests.rs
madadam: I think I originally wanted to also nave `new` which would return a default or random address, but ended up not needing it. I'm fine with renaming it to `new`, but I think @pierrechevalier83 has a valid point.

src/states/elder/tests.rs
madadam: Yes, in the `accept_previously_rejected_client_after_reaching_min_section_size` test.

src/states/elder/tests.rs
madadam: This flag control whether client IP limiting is enabled (`true` = disable, `false` = not disable). I needed it to be enabled for the new tests to work. It doesn't affect the tests that were already there.

PR #1699, opened 2019-06-26
Print rng seed before tests start

madadam:
So we still get the seed in case the tests get stuck.

Also modify the CI scripts to run tests with --features=mock instead of --features=mock_parsec to speed them up.


pierrechevalier83:
This is pretty uncontroversial. Merging it with only my review.

PR #1719, opened 2019-07-26
feat/routing: pull latest Parsec with genesis info and DKG

### LINKED ISSUE #1702: Pull latest Parsec and update to fix API breaks (Any time).
(Ignore all DKG events in routing for now)

jeanphilippeD:
Closes #1702 

-Update maidsafe_utilities & safe_crypto to fix conflict.
-Implement encrypt/decrypt similarly to relocated implementation.
-Use OsRng for the secure Rng everywhere (this may cause issue when
 to create reproductible tests when we start using the DKG functionality
 but can be better addressed then)
-StartDkg & DkgMessage are internal observation so log_or_panic if we
 see blocks with it. DkgResult will only come out once we start the
 DKG, we will use it when we do so.

 Test:
 Run `cargo test --release --features=mock_serialise` (use real parsec)
 Soak test with `cargo test --release --features=mock` (mock parsec)


src/mock/parsec/mod.rs
```diff
@@ -53,7 +53,9 @@ where
         section_hash: Digest256,
         our_id: S,
         genesis_group: &BTreeSet<S::PublicId>,
+        genesis_related_info: Vec<u8>,
         consensus_mode: ConsensusMode,
+        _secure_rng: Box<dyn rand::Rng>,
```
maqi: hmm... if it's not used, why added it in?
same to the other place

src/states/common/approved.rs
```diff
@@ -170,6 +171,17 @@ pub trait Approved: Relocated {
...
                 }
+                obs @ Observation::StartDkg(_) | obs @ Observation::DkgMessage(_) => {
+                    log_or_panic!(
+                        LogLevel::Error,
+                        "parsec_poll polled internal Observation: {:?}",
+                        obs
+                    );
+                }
+                Observation::DkgResult { .. } => {
+                    // FIXME: Handle properly
```
maqi: shall we use `TODO` ?
btw. any policy here that which key word we shall choose?

jeanphilippeD: I'll just remove it (it was a copy paste).

src/mock/parsec/mod.rs
jeanphilippeD: It is used by the real parsec, so the mock has the same interface (except it needs an extra param).

jeanphilippeD:
Fixed comment

PR #1723, opened 2019-07-30
WIP/FIRST DRAFT: Update shared state on genesis

jeanphilippeD:
Many thing will need to be cleaned and changed as well as test fixed

Version is needed in update_their_keys otherwise we try to update a key that had a latter version.
(their_keys != neighour_infos).

Moved calling update_their_keys into chain that is called also in adult state.
This is probably what we would want to do if we had another NetworkEvent, we would handle it in chain poll.

The following are passing:
loop -q target/release/deps/mock_network_tests-7b7d27ed1ac1950c --skip merge --skip churn

merge tests and churn that also do some merge are not passing.
This is expected as after merge, the 2 side would vote for different genesis block related_info.

Also, did not fix shared_state tests.


jeanphilippeD:
superseeded by other PRs

PR #1725, opened 2019-08-01
feat/routing: ensure part of our state is shared our_infos/our_history

jeanphilippeD:
This place the building block of sharing state with only part of
SharedState. Follow up will add their_keys, and neighbour_infos as
weel as their_knowledge.

Disable prunning our_infos, this is required until their_knowledge
is solid and shared state as we require updates to our_infos to be
shared by all peers in exactly the same way.

Test:
Run soak tests.

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -182,7 +182,7 @@ fn create(full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec {
             *gen_pfx_info.first_info.hash(),
             full_id,
             &gen_pfx_info.first_info.members(),
-            vec![],
+            gen_pfx_info.first_state_serialized.clone(),
             ConsensusMode::Single,
             Box::new(rand::os::OsRng::new().unwrap()),
         )
--- a/src/states/common/approved.rs
+++ b/src/states/common/approved.rs
@@ -137,8 +137,13 @@ pub trait Approved: Relocated {
                     // FIXME: Handle properly
                     unreachable!("...")
                 }
-                Observation::Genesis { .. } => {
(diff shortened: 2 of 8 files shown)


jeanphilippeD:
Fixed self review: avoid using unwrap (was left from debugging issue), avoid using assert_eq in code, use log_or_panic.

src/parsec.rs
```diff
@@ -182,7 +186,7 @@ fn create(full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec {
             *gen_pfx_info.first_info.hash(),
             full_id,
             &gen_pfx_info.first_info.members(),
-            vec![],
+            related_info,
```
madadam: Why not take the related_info from the `gen_pfx_info.first_state_serialised` instead of passing it as argument?

src/states/common/approved.rs
```diff
@@ -137,8 +137,13 @@ pub trait Approved: Relocated {
                     // FIXME: Handle properly
                     unreachable!("...")
                 }
-                Observation::Genesis { .. } => {
+                Observation::Genesis {
+                    group,
+                    related_info,
+                } => {
                     // FIXME: Validate with Chain info.
```
madadam: Is this `FIXME` still relevant?

jeanphilippeD: This is about validating the group value. Probably still relevant

src/parsec.rs
jeanphilippeD: Yes, left over from an earlier attempt.

jeanphilippeD:
Fixed review comments.

PR #1726, opened 2019-08-02
neighbour_info in shared state

jeanphilippeD:
-Use the genesis block to ensure `neighbour_infos` are shared state.
-No more need to re-vote for these sections on prefix change.
-Remove NeighbourSigs as it was only used by previous iteration of Secure messaging with ProvingSections.


jeanphilippeD:
force push rebased changes with minor conflict resolved.

Issue #1670, opened 2019-05-31
Update their_knowledge (**) on consensus on receiving an AckMessage from that section


nadiaburborough:
=> On consensus on an AckMessage
=> Note: AckMessage is sent on receiving a message from a section where the SectionProofChain contained in that message has len > 1

Dependent on https://github.com/maidsafe/routing/issues/1673

jeanphilippeD:
When receiving an AckMessage, vote for a new NetworkEvent that contains the needed information for all peers to update `their_knowledge` on consensus from their shared state: SectionInfo / Signatures / Proof that signature are valid decendants from the one we know in `their_keys`.
On consensus, Update `their_knowledge` after verifying validity (Should always be valid unless malice).
Like currently existing Proving sections, a single vote is needed to accumulate.

(We probably should split into 2 PR: 1 - Add NetworkEvent containing only SectionInfo, and implicitly trust these AckMessage consensus are valid and update the their_knowledge accordingly. 2 - Add necessary info to consensused event so all peers can validate trust.)

jeanphilippeD:
The second part of this issue is no longer needed:
> 2 - Add necessary info to consensused event so all peers can validate trust.)

The AckMessage is a section signed message. Section signed message always contain sufficient proof for all nodes at the destination to validate them since PR #1739 

As a result, we do not need to add additional information to the consensused event beside that it was received as a message that was trusted by the 2/3 quorum (which it will have since the destination is a section so all node will get the messages).

@maqi @fizyk20 
Could you both confirm and I'll close this issue as done.

jeanphilippeD:
Closing as per https://github.com/maidsafe/routing/issues/1670#issuecomment-525751754

PR #1733, opened 2019-08-07
fix/routing: update their_knowledge for correct prefix

jeanphilippeD:
Partial fix for #1670

Their knowledge is a map of their prefix to our version that they know.
The AckMessage need to contain the prefix, and the consensused message
need to consensus that prefix with the version.

We incorrectly used our own section that we consensused and then update
our prefix and version.

Test:
Update check_section_info_ack so it fails without this fix.
Run clippy+ soak test

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -574,7 +574,7 @@ impl Chain {
             | NetworkEvent::Offline(_)
             | NetworkEvent::ExpectCandidate(_)
             | NetworkEvent::PurgeCandidate(_)
-            | NetworkEvent::AckMessage(_) => {
+            | NetworkEvent::AckMessage { .. } => {
                 self.state.change == PrefixChange::None && self.our_info().is_quorum(proofs)
             }
             NetworkEvent::ProvingSections(_, _) => true,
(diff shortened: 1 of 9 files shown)


src/chain/chain.rs
```diff
@@ -1274,8 +1284,8 @@ impl Chain {
 #[cfg(feature = "mock_base")]
 impl Chain {
     /// returns the list of versions in `their_knowledge`
```
maqi: shall the comment be changed as well?

src/chain/network_event.rs
```diff
@@ -38,6 +39,14 @@ pub struct OnlinePayload {
     pub client_auth: Authority<XorName>,
 }
 
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
+pub struct AckMessagePayload {
+    /// The prefix acknowledging knowledge
```
maqi: I thought it shall be consistent with the comment of `ack_version`,
i.e. the src_prefix means `the prefix acknowledged`?

src/states/common/approved.rs
```diff
@@ -235,7 +238,7 @@ pub trait Approved: Relocated {
                         transition => return Ok(transition),
                     }
                 }
-                NetworkEvent::AckMessage(sec_info) => self.handle_ack_message_event(sec_info)?,
+                NetworkEvent::AckMessage(info) => self.handle_ack_message_event(info)?,
```
maqi: maybe use `ack_payload` or just `payload` instead of `info`, to be consistent with the variable name in the function definition?

src/chain/network_event.rs
jeanphilippeD: `the prefix acknowledged` for me would mean the prefix of the section info that ack_version is equal to. That would not be correct.
However, This is the prefix of our section when we acknowledge their section info that ack_version is equal to.

I'm happy to reword, but I want this distinction to be clear since that was the source of the bug.

maqi: >This is the prefix of our section when we acknowledge their section info that ack_version is equal to

Ah, yes. This is much clear. Maybe we shall ues this as the comment.

jeanphilippeD:
Fixed review comments.

Issue #1731, opened 2019-08-05
Move their_knowledge to shared state


jeanphilippeD:
Following #1717 
May be best addressed after #1668 #1669 

Part 1 (PR1):
`their_knowledge` is updated on consensus events, it should be shared state, and shared with joining nodes (serialized with rest of SharedState into Genesis event).
 - Move their_knowledge to SharedState
 - Call update_their_knowledge from the chain `poll` like we do for `update_their_keys`so it happens for both Adult and Elder
 - Update Genesis block to contain serialized their_knowledge as part of the tuple and check it/update it like the other fields.

Moved Part 2 to: https://github.com/maidsafe/routing/issues/1740

jeanphilippeD:
Closed Part 1 with #1735, Moved part 2 to #1740

PR #1735, opened 2019-08-08
feat/node: move their_knowledge into shared_state

maqi:
Address Part 1 of #1731

### DIFF
--- a/src/states/common/approved.rs
+++ b/src/states/common/approved.rs
@@ -9,8 +9,8 @@
 use super::Relocated;
 use crate::{
     chain::{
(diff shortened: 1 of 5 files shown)


src/states/common/approved.rs
```diff
@@ -235,7 +232,7 @@ pub trait Approved: Relocated {
                         transition => return Ok(transition),
                     }
                 }
-                NetworkEvent::AckMessage(sec_info) => self.handle_ack_message_event(sec_info)?,
+                NetworkEvent::AckMessage(_sec_info) => {}
```
jeanphilippeD: Can you add a comment like `// Handled within the chain.` within those `{}`.

PR #1739, opened 2019-08-09
feat/routing: use their_knowledge ensuring it is never ahead of a node

### LINKED ISSUE #1667: Handled failed verification of message
=> TODO: answer this: “Do we need to cache the message or is it definitely malicious?”

Risk:
9- Need decision regardless of decision little impact(low)

Dependent on:
• https://github.com/maidsafe/routing/issues/1668

### LINKED ISSUE #1737: Use their_knowledge to filter the proof that we send with each messages
- Use their_knowledge to only give the part of our history the destination needs.
- Ensure their_knowledge will only be updated when all the section nodes have seen it (Alternative ensure trust is check only once the node would have seen their knowledge i.e all incoming messages through parsec).

jeanphilippeD:
Closes #1737 
Closes #1667 

When updating their_knowledge, we need to ensure that all nodes we
will communicate with will have enough information to check trust.

In order to achieve this, we have to have all nodes in a section
acknowledging updating their_keys before sending the AckMessage.
This is not something we could do at the signature accumulation stage
as a malicious node could still send the message early.

The failure mode here is that we do not send AckMessage and we receive
more information than needed. It is much better than what we have
currently implemented as we currently send all proofs.

Additionally, we will be able in the future to detect this as Malice
and punish nodes. We will in the future also need to ensure this
SendAckMessage consensus even in case of membership change, this would
require Adult transitionning to Elder to vote, or limiting ourselves to
all the elder that should have voted intially (other will necessarily
have that information when they reach the parsec point that added them).

Note, this issue could also have been addressed by ensuring all
incomming messages go through Parsec, but it is a bigger change and
also have its own draw backs.

Test:
Run soak test

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -536,7 +536,6 @@ impl Chain {
 
     /// Returns the index of the public key in our_history that will be trusted by the target
     /// Authority
-    #[allow(unused)]
     fn proving_index(&self, target: &Authority<XorName>) -> u64 {
         self.state
             .their_knowledge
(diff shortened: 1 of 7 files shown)


octol:
I had a look at the last commit and it seems reasonable. Been out of the loop with routing for awhile though so I'd need a bit more time if you need a more thorough review

PR #1741, opened 2019-08-13
ensure old key versions are ignored

### LINKED ISSUE #1742: Ensure their_keys and their_knowledge do not go back to a previous version.
(init est 2-3 days)

jeanphilippeD:
Closes #1742 

Because of message re-ordering, some older key version could consensus after the new ones.

Ensure we ignore these so section would not loose the ability to trust each other:
 - if we notified with AckMessage that we trust a version, and then reverted their_keys to a previous version, we would no longer trust messages coming from that section.
 - Additionally the code does not handle merge, so if a previous version was for a shorter prefix, this would cause additional issues.

### DIFF
--- a/src/chain/shared_state.rs
+++ b/src/chain/shared_state.rs
@@ -7,9 +7,7 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use super::{ProofSet, SectionInfo};
-use crate::{
-    error::RoutingError, id::PublicId, sha3::Digest256, BlsPublicKey, BlsSignature, Prefix, XorName,
-};
+use crate::{error::RoutingError, sha3::Digest256, BlsPublicKey, BlsSignature, Prefix, XorName};
 use itertools::Itertools;
 use log::LogLevel;
 use maidsafe_utilities::serialisation;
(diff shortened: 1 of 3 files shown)


src/chain/shared_state.rs
```diff
@@ -242,63 +240,63 @@ impl SharedState {
...
             .their_keys
-            .keys()
-            .find(|pfx| pfx.is_compatible(&prefix))
+            .iter()
+            .find(|(pfx, _)| pfx.is_compatible(&key_info.prefix))
+            .map(|(pfx, info)| (pfx, info.version))
         {
-            let old_key = unwrap!(self.their_keys.remove(&pfx));
-            self.their_recent_keys.push_front((pfx, old_key.clone()));
+            if old_version >= key_info.version || old_pfx.is_extension_of(&key_info.prefix) {
```
maqi: I thought it was suggested that currently any `merge` related code can be skipped ?
so, maybe the extension check here, and the words in the PR description of `in case of merge`, are not necessary?

src/chain/shared_state.rs
```diff
@@ -493,32 +498,54 @@ mod test {
     //           the index is the index in the `updates` vector, which should have generated the
     //           key we expect to get for the given prefix
     fn update_keys_and_check(start_pfx: &str, updates: Vec<&str>, expected: Vec<(&str, usize)>) {
+        update_keys_and_check_with_version(
```
maqi: why need this extra function level?

jeanphilippeD: We use both, one that does not care about version, and one to test out of order.

src/chain/shared_state.rs
jeanphilippeD: This is not to handle merge, it is to handle prefix before a split coming after the split. (i.e like in the new test case that would fail without it).

src/chain/shared_state.rs
maqi: But in the places of currently calling `update_keys_and_check`, you can also just call `update_keys_and_check_with_version` directly.
It's just an extra covertion into enumerate required when you do so

jeanphilippeD: Yes, it automatically generate the version number so the test cases are not polluted with extra information that is not relevant to it.

PR #1743, opened 2019-08-13
fix split their knowledge

### LINKED ISSUE #1740: Move their_knowledge to shared state Part 2.
Direct follow up to https://github.com/maidsafe/routing/issues/1731. (init est: 3-4 days)

Part 2 (PR2 - JP looked at this a bit):
This issue should also ensure that on split both section have an updated their_knowledge.

    Problem: The SectionInfo that consensus last during split is used to trigger sending the message, so only one half of the split will send that Ack.

Things that could help:

    Avoid doing anything on prefix change that is tied to the SectionInfo that comes out.
    => Make it symmetrical, but also does not send Ack,
    Use a separate NetworkEvent to get consensus on SecurityMetadata so it is not lost like another SectionInfo event might be.
    Maybe the test should cover the split, and another adding an extra node.

### LINKED ISSUE #1744: Split consensus event for SectionInfo and security_metadata.
(init est: 3-4 days)

### LINKED ISSUE #1746: Only update other SectionInfo after verifying the trust
(i.e not do it on hops). (init est: 2-3 days (risk/test fail))
=> Yes large vuln. 
   => Destination is not an issue, but src need full section

jeanphilippeD:
Closes #1740 
Closes #1744 
Closes #1746 

- Ensure we update their_keys/their_knowledge for ourselves so we can trust message from self section, and send minimal proof to self section: prune proof similarly when we receive an AckMessage.
- Move adding received messages src SectionInfo/SecurityMetadata only when we are the final recipient and have check the signature.
- Use new NetworkEvent to handle receiving Section Signatures:
  => Fix their_knowledge after split since we no longer ignore that version when sending Ack.
  => Ensure we get their_keys, and send AckMessage for non neighbour sections that were filtered in `add_new_section`.

### DIFF
--- a/src/states/elder/mod.rs
+++ b/src/states/elder/mod.rs
@@ -15,7 +15,7 @@ use crate::{
     chain::{
         delivery_group_size, AckMessagePayload, Chain, ExpectCandidatePayload, GenesisPfxInfo,
         NetworkEvent, OnlinePayload, PrefixChange, PrefixChangeOutcome, ProvingSection,
-        SectionInfo, SendAckMessagePayload,
+        SectionInfo, SendAckMessagePayload, TheirKeyInfo,
     },
     config_handler,
     error::{BootstrapResponseError, InterfaceError, RoutingError},
@@ -41,7 +41,7 @@ use crate::{
     types::MessageId,
     utils::{self, DisplayDuration, XorTargetInterval},
     xor_name::XorName,
-    BlsPublicKeySet, ConnectionInfo, NetworkService,
+    BlsPublicKey, BlsPublicKeySet, ConnectionInfo, NetworkService,
 };
 use itertools::Itertools;
 use log::LogLevel;
@@ -435,6 +435,7 @@ impl Elder {
                 // Keep: Still relevant after prefix change.
                 NetworkEvent::NeighbourMerge(_)
                 | NetworkEvent::ProvingSections(_, _)
+                | NetworkEvent::TheirKeyInfo(_)
                 | NetworkEvent::AckMessage(_)
                 | NetworkEvent::SendAckMessage(_) => true,
             })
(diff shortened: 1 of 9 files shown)


jeanphilippeD:
Rebased: No longer WIP

src/states/elder/mod.rs
```diff
@@ -1202,6 +1192,32 @@ impl Elder {
...
+            return;
+        }
+
+        let sec_info = if let Some(si) = signed_msg.source_section() {
+            si
+        } else {
+            return;
+        };
+
+        let new_key_info = self.chain.get_their_keys_info().any(|(_, info)| {
```
maqi: just to confirm that is this `any` check correct?

src/states/elder/mod.rs
```diff
@@ -2090,22 +2106,35 @@ impl Approved for Elder {
 
         if self_sec_update {
             self.chain.reset_candidate_if_member_of(sec_info.members());
+
+            // Vote to update our self messages proof
+            self.vote_send_section_info_ack(SendAckMessagePayload {
```
maqi: Is this a new requirement? Previously we only send out section_info ack for non-self-section.

maqi:
Looking at the commit history, I think those `fix` commits could be squashed?
We could have the current separated commits' titles as entries in the commit message of the squashed one?

src/states/elder/mod.rs
jeanphilippeD: Yes it is a new requirement so message to self do not need proof since genesis.
If you look at the commit history on this PR, one of the commit is making it so we send to all. Also PR message:
```
Ensure we update their_keys/their_knowledge for ourselves so we can trust message from self section, and send minimal proof to self section: prune proof similarly when we receive an AckMessage.
```

src/states/elder/mod.rs
jeanphilippeD: In what way would it be not correct?

jeanphilippeD:
> Looking at the commit history, I think those `fix` commits could be squashed?
> We could have the current separated commits' titles as entries in the commit message of the squashed one?

They each do one specific thing each not 100% related to each other, so I'm not sure why we would squash them particularly since the last commit is already quite big.

src/states/elder/mod.rs
maqi: Well, I gave it a second thought, and cann't raise a concrete scenario of failue.
So it's just my personal nervous against the `any newer version of compatible their keys`   :)

jeanphilippeD: I think this is what we want, it is similar to `is_new_neighbour`.
At this point this is a trusted message, we could actually remove the condition all together, but we would end up doing this filtering in Parsec (where we would see we already voted for it).

Issue #1748, opened 2019-08-14
Remove `src_section` from SignedMessage / redundant with security_metadata


jeanphilippeD:
Dependant on https://github.com/maidsafe/routing/issues/1744

(init est: ~-3-4-~  2-3 days)

   => NeighbourInfo need to contain that so that we can still update our neighbour_info 
   => Remove src_section from SignedMessage.

PR #1749, opened 2019-08-15
Clean ups related to neighbour info

jeanphilippeD:
Prepare to address https://github.com/maidsafe/routing/issues/1748.

-NeighbourInfo sent as a Section message like other messages so we can trust the content and reduce number of messages
-Rename to SectionKeyInfo as it will be used in multiple context to address #1748 
-Clean up handling of message from self, ensuring their_keys are propulated with our keys from the very first block.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -8,7 +8,7 @@
 
 use super::{
     candidate::Candidate,
-    shared_state::{PrefixChange, SectionKeyInfo, SectionProofBlock, SharedState},
+    shared_state::{PrefixChange, SectionKeyInfo, SharedState},
     GenesisPfxInfo, NetworkEvent, OnlinePayload, Proof, ProofSet, SectionInfo, SectionProofChain,
 };
 use crate::{
@@ -502,15 +502,12 @@ impl Chain {
     /// Returns `true` if the `proof_chain` contains a key we have in `their_keys` and that key is
     /// for a prefix compatible with `prefix`
     pub fn check_trust(&self, prefix: &Prefix<XorName>, proof_chain: &SectionProofChain) -> bool {
-        let filtered_keys: BTreeSet<_> = if prefix.is_compatible(self.our_prefix()) {
-            self.state.our_history.all_keys().collect()
-        } else {
-            self.state
-                .get_their_keys_info()
-                .filter(|&(pfx, _)| prefix.is_compatible(pfx))
-                .map(|(_, info)| &info.key)
-                .collect()
-        };
+        let filtered_keys: BTreeSet<_> = self
+            .state
+            .get_their_keys_info()
+            .filter(|&(pfx, _)| prefix.is_compatible(pfx))
+            .map(|(_, info)| &info.key)
+            .collect();
         proof_chain
             .all_keys()
             .any(|key| filtered_keys.contains(key))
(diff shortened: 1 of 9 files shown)


src/chain/chain.rs
```diff
@@ -495,22 +495,19 @@ impl Chain {
...
-            self.state
-                .get_their_keys_info()
-                .filter(|&(pfx, _)| prefix.is_compatible(pfx))
-                .map(|(_, info)| &info.key)
-                .collect()
-        };
+        let filtered_keys: BTreeSet<_> = self
+            .state
+            .get_their_keys_info()
+            .filter(|&(pfx, _)| prefix.is_compatible(pfx))
```
maqi: just to confirm, so this means when checking trust, we always use their_keys from the shared_state.
It seems this is the main place that `our_history.all_keys()` to be used.
If so, do we still need it?

src/chain/section_info.rs
```diff
@@ -142,6 +142,15 @@ impl SectionInfo {
         NetworkEvent::SectionInfo(self)
     }
 
+    #[cfg(test)]
+    pub fn new_for_test(
```
maqi: the name of the test is bit general. Also, better have a comment explains what's the purpose of the test?

src/chain/shared_state.rs
```diff
@@ -47,17 +47,20 @@ pub struct SharedState {
...
     /// Recent keys removed from their_keys
-    pub their_recent_keys: VecDeque<(Prefix<XorName>, TheirKeyInfo)>,
+    pub their_recent_keys: VecDeque<(Prefix<XorName>, SectionKeyInfo)>,
 }
 
 impl SharedState {
     pub fn new(section_info: SectionInfo) -> Self {
         let pk = BlsPublicKey::from_section_info(&section_info);
-        let our_history = SectionProofChain::from_genesis(pk);
+        let our_history = SectionProofChain::from_genesis(pk.clone());
```
maqi: don't see other place using this `pk`
is the `clone` really needed?

src/chain/chain.rs
jeanphilippeD: proof_chain just bellow here is the same type and we use `all_keys()`

src/chain/section_info.rs
jeanphilippeD: This is not a test, this is a constructor function that should only be used by tests, so that we can construct it as needed.

src/chain/shared_state.rs
jeanphilippeD: Will fix.

jeanphilippeD:
Rebase ontop of rust 1.37 build fix.
Add additional commit ` fix/chain: ensure we do not remove neighbour when getting older version ` at the begining.
Fix review comments

maqi:
Regarding the new fix commit, not sure if it's a good idea to inlcude the detailed log msg in the commit msg.
However, if do want to do so:
1, instead of saying `On top of PR 1750`, better mention the exact commit ID that the seed will work?
2, trim down the log little bit, for example, the time tag and file path can be removed, `PublicId(name: xxxxxx...)` can be `Id(xxx)` ?

jeanphilippeD:
> Regarding the new fix commit, not sure if it's a good idea to inlcude the detailed log msg in the commit msg.
> However, if do want to do so:
> 1, instead of saying `On top of PR 1750`, better mention the exact commit ID that the seed will work?
> 2, trim down the log little bit, for example, the time tag and file path can be removed, `PublicId(name: xxxxxx...)` can be `Id(xxx)` ?

The reason I included the very specific part of the log, is that it seem the best way to understand the problem I try to fix, though I also provide an explanation so someone can skip the log.

I cannot provide a specific commit id since that commit ID will not exists anymore, however, one can get to the state where the seed would reproduce from the commit that would merge PR 1750.

It did not seem particularly good to trim that message, since any modification will make it less familiar to routing developers that have to read routing logs.

Issue #1756, opened 2019-08-19
Only accept as trusted messages with a valid signature/proof.


jeanphilippeD:
(Init estimate 3-4 days)
On starting: 1 day.

We currently accept messages implicitly if they are not signed.
We need to ensure that we accept only signed messages, or do not treat unsigned messages as if they were signed.

 => Decide what to do
 => Implement it.

Issue #1751, opened 2019-08-15
Follow up for single source message (Optional)


jeanphilippeD:
(init est 3-4 days)
This follows https://github.com/maidsafe/routing/issues/1738 (May require some part of real BLS).

If we want to retain proof that the message is from a node that actually exists in the network, we have the option of doing it:
- When signing BlsPublicKey after a churn event:
  Within the same Parsec event that contain the signature of BlsPublicKey_v1 by the signature that validate with BlsPublicKey_v0, also include the signature of PublicId of each members by the signature that validate with BlsPublicKey_v1 (NodeIdSig).

- When sending a message with a single node as source:
  The message is signed by the private key matching the `src` PublicId.
  The signature also contain an additional field `NodeIdSig`, the signature that validate with the BlsPublicKey in the `proof : SectionProofChain` providing proof that the section with the BlsPublicKey considered the `src` node to be part of its section.

PR #1757, opened 2019-08-19
feat/routing: sign and check single source message with ED key

### LINKED ISSUE #1738: Send SecurityMetadata and check validity for Single source Messages.
(init est: 3-4 days)
Needs fleshing out.
Solution seem to be:
"As a start use security_metadata variant use single signature from single ED key. at least know valid but not if real member of network."

Extract from conversation:
```
- Messages with a single source (no security right now.)
  => RFC should be using Parsec for that hence
     => could also need to do that 1660 for that.
     => could also just notify Elder that they need to send signed message with this content.
  => Depend on verifying existence of the sender on the network (if we can leave with invented identity can be on the network).
  => Hard part may or may not have benefit, cannot protect against invented client.
  => Managed node need to prove exist on the network.
  ====> As a start use security_metadata variant use single signature from single ED key. at least know valid but not if real member of network.
        (possibly the way)
```

jeanphilippeD:
Closes #1738 

- This only provide moderate additional security: We validate that the
  'src' of the message has signed the message (src+dst+content).
- Also use that to check that new connection are comming from the 'src'.

Note: For ManagedNodes (but not for Client Authority), it would be
possible to ensure the node is a valid member of a section
(https://github.com/maidsafe/routing/issues/1751).

Note: We are still allowing untrusted communication. We will need to
ensure that we do not treat untrusted communication as trusted (see
https://github.com/maidsafe/routing/issues/1756)

Test:
Clippy + soak test

### DIFF
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -135,12 +135,38 @@ impl Debug for FullSecurityMetadata {
     }
 }
 
+/// Metadata needed for verification of the single node sender.
+#[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Hash, Serialize, Deserialize)]
+pub struct SingleSrcSecurityMetadata {
+    public_id: PublicId,
+    signature: Signature,
+}
+
+impl SingleSrcSecurityMetadata {
+    pub fn verify_sig(&self, bytes: &[u8]) -> bool {
+        self.public_id
+            .signing_public_key()
+            .verify_detached(&self.signature, bytes)
+    }
+}
+
+impl Debug for SingleSrcSecurityMetadata {
+    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
+        write!(
+            formatter,
+            "SingleSrcSecurityMetadata {{ public_id: {:?}, .. }}",
+            self.public_id
+        )
+    }
+}
+
 #[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Hash, Serialize, Deserialize)]
 #[allow(clippy::large_enum_variant)]
 pub enum SecurityMetadata {
     None,
     Partial(PartialSecurityMetadata),
     Full(FullSecurityMetadata),
+    Single(SingleSrcSecurityMetadata),
 }
 
 impl Debug for SecurityMetadata {
@@ -149,6 +175,7 @@ impl Debug for SecurityMetadata {
             SecurityMetadata::None => write!(formatter, "None"),
             SecurityMetadata::Partial(pmd) => write!(formatter, "{:?}", pmd),
             SecurityMetadata::Full(smd) => write!(formatter, "{:?}", smd),
+            SecurityMetadata::Single(smd) => write!(formatter, "{:?}", smd),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/messages/mod.rs
```diff
@@ -219,16 +277,17 @@ impl SignedRoutingMessage {
...
             SecurityMetadata::Partial(_) => false,
         }
     }
 
     /// Returns the security metadata validating the message.
     pub fn source_section_key_info(&self) -> Option<&SectionKeyInfo> {
         match self.security_metadata {
-            SecurityMetadata::None => None,
-            SecurityMetadata::Partial(_) => None,
+            SecurityMetadata::None | SecurityMetadata::Partial(_) | SecurityMetadata::Single(_) => {
```
maqi: I thought even it is a `single`, it's still a security metadata from source ?

src/messages/mod.rs
```diff
@@ -319,7 +378,7 @@ impl SignedRoutingMessage {
         let invalid_sigs = match self.security_metadata {
             // unfortunately, `match`es had to be split because of the borrow checker;
             // the two cases below can return early as they have nothing left to do
```
maqi: three cases now?

src/messages/mod.rs
```diff
@@ -375,7 +436,7 @@ impl SignedRoutingMessage {
         match &self.security_metadata {
             SecurityMetadata::None => !self.content.src.is_multiple(),
             SecurityMetadata::Partial(partial) => partial.shares.len() > partial.pk_set.threshold(),
-            SecurityMetadata::Full(_) => true,
+            SecurityMetadata::Full(_) | SecurityMetadata::Single(_) => true,
```
maqi: do we need to have the same check (`!self.content.src.is_multiple()`) as in the `None` branch ?

src/routing_table/authority.rs
```diff
@@ -96,6 +97,21 @@ impl<N: Xorable + Clone + Copy + Binary + Default> Authority<N> {
...
 
+impl Authority<XorName> {
+    /// provide the name mathching a single node's public key
+    pub fn single_signing_name(&self) -> Option<&XorName> {
+        match *self {
+            Authority::ClientManager(_)
+            | Authority::NaeManager(_)
+            | Authority::Section(_)
+            | Authority::PrefixSection(_)
+            | Authority::NodeManager(_) => None,
```
maqi: wondering that in case of other authority, shall we log_or_panic directly?

src/messages/mod.rs
jeanphilippeD: We do not have a mechanism to tie the BLS SectionKey to the specific ED key of a member.
This would be what this https://github.com/maidsafe/routing/issues/1751 is about, but I don't think we can really do it for now (since we still use Parsec signatures for SectionInfo) until swtiching to real BLS (Whay I mention in the commit comment).

src/routing_table/authority.rs
jeanphilippeD: That does not seem right, I actually added a test task to cover sending messages that are not correctly signed. This seems more about sanitizing user input, and so should not panic (since we want to test it).

src/messages/mod.rs
jeanphilippeD: I am not sure.
It could have been a bug since user could sign a Authority::Section with a single signature, however, we do check that the signature is for the valid authority Client or ManagedNode. Moreover, a malicious node doing that could simply send that signed message directly, so there does not seem to be a need for any checks.

I guess the reason `SecurityMetadata::None => !self.content.src.is_multiple()` is because before we had a limitation where only Section messages were signed. I guess it could now be set to always false, but also there is an issue to address this properly everywhere: https://github.com/maidsafe/routing/issues/1756. So it does not seem we want to change `SecurityMetadata::None` here, and `SecurityMetadata::Single(_) => true` seem accurate too.

jeanphilippeD:
Rebase a fix the review comments.

Issue #1762, opened 2019-08-26
Add Lines of Code badge to README.md


willief:
Suggesting a trivial little enhancement to the README. At the time of writing the server that provides the LoC functionality is down [tokei.rs](http://tokei.rs). Hopefully this is merely temporary and we can have badges on all Maidsafe projects showing the lines of code in each component.

madadam:
Closed via https://github.com/maidsafe/routing/pull/1763

PR #1763, opened 2019-08-26
Add a Lines of Code badge

willief:
issue #1762

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 Routing - a specialised storage DHT
 
-|Crate|Documentation|Linux/OS X|Windows|Issues|
-|:---:|:-----------:|:--------:|:-----:|:----:|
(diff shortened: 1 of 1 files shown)


S-Coyle:
@willief As you mentioned in the issue raised for this, tokei.rs seems to be down at the moment. Note that this has been added into the SCL repository (https://github.com/maidsafe/safe_client_libs#safe_client_libs) earlier this year. It would be good to know that it was stable before we started adding it to more repositories. I'll check up on it daily to see if it's up and it stays up before deciding whether to approve or close this PR.

willief:
Yes, shame tokei.rs is down. I had been meaning to submit LoC PRs for all projects but decided just to submit this one until we see if its a temporary blip or if tokei.rs is no more.
I think there are other LoC providers but have not looked too hard yet.
So, yes, fully agree, lets leave this one for a while the tokei.rs status becomes clearer.

willief:
> @willief As you mentioned in the issue raised for this, tokei.rs seems to be down at the moment. Note that this has been added into the SCL repository (https://github.com/maidsafe/safe_client_libs#safe_client_libs) earlier this year. 


Yes that was me under a different GitHub persona.    No longer appropriate for me to use the trecement account.

S-Coyle:
Seems tokei.rs has been back up the last few days so at least we know it's not permanently down - I think let's leave it for another week or so and if it remains stable during that time then we can go ahead & merge (I'm checking its status daily)

willief:
its been up - it's been down...   I agree, lets leave it until we see how stable it remains.

willief:
Yes indeed, @S-Coyle  what seems like a trivial wee enhancement can easily turn into a maintenance nightmare.  As you say this is probably  best suited for the larger "flagship" repos only.

PR #1786, opened 2019-09-11
Detect if PARSEC needs pruning

### LINKED ISSUE #1784: Prune: Detect the right time for Parsec graph pruning
Detect the right time for Parsec graph pruning + Test

How to decide to vote for that prune event:
Can be pretty arbitrary, anything on the routing level. (timer (no), Could be a number of requests/responses, or accumulated serialised byte count from requests+responses - this overestimates the size of the graph and can be spammed, though)

- [x] Track byte count for votes
- [x] Detect when it's time to prune
- [x] Test

octol:
Resolves #1784

Keep track of size and number of requests/responses so we can prune/reset the parsec gossip graph before it becomes too bloated.

Note that we use `serialised_size` to accumulate the size. This may turn out to have a performance impact so we might need to address this properly in the future. Maybe to have parsec tell us the request/response size.

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -15,25 +15,59 @@ use crate::{
     utils::LogIdent,
 };
 use log::LogLevel;
+use maidsafe_utilities::serialisation;
 #[cfg(not(feature = "mock_parsec"))]
 use parsec as inner;
 use rand::Rng;
-use std::collections::{btree_map::Entry, BTreeMap};
+use std::{
+    collections::{btree_map::Entry, BTreeMap},
+    fmt,
+};
 
 #[cfg(feature = "mock_parsec")]
 pub use crate::mock::parsec::{
-    init_mock, ConsensusMode, NetworkEvent, Observation, Proof, PublicId, SecretId,
+    init_mock, ConsensusMode, Error, NetworkEvent, Observation, Proof, PublicId, SecretId,
 };
 #[cfg(not(feature = "mock_parsec"))]
-pub use parsec::{ConsensusMode, NetworkEvent, Observation, Proof, PublicId, SecretId};
+pub use parsec::{ConsensusMode, Error, NetworkEvent, Observation, Proof, PublicId, SecretId};
 
 pub type Block = inner::Block<chain::NetworkEvent, id::PublicId>;
 pub type Parsec = inner::Parsec<chain::NetworkEvent, FullId>;
 pub type Request = inner::Request<chain::NetworkEvent, id::PublicId>;
 pub type Response = inner::Response<chain::NetworkEvent, id::PublicId>;
 
+// TODO: we'll set PARSEC_SIZE_LIMIT to 1 GB once it's used outside of mock_parsec
+//#[cfg(not(feature = "mock_parsec"))]
+//const PARSEC_SIZE_LIMIT: u64 = 1_000_000_000;
+// Mock parsec request/responses are much smaller, so we need a lower limit.
+// TODO: once it's used outside of tests, this should be changed to cfg(feature = "mock_parsec")
+#[cfg(all(test, feature = "mock_parsec"))]
+const PARSEC_SIZE_LIMIT: u64 = 100;
+
+// Keep track of size in case we need to prune.
+#[derive(Default, Debug, PartialEq, Eq)]
+struct ParsecSizeCounter(u64);
+
+impl ParsecSizeCounter {
+    fn increase_size(&mut self, size: u64) {
+        self.0 += size;
+    }
+
+    #[cfg(all(test, feature = "mock_parsec"))]
+    fn needs_pruning(&self) -> bool {
+        self.0 > PARSEC_SIZE_LIMIT
+    }
+}
+
+impl fmt::Display for ParsecSizeCounter {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "size: {}", self.0)
+    }
+}
+
 pub struct ParsecMap {
     map: BTreeMap<u64, Parsec>,
+    size_counter: ParsecSizeCounter,
 }
 
 impl ParsecMap {
@@ -43,13 +77,15 @@ impl ParsecMap {
             *gen_pfx_info.first_info.version(),
             create(full_id, gen_pfx_info),
         );
+        let size_counter = ParsecSizeCounter::default();
 
-        Self { map }
+        Self { map, size_counter }
     }
 
     pub fn init(&mut self, full_id: FullId, gen_pfx_info: &GenesisPfxInfo, log_ident: &LogIdent) {
         if let Entry::Vacant(entry) = self.map.entry(*gen_pfx_info.first_info.version()) {
             let _ = entry.insert(create(full_id, gen_pfx_info));
+            self.size_counter = ParsecSizeCounter::default();
             info!(
                 "{}: Init new Parsec, genesis = {:?}",
                 log_ident, gen_pfx_info
@@ -64,6 +100,13 @@ impl ParsecMap {
         pub_id: id::PublicId,
         log_ident: &LogIdent,
     ) -> (Option<DirectMessage>, bool) {
+        // Increase the size before fetching the parsec to satisfy the borrow checker
+        self.count_size(
+            serialisation::serialised_size(&request),
+            msg_version,
+            log_ident,
+        );
+
         let parsec = if let Some(parsec) = self.map.get_mut(&msg_version) {
             parsec
         } else {
@@ -90,6 +133,13 @@ impl ParsecMap {
         pub_id: id::PublicId,
         log_ident: &LogIdent,
     ) -> bool {
+        // Increase the size before fetching the parsec to satisfy the borrow checker
+        self.count_size(
+            serialisation::serialised_size(&response),
+            msg_version,
+            log_ident,
+        );
+
         let parsec = if let Some(parsec) = self.map.get_mut(&msg_version) {
             parsec
         } else {
@@ -169,6 +219,22 @@ impl ParsecMap {
 
         parsec.has_unpolled_observations()
     }
+
+    fn count_size(&mut self, size: u64, msg_version: u64, log_ident: &LogIdent) {
+        if self.last_version() == msg_version && self.map.contains_key(&msg_version) {
+            self.size_counter.increase_size(size);
+            trace!(
+                "{} - Parsec size is now estimated to: {}.",
+                log_ident,
+                self.size_counter,
+            );
+        }
+    }
+
+    #[cfg(all(test, feature = "mock_parsec"))]
+    fn needs_pruning(&self) -> bool {
+        self.size_counter.needs_pruning()
+    }
 }
 
 /// Create Parsec instance.
@@ -222,5 +288,173 @@ fn new_rng() -> Box<dyn Rng> {
         };
 
         Box::new(rng)
+    }
+}
+
+#[cfg(all(test, feature = "mock_parsec"))]
+mod tests {
+    use super::*;
+    use crate::{chain::SectionInfo, routing_table::Prefix, xor_name::XorName};
+    use serde::Serialize;
+    use unwrap::unwrap;
+
+    const DEFAULT_MIN_SECTION_SIZE: usize = 4;
+
+    #[test]
+    fn parsec_size_counter() {
+        let mut counter = ParsecSizeCounter::default();
+        assert!(!counter.needs_pruning());
+        counter.increase_size(PARSEC_SIZE_LIMIT);
+        assert!(!counter.needs_pruning());
+        counter.increase_size(1);
+        assert!(counter.needs_pruning());
+    }
+
+    fn create_full_ids() -> Vec<FullId> {
+        (0..DEFAULT_MIN_SECTION_SIZE)
+            .map(|_| FullId::new())
+            .collect()
+    }
+
+    fn create_gen_pfx_info(full_ids: Vec<FullId>, version: u64) -> GenesisPfxInfo {
+        let members = full_ids.iter().map(|id| *id.public_id()).collect();
+        let section_info = unwrap!(SectionInfo::new_for_test(
+            members,
+            Prefix::<XorName>::default(),
+            version
+        ));
+        GenesisPfxInfo {
+            first_info: section_info,
+            first_state_serialized: Vec::new(),
+            latest_info: SectionInfo::default(),
+        }
+    }
+
+    fn create_parsec_map(size: u64) -> ParsecMap {
+        let log_ident = LogIdent::new("node");
+        let full_ids = create_full_ids();
+        let full_id = full_ids[0].clone();
+
+        let gen_pfx_info = create_gen_pfx_info(full_ids.clone(), 0);
+        let mut parsec_map = ParsecMap::new(full_id.clone(), &gen_pfx_info);
+
+        for parsec_no in 1..=size {
+            let gen_pfx_info = create_gen_pfx_info(full_ids.clone(), parsec_no);
+            parsec_map.init(full_id.clone(), &gen_pfx_info, &log_ident);
+        }
+
+        parsec_map
+    }
+
+    fn add_to_parsec_map(parsec_map: &mut ParsecMap, version: u64) {
+        let log_ident = LogIdent::new("node");
+        let full_ids = create_full_ids();
+        let full_id = full_ids[0].clone();
+
+        let gen_pfx_info = create_gen_pfx_info(full_ids, version);
+        parsec_map.init(full_id, &gen_pfx_info, &log_ident);
+    }
+
+    trait HandleRequestResponse {
+        fn handle(
+            &self,
+            parsec_map: &mut ParsecMap,
+            msg_version: u64,
+            pub_id: &id::PublicId,
+            log_ident: &LogIdent,
+        );
+    }
+
+    impl HandleRequestResponse for Request {
+        fn handle(
+            &self,
+            parsec_map: &mut ParsecMap,
+            msg_version: u64,
+            pub_id: &id::PublicId,
+            log_ident: &LogIdent,
+        ) {
+            let _ = parsec_map.handle_request(msg_version, self.clone(), *pub_id, &log_ident);
+        }
+    }
+
+    impl HandleRequestResponse for Response {
+        fn handle(
+            &self,
+            parsec_map: &mut ParsecMap,
+            msg_version: u64,
+            pub_id: &id::PublicId,
+            log_ident: &LogIdent,
+        ) {
+            let _ = parsec_map.handle_response(msg_version, self.clone(), *pub_id, &log_ident);
+        }
+    }
+
+    fn handle_msgs_just_below_prune_limit<T: HandleRequestResponse + Serialize>(
+        parsec_map: &mut ParsecMap,
+        msg_version: u64,
+        msg: &T,
+        pub_id: &id::PublicId,
+        log_ident: &LogIdent,
+    ) {
+        let msg_size = serialisation::serialised_size(&msg);
+        let msg_size_limit = PARSEC_SIZE_LIMIT / msg_size;
+
+        // Handle msg_size_limit msgs which should trigger pruning needed on the next
+        // msg if it's against the latest parsec.
+        for _ in 0..msg_size_limit {
+            msg.handle(parsec_map, msg_version, pub_id, log_ident);
+        }
+
+        // Make sure we don't cross the prune limit
+        assert_eq!(parsec_map.needs_pruning(), false);
+    }
+
+    fn check_prune_needed_after_msg<T: HandleRequestResponse + Serialize>(
+        msg: T,
+        parsec_age: u64,
+        prune_needed: bool,
+    ) {
+        let full_id = FullId::new();
+        let pub_id = full_id.public_id();
+        let number_of_parsecs = 2;
+
+        let log_ident = LogIdent::new("node");
+        let mut parsec_map = create_parsec_map(number_of_parsecs);
+
+        // Sometimes send to an old parsec
+        let msg_version = number_of_parsecs - parsec_age;
+
+        handle_msgs_just_below_prune_limit(&mut parsec_map, msg_version, &msg, &pub_id, &log_ident);
+
+        msg.handle(&mut parsec_map, msg_version, pub_id, &log_ident);
+        assert_eq!(parsec_map.needs_pruning(), prune_needed);
+
+        // Add another parsec should always reset `needs_pruning()`
+        add_to_parsec_map(&mut parsec_map, number_of_parsecs + 1);
+        assert_eq!(parsec_map.needs_pruning(), false);
+    }
+
+    #[test]
+    fn prune_not_required_for_resp_to_old_parsec() {
+        let parsec_age = 1;
+        check_prune_needed_after_msg(Response::new(), parsec_age, false);
+    }
+
+    #[test]
+    fn prune_required_for_resp_to_latest_parsec() {
+        let parsec_age = 0;
+        check_prune_needed_after_msg(Response::new(), parsec_age, true);
+    }
+
+    #[test]
+    fn prune_not_required_for_req_to_old_parsec() {
+        let parsec_age = 1;
+        check_prune_needed_after_msg(Request::new(), parsec_age, false);
+    }
+
+    #[test]
+    fn prune_required_for_req_to_latest_parsec() {
+        let parsec_age = 0;
+        check_prune_needed_after_msg(Request::new(), parsec_age, true);
     }
 }
--- a/src/mock/parsec/mod.rs
+++ b/src/mock/parsec/mod.rs
@@ -256,21 +256,23 @@ where
     }
 }
 
-#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Debug)]
-pub struct Request<T: NetworkEvent, P: PublicId>(PhantomData<(T, P)>);
+// Contains an additional `u8` so that the size is > 0. This is needed when counting sizes to
+// determine parsec graph pruning.
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Debug, Default)]
+pub struct Request<T: NetworkEvent, P: PublicId>(PhantomData<(T, P)>, u8);
 
 impl<T: NetworkEvent, P: PublicId> Request<T, P> {
-    fn new() -> Self {
-        Request(PhantomData)
+    pub fn new() -> Self {
+        Request(PhantomData, 1)
     }
 }
 
-#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Debug)]
-pub struct Response<T: NetworkEvent, P: PublicId>(PhantomData<(T, P)>);
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Debug, Default)]
+pub struct Response<T: NetworkEvent, P: PublicId>(PhantomData<(T, P)>, u8);
 
 impl<T: NetworkEvent, P: PublicId> Response<T, P> {
-    fn new() -> Self {
-        Response(PhantomData)
+    pub fn new() -> Self {
+        Response(PhantomData, 1)
     }
 }
 


src/parsec.rs
```diff
@@ -63,11 +101,19 @@ impl ParsecMap {
...
-            return (None, false);
-        };
+        let (parsec, size_counter) =
+            if let Some((parsec, size_counter)) = self.map.get_mut(&msg_version) {
+                (parsec, size_counter)
+            } else {
+                return (None, false);
+            };
+
+        size_counter.increment_size(serialisation::serialised_size(&request));
```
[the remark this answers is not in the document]
octol: The two blocks (104 - 116) and (138 - 150) are slightly different though (different return values and request vs response)

src/chain/chain.rs
```diff
@@ -377,9 +380,9 @@ impl Chain {
...
     }
 
-    /// Finalises a split or merge - creates a `GenesisPfxInfo` for the new graph and returns the
-    /// cached and currently accumulated events.
-    pub fn finalise_prefix_change(&mut self) -> Result<PrefixChangeOutcome, RoutingError> {
+    /// Finalises a split, merge or parsec prune by resetting state - creates a `GenesisPfxInfo`
+    /// for the new graph and returns the cached and currently accumulated events.
+    pub fn reset_and_get_parsec_init_info(&mut self) -> Result<ResetParsecOutcome, RoutingError> {
         // TODO: Bring back using their_knowledge to clean_older section in our_infos
         self.check_and_clean_neighbour_infos(None);
```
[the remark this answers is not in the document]
octol: Good point, probably we don't?

src/parsec.rs
```diff
@@ -31,24 +33,60 @@ pub type Parsec = inner::Parsec<chain::NetworkEvent, FullId>;
 pub type Request = inner::Request<chain::NetworkEvent, id::PublicId>;
 pub type Response = inner::Response<chain::NetworkEvent, id::PublicId>;
 
+// TODO: These two needs to be determined from better data.
+// For now: 640kb ought to be enough for anybody!
+const PARSEC_SIZE_LIMIT: u64 = 640_000;
```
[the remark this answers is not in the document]
octol: Nah both those two numbers I just made up on the spot until I update them to reflect what's observed in tests.

About 640kb and the comment preceding it: 
https://www.computerworld.com/article/2534312/the--640k--quote-won-t-go-away----but-did-gates-really-say-it-.html

src/parsec.rs
```diff
@@ -160,14 +220,33 @@ impl ParsecMap {
...
+        self.map.get_mut(&version).map(|(p, _)| p)
+    }
+
+    #[allow(unused)]
+    fn needs_pruning(&self, version: u64) -> Option<bool> {
+        self.map.get(&version).map(|(_, s)| s.needs_pruning())
+    }
+
+    #[allow(unused)]
+    fn last_needs_pruning(&self) -> bool {
```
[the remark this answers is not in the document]
octol: No not yet, I'll add that!

src/states/elder/mod.rs
```diff
@@ -373,17 +373,21 @@ impl Elder {
...
-        let PrefixChangeOutcome {
+        let ResetParsecOutcome {
             gen_pfx_info,
             mut cached_events,
             completed_events,
-        } = self.chain.finalise_prefix_change()?;
+        } = self.chain.reset_and_get_parsec_init_info()?;
         self.gen_pfx_info = gen_pfx_info;
         self.chain.reset_candidate();
         self.peer_mgr.reset_candidate();
```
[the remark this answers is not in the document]
octol: Yes you're right I think these particular changes are not fully refined yet so I'm thinking of removing them from the PR

octol:
>  skimmed this quickly, overall I'm a tiny bit confused, the 3 commits seem to be unrelated as in they could have been 3 PR not rebased on each others.

> It seem strange in first commit to have that Prune event that we won't use. It is probably meant to be in another PR or even in a commit that does not have unimplemented code.

Correct, I'm actually now just consider splitting it into 2 PRs (ParsecMap changes + ParsecPrune event) and drop the finalise method changes for now

src/parsec.rs
```diff
@@ -31,24 +33,58 @@ pub type Parsec = inner::Parsec<chain::NetworkEvent, FullId>;
 pub type Request = inner::Request<chain::NetworkEvent, id::PublicId>;
 pub type Response = inner::Response<chain::NetworkEvent, id::PublicId>;
 
+// TODO: These two needs to be determined from better data.
+// For now: 640kb ought to be enough for anybody!
+const PARSEC_SIZE_LIMIT: u64 = 640_000;
+// The maximum number of handled requests/responses
+const PARSEC_EVENT_LIMIT: u64 = 1000;
```
jeanphilippeD: Do we need both an event limit and a size, why not use one or the other (Maybe more the size).

src/parsec.rs
```diff
@@ -31,24 +33,58 @@ pub type Parsec = inner::Parsec<chain::NetworkEvent, FullId>;
...
+
+impl fmt::Display for ParsecSizeCounter {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "size: {}, count: {}", self.size_count, self.event_count)
+    }
+}
+
 pub struct ParsecMap {
-    map: BTreeMap<u64, Parsec>,
+    map: BTreeMap<u64, (Parsec, ParsecSizeCounter)>,
```
jeanphilippeD: would it be worth using a struct instead of tuple for the value?

src/parsec.rs
```diff
@@ -89,11 +133,19 @@ impl ParsecMap {
...
-            return false;
-        };
+        let (parsec, size_counter) =
+            if let Some((parsec, size_counter)) = self.map.get_mut(&msg_version) {
+                (parsec, size_counter)
+            } else {
+                return false;
+            };
+
+        size_counter.increase_size(serialisation::serialised_size(&response));
```
jeanphilippeD: Are we processing the response again akin to a serialization to get the size?
If so, maybe that is not ideal.

src/parsec.rs
```diff
@@ -63,11 +99,19 @@ impl ParsecMap {
         pub_id: id::PublicId,
         log_ident: &LogIdent,
     ) -> (Option<DirectMessage>, bool) {
-        let parsec = if let Some(parsec) = self.map.get_mut(&msg_version) {
-            parsec
-        } else {
-            return (None, false);
-        };
+        let (parsec, size_counter) =
+            if let Some((parsec, size_counter)) = self.map.get_mut(&msg_version) {
```
maqi: better have a helper function to de-duplicate the this block (102-114) of code and the below.

src/parsec.rs
jeanphilippeD: A struct for (parsec, size_counter) would probably allow you to avoid duplication here.

src/parsec.rs
```diff
@@ -160,14 +218,33 @@ impl ParsecMap {
...
+    }
+
+    #[allow(unused)]
+    fn needs_pruning(&self, version: u64) -> Option<bool> {
+        self.map.get(&version).map(|(_, s)| s.needs_pruning())
+    }
+
+    #[allow(unused)]
+    fn last_needs_pruning(&self) -> bool {
+        if let Some(s) = self.map.values().last().map(|(_, s)| s.needs_pruning()) {
```
maqi: do we really need the log_or_panic? or maybe we shall use `unwrap_or` to return `false` in case of empty?

src/parsec.rs
```diff
@@ -31,24 +33,58 @@ pub type Parsec = inner::Parsec<chain::NetworkEvent, FullId>;
...
+        Default::default()
+    }
+
+    fn increase_size(&mut self, serialised_size: u64) {
+        self.event_count += 1;
+        self.size_count += serialised_size;
+    }
+
+    fn needs_pruning(&self) -> bool {
+        (self.size_count > PARSEC_SIZE_LIMIT) || (self.event_count > PARSEC_EVENT_LIMIT)
```
jeanphilippeD: Might it be good to have PARSEC_SIZE_LIMIT be the default initialization for a member here, so you can use different limit in your unit test (also maybe it could be dynamic in the future), but also that mean the the test that change logic do not need to change when limit change.

not sure about that.

Also it might be good in tests to reduce that limit a lot so pruning happen more regularly (and also we got a bunch of nodes running so we have less memory for each).

src/parsec.rs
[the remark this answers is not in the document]
jeanphilippeD: I'd rather your limit be 1GB then.

src/parsec.rs
```diff
@@ -198,3 +275,146 @@ fn create(full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec {
...
+
+        let log_ident = LogIdent::new("node");
+        let mut parsec_map = create_parsec_map();
+
+        // First handle responses with parsec 1
+        for _ in 0..PARSEC_EVENT_LIMIT {
+            assert!(!parsec_map.handle_response(1, response.clone(), *pub_id, &log_ident));
+        }
+
+        assert_eq!(parsec_map.needs_pruning(0), Some(false));
```
jeanphilippeD: Probably can be more compact:

```
assert_eq!(
    need_pruning(&parsec_map, 4),
    vec![Some(false), Some(false), Some(false), None] );
assert_eq!(parsec_map.last_needs_pruning(), false);

File: src/parsec.rs [hunk 13]
```diff
@@ -160,14 +218,33 @@ impl ParsecMap {
...
 
         parsec.has_unpolled_observations()
     }
+
+    fn get_parsec_mut(&mut self, version: u64) -> Option<&mut Parsec> {
+        self.map.get_mut(&version).map(|(p, _)| p)
+    }
+
+    #[allow(unused)]
+    fn needs_pruning(&self, version: u64) -> Option<bool> {
```
jeanphilippeD: Why do you need to specify the version, when using it, why would you care for anything but the last one?

src/parsec.rs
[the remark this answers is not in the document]
maqi: maybe 640k is good for the testing purpose.  :)
maybe we shall have a low number under the `test` cfg guard, and a larger number for the production

src/parsec.rs
maqi: yeah, same doubt. And I will prefer just use size.

src/parsec.rs
maqi: similiar idea as in my comment of https://github.com/maidsafe/routing/pull/1786/files#r324227079

src/parsec.rs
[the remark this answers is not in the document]
jeanphilippeD: Not fussed, whatever option needs to allow us to test with real parsec and mock parsec.

src/parsec.rs
octol: Well. On one hand pruning an older parsec seems highly infeasible, on the other hand not keeping track of the size of older parsec seems like it could be an exploitable loophole...

src/parsec.rs
octol: Agree that's a nice one!

src/parsec.rs
octol: I've wrapped those two in a struct in a fixup commit

src/parsec.rs
jeanphilippeD: In what way? The previous parsec is on its way out, either because of prune or prefix change. What could we do with that information even if we had it?

octol: For example refuse to handle further requests for an older parsec if a client is deliberately spamming it?

src/parsec.rs
octol: In the latest update I've remove the event limit and only kept the size limit

src/parsec.rs
jeanphilippeD: The parsec will be gone, also client cannot spam parsec, only other node sending gossip request/response.

src/parsec.rs
octol: Good point, I've changed it to `unwrap_or(false)`

src/parsec.rs
octol: > The parsec will be gone
Not sure I follow, don't we keep at least one old parsec, which can potentially stay for a long time?
> also client cannot spam parsec, only other node sending gossip request/response.
You're right, what about other (malicious) nodes spamming the old parsec though?

jeanphilippeD: > don't we keep at least one old parsec, which can potentially stay for a long time?

jeanphilippeD: > don't we keep at least one old parsec, which can potentially stay for a long time?
> You're right, what about other (malicious) nodes spamming the old parsec though?

Yes, that could be a problem, but I'm not sure keeping track size would help. What can you do with that information. (is it a solution in search of a problem?).

Probably what needs to happen to stop that loophole is ensuring that we don't feed parsec more data when it is complete (i.e maybe when we get a request, we just send a request back, and no longer reply to request from that node or something like that).

But we can deal with it separately (this is not a new problem, i.e it happened before with split).

src/parsec.rs
```diff
@@ -17,38 +17,110 @@ use crate::{
...
+}
+
+impl ParsecAndCounter {
+    fn handle_request(
+        &mut self,
+        src: &id::PublicId,
+        req: Request,
+        log_ident: &LogIdent,
+    ) -> Result<Response, Error> {
+        self.count_size(mem::size_of_val(&req), log_ident);
```
jeanphilippeD: Is `mem::size_of_val` doing the right thing? i.e is it taking into account that there is heap memory here?

src/parsec.rs
```diff
@@ -17,38 +17,110 @@ use crate::{
...
+            },
         );
 
         Self { map }
     }
 
     pub fn init(&mut self, full_id: FullId, gen_pfx_info: &GenesisPfxInfo, log_ident: &LogIdent) {
         if let Entry::Vacant(entry) = self.map.entry(*gen_pfx_info.first_info.version()) {
-            let _ = entry.insert(create(full_id, gen_pfx_info));
+            let _ = entry.insert(ParsecAndCounter {
```
jeanphilippeD: Seem like you could either update create or add a new wrapping function?

src/parsec.rs
```diff
@@ -198,3 +296,130 @@ fn create(full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec {
...
+    }
+
+    fn need_pruning(parsec_map: &ParsecMap, up_to_version: u64) -> Vec<Option<bool>> {
+        (0..=up_to_version)
+            .map(|ii| parsec_map.needs_pruning(ii))
+            .collect()
+    }
+
+    #[test]
+    fn parsec_map_check_prune_required() {
```
jeanphilippeD: wonder if that very long test could be split into mutltiple simple test (maybe following AAA testing pattern.)

src/parsec.rs
```diff
@@ -136,22 +208,20 @@ impl ParsecMap {
     }
 
     pub fn gossip_recipients(&self) -> Vec<&id::PublicId> {
-        self.map
-            .values()
+        self.get_parsec_values()
             .last()
-            .map(|parsec| parsec.gossip_recipients().collect())
+            .map(|p| p.gossip_recipients().collect())
```
maqi: prefer remain with the `parsec` ?

src/parsec.rs
```diff
@@ -160,14 +230,42 @@ impl ParsecMap {
...
         };
 
         parsec.has_unpolled_observations()
     }
+
+    fn get_parsec_mut(&mut self, version: u64) -> Option<&mut Parsec> {
+        self.map.get_mut(&version).map(|p| &mut p.parsec)
+    }
+
+    fn get_parsec_values(&self) -> impl Iterator<Item = &Parsec> {
```
maqi: get_parsecs ?

src/parsec.rs
```diff
@@ -17,38 +17,110 @@ use crate::{
...
+    ) -> Result<(), Error> {
+        self.count_size(mem::size_of_val(&resp), log_ident);
+        self.parsec.handle_response(src, resp)
+    }
+
+    fn count_size(&mut self, size: usize, log_ident: &LogIdent) {
+        self.size_counter.increase_size(size);
+        trace!(
+            "{} - Parsec size is now estimated to: {}.",
+            log_ident,
```
maqi: not sure whether it's worth to pass down this LogIdent via such couple of function interface.
we can still log the size info (if needed) on elder level to avoid pass it down?

src/parsec.rs
octol: I'll remove the `needs_pruning(version)` and that should cut the test in half

src/parsec.rs
octol: Yes good catch! That one actually slipped through by mistake

src/parsec.rs
octol: Do you mean move the `trace` to `Approved::handle_parsec_request/response`?

maqi: well, I don't think that trace is mandatory. 
However, if you think such log is useful, then, yes.

jeanphilippeD: I prefer it like it is now though. Either remove the trace or keep it in the subfunction instead of duplicating the code and adding noise to the upper functions.

src/parsec.rs
octol: Well unlike `size_of` it's runtime check, so you'd think so. It's not super clear from the docs though if it's done recursively...

jeanphilippeD: Seems to be for array size, but if you can write a test or a playground example we can see.

octol: Yes you're right: https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=0dc9cb45007defdb046830cf541d0dad

jeanphilippeD: Updated: https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=6b13a34e5e9b498c4d8260cafd28f3bd

src/parsec.rs
```diff
@@ -161,14 +236,35 @@ impl ParsecMap {
...
+        let parsec = if let Some(parsec) = self.get_parsec_values().last() {
             parsec
         } else {
             return false;
         };
 
         parsec.has_unpolled_observations()
     }
+
+    fn get_parsec_mut(&mut self, version: u64) -> Option<&mut Parsec> {
```
jeanphilippeD: Might as well be `get_last_parsec_values`\`get_last_parsec_values_mut`

src/parsec.rs
```diff
@@ -161,14 +236,35 @@ impl ParsecMap {
...
+
+    fn get_parsec_values(&self) -> impl Iterator<Item = &Parsec> {
+        self.map.values().map(|p| &p.parsec)
+    }
+
+    fn get_parsec_values_mut(&mut self) -> impl Iterator<Item = &mut Parsec> {
+        self.map.values_mut().map(|p| &mut p.parsec)
+    }
+
+    #[allow(unused)]
```
jeanphilippeD: is it best to use `#[cfg(all(test, feature = "mock_parsec"))]`, this way we are not silencing dead code warning

src/parsec.rs
```diff
@@ -224,3 +320,100 @@ fn new_rng() -> Box<dyn Rng> {
...
+#[cfg(all(test, feature = "mock_parsec"))]
+mod tests {
+    use super::*;
+    use crate::{chain::SectionInfo, routing_table::Prefix, xor_name::XorName};
+    use unwrap::unwrap;
+
+    const DEFAULT_MIN_SECTION_SIZE: usize = 4;
+
+    #[test]
+    fn parsec_size_counter() {
```
jeanphilippeD: nit pick probably try to group `#[test]` together and not insert helper function between them

src/parsec.rs
```diff
@@ -224,3 +320,100 @@ fn new_rng() -> Box<dyn Rng> {
...
+            assert!(!parsec_map.handle_response(1, response.clone(), *pub_id, &log_ident));
+        }
+        assert_eq!(parsec_map.needs_pruning(), false);
+
+        // Then handle responses with parsec 2 which should trigger pruning
+        for _ in 0..PARSEC_SIZE_LIMIT {
+            assert!(parsec_map.handle_response(2, response.clone(), *pub_id, &log_ident));
+        }
+        assert_eq!(parsec_map.needs_pruning(), false);
+        assert!(parsec_map.handle_response(2, response.clone(), *pub_id, &log_ident));
```
jeanphilippeD: I'm not 100% clear what you are testing with that 1 instance, that it does not affect need_pruning (then it does not test it) or that it does not affect the count on second instance, then it does?

src/parsec.rs
```diff
@@ -224,3 +320,100 @@ fn new_rng() -> Box<dyn Rng> {
...
+        assert_eq!(parsec_map.needs_pruning(), true);
+
+        let parsec_map_vec = parsec_map_to_vec(parsec_map.map);
+        assert_eq!(
+            parsec_map_vec,
+            vec![
+                (0u64, ParsecSizeCounter(0)),
+                (1u64, ParsecSizeCounter(PARSEC_SIZE_LIMIT + 1)),
+                (2u64, ParsecSizeCounter(PARSEC_SIZE_LIMIT + 1)),
```
jeanphilippeD: I don't get it, why is this test passing, for 2, we are calling handle_response PARSEC_SIZE_LIMIT +1, but for 1 are we not only calling it PARSEC_SIZE_LIMIT?

src/parsec.rs
```diff
@@ -224,3 +320,100 @@ fn new_rng() -> Box<dyn Rng> {
...
+        let response = Response::new();
+        let full_id = FullId::new();
+        let pub_id = full_id.public_id();
+
+        let log_ident = LogIdent::new("node");
+        let mut parsec_map = create_parsec_map();
+
+        // First handle responses with parsec 1 which should not trigger a need for pruning.
+        for _ in 0..=PARSEC_SIZE_LIMIT {
+            assert!(!parsec_map.handle_response(1, response.clone(), *pub_id, &log_ident));
```
jeanphilippeD: Should we send responses and requests to test both?

jeanphilippeD: Why do we assert on the handle_response result? Is it not assume it will be true for this test?

src/parsec.rs
```diff
@@ -224,3 +320,100 @@ fn new_rng() -> Box<dyn Rng> {
...
+    fn parsec_map_check_prune_required() {
+        let response = Response::new();
+        let full_id = FullId::new();
+        let pub_id = full_id.public_id();
+
+        let log_ident = LogIdent::new("node");
+        let mut parsec_map = create_parsec_map();
+
+        // First handle responses with parsec 1 which should not trigger a need for pruning.
+        for _ in 0..=PARSEC_SIZE_LIMIT {
```
jeanphilippeD: Is the serialized size of the Response 1 then?

src/parsec.rs
maqi: I will vote for removing the trace then.

src/parsec.rs
octol: It's not the last one though?

src/parsec.rs
octol: Agree good point

src/parsec.rs
octol: Hmm I'm not sure. In general yes I agree, but it's nice to keep functions local to where they are used

src/parsec.rs
```diff
@@ -15,41 +15,118 @@ use crate::{
...
+            "{} - Parsec size is now estimated to: {}.",
+            log_ident,
+            self.size_counter,
+        );
+    }
+}
+
 pub struct ParsecMap {
-    map: BTreeMap<u64, Parsec>,
+    map: BTreeMap<u64, ParsecAndCounter>,
```
jeanphilippeD: Sorry for that coming so late in the design, do we need `ParsecAndCounter`?
why is it not 
pub struct ParsecMap {
   map: BTreeMap<u64, Parsec>,
   size_counter: ParsecSizeCounter,
}

Essentially we reset it to 0 when a new parsec is added, and increment it only if it is the last parsec instance.
We are already doing special thing in this case `let poll = self.last_version() == msg_version;` It seem it would be reasonable to add an if condition, or something similar.

src/parsec.rs
maqi: For the 1 instance, the loop is `0..=PARSEC_SIZE_LIMIT`, which has an extra handle.
Though, I normally won't prefer using such `trick`.

src/parsec.rs
maqi: For the 1 instance, the loop is `0..=PARSEC_SIZE_LIMIT`, which has an extra handle.
Though, I normally won't prefer using such `trick`.

jeanphilippeD: ahhh, I missed that bit, so these for loops are not 100% explicit, probably better to have sub functions that take a number so it is a lot clearer what is happening.

src/parsec.rs
octol: That it goes from not needing pruning, to needing pruning on `PARSEC_SIZE_LIMIT + 1`

src/parsec.rs
octol: We are calling handle_response `PARSEC_SIZE_LIMIT + 1` times for both cases

src/parsec.rs
octol: Yes that sounds like a good idea

octol: Just to assert that we're not sending to the latest parsec

src/parsec.rs
octol: Yes, it's a bit implicit assumption actually, maybe one could make that explicit

src/parsec.rs
octol: Well it was added mostly as a response to previous review comments: "A struct for (parsec, size_counter) would probably allow you to avoid duplication here."

I'm still of the opinion that keeping a counter per parsec seems the cleanest option. Especially since it sounded like we might end up pruning frequently. Some of the code that uses this though was removed in an earlier revision though due to review comments

src/parsec.rs
octol: Not sure I follow. The two are testing different things. What sub functions are you suggesting? The `0..=PARSEC_SIZE_LIMIT` was done like this as `0..PARSEC_SIZE_LIMIT + 1` was rejected by clippy

src/parsec.rs
octol: I like having these `trace` statements, but I can remove them if people feel strongly about it

src/parsec.rs
jeanphilippeD: > Well it was added mostly as a response to previous review comments: "A struct for (parsec, size_counter) would probably allow you to avoid duplication here."

True, that comment made sense in the context of we want counter per parsec, that new comment is taking a step back, and so make that less relevant.

> I'm still of the opinion that keeping a counter per parsec seems the cleanest option. 

Why, what is the purpose of the old counters?

src/parsec.rs
jeanphilippeD: I put the comment at the wrong place, it was refering to `get_parsec_values` and the nut version

src/parsec.rs
jeanphilippeD: I'm suggesting 
`handle_n_response(PARSEC_SIZE_LIMIT + 1, ...);`
`handle_n_response(PARSEC_SIZE_LIMIT, ...);`

src/parsec.rs
jeanphilippeD: Completely missed that bit can you clarify that as well.
Do we really care about it? I mean if we do send it the the latest parsec we should get a failure because need_pruning would be false.

src/parsec.rs
jeanphilippeD: > Especially since it sounded like we might end up pruning frequently.
Why would that matter?

> Some of the code that uses this though was removed in an earlier revision though due to review comments
Yes, a contributing factor in me not questioning that design initially.

src/parsec.rs
octol: The `get_parsec_values` method returns all parsecs though, not the last one

src/parsec.rs
octol: Yes the assert afterwards would also fail (but not necessarily for that reason if something else is also broken). Why would we not check the return value for `handle_response`?

src/parsec.rs
octol: > Why, what is the purpose of the old counters?

I mean we add to all parsecs, so doesn't it make sense to keep track how much we add to all of them?

It seems to me that for example we could use that to keep adding to the older parsecs up until they are full at which we use the more strict response suggested earlier (e.g. not replying at all)

src/parsec.rs
jeanphilippeD: Yes, but all the uses follow it with `.last()` so you may as well fold that call within that function and clarify that we only care about the last one.

src/parsec.rs
jeanphilippeD: The reason, is that is probably a different unit test, i.e it is a handle_response test. As a result it is less clear what you are actually testing in this code. The AAA pattern is a guide that allow to clarify test, the assert should not be used to validate an assumption in the test, but what we want to validate.

So what you could do is add a small test that call handle_response on the different parsec and validate it is correct, and then just assume in this test it is.

src/parsec.rs
jeanphilippeD: > It seems to me that for example we could use that to keep adding to the older parsecs up until they are full at which we use the more strict response suggested earlier (e.g. not replying at all)

But why not just do that immediately when it become not the last, as oppose to have that intermediate state where for some node that have reach size and do something different than the other nodes? (solution in search of a problem?)

src/parsec.rs
octol: Well, the assert here isn't testing that `handle_response` returns false when sending to and old parsec, it's verifying for the purpose of the later asserts that we didn't send to the latest parsec

jeanphilippeD: If that is true, would not not be better done declaratively?
(i.e there is an implicit assumption here that last parsec means return true, and that is tested nowhere).

```
const INITIAL_PARSEC = 0;
const FIRST_PARSEC = 1;
const LAST_PARSEC = 2;

handle_n_parsec_responses(PARSEC_SIZE_LIMIT, FIRST_PARSEC, ... );
```

octol: I like having these named constants, so I'll add that. But having a general function like this still doesn't confirm that FIRST_PARSEC isn't handled as the last parsec

jeanphilippeD: No, but you have to make choices on what your test do validate otherwise very quickly become difficult to read.

You want to validate that assumption make it the other test I suggested (i.e test handle_parsec_response(FIRST_PARSEC ... ) is false and LAST_PARSEC is true.

What I am saying is you are testing multiple things in your test. That is why guildline like the AAA pattern where created, to help keep test clean.

Now it is up to you to decide.

octol: Even in a strict AAA pattern, checking that the steps in the ACT phase are doing what you think it's doing makes things easier I think. The earlier the failure happens the easier it's to spot what's wrong.

That said, following this discussion I've split the test into 2 to follow AAA, and added a third test to test requests too

src/parsec.rs
```diff
@@ -224,3 +284,160 @@ fn new_rng() -> Box<dyn Rng> {
...
+        // msg if it's against the latest parsec.
+        for _ in 0..msg_size_limit {
+            msg.handle(parsec_map, msg_version, pub_id, log_ident);
+        }
+
+        // Make sure we don't cross the prune limit
+        assert_eq!(parsec_map.needs_pruning(), false);
+    }
+
+    #[allow(unused)]
```
jeanphilippeD: Why allow unsused?

octol: Oops that one snuck through the cracks, thanks for spotting!

octol: Removed

src/parsec.rs
```diff
@@ -43,13 +77,15 @@ impl ParsecMap {
...
+        let size_counter = ParsecSizeCounter::default();
 
-        Self { map }
+        Self { map, size_counter }
     }
 
     pub fn init(&mut self, full_id: FullId, gen_pfx_info: &GenesisPfxInfo, log_ident: &LogIdent) {
         if let Entry::Vacant(entry) = self.map.entry(*gen_pfx_info.first_info.version()) {
             let _ = entry.insert(create(full_id, gen_pfx_info));
+            self.size_counter = ParsecSizeCounter::default();
```
jeanphilippeD: I'm not sure if any of your test valid this?

src/parsec.rs
```diff
@@ -15,25 +15,59 @@ use crate::{
...
+#[derive(Default, Debug, PartialEq, Eq)]
+struct ParsecSizeCounter(u64);
+
+impl ParsecSizeCounter {
+    fn increase_size(&mut self, size: u64) {
+        self.0 += size;
+    }
+
+    #[cfg(all(test, feature = "mock_parsec"))]
+    fn needs_pruning(&self) -> bool {
```
maqi: The feature guards reminds me that whether This PR shall contain the part of `returning` the counter status for the later on `vote for prune` ?
That is, the return of the `ParsecSiceCounter::increase_size()` and `Parsec::count_size()` shall incidicates a `overflow` and trigering an empty vote function to be called?
As the titile of the PR indicates this shall be part of the `detection` ?

src/parsec.rs
```diff
@@ -15,25 +15,59 @@ use crate::{
...
+pub use parsec::{ConsensusMode, Error, NetworkEvent, Observation, Proof, PublicId, SecretId};
 
 pub type Block = inner::Block<chain::NetworkEvent, id::PublicId>;
 pub type Parsec = inner::Parsec<chain::NetworkEvent, FullId>;
 pub type Request = inner::Request<chain::NetworkEvent, id::PublicId>;
 pub type Response = inner::Response<chain::NetworkEvent, id::PublicId>;
 
+// TODO: we'll set PARSEC_SIZE_LIMIT to 1 GB once it's used outside of mock_parsec
+//#[cfg(not(feature = "mock_parsec"))]
+//const PARSEC_SIZE_LIMIT: u64 = 1_000_000_000;
```
maqi: shall this comment out const be put back?

octol: Well it's not used so would have to be under a `allow(unused)` guard?

maqi: Ah, I think this relates to my comment at https://github.com/maidsafe/routing/pull/1786/files#r326137333
i.e. whether we shall put a triggered empty call

src/parsec.rs
```diff
@@ -90,6 +131,11 @@ impl ParsecMap {
         pub_id: id::PublicId,
         log_ident: &LogIdent,
     ) -> bool {
+        // Increase the size before fetching the parsec to satisfy the borrow checker
+        if self.last_version() == msg_version && self.map.contains_key(&msg_version) {
```
maqi: the comment and the if clause can be put into the `count_size` function?
there is duplication of the code.
I think you only need to pass the `msg_version` into `count_size` function ?

src/parsec.rs
octol: Do you mean something like
```
    fn increase_size(&mut self, size: u64) -> bool {
        self.0 += size;
        self.needs_pruninig()
    }
```
?

jeanphilippeD: I'm not seeing what that would be coupled, can look at the follow up PR to see how it is used.

src/parsec.rs
octol: Good point, I added a fixup commit that adds a check for this

src/parsec.rs
octol: Pushed a fixup that moves the guard inside count_size. Although I think strictly speaking it's not counting the size anymore...now

src/parsec.rs
octol: Well I'm of the opinion that if these needs to tweaked depending on how they're consumed then that should happen in that PR

jeanphilippeD: What I mean is, looking at the follow up PR, which exists, we can see that the existing API is fine?

jeanphilippeD: i.e no need to return a bool?

PR #1789, opened 2019-09-12
Vote to prune PARSEC gossip graph

### LINKED ISSUE #1782: Prune: New Parsec prune event and vote
New Parsec prune event with voting for it at the right time + test (maybe)

- [x] New Parsec prune event
- [x] Vote for it at the right time
- [x] Tests (maybe)

octol:
Resolves #1782 

- Add new `ParsecPrune` event
- Vote for pruning at the right time
- Unit and integration test

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -36,32 +36,36 @@ pub type Parsec = inner::Parsec<chain::NetworkEvent, FullId>;
 pub type Request = inner::Request<chain::NetworkEvent, id::PublicId>;
 pub type Response = inner::Response<chain::NetworkEvent, id::PublicId>;
 
-// TODO: we'll set PARSEC_SIZE_LIMIT to 1 GB once it's used outside of mock_parsec
-//#[cfg(not(feature = "mock_parsec"))]
-//const PARSEC_SIZE_LIMIT: u64 = 1_000_000_000;
+#[cfg(not(feature = "mock_parsec"))]
+const PARSEC_SIZE_LIMIT: u64 = 1_000_000_000;
 // Mock parsec request/responses are much smaller, so we need a lower limit.
-// TODO: once it's used outside of tests, this should be changed to cfg(feature = "mock_parsec")
-#[cfg(all(test, feature = "mock_parsec"))]
+#[cfg(feature = "mock_parsec")]
 const PARSEC_SIZE_LIMIT: u64 = 100;
 
 // Keep track of size in case we need to prune.
 #[derive(Default, Debug, PartialEq, Eq)]
-struct ParsecSizeCounter(u64);
+struct ParsecSizeCounter {
+    size_counter: u64,
+    pruning_voted_for: bool,
+}
 
 impl ParsecSizeCounter {
     fn increase_size(&mut self, size: u64) {
-        self.0 += size;
+        self.size_counter += size;
     }
 
-    #[cfg(all(test, feature = "mock_parsec"))]
     fn needs_pruning(&self) -> bool {
-        self.0 > PARSEC_SIZE_LIMIT
+        self.size_counter > PARSEC_SIZE_LIMIT && !self.pruning_voted_for
+    }
+
+    fn set_pruning_voted_for(&mut self) {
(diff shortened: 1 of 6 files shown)


src/states/common/approved.rs
```diff
@@ -248,6 +248,12 @@ pub trait Approved: Relocated {
                 NetworkEvent::PurgeCandidate(old_public_id) => {
                     self.handle_purge_candidate_event(old_public_id)?
                 }
+                NetworkEvent::ParsecPrune => {
+                    info!(
```
jeanphilippeD: Cannot really test logs:-), as a step through can add a counter and a getter (did that for message Ack).
so you can write the test case that exercise it even if prunning does not actually happen on consensus,

src/parsec.rs
```diff
@@ -257,14 +257,23 @@ impl ParsecMap {
...
             .unwrap_or(false)
     }
+
+    pub fn pruning_voted_for(&mut self) {
+        let _ = self
+            .map
+            .values_mut()
+            .last()
+            .map(|ParsecAndCounter { size_counter, .. }| {
+                *size_counter = ParsecSizeCounter::default()
```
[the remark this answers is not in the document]
octol: That's correct. My thinking was that it could be useful to vote again if for some strange reason the first round didn't result in a prune being done

src/parsec.rs
```diff
@@ -416,4 +425,41 @@ mod tests {
...
+
+        // Then handle responses with parsec 2 which should trigger pruning
+        for _ in 0..PARSEC_SIZE_LIMIT {
+            assert!(parsec_map.handle_response(2, response.clone(), *pub_id, &log_ident));
+        }
+        assert_eq!(parsec_map.needs_pruning(), false);
+        assert!(parsec_map.handle_response(2, response.clone(), *pub_id, &log_ident));
+        assert_eq!(parsec_map.needs_pruning(), true);
+
+        parsec_map.pruning_voted_for();
```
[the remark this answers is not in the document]
octol: Yes you're right there is probably a bit too much overlap, I'll remove some stuff

src/parsec.rs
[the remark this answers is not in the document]
jeanphilippeD: duplicate vote will be ignored by parsec, so I'm not sure it solve the issue?

src/parsec.rs
```diff
@@ -413,7 +420,7 @@ mod tests {
         msg: T,
         parsec_age: u64,
         prune_needed: bool,
-    ) {
+    ) -> ParsecMap {
```
maqi: The purpose of the function is just to check wheth prune is required, why we need to return a new parsec_map instance?
I thought at this stage, we only need to cast the vote?

octol: It's to test that that `needs_pruning` reports false after marking the vote as cast

maqi: I am wondering if this is only for the testing purpose, then we shall not make it as a production code?
Maybe we shall find another way to test, which is not so intrusive to the production code?

octol: Yep I agree. The above function is not production code though, it's a helper function for the tests and inside `mod tests`

src/states/elder/mod.rs
```diff
@@ -1064,6 +1065,11 @@ impl Elder {
         if let Some(msg) = self.parsec_map.create_gossip(version, &gossip_target) {
             self.send_direct_message(&gossip_target, msg);
         }
+
+        if self.parsec_map_mut().needs_pruning() {
```
maqi: not a big issue, however do we need `mut` here?
maybe `self.parsec_map.needs_pruning` will be enough?

octol: I don't there is a `parsec_map()` method

maqi: yeah, but here you have the access to the parsec_map so don't `have to` use that method?

octol: Ah yes I see what you mean, yep that's a good point. We do mutate it 2 lines below but I'll update

PR #1794, opened 2019-09-13
Fix premature consensus in mock parsec

madadam:
The bug happened in a situation like this:

Say there are 8 nodes in the section and the genesis group size is 4. A new node X joins. Another node Y is in the process of being voted in, but so far has only 4 votes. X receives a gossip which triggers the consensus calculation. In mock parsec, this means reaching into the shared state and picking the first unconsensused observation. That happens to be Add(Y). There are 4 votes for it. X only knows about 4 valid voters at this point (the genesis group), so it thinks there are enough votes and marks the observation as consensused. Thus Add(Y) is consensused prematurelly.

The fix is to make sure the new node catches up to all the previously consensused blocks before it itself performs the consensus calculation.

### DIFF
--- a/src/mock/parsec/mod.rs
+++ b/src/mock/parsec/mod.rs
@@ -192,6 +192,10 @@ where
     }
 
     fn compute_consensus(&mut self, state: &mut SectionState<T, S::PublicId>) {
+        // Call `update_blocks` once, to allow this node to catch up to the previously consensused
+        // blocks.
+        let _ = self.update_blocks(state);
+
         loop {
             state.compute_consensus(&self.peer_list, self.consensus_mode);
 
--- a/src/mock/parsec/tests.rs
+++ b/src/mock/parsec/tests.rs
@@ -27,9 +27,7 @@ fn smoke() {
     let alice_id = PeerId(0);
     let bob_id = PeerId(1);
 
-    let mut genesis_group = BTreeSet::new();
-    let _ = genesis_group.insert(alice_id);
-    let _ = genesis_group.insert(bob_id);
+    let genesis_group: BTreeSet<_> = vec![alice_id, bob_id].into_iter().collect();
 
     let mut alice = from_genesis(alice_id, &genesis_group, ConsensusMode::Supermajority);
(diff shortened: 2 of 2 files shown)


src/mock/parsec/mod.rs
```diff
@@ -192,6 +192,10 @@ where
     }
 
     fn compute_consensus(&mut self, state: &mut SectionState<T, S::PublicId>) {
+        // Call `update_blocks` once, to allow this node to catch up to the previously consensused
+        // blocks.
+        let _ = self.update_blocks(state);
```
jeanphilippeD: Worth adding a test?

madadam: There is already a failing test that is fixed by this change. But maybe better to have one focused specifically on this? I'll let you decide.

jeanphilippeD: Yes, that was a very difficult to reproduce random test, so probably best to have a unit test for it.

madadam: Ok.

src/mock/parsec/tests.rs
```diff
@@ -226,6 +226,75 @@ fn reevaluate_previously_insufficient_votes() {
     assert_consensus_on_remove(&mut nodes, dave);
 }
 
+#[test]
+fn newly_joined_node_does_not_cause_premature_consensus() {
+    init_mock();
+
+    // Create the genesis group.
+    let mut genesis_group = BTreeSet::new();
```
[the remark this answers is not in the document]
madadam: Sure, I can do that. Ideally I would use this: https://docs.rs/maplit/1.0.2/maplit/, but not sure it's worth bringing another dependency for it (although it would be only dev-dependency).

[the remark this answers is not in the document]
jeanphilippeD: Maybe not yet bring the dependency, we can always do it later, should be easier to refactor from the suggested pattern. (in parsec we also had a btreeset! macro I think, but lets keep it simple for now.)

PR #1796, opened 2019-09-16
Fix too many open files in tests

### LINKED ISSUE #1795: Ensure no Rng failure in test for mac os.
See error in log file:
[log_too_many_files.txt](https://github.com/maidsafe/routing/files/3610222/log_too_many_files.txt)

```
thread 'mock_network::multiple_joining_nodes' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 24, kind: Other, message: "Too many open files" }', src/libcore/result.rs:999:5

stack backtrace:
   6: core::result::unwrap_failed
   7: routing::parsec::create
   8: routing::parsec::ParsecMap::init
   9: routing::states::elder::Elder::finalise_prefix_change
```

Address by:
- panic with good message: Cannot proceed further if Rng is not valid.
- Avoid using OsRng in `mock` tests as it is too taxing (many nodes at once).

madadam:
Closes #1795

### DIFF
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -17,6 +17,7 @@ use crate::{
 use log::LogLevel;
 #[cfg(not(feature = "mock_parsec"))]
 use parsec as inner;
+use rand::Rng;
 use std::collections::{btree_map::Entry, BTreeMap};
 
 #[cfg(feature = "mock_parsec")]
@@ -172,6 +173,8 @@ impl ParsecMap {
 
 /// Create Parsec instance.
 fn create(full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec {
+    let rng = new_rng();
+
     if gen_pfx_info
         .first_info
         .members()
@@ -184,7 +187,7 @@ fn create(full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec {
             &gen_pfx_info.first_info.members(),
             gen_pfx_info.first_state_serialized.clone(),
             ConsensusMode::Single,
-            Box::new(rand::os::OsRng::new().unwrap()),
+            rng,
         )
     } else {
         Parsec::from_existing(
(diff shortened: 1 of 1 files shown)


src/parsec.rs
```diff
@@ -172,6 +172,19 @@ impl ParsecMap {
...
 fn create(full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec {
+    use maidsafe_utilities::SeededRng;
+    use rand::Rng;
+    use unwrap::unwrap;
+
+    // Note: in production, use `OsRng` for maximum cryptographic security. In tests, however, this
+    // could cause "Too many open files" error because some tests create lot of nodes and each node
+    // has its own rng. To prevent that we use a weaker rng in tests.
+    let rng: Box<dyn Rng> = if cfg!(feature = "mock_base") {
+        Box::new(SeededRng::new())
```
jeanphilippeD: I'm not sure this is quite right, as this will create the same generator for all parsec instances.
Can we use another of the rand Rng?

src/parsec.rs
```diff
@@ -172,6 +172,19 @@ impl ParsecMap {
...
 /// Create Parsec instance.
 fn create(full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec {
+    use maidsafe_utilities::SeededRng;
+    use rand::Rng;
+    use unwrap::unwrap;
+
+    // Note: in production, use `OsRng` for maximum cryptographic security. In tests, however, this
+    // could cause "Too many open files" error because some tests create lot of nodes and each node
+    // has its own rng. To prevent that we use a weaker rng in tests.
+    let rng: Box<dyn Rng> = if cfg!(feature = "mock_base") {
```
jeanphilippeD: Maybe worth putting in its own sub function since there is a need to document it?

src/parsec.rs
```diff
@@ -172,6 +172,19 @@ impl ParsecMap {
...
+        Box::new(SeededRng::new())
+    } else {
+        Box::new(unwrap!(rand::os::OsRng::new()))
```
jeanphilippeD: Maybe could avoid using unwrap! here, and instead handle the error with a panic with dedicated error message?

src/parsec.rs
```diff
@@ -194,7 +197,30 @@ fn create(full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec {
...
+//
+// In test, use a weaker rng to prevent creating too many open file handles (as each `OsRng`
+// instance might internally contain a handle to /dev/random or similar). This avoid a bug with too
+// many open files which sometimes happens because some test create lot of nodes and each node has its own rng.
+fn new_rng() -> Box<dyn Rng> {
+    use maidsafe_utilities::SeededRng;
+    use rand::os::OsRng;
+
+    if cfg!(feature = "mock_base") {
+        Box::new(SeededRng::thread_rng().new_rng())
```
jeanphilippeD: why the call to `new_rng()` is it not what `thread_rng()` already does?

src/parsec.rs
```diff
@@ -194,7 +197,30 @@ fn create(full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec {
...
     }
 }
+
+// Create new Rng instance.
+//
+// In production, use `OsRng` for maximum cryptographic security.
+//
+// In test, use a weaker rng to prevent creating too many open file handles (as each `OsRng`
+// instance might internally contain a handle to /dev/random or similar). This avoid a bug with too
+// many open files which sometimes happens because some test create lot of nodes and each node has its own rng.
```
maqi: what's your setting for the number of chars each line?
I think we have the perference of 100 and seems the last line exceeds it?

src/parsec.rs
```diff
@@ -194,7 +197,30 @@ fn create(full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec {
...
+fn new_rng() -> Box<dyn Rng> {
+    use maidsafe_utilities::SeededRng;
+    use rand::os::OsRng;
```
maqi: shall these above two imports be put under the correspondent cfg guard as well?

PR #1798, opened 2019-09-16
feat/routing: prune parsec as requested

maqi:
This PR contains the detailed operation when a ParsecPrune event reached consensus, and the test to confirm it.
The operation utilize the current split approach, so the most part of existing `finalise_prefix_change` handling can be retained.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -60,9 +60,6 @@ pub struct Chain {
     chain_accumulator: ChainAccumulator,
     /// Pending events whose handling has been deferred due to an ongoing split or merge.
     event_cache: BTreeSet<NetworkEvent>,
-    /// Temporary. Counting the accumulated prune events. Only used in tests until tests that
-    /// actually tests pruning is in place.
-    parsec_prune_accumulated: usize,
     /// Marker indicating we are processing churn event
     churn_in_progress: bool,
 }
@@ -116,7 +113,6 @@ impl Chain {
             is_elder,
             chain_accumulator: Default::default(),
             event_cache: Default::default(),
-            parsec_prune_accumulated: 0,
             churn_in_progress: false,
         }
     }
@@ -275,14 +271,6 @@ impl Chain {
             AccumulatingEvent::NeighbourMerge(digest) => {
                 // TODO: Check that the section is known and not already merged.
                 let _ = self.state.merging.insert(digest);
-            }
-            AccumulatingEvent::ParsecPrune => {
-                info!(
-                    "{} Handling accumulated {:?} not yet implemented, ignoring.",
-                    self, event
-                );
-                // TODO: remove once we have real integration tests of `ParsecPrune` accumulating.
-                self.parsec_prune_accumulated += 1;
             }
             AccumulatingEvent::Relocate(_) => {
                 self.churn_in_progress = false;
@@ -1530,12 +1518,6 @@ impl Chain {
     /// Returns their_knowledge
     pub fn get_their_knowledge(&self) -> &BTreeMap<Prefix<XorName>, u64> {
         &self.state.get_their_knowledge()
-    }
-
-    /// Get the number of accumulated `ParsecPrune` events. This is only used until we have
-    /// implemented acting on the accumulated events.
-    pub fn parsec_prune_accumulated(&self) -> usize {
-        self.parsec_prune_accumulated
     }
 
     /// Return a minimum length prefix, favouring our prefix if it is one of the shortest.
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -293,6 +293,7 @@ impl Chain {
             | AccumulatingEvent::Offline(_)
             | AccumulatingEvent::StartDkg(_)
             | AccumulatingEvent::User(_)
+            | AccumulatingEvent::ParsecPrune
             | AccumulatingEvent::SendAckMessage(_) => (),
         }
 
@@ -569,7 +570,10 @@ impl Chain {
     }
 
     /// Gets the data needed to initialise a new Parsec instance
-    pub fn prepare_parsec_reset(&mut self) -> Result<ParsecResetData, RoutingError> {
+    pub fn prepare_parsec_reset(
+        &mut self,
+        parsec_version: u64,
+    ) -> Result<ParsecResetData, RoutingError> {
         let remaining = self.chain_accumulator.reset_accumulator(&self.our_id);
         let event_cache = mem::replace(&mut self.event_cache, Default::default());
         let merges = mem::replace(&mut self.state.merging, Default::default())
@@ -584,6 +588,7 @@ impl Chain {
                 first_state_serialized: self.get_genesis_related_info()?,
                 first_ages: self.get_age_counters(),
                 latest_info: Default::default(),
+                parsec_version,
             },
             cached_events: remaining
                 .cached_events
@@ -597,7 +602,10 @@ impl Chain {
 
     /// Finalises a split or merge - creates a `GenesisPfxInfo` for the new graph and returns the
     /// cached and currently accumulated events.
-    pub fn finalise_prefix_change(&mut self) -> Result<ParsecResetData, RoutingError> {
+    pub fn finalise_prefix_change(
+        &mut self,
+        parsec_version: u64,
+    ) -> Result<ParsecResetData, RoutingError> {
         // Clear any relocation overrides
         #[cfg(feature = "mock_base")]
         {
@@ -612,7 +620,7 @@ impl Chain {
         info!("{} - finalise_prefix_change: {:?}", self, self.our_prefix());
         trace!("{} - finalise_prefix_change state: {:?}", self, self.state);
 
-        self.prepare_parsec_reset()
+        self.prepare_parsec_reset(parsec_version)
     }
 
     /// Returns our public ID
@@ -819,6 +827,11 @@ impl Chain {
     pub fn check_vote_status(&mut self) -> BTreeSet<PublicId> {
         let members = self.our_info().member_ids();
         self.chain_accumulator.check_vote_status(members)
+    }
+
+    /// Returns whether a churning is in progress.
+    pub fn is_churn_in_progress(&self) -> bool {
+        self.churn_in_progress
     }
 
     /// Returns `true` if the given `NetworkEvent` is already accumulated and can be skipped.
@@ -1680,6 +1693,7 @@ mod tests {
             first_state_serialized: Vec::new(),
             first_ages,
             latest_info: Default::default(),
+            parsec_version: 0,
         };
 
         let mut chain = Chain::new(
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -49,10 +49,10 @@ const MAX_PARSECS: usize = 10;
 const PARSEC_SIZE_LIMIT: u64 = 1_000_000_000;
 // Limit in integration tests
 #[cfg(all(feature = "mock_base", not(feature = "mock_parsec")))]
-const PARSEC_SIZE_LIMIT: u64 = 10_000_000;
+const PARSEC_SIZE_LIMIT: u64 = 20_000_000;
 // Limit for integration tests with mock-parsec
 #[cfg(feature = "mock_parsec")]
-const PARSEC_SIZE_LIMIT: u64 = 100;
+const PARSEC_SIZE_LIMIT: u64 = 500;
 
 // Keep track of size in case we need to prune.
 #[derive(Default, Debug, PartialEq, Eq)]
@@ -228,17 +228,6 @@ impl ParsecMap {
             .flatten()
     }
 
-    #[cfg(feature = "mock_base")]
-    pub fn has_unpolled_observations(&self) -> bool {
-        let parsec = if let Some(parsec) = self.map.values().last() {
-            parsec
-        } else {
-            return false;
-        };
-
-        parsec.has_unpolled_observations()
-    }
-
     pub fn needs_pruning(&self) -> bool {
         self.size_counter.needs_pruning()
     }
@@ -284,6 +273,23 @@ impl ParsecMap {
             .take(MAX_PARSECS)
             .rev()
             .collect();
+    }
+}
+
+#[cfg(feature = "mock_base")]
+impl ParsecMap {
+    pub fn has_unpolled_observations(&self) -> bool {
+        let parsec = if let Some(parsec) = self.map.values().last() {
+            parsec
+        } else {
+            return false;
+        };
+
+        parsec.has_unpolled_observations()
+    }
+
+    pub fn get_size(&self) -> u64 {
+        self.size_counter.size_counter
     }
 }
 
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -7,6 +7,8 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 #[cfg(feature = "mock_parsec")]
+use crate::crypto;
+#[cfg(feature = "mock_parsec")]
 use crate::mock::parsec as inner;
 use crate::{
     chain::{self, GenesisPfxInfo},
@@ -23,6 +25,8 @@ use std::{
     collections::{btree_map::Entry, BTreeMap},
     fmt, mem,
 };
+#[cfg(feature = "mock_parsec")]
+use unwrap::unwrap;
 
 #[cfg(feature = "mock_parsec")]
 pub use crate::mock::parsec::{
@@ -45,7 +49,7 @@ const MAX_PARSECS: usize = 10;
 const PARSEC_SIZE_LIMIT: u64 = 1_000_000_000;
 // Limit in integration tests
 #[cfg(all(feature = "mock_base", not(feature = "mock_parsec")))]
-const PARSEC_SIZE_LIMIT: u64 = 200_000;
+const PARSEC_SIZE_LIMIT: u64 = 10_000_000;
 // Limit for integration tests with mock-parsec
 #[cfg(feature = "mock_parsec")]
 const PARSEC_SIZE_LIMIT: u64 = 100;
@@ -86,7 +90,7 @@ impl ParsecMap {
     pub fn new(rng: &mut MainRng, full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Self {
         let mut map = BTreeMap::new();
         let _ = map.insert(
-            *gen_pfx_info.first_info.version(),
+            gen_pfx_info.parsec_version,
             create(rng, full_id, gen_pfx_info),
         );
         let size_counter = ParsecSizeCounter::default();
@@ -262,7 +266,7 @@ impl ParsecMap {
         gen_pfx_info: &GenesisPfxInfo,
         log_ident: &LogIdent,
     ) {
-        if let Entry::Vacant(entry) = self.map.entry(*gen_pfx_info.first_info.version()) {
+        if let Entry::Vacant(entry) = self.map.entry(gen_pfx_info.parsec_version) {
             let _ = entry.insert(create(rng, full_id, gen_pfx_info));
             self.size_counter = ParsecSizeCounter::default();
             info!(
@@ -285,10 +289,16 @@ impl ParsecMap {
 
 /// Create Parsec instance.
 fn create(rng: &mut MainRng, full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> Parsec {
+    #[cfg(feature = "mock_parsec")]
+    let hash = {
+        let fields = (gen_pfx_info.first_info.hash(), gen_pfx_info.parsec_version);
+        crypto::sha3_256(&unwrap!(serialisation::serialise(&fields)))
+    };
+
     if gen_pfx_info.first_info.is_member(full_id.public_id()) {
         Parsec::from_genesis(
             #[cfg(feature = "mock_parsec")]
-            *gen_pfx_info.first_info.hash(),
+            hash,
             full_id,
             &gen_pfx_info.first_info.member_ids().copied().collect(),
             gen_pfx_info.first_state_serialized.clone(),
@@ -298,7 +308,7 @@ fn create(rng: &mut MainRng, full_id: FullId, gen_pfx_info: &GenesisPfxInfo) -> 
     } else {
         Parsec::from_existing(
             #[cfg(feature = "mock_parsec")]
-            *gen_pfx_info.first_info.hash(),
+            hash,
             full_id,
             &gen_pfx_info.first_info.member_ids().copied().collect(),
             &gen_pfx_info.latest_info.member_ids().copied().collect(),
@@ -367,6 +377,7 @@ mod tests {
             first_state_serialized: Vec::new(),
             first_ages,
             latest_info: EldersInfo::default(),
+            parsec_version: version,
         }
     }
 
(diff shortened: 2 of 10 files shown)


src/states/elder/mod.rs
```diff
@@ -2048,6 +2055,10 @@ impl Approved for Elder {
         } else if old_pfx.is_extension_of(sec_info.prefix()) {
             self.finalise_prefix_change()?;
             self.send_event(Event::SectionMerged(*sec_info.prefix()), outbox);
+        } else if old_pfx == *sec_info.prefix() {
+            // A pruning happens.
```
jeanphilippeD: This does not seem accurate, the prefix will stay the same in most churn events / SectionInfo event since adding/removing a node will generally not lead to a split.

Are the test passing?

src/chain/chain.rs
```diff
@@ -432,7 +444,7 @@ impl Chain {
         self.state.our_prefix()
     }
 
-    /// Returns whether our section is in the process of changing (splitting or merging).
+    /// Returns whether our section is in the process of changing (splitting, merging or pruning).
     pub fn prefix_change(&self) -> PrefixChange {
```
octol: Does this method name now needs updating to reflect the new responsibility?

src/states/elder/mod.rs
```diff
@@ -2030,6 +2036,12 @@ impl Approved for Elder {
         Ok(())
     }
 
+    fn handle_prune(&mut self) -> Result<(), RoutingError> {
+        let _sec_info = NetworkEvent::SectionInfo(self.chain.prune()?);
+        self.finalise_prefix_change()?;
```
octol: This calls `Chain::check_and_clean_neighbour_infos` I think, do we want that?

src/chain/chain.rs
maqi: This follows the enum name. unless we want change them all ?

src/states/elder/mod.rs
maqi: I will go to separate it out.

src/chain/chain.rs
octol: Yeah that's a good question. I think some of these names needs updating at least. But maybe we can do that once more pieces are in place?

src/chain/chain.rs
```diff
@@ -307,6 +308,17 @@ impl Chain {
         Ok(vec![self.state.new_info.clone()])
     }
 
+    /// Starts pruning as requested. Returns a new `SectionInfo`.
+    pub fn prune(&mut self) -> Result<SectionInfo, RoutingError> {
+        if self.state.change != PrefixChange::None {
+            log_or_panic!(LogLevel::Warn, "Tring to prune during pfx change.",);
+        }
+
+        self.state.change = PrefixChange::Pruning;
```
jeanphilippeD: What is the purpose of  PrefixChange::Pruning in the context of this PR? are we not setting it to immediatly reset it?

src/states/elder/mod.rs
```diff
@@ -2030,6 +2044,12 @@ impl Approved for Elder {
         Ok(())
     }
 
+    fn handle_prune(&mut self) -> Result<(), RoutingError> {
+        let _sec_info = NetworkEvent::SectionInfo(self.chain.prune()?);
```
jeanphilippeD: Why are we creating a NetworkEvent we do not use?

src/chain/chain.rs
```diff
@@ -307,6 +308,17 @@ impl Chain {
...
+
+        Ok(self.state.new_info.clone())
```
jeanphilippeD: Why is prune returning a sectionInfo?

src/chain/chain.rs
maqi: Ah, this is a leftover from my previous approach.
I shall remove it in the context of `immediately reset`.

src/states/elder/mod.rs
```diff
@@ -368,6 +368,14 @@ impl Elder {
         }
     }
 
+    fn finalise_prune(&mut self) -> Result<(), RoutingError> {
+        let _ = self.chain.finalise_prune();
+        self.chain.reset_candidate();
```
jeanphilippeD: Why are we reseting candidate? does pruning impact how we accept candidates?

src/states/elder/mod.rs
```diff
@@ -368,6 +368,14 @@ impl Elder {
...
+        self.peer_mgr.reset_candidate();
+        self.init_parsec(); // We don't reset the chain on pruning.
```
jeanphilippeD: Prunning does not seem to carry over any events the way prefix change does, how does that work?

src/states/elder/mod.rs
```diff
@@ -1812,6 +1821,11 @@ impl Elder {
     ) {
         self.send_message_to_targets(dst_targets, dg_size, message)
     }
+
+    pub fn vote_for_prune(&mut self) {
```
jeanphilippeD: Use `vote_for_event` instead of this new funcion?

tests/mock_network/mod.rs
```diff
@@ -464,3 +464,22 @@ fn check_section_info_ack() {
...
+            .inner
+            .node_state_mut()
+            .map(|state| state.vote_for_prune());
+    }
+    poll_and_resend(&mut nodes, &mut []);
+
+    nodes = create_connected_nodes_until_split(&network, vec![1, 1], true);
+    poll_and_resend(&mut nodes, &mut []);
+
+    verify_invariant_for_all_nodes(&network, &mut nodes);
```
jeanphilippeD: That is pretty good, Ideally we would also test that prunning catually happen (i.e it seem this test would pass even if prunning does nothing.

src/chain/chain.rs
maqi: Yeah, an empty return shall be enough, I will correct it.

src/states/elder/mod.rs
maqi: You are right, it shall not reset candidate. I will correct it.

src/states/elder/mod.rs
maqi: It happens immediately when got the consensused `ParsecPrune` event.
So there is no `cached` events to handle as the prefix change.
Unless, the comment (code, i.e. set the state to pruning) at https://github.com/maidsafe/routing/pull/1798/files#r326167029 shall be carried out immediately when `casting the ParsecPrune` vote. However, this will be problematic I think?

jeanphilippeD: Things are still in the chain_accumulator, or even in parsec not yet consensused

src/states/elder/mod.rs
maqi: using `vote_for_event` means need to expose extra NetworkEvent and LogIndent, which I think don't worth it?

src/states/elder/mod.rs
maqi: Yeah, I shall remove it. A left over from the previous approach

src/states/elder/mod.rs
jeanphilippeD: No make sense, can use it from within vote_for_prune though.

tests/mock_network/mod.rs
maqi: agreed, I will expand the test

src/states/elder/mod.rs
maqi: will have a double check

src/chain/chain.rs
```diff
@@ -481,6 +470,21 @@ impl Chain {
             .should_vote_for_merge(self.elder_size(), self.neighbour_infos())
     }
 
+    /// Increase version by one due to the parsec pruning.
```
jeanphilippeD: Sorry, I'm not sure why we increase the ElderInfo version, this should be an internal thing only, shouldn't it?

src/states/elder/mod.rs
```diff
@@ -381,18 +381,19 @@ impl Elder {
...
-
         let ParsecResetData {
             gen_pfx_info,
             mut cached_events,
             completed_events,
         } = reset_data;
+
         self.gen_pfx_info = gen_pfx_info;
+
+        let drained_obs: Vec<_> = self
```
jeanphilippeD: Any reason for this change?

src/states/elder/mod.rs
```diff
@@ -664,6 +665,7 @@ impl Elder {
 
         let src = Authority::PrefixSection(*trimmed_info.first_info.prefix());
         let content = MessageContent::NodeApproval(trimmed_info);
+
```
jeanphilippeD: Why the change?

tests/mock_network/mod.rs
```diff
@@ -645,3 +607,50 @@ fn node_pause_and_resume() {
     poll_and_resend(&mut nodes);
     verify_invariant_for_all_nodes(&network, &mut nodes);
 }
+
+#[test]
+fn carry_out_parsec_pruning() {
```
jeanphilippeD: would it make sense to put this test in the same position as vote_prune was since it replaces it? (for history).

tests/mock_network/mod.rs
```diff
@@ -645,3 +607,50 @@ fn node_pause_and_resume() {
...
+    let mut nodes = create_connected_nodes(&network, init_network_size);
+    poll_and_resend(&mut nodes);
+
+    let mut parsec_versions = BTreeMap::new();
+
+    for node in nodes.iter_mut() {
+        let id = node.name();
+        let _ = node.inner.elder_state_mut().map(|state| {
+            let _ = parsec_versions.insert(id, state.parsec_last_version());
+            state.vote_for_prune()
```
jeanphilippeD: Should that be how this test happen, that we vote_prune explicitly?
How does this test the mechanism that trigger the vote_prune?

src/chain/chain.rs
```diff
@@ -481,6 +470,21 @@ impl Chain {
...
+    /// Should not be called while a pfx change is in progress.
```
jeanphilippeD: Probably should not be called anytime churn is in progress since we are already reseting Parsec graph.
Also how is that prevented, do we filter prune votes somewhere?

maqi: > do we filter prune votes somewhere

when PrefixChange is not None, there will be no vote for ParsecPrune

> Probably should not be called anytime churn is in progress since we are already reseting Parsec graph

yeah, good idea. will think about it.

src/chain/chain.rs
maqi: The parsec_map creates parsec instance based on version which is deduced from the ElderInfo version.
If we don't create a new ElderInfo (with version increased by one here), there is no new parsec instance will got created.

src/states/elder/mod.rs
maqi: moved position when resolve conflicts due to rebase. Will change it back.

src/states/elder/mod.rs
maqi: same as above, changed when resolve conflicts due to rebase. Will change it back.

tests/mock_network/mod.rs
maqi: ok. will move the test to that position.

tests/mock_network/mod.rs
maqi: I think with the introduction of the reset parsec during churn, here we can only dispatch a bunch of user-data to bump up the graph size?

jeanphilippeD: Yes, or let it gossip.

src/chain/chain.rs
jeanphilippeD: Yes, it seem to be the case.
This is a short cut that is no longer reasonable in the context of Parsec Prune, that value should probably be in GenesisPfxInfo, and incremented independently of SectionInfo.

(We do not want to notify our neighbour that things have changed just because we locally are gossiping to much data. (It would make us gossip even more data which is not ideal in a prune scenario).

maqi: I was just want to limit the affects to the existing structs & approaches.
I agree with the concern of notify neighbours.
However, without a version increased, the parsec_map has to `replace` the instance, which could cause other issues? (as there are many other stuff depends on `version` I think? )

tests/mock_network/mod.rs
```diff
@@ -582,42 +582,62 @@ fn check_section_info_ack() {
...
+            .map(|state| parsec_versions.insert(id, state.parsec_last_version()));
+    }
+
+    // Keeps polling (which causes parsec gossiping) till trigger a pruning.
+    let max_gossips = 6_000;
+    for i in 0..max_gossips {
         poll_and_resend(&mut nodes);
+
+        let mut version_increased = 0;
+        for node in nodes.iter() {
```
jeanphilippeD: probably easier as:
```
let parsec_versions = |nodes: &TestNodes| {
  nodes.iter()
    .map(|node| unwrap!(node.inner.elder_state()).parsec_last_version())
    .collect::Vec<_>()
};

let initial_parsec_versions = parsec_versions(nodes);
for ... {

    let new_parsec_versions = parsec_versions(nodes);
    if initial_parsec_versions.zip(new_parsec_versions).all(|vi, vn)| vi < vn) {
        break;
    }
}
```

src/states/elder/mod.rs
```diff
@@ -664,6 +735,7 @@ impl Elder {
             first_state_serialized: self.gen_pfx_info.first_state_serialized.clone(),
             first_ages: self.gen_pfx_info.first_ages.clone(),
             latest_info: self.chain.our_info().clone(),
+            parsec_version: *self.gen_pfx_info.first_info.version(),
```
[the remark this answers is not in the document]
maqi: Yeah, I agree it shall use `self.gen_pfx_info.parsec_info`
There is pruning happens in tests. 
I guess the soak tests passes because this is `starting version` of the parsec instances, and every nodes can having independ version-indexing.

src/states/elder/mod.rs
```diff
@@ -383,6 +386,73 @@ impl Elder {
...
+            .cloned()
+            .collect();
+
+        let ParsecResetData {
+            mut gen_pfx_info,
+            mut cached_events,
+            completed_events,
+        } = reset_data;
+
+        gen_pfx_info.increase_parsec_version();
```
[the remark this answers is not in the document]
maqi: the reset_data from `chain::prepare_parsec_reset` doesn't increase the parsec_version.
maybe it should. I will have a check

src/states/elder/mod.rs
```diff
@@ -919,7 +925,7 @@ impl Elder {
     }
 
     fn maintain_parsec(&mut self) {
-        if self.parsec_map.needs_pruning() {
+        if self.parsec_map.needs_pruning() && !self.chain.is_churn_in_progress() {
```
jeanphilippeD: This is probably not where we should check for `!self.chain.is_churn_in_progress()` but probably in `handle_prune`.
Additionally, we need to make sure that ParsecPrune event are not carried over when a new parsec is created.

tests/mock_network/mod.rs
```diff
@@ -582,42 +582,53 @@ fn check_section_info_ack() {
...
+        let new_parsec_versions = parsec_versions(&nodes);
+        if initial_parsec_versions
+            .iter()
+            .zip(new_parsec_versions.iter())
+            .all(|(vi, vn)| vi < vn)
+        {
+            break;
+        }
+
+        if i == max_gossips - 1 {
```
jeanphilippeD: probably best to leave that as an assertion at the end.

let expected = initial_parsec_versions.iter().map(|v| v + 1).collect_vec();
let actual = parsec_versions(&nodes);
assert(expected, actual);

tests/mock_network/mod.rs
```diff
@@ -582,42 +582,53 @@ fn check_section_info_ack() {
...
+            panic!("Not all nodes carried out pruning");
+        }
     }
 
-    assert!(nodes
-        .iter()
-        .all(|node| unwrap!(node.inner.parsec_prune_accumulated()) > 0));
+    nodes = create_connected_nodes_until_split(&network, vec![1, 1]);
```
jeanphilippeD: Is that reasonable? can you call create_connected_nodes_until_split when a network already has node, and why do you do that, we seem to ignore all nodes we had before.

I guess you only wanted verify_invariant_for_all_nodes?

src/states/elder/mod.rs
maqi: `handle_prune` is kind of a local state, which means once received consensused ParsecPrune, different node may carry out or not to carry out the pruning, depends on its churn_in_progress status.
Meanwhile, the `voting for ParsecPrune` is a shared state, and once a node received the consensused vote. The action shall be carried out.

> ParsecPrune event are not carried over when a new parsec is created
`ParsecPrune` will be filterred out when carry out revote, within reset_parsec_with_data

tests/mock_network/mod.rs
maqi: Ah, my fault. it is `add_connected_nodes_until_split` shall be called, not `create_connected_nodes_until_split`

src/states/elder/mod.rs
jeanphilippeD: Sorry I do not understand `handle_prune is kind of a local...` and `Meanwhile, the voting for ParsecPrune is a sh...` sentences.

maqi: The current approach may make some nodes vote for ParsecPrune, some not. However, once the ParsecPrune got consensused. They all will carry out the pruning.

If we move the check of `self.chain.is_churn_in_progress()` into `handle_churn`, some nodes may decide to carry out the pruning, some nodes may decide not.

jeanphilippeD: I do not think so, as is_rn_in_progress() should be shared state so all should do the same.
But mostly, if you all voted for prune, and now churn is happening, and now Prune get consensus, we should not prune. so it would seem incorrect to not check in handle_prune whether we should carry out the prune or not (we do check for is split in there already).

maqi: right.
then shall we do a double check? that is:
keep the check here, and in `handle_prune` carries out one more check?

jeanphilippeD: Probably, not, what would be the purpose of this one with there is one at the end?

maqi: avoiding the extra vote (which incurs traffic) when there is churn already happens?

jeanphilippeD: A single vote does not seem to matter much, there will be other traffic going on.

maqi: ok then. I will move it to only check inside handle_churn

src/parsec.rs
```diff
@@ -44,10 +48,10 @@ const MAX_PARSECS: usize = 10;
 const PARSEC_SIZE_LIMIT: u64 = 1_000_000_000;
 // Limit in integration tests
 #[cfg(all(feature = "mock_base", not(feature = "mock_parsec")))]
-const PARSEC_SIZE_LIMIT: u64 = 200_000;
+const PARSEC_SIZE_LIMIT: u64 = 10_000_000;
 // Limit for integration tests with mock-parsec
 #[cfg(feature = "mock_parsec")]
-const PARSEC_SIZE_LIMIT: u64 = 100;
+const PARSEC_SIZE_LIMIT: u64 = 10_000;
```
jeanphilippeD: Why were the 2 limits increased, is that still relevant?
Does that ensure that other tests never prune? is that a good thing. it could be for now anyway.

src/states/adult.rs
```diff
@@ -195,7 +195,7 @@ impl Adult {
     // Can also just be a single target once node-ageing makes Offline votes Opaque which should
     // remove invalid test failures for unaccumulated parsec::Remove blocks.
     fn send_parsec_poke(&mut self) {
-        let version = *self.gen_pfx_info.first_info.version();
+        let version = self.gen_pfx_info.parsec_version;
```
jeanphilippeD: can you search in code if there are other places using `gen_pfx_info` or the struct that we did not update?

src/states/elder/mod.rs
```diff
@@ -1567,6 +1577,25 @@ impl Approved for Elder {
         self.merge_if_necessary()
     }
 
+    fn handle_prune(&mut self) -> Result<(), RoutingError> {
+        if self.chain.prefix_change() != PrefixChange::None {
+            log_or_panic!(
```
jeanphilippeD: probably want to return Ok(()) if we log or panic here.

src/parsec.rs
maqi: Ah, the increase is to ensure the pruning not happen too much previously (quite long time ago)
However, with the change that we now creates a new parsec instance each time there is a churn.
The change is not relevant anymore. The previous number now only cause pruning with reasonable frequency.
I will change them back.

src/states/adult.rs
maqi: had searches with `gen_pfx_info`, `GenesisPrxInfo`, `ParsecPoke`, `first_info.version()`, `ParsecRequest` & `ParsecResponse`
all got updated.

tests/mock_network/mod.rs
```diff
@@ -582,42 +582,55 @@ fn check_section_info_ack() {
...
-
+fn carry_out_parsec_pruning() {
+    let init_network_size = 7;
+    let elder_size = 8;
+    let safe_section_size = 8;
     let network = Network::new(
         NetworkParams {
-            elder_size: LOWERED_ELDER_SIZE,
-            safe_section_size: LOWERED_ELDER_SIZE,
+            elder_size,
```
jeanphilippeD: Any reason to change that? does that test need to run with more nodes, (i.e would it be less effective on reduced size network?)
Did you validate the test pass with mock & mock_base?

maqi: `add_connected_nodes_until_split` requires the initial network size NOT being greater than the min_section_size.
The previous setup is incorrect.

Test is passing with mock. 
Was disabled for mock_base as it takes long time to complete. Will have a local tweak of the graph_size parameter to verify it.

jeanphilippeD: > Was disabled for mock_base as it takes long time to complete.

=> That is really not ideal that test should be able to pass with mock_base in a reasonable time: We need it because Prune is all about impacting Parsec, we want to be able to test it with real Parsec.
=> Do we know why? would it make sense to vote for User payloads for mock_base would that speed it up?

> add_connected_nodes_until_split requires the initial network size NOT being greater than the min_section_size.

=> Then what not have min_section_size nodes initially?
=> Also any reason to go all the way to split after prune? why not just add 1 or 2 nodes?

maqi: > would it make sense to vote for User payloads for mock_base would that speed it up

I will have a try to see if this can speed things up.

> Also any reason to go all the way to split after prune? why not just add 1 or 2 nodes

I thought a split shall expose more potential issue if there is any?

src/parsec.rs
jeanphilippeD: OK, new value seem reasonable, values seem quite high but we can update them if needed.

tests/mock_network/mod.rs
jeanphilippeD: > I thought a split shall expose more potential issue if there is any?

Could be, but since we are already doing Parsec reset at every churn probably not needed.

> those nodes are required to carry out the pruning?

Sorry I don't understand, why would we not be able to prune if we only had min_section_size nodes?

maqi: > why would we not be able to prune if we only had min_section_size nodes

I probably mis-understood your question.
You were asking ` Then what not have min_section_size nodes initially?`  
If the initial size is less than the min_sec_size, then it's ok. If is greater than that, there is a chance a panic will be raised within the call to `add_connected_nodes_until_split `   https://github.com/maidsafe/routing/blob/fleming/tests/mock_network/utils.rs#L540-L549

jeanphilippeD: So we could have min_section_size, and just add one node afterward and be ok.
That would probably make the test significantly faster / simpler

maqi: ok. I will change to that

PR #1799, opened 2019-09-18
Approve candidate earlier to allow them to participate in DKG

### LINKED ISSUE #1773: New node needs to participate in DKG. (2-4 days should cover risk, but possible unkown)
(should already be part of `peer_manager` connected, so should receive gossip once DkgStart consensus).
   Will it accept it?
   Adult state is handling parsec gossip.
   => But need to be in adult state (Maybe will just work).
   => Maybe can call `handle_candidate_approval` in `handle_online_event`
   => Maybe in this task, in `handle_online_event` also vote to start DKG as well as Add, but ignore output.
   => Test
   => Risk (small change in order of things, but not really impact elder, and ProvingNode becoming an adult receive a message in same order so should not have impact either).

madadam:
This PR changes the candidate approval process so that it now happens on consensusing `Online`. This promotes the candidate to Adult which allows them to participate in parsec gossip and thus in the DKG. This PR does not implement the DKG handling itself though.

Closes #1773

### DIFF
--- a/src/states/elder/mod.rs
+++ b/src/states/elder/mod.rs
@@ -665,39 +665,35 @@ impl Elder {
         }
     }
 
+    // Send NodeApproval to the current candidate which promotes them to Adult and allows them to
+    // passively participate in parsec consensus (that is, they can receive gossip and poll
+    // consensused blocks out of parsec, but they can't vote yet)
     fn handle_candidate_approval(
         &mut self,
         new_pub_id: PublicId,
         new_client_auth: Authority<XorName>,
         outbox: &mut dyn EventBox,
-    ) -> Result<(), RoutingError> {
-        // Once the joining node joined, it may receive the vote regarding itself.
-        // Or a node may receive CandidateApproval before connection established.
-        // If we are not connected to the candidate, we do not want to add them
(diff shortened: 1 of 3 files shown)


src/states/elder/mod.rs
```diff
@@ -665,40 +665,23 @@ impl Elder {
...
         new_client_auth: Authority<XorName>,
-        outbox: &mut dyn EventBox,
-    ) -> Result<(), RoutingError> {
-        // Once the joining node joined, it may receive the vote regarding itself.
-        // Or a node may receive CandidateApproval before connection established.
-        // If we are not connected to the candidate, we do not want to add them
-        // to our RT.
-        // This will flag peer as valid if its found in peer_mgr regardless of their
-        // connection status to us.
-        let is_connected = if self.peer_mgr.get_peer(&new_pub_id).is_some() {
```
jeanphilippeD: Should we not still make sure we are connected.
If connection did not happen before and we get the parsec event Online consensus, we want to be connected so we can send the gossip?

madadam: That's a good point!

PR #1803, opened 2019-09-20
Remove resource proving

madadam:
Note that the `ProvingNode` state is not removed yet as it was doing more than just resource proof - namely it establishes connections to all the members of the section, it sends them `CandidateInfo` and waits for `NodeApproval`. These responsibilities might potentially be transferred to other states in a future work.


src/chain/candidate.rs
```diff
@@ -7,26 +7,36 @@
...
     None,
-    /// We accepted a candidate to perform resource proof in this section. We are waiting for
-    /// them to send their `CandidateInfo` before starting the actual resource proof.
-    AcceptedForResourceProof {
+    /// We accepted a candidate for joining this section. We are waiting for them to send their
+    /// `CandidateInfo`.
+    Accepted {
         target_interval: XorTargetInterval,
         old_public_id: PublicId,
+        timestamp: Instant,
```
jeanphilippeD: That does not seem right, this is for shared state stuff, cannot have a timestamp.
This is why there was the candidate in peer manager that is different: In peer manager this is the view of each node. Here it is the state shared by all nodes.

Comment apply to other place that use it.

src/peer_manager.rs
```diff
@@ -271,170 +174,6 @@ impl PeerManager {
...
-        self.candidate = Candidate::AcceptedForResourceProof {
-            res_proof_start: Instant::now(),
-            expired_once: false,
-        };
-    }
-
-    /// Verifies proof of resource.  If the response is not the current candidate, or if it fails
-    /// validation, returns `Err`.  Otherwise returns the target size, difficulty and the time
-    /// elapsed since the candidate was inserted.
-    pub fn verify_candidate(
```
jeanphilippeD: Can you clarify in PR comment that in addition to not resource proof, we are also not checking that the node is in the correct interval.

Are we doing no checks at all, what if it comes with a public id that is not part of our section?
=> answering my own question: if !target_interval.contains(new_pub_id.name()) {

src/chain/candidate.rs
dirvine: This would clear up with the replacement of the candidate/prover parts that the simpler BootStrap process will give us. I suggest we impl the bootstrap process in this PR to show the whole logic and allow us to remove temporary code to keep tests working.

jeanphilippeD: If it is going away soon, a TODO comment to say timestamp should not be in chain  is probably enough to address my comment.

madadam: I tried to see if I can remove the local candidate state (`Chain::candidate` and `PeerManager::candidate`) but there is a problem with the relocation interval which needs to be stored somewhere (so the section members can verify the new id of the joining node fits withing the interval). And as we are storing it we also need a mechanism to expire it so non-responding candidate won't block another nodes that want to join. Because of this I'm going to move the expiration info from `Chain` back to `PeerManager` for now.

PR #1804, opened 2019-09-20
sign SectionInfo with share

### LINKED ISSUE #1769: Use new mechanism to sign SectionInfo.
(estimate 2-3 days)

jeanphilippeD:
Closes #1769 

Building on PR #1802 , we are now adding a signature share to the parsec payload containing the SectionInfo. When ready, we will be able to also/instead sign the SectionKeyInfo when we add it as a result of the DKG (i.e `NetworkEvent(SectionInfo,SectionKeyInfo)).

The initial commits clean up code that was no longer used.

### DIFF
--- a/src/chain/network_event.rs
+++ b/src/chain/network_event.rs
@@ -6,8 +6,8 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use super::{SectionInfo, SectionKeyInfo};
-use crate::id::PublicId;
+use super::{Proof, SectionInfo, SectionKeyInfo};
+use crate::id::{FullId, PublicId};
 use crate::parsec;
 use crate::routing_table::Prefix;
 use crate::sha3::Digest256;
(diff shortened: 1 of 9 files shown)


src/chain/network_event.rs
```diff
@@ -59,10 +57,24 @@ pub struct SendAckMessagePayload {
...
+
+impl SectionInfoSigPayload {
+    pub fn new(
+        info: &SectionInfo,
+        full_id: &FullId,
+    ) -> Result<SectionInfoSigPayload, RoutingError> {
+        let proof = Proof::new(*full_id.public_id(), full_id.signing_private_key(), &info)?;
+
+        Ok(SectionInfoSigPayload {
+            pub_key_share: BlsPublicKeyShare(proof.pub_id),
```
maqi: just a curious question, if we only have the singing pub_id & sig included in this struct, why still call it `share`?

src/error.rs
```diff
@@ -149,6 +149,8 @@ pub enum RoutingError {
     UntrustedMessage,
     /// Crypto related error.
     Crypto(safe_crypto::Error),
+    /// a new SectionInfo could not be pushed.
```
maqi: The comment sounds like the error is not because the `New section info` is invalid, but ourself cannot accept it?

src/chain/network_event.rs
jeanphilippeD: because it will be a BLS signature share and public key share (i.e a part of the sig needed to identify it as signed by the section).

jeanphilippeD:
Rebased and fixed `/// a new SectionInfo could not be pushed.` comment.

PR #1809, opened 2019-09-24
New crypto

### LINKED ISSUE #1808: Consts required for node age
We will define some const values to allow node age to calculate the section membership and node Persona's. These are as follows 

`SAFE_SECTION_SIZE :u8 = 100`
`ELDER_SIZE :u8 = 7`

Note (thanks @ustulation ) that a split occurs when we can guarantee SAFE_SECTION_SIZE per new section. 

These const values are used to determine a nodes persona when we split the section. Routing will add nodes up to the `SAFE_SECTION_SIZE` but then the consumer of routing (Vaults in SAFE case) will request to add new nodes as the existing nodes become full or otherwise incapable. Whether the consumer indeed wishes a node `Killed` or just to leave it full is the choice of the consumer library.

madadam:
Replace the cryptographic primitives from the safe_crypto crate with the corresponding ones from tiny-keccak, ed25519-dalek and threshold-crypto.

- Hashing is implemented with tiny_keccak
- Signing and verification with ed25519-dalek
- Encryption and decryption with the non-threshold primitives from threshold-crypto.

Also add some constants that will be useful for node ageing (Closes #1808 )


src/id.rs
```diff
@@ -146,98 +152,121 @@ impl Display for PublicId {
...
+fn gen_secret_signing_key() -> signing::SecretKey {
+    use rand::OsRng;
+    let rng = OsRng::new().expect("Cannot generate random data, unsafe to continue");
+    signing::SecretKey::generate(&mut RngCompat(rng))
+}
+
+#[cfg(any(test, feature = "mock_base"))]
+fn gen_secret_signing_key() -> signing::SecretKey {
+    use maidsafe_utilities::SeededRng;
+    let rng = SeededRng::thread_rng();
```
jeanphilippeD: Does that ensure reproducible test seeds?

src/id.rs
```diff
@@ -96,21 +98,20 @@ impl parsec::SecretId for FullId {
...
+        self.sign(data)
     }
 
-    fn encrypt<M: AsRef<[u8]>>(&self, to: &Self::PublicId, msg: M) -> Option<Vec<u8>> {
-        let shared_secret = self
-            .encrypting_private_key()
-            .shared_secret(to.encrypting_public_key());
-        shared_secret.encrypt_bytes(msg.as_ref()).ok()
+    fn encrypt<M: AsRef<[u8]>>(&self, to: &Self::PublicId, plaintext: M) -> Option<Vec<u8>> {
+        serialise(&to.public_encryption_key().encrypt(plaintext)).ok()
```
jeanphilippeD: Is it ok to create a new OsRng every time we encrypt something?
Does that allow reproducible seeds? 
```
    /// Encrypts the message using the OS random number generator.
    ///
    /// Uses the `OsRng` by default. To pass in a custom random number generator, use
    /// `encrypt_with_rng()`.
    pub fn encrypt<M: AsRef<[u8]>>(&self, msg: M) -> Ciphertext {
        self.encrypt_with_rng(&mut OsRng::new().expect(ERR_OS_RNG), msg)
    }
```

src/id.rs
madadam: It does. `SeededRng` uses a global shared seed so as long as you create one `SeededRng` instance using `from_seed` somewhere at the beginning of your tests, all the other instances will be created from that seed.

src/id.rs
madadam: Good catch! I didn't realize it uses rng under the hood. It makes sense though given this thing doesn't use nonces. I'll have to think about how to handle this.

dirvine: Think we are OK here encrypt/decrypt is tested in the BLS crate itself. It is unlikely to fail in any reproducible manner that would be useful.

src/mock/crypto.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    impl SecretKey {
+        pub fn public_key(&self) -> PublicKey {
+            PublicKey(self.0)
+        }
+
+        pub fn decrypt(&self, ciphertext: &Ciphertext) -> Option<Vec<u8>> {
+            let offset = ciphertext.0.len() - self.0.len();
+
+            if ciphertext.0[offset..] == self.0 {
```
dirvine: :+1:

src/id.rs
```diff
@@ -96,21 +98,16 @@ impl parsec::SecretId for FullId {
...
-        self.signing_private_key().sign_detached(data)
+        self.sign(data)
     }
 
-    fn encrypt<M: AsRef<[u8]>>(&self, to: &Self::PublicId, msg: M) -> Option<Vec<u8>> {
-        let shared_secret = self
-            .encrypting_private_key()
-            .shared_secret(to.encrypting_public_key());
-        shared_secret.encrypt_bytes(msg.as_ref()).ok()
+    fn encrypt<M: AsRef<[u8]>>(&self, to: &Self::PublicId, plaintext: M) -> Option<Vec<u8>> {
```
dirvine: Yes, pity this trait demands type erasure.  Later we can fix that and use a generic for encrypted data types. That will be  a PARSEC update though.

src/id.rs
```diff
@@ -6,64 +6,67 @@
...
+use maidsafe_utilities::serialisation::{deserialise, serialise};
 use serde::de::Deserialize;
 use serde::{Deserializer, Serialize, Serializer};
 use std::fmt::{self, Debug, Display, Formatter};
-use std::ops::RangeInclusive;
+use std::{ops::RangeInclusive, rc::Rc};
 
 /// Network identity component containing name, and public and private keys.
-// FIXME Remove the Clone-ability
 #[derive(Clone)]
```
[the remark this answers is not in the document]
madadam: Sure, but I would do that only when we actually need it.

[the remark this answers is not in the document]
madadam: Unfortunately no - we need it because `FullId` is passed to Parsec. I tried to make it safer by putting the secret keys into `Rc`, so no copies are actually made in memory.

src/id.rs
```diff
@@ -146,98 +148,141 @@ impl Display for PublicId {
...
     fn verify_signature(&self, signature: &Self::Signature, data: &[u8]) -> bool {
-        self.signing_public_key().verify_detached(signature, data)
+        self.verify(data, signature)
     }
 }
 
 impl PublicId {
-    /// Return age.
+    /// Returns age.
     pub fn age(&self) -> u8 {
```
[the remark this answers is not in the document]
madadam: OK, I'll remove it then.

src/id.rs
jeanphilippeD: Why is that a pity?

madadam: Because it forces us to serialise the `Ciphertext` type which is additional work that wouldn't be necessary otherwise.

jeanphilippeD: It is a low occurring event so performance is not a problem, also it will be serialized on the wire so probably not make much gain.
On the other hand, making it generic means that parsec code & DKG has to cater for generic parameter where it only really care about a blob of data from this encrypt function.

So I'm not sure it will be a worthwhile trade off in the end but we'll see.

src/crypto.rs
```diff
@@ -0,0 +1,31 @@
...
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+//! Cryptographic primitives.
+
+/// SHA3-256 hash function.
+pub use tiny_keccak::sha3_256 as hash;
```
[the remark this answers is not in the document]
jeanphilippeD: Not sure what use case that would serve.
Would it make sense to do that rename when we start needed 2 different hashes?

src/states/common/relocated.rs
```diff
@@ -49,21 +50,9 @@ pub trait Relocated: Bootstrapped {
...
-                "{} - Failed to serialise our connection info for {:?}: {:?}.",
-                self, their_pub_id, err
-            );
-            err
-        })?;
-
+        let conn_info = serialise(&self.our_connection_info()?)?;
         let content = MessageContent::ConnectionRequest {
-            encrypted_conn_info,
+            encrypted_conn_info: conn_info,
```
fizyk20: Is the connection info no longer encrypted, then?

src/mock/quic_p2p/tests.rs
```diff
@@ -692,17 +692,16 @@ fn gen_message() -> NetworkBytes {
             id::FullId,
             messages::{DirectMessage, Message, SignedDirectMessage},
         };
-        use lazy_static::lazy_static;
 
-        lazy_static! {
-            static ref FULL_ID: FullId = FullId::new();
+        thread_local! {
```
jeanphilippeD: any reason for this change?

src/routing_table/authority.rs
```diff
@@ -20,6 +20,7 @@ use std::fmt::{self, Binary, Debug, Display, Formatter};
 /// `PrefixSection` use _section_ verification: the set from which a quorum is required is all
 /// members of the section (`Section`) or of all sections matching the prefix (`PrefixSection`).
 #[derive(Serialize, Deserialize, PartialEq, PartialOrd, Eq, Ord, Clone, Copy, Hash)]
+#[allow(clippy::large_enum_variant)]
```
jeanphilippeD: What changed, is it the PublicId now bigger?

src/states/common/relocated.rs
jeanphilippeD: Rename `encrypted_conn_info` ?

src/crypto.rs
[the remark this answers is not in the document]
dirvine: i.e. we could use the internals of ed25519 and pass SAH512 hashes for signature size/generic. We could also end up using 512 and 256 ourselves. We can rename at that point, however though I agree, however, hash itself is not very descriptive of the type/length etc. Especially if we then output a DIGEST256 alias. I am OK either way though

src/id.rs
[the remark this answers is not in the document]
dirvine: No worries we can raise a separate PR for that.  :+1:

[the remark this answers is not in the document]
dirvine: Ah yes, we will need to probably address that later on but OK for now. I would hope that is not the case fro too much longer mind you. Let's see, but for here we are good for now.

src/id.rs
```diff
@@ -86,12 +101,17 @@ impl parsec::SecretId for FullId {
...
-        unimplemented!()
+    fn encrypt<M: AsRef<[u8]>>(&self, to: &Self::PublicId, plaintext: M) -> Option<Vec<u8>> {
+        serialise(&to.public_encryption_key().encrypt(plaintext)).ok()
     }
 
-    fn decrypt(&self, _from: &Self::PublicId, _ct: &[u8]) -> Option<Vec<u8>> {
-        unimplemented!()
+    fn decrypt(&self, _from: &Self::PublicId, ciphertext: &[u8]) -> Option<Vec<u8>> {
+        let ciphertext: encryption::Ciphertext = deserialise(ciphertext).ok()?;
+        if !ciphertext.verify() {
```
jeanphilippeD: Why are we doing that?
I expected it would be done by the decrypt function as otherwise it would not be safe.

```
    /// Returns the decrypted text, or `None`, if the ciphertext isn't valid.
    pub fn decrypt(&self, ct: &Ciphertext) -> Option<Vec<u8>> {
        if !ct.verify() {
            return None;
        }
        let Ciphertext(ref u, ref v, _) = *ct;
        let g = u.into_affine().mul(*self.0);
        Some(xor_with_hash(g, v))
    }
```

src/id.rs
dirvine: I see type erasure always as a pity ;-) Here I am also still cool to merge though. In this case for parsec at least we can go back to Ciphertext and validate before trying to decrypt. So the overhead is on both sides of the encrypt/decrypt setup here.

src/states/common/relocated.rs
dirvine: No, it is open to everyone. It is just how it all works now, each node can connect to any other node direct. quic-p2p checks for 2 way connections to prevent network imbalance of some nodes connecting and some not to random other nodes. 
This is why we can hold connection info for any node and give it to any other. Later (post-launch we will enable hole punch and that will be an indirection, but complex for now)

src/id.rs
```diff
@@ -126,76 +152,114 @@ impl Display for PublicId {
...
-    SecretKey::generate(&mut RngCompat(rng))
+    signing::SecretKey::generate(&mut RngCompat(rng))
+}
+
+#[cfg(not(any(test, feature = "mock_base")))]
+fn gen_secret_encryption_key() -> encryption::SecretKey {
+    use rand::OsRng;
+    use rand_crypto::Rng;
+
+    let rng = OsRng::new().expect("Cannot generate random data, unsafe to continue");
```
jeanphilippeD: Is that re-implementing `new_rng()` from parsec.rs?

src/chain/network_event.rs
```diff
@@ -69,7 +69,7 @@ impl SectionInfoSigPayload {
         info: &SectionInfo,
         full_id: &FullId,
     ) -> Result<SectionInfoSigPayload, RoutingError> {
-        let proof = Proof::new(*full_id.public_id(), full_id.signing_private_key(), &info)?;
+        let proof = Proof::new(full_id, &info)?;
```
jeanphilippeD: Should that not have been part of a previous commit?

src/mock/crypto.rs
```diff
@@ -0,0 +1,196 @@
...
+                Err(SignatureError)
+            } else {
+                let mut signature = Self([0; SIGNATURE_LENGTH]);
+                signature.0.copy_from_slice(bytes);
+                Ok(signature)
+            }
+        }
+    }
+
+    impl PartialEq for Signature {
```
jeanphilippeD: is that not deriving all these?

src/id.rs
```diff
@@ -253,6 +253,26 @@ fn gen_secret_encryption_key() -> encryption::SecretKey {
...
+fn encrypt<M: AsRef<[u8]>>(
+    public_key: &encryption::PublicKey,
+    plaintext: M,
+) -> encryption::Ciphertext {
+    // Note: this uses OsRng under the hood.
+    public_key.encrypt(plaintext)
+}
+
+#[cfg(any(test, feature = "mock_base"))]
+fn encrypt<M: AsRef<[u8]>>(
```
jeanphilippeD: Is it somewhat re-implementing new_rng again? (maybe different here)

src/mock/quic_p2p/tests.rs
madadam: `FullId` is no longer `Sync` because it now holds `Rc` and `lazy_static` requires `Sync`.

src/id.rs
jeanphilippeD: Fixed in later commit

src/states/common/relocated.rs
fizyk20: We should probably rename the field, then (JP mentioned this in another comment).

src/routing_table/authority.rs
madadam: Hm, I just checked. In fleming, `PublicId` is 97 bytes, in this PR it's 96. So it's actually smaller. Then I tried to run clippy on fleming and that lint got triggered also there. No idea how that get through CI :thinking:

src/id.rs
madadam: This is outdated change.

src/mock/crypto.rs
madadam: Unfortunately no, as the array in `Signature` is 64 elements long and `derive` of `Eq`, `Ord`, `Hash`, etc... works only up to 32 elements. This will get resolved when const generics get stabilized in rust, which should hopefully be soon.

src/chain/network_event.rs
madadam: It looks like that. Not sure how it happened. Do you want me to do something about it?

jeanphilippeD: Could move it to the right commit, but I'll let you decide if it is worth it.

PR #1814, opened 2019-09-26
Add section member infos to shared state

### LINKED ISSUE #1807: Node Container
All nodes in our section are held in the `Chain` currently.  Here we introduce a new node container for all nodes. This container is for internal section management held by `Elders`. The container is a list of 

```
enum NodePersona {
  Elder(PublicId),
  Adult(PublicId),
  Infant(PublicId)
}

enum NodeNetworkStatus {
 Joined,
Left,
Relocated,
}

struct NodeStatus {
  age : u8,
  status : NodeStatus,
  persona : NodePersona,
}

```

As this information is not inter-section but intra-section then only this section requires to make it non-volatile. This information will be written to disk by the section Elders.

Any `NodePersona` that is an `Elder` will be written into the chain.  The length of the `NodeStatus` list should be capped at 20,000 entries for now. If a rejoining node cannot be found then it will start again as an `Infant`.  If a node has been `Relocated` if will be removed fom this list.

madadam:
It will be used for node ageing. Closes #1807 .

### DIFF
--- a/src/chain/shared_state.rs
+++ b/src/chain/shared_state.rs
@@ -6,7 +6,10 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use super::{bls_emu::BlsPublicKeyForSectionKeyInfo, AccumulatingProof, EldersInfo, MemberInfo};
+use super::{
+    bls_emu::BlsPublicKeyForSectionKeyInfo, AccumulatingProof, EldersInfo, MemberInfo,
+    MemberPersona, MemberState,
+};
 use crate::{
     crypto::Digest256, error::RoutingError, id::PublicId, BlsPublicKey, BlsPublicKeySet,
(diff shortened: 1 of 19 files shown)


src/chain/shared_state.rs
```diff
@@ -34,6 +34,8 @@ pub struct SharedState {
     /// The latest few fully signed infos of our own sections.
     /// This is not a `BTreeSet` as it is ordered according to the sequence of pushes into it.
     pub our_infos: NonEmptyList<SectionInfo>,
+    /// Info about all members of our section - elders, adults and infants.
+    pub our_members: BTreeMap<PublicId, MemberInfo>,
```
jeanphilippeD: It seem we've added dead code and an empty collection.

Would it make sense to populate it and use it as part of this PR. i.e:
- When a node drop put it in the `Left` state.
- Currently we only have elder, but getting consensus on it with the NetworkEvent::SectionInfo(....).

PR #1816, opened 2019-09-26
Tune prune limit for real PARSEC

### LINKED ISSUE #1815: Parse prune test vote_prune fails with real parsec
The `vote_prune` fails when run with `mock_base` or `mock_serialise`

octol:
Fixes the vote_prune test when runnig with `mock_base` or `mock_serialise`.

NOTE: test still fails for unrelated reasons until #1813 is merged

Resolves #1815


octol:
> Nice one, is it best to block that one until 1813 is merged (and probably 1809 as well), so we can run soak testing on this specific test with real parsec?

Yes I think that makes sense

PR #1822, opened 2019-09-30
Update constructor (builder)

### LINKED ISSUE #1819: Update constructor (builder)
Routing must allow the consumer to have several options to construct a node. These options are 

1. Start with no `FullId` (node guaranteed Infant)
2. Start with existing `FullId` (node is rejoining)
3. Start with a single boostrap node (`NodeInfo`)
4. Start with several bootstrap nodes. 
5. Start with existing chain & gossip graph (i.e. when a node is restarting, but not necessarily rejoining the network, such as when a node upgrades and simply restarts)

Currently, there are not more options for the constructor (builder).

madadam:
This PR expands the ways to construct a `Node`:
1. Using randomly generated full id, will join a random section as infant (pending node ageing)
2. Using explicit full id, will try to rejoin previously joined section, keeping the age and persona
3. Resuming a previously paused node. Used for restarts and upgrades.

It also includes small refactoring of the testing framework. This was done with the intention of making the pause/resume implementation and testing easier. It turned out to not be needed in the end, but as it simplifies the testing framework a bit, I believe it is worthwhile to keep it.

Finally, it removed the config and config_handler as it was no longer needed.

Closes #1819


sample_config/sample.routing.config
```diff
@@ -1,8 +1,6 @@
 {
```
dirvine: I suspect we can get rid of this file and config handler altogether. If we wish to set Elder_size etc. we cah have `cfg` flags for tests or even add that to the builder? I think though cfg flags are easy.  Then we don't have routing config files at all.

PR #1838, opened 2019-10-10
inform vaults (users) about client events

ustulation:
Routing doesn't handle these. Simply off-load to vaults.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -9,8 +9,11 @@
 use crate::routing_table::Authority;
 use crate::routing_table::Prefix;
 use crate::xor_name::XorName;
+use crate::NetworkBytes;
 use hex_fmt::HexFmt;
+use quic_p2p::Token;
 use std::fmt::{self, Debug, Formatter};
+use std::net::SocketAddr;
 
 /// An Event raised by a `Node` or `Client` via its event sender.
 ///
@@ -23,6 +26,43 @@ use std::fmt::{self, Debug, Formatter};
 // FIXME - See https://maidsafe.atlassian.net/browse/MAID-2026 for info on removing this exclusion.
 #[allow(clippy::large_enum_variant)]
 pub enum Event {
+    // =========== Client Events - Routing won't handle these ===========
+    /// Inform the user (library) that we are connected to a client
+    ConnectedToClient {
+        /// Client's endpoint
+        peer_addr: SocketAddr,
+    },
+    /// Inform the user (library) that we are disconnected from a client
+    ConnectionFailureToClient {
+        /// Client's endpoint
+        peer_addr: SocketAddr,
+    },
+    /// Inform the user (library) that we have a new message from a client
+    NewMessageFromClient {
+        /// Client's endpoint
+        peer_addr: SocketAddr,
+        /// Client's message
+        msg: NetworkBytes,
+    },
+    /// Inform the user (library) that we couldn't send this message to a client
+    UnsentUserMsgToClient {
+        /// Client's endpoint
+        peer_addr: SocketAddr,
+        /// Message we had tried to send to the client
+        msg: NetworkBytes,
+        /// Token that we had used to identify this message
+        token: Token,
+    },
+    /// Inform the user (library) that we have successfully sent this message to a client
+    SentUserMsgToClient {
+        /// Client's endpoint
+        peer_addr: SocketAddr,
+        /// Message we had tried to send to the client
+        msg: NetworkBytes,
+        /// Token that we had used to identify this message
+        token: Token,
+    },
+    // ==================================================================
     /// Received a message.
     MessageReceived {
         /// The content of the message.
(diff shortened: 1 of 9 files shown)


src/event.rs
```diff
@@ -23,6 +26,43 @@ use std::fmt::{self, Debug, Formatter};
 // FIXME - See https://maidsafe.atlassian.net/browse/MAID-2026 for info on removing this exclusion.
 #[allow(clippy::large_enum_variant)]
 pub enum Event {
+    // =========== Client Events - Routing won't handle these ===========
```
dirvine: Could this be called CleintEvent and remove all the Client refs inside?

src/peer_map.rs
```diff
@@ -129,6 +129,12 @@ impl PeerMap {
     pub fn has<N: AsRef<XorName>>(&self, name: N) -> bool {
         self.forward.contains_key(name.as_ref())
     }
+
+    // If we don't have any information about this peer when this query is made then in certain
+    // context it could mean it's likely a client
+    pub fn is_peer_known_to_routing(&self, peer_addr: &SocketAddr) -> bool {
```
madadam: I'd call this `is_node` because I think it better describes the purpose of the method (whether the connection is from client or node). Also it's shorter :D

src/event.rs
```diff
@@ -23,6 +26,43 @@ use std::fmt::{self, Debug, Formatter};
...
+        /// Message we had tried to send to the client
+        msg: NetworkBytes,
+        /// Token that we had used to identify this message
+        token: Token,
+    },
+    /// Inform the user (library) that we have successfully sent this message to a client
+    SentUserMsgToClient {
+        /// Client's endpoint
+        peer_addr: SocketAddr,
+        /// Message we had tried to send to the client
```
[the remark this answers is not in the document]
ustulation: If they have a logic that waits of success or failure i figured it could be useful. Since we have that for node-node and routing uses it, kinda unfair to not give that freedom to vault to client esp when it's coming from quic-p2p universally for all connections regardless.

I'm OK to remove it if you want though

src/event.rs
```diff
@@ -23,6 +26,43 @@ use std::fmt::{self, Debug, Formatter};
...
+    },
+    /// Inform the user (library) that we have a new message from a client
+    NewMessageFromClient {
+        /// Client's endpoint
+        peer_addr: SocketAddr,
+        /// Client's message
+        msg: NetworkBytes,
+    },
+    /// Inform the user (library) that we couldn't send this message to a client
+    UnsentUserMsgToClient {
```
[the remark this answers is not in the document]
ustulation: Thought would maintain symmetry with quic-p2p and routing types so that it's not named in different ways in different places. It's already `UnsentUserMsg` between routing and quic-p2p and hence this for the vaults.

Again, I'm OK to call it SendError if you want (i don't feel strongly either way)

src/event.rs
ustulation: A separate event enum you mean ? If so I think that's fine.

dirvine: Yes that is what I mean. :+1:

src/event.rs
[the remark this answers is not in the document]
dirvine: Let's leave, for now, it may be an optimisation.

src/event.rs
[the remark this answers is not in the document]
dirvine: No worries let's keep that pattern for now and if needed we can change in all crates.

src/event.rs
```diff
@@ -9,8 +9,59 @@
...
+        /// Token that we had used to identify this message
+        token: Token,
+    },
+}
+
+impl Into<Event> for ClientEvent {
+    fn into(self) -> Event {
+        Event::ClientEvent(self)
+    }
+}
```
madadam: Nitpick, but it's more idiomatic to define this as `impl From<ClientEvent> for Event`. Then you get the `Into` impl for free.

src/states/common/base.rs
```diff
@@ -99,6 +115,21 @@ pub trait Base: Display {
...
     }
 
+    fn handle_send_msg_to_client(
+        &mut self,
+        _peer_addr: SocketAddr,
+        _msg: NetworkBytes,
+        _token: Token,
+    ) -> Result<(), InterfaceError> {
+        warn!("{} - Cannot send message to client- invalid state.", self);
+        Err(InterfaceError::InvalidState)
```
madadam: Why not just send the message here? You don't need to be elder to do it.

src/states/common/base.rs
```diff
@@ -99,6 +115,21 @@ pub trait Base: Display {
...
+    }
+
+    fn handle_disconnect_client(&mut self, _peer_addr: SocketAddr) -> Result<(), InterfaceError> {
+        warn!("{} - Cannot handle DisconnectClient - invalid state.", self);
+        Err(InterfaceError::InvalidState)
```
madadam: Same here (see https://github.com/maidsafe/routing/pull/1838/files#r333556207)

src/states/common/base.rs
ustulation: You are sending to a client. It doens't make sense unless you are an elder

src/states/common/base.rs
ustulation: Same reply as there :)

ustulation: Note this is also allowed for `adults` as you would want to disconnect if there was a demotion.

PR #1841, opened 2019-10-11
Fix SectionProofChain comparison

madadam:
It now takes into account that two emulated-BLS signatures might not be equal according to their `Eq` impl even though they are each created from threshold + 1 signature shares from the same set of nodes. This would possibly not be the case with real BLS.

Also modifies the simultaneous_joining_nodes test to stop polling once relevant conditions are met instead of waiting until all nodes become idle which might not always happen before the max number of poll iterations is exhausted.

### DIFF
--- a/src/chain/bls_emu.rs
+++ b/src/chain/bls_emu.rs
@@ -31,7 +31,7 @@ pub struct SecretKeyShare(FullId);
 #[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Copy, Hash, Serialize, Deserialize, Debug)]
 pub struct PublicKeyShare(pub PublicId);
 
-#[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Hash, Serialize, Deserialize)]
+#[derive(Eq, PartialEq, Clone, Hash, Serialize, Deserialize)]
 pub struct Signature {
     sigs: BTreeMap<PublicId, SignatureShare>,
 }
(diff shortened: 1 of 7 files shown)


jeanphilippeD:
> This would possibly not be the case with real BLS.

This would not be the case with real BLS, I've added a test for it in Parsec.

jeanphilippeD:
> Also modifies the simultaneous_joining_nodes test to stop polling once relevant conditions are met instead of waiting until all nodes become idle which might not always happen before the max number of poll iterations is exhausted.

Why is it that we would exhaust the max number of iteration, is this maximum incorrect?
We add a lot of code in this test, but that pattern is used in many tests (i think), what makes that test different?

madadam:
> Why is it that we would exhaust the max number of iteration, is this maximum incorrect?
> We add a lot of code in this test, but that pattern is used in many tests (i think), what makes that test different?

It's a good question. It could be that the max number of iterations is simply not sufficient for this test. It could also be some other issue. My guess (which I haven't verified) is that perhaps some node cast a vote which  never got consensused for some reason which then blocks the polling because `has_unpolled_observations` keeps returning `true`. I think it might be worth investigating this issue more closely to figure out what's really going on.

On the other hand, the change I did in the test is how I think we should do all tests eventually - that is - do not wait until all the nodes become "idle". Instead poll only until some condition that we care about it fulfilled (In this case we care about the new nodes joining, becoming elders and all the other nodes registering them as such). This change made the test pass while we still test the thing we actually care about, so I considered it good enough.

PR #1851, opened 2019-10-17
fix/routing: correctly identify messages as non client

jeanphilippeD:
Fix easy to reproduce failure in messages_during_churn:
-When connection is dropped by the destination because of split for
 example, we get a quic_p2p error and we need to resend the message.
-Unfortunatly, we detected that message as from a client because we
 dropped the node information before getting this error.

 => Use a straightforward client detection by storing all client address
    in hash map. There may be many client but we can optimize as needed.

Also clean up interface: We were half delegating client handling to
vault and half handling it depending on our state.
=> Always delegate client handling to Vault regardless of state.
   Vault can decide what to do as an adult.

Test:
Soak tests pass
Clippy pass

### DIFF
--- a/src/peer_map.rs
+++ b/src/peer_map.rs
@@ -7,7 +7,7 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::{id::PublicId, xor_name::XorName, ConnectionInfo};
-use fxhash::FxHashMap;
+use fxhash::{FxHashMap, FxHashSet};
 use std::net::SocketAddr;
 
 /// This structure holds the bi-directional association between peers public id and their network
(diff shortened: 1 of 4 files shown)


jeanphilippeD:
> At a fundamental level, if the dest forgot about you for whatever reason and you still need to resend the msg to the dest, it means you still are interested in the dest. So the dest should still figure out in your peer-map as a node. If that is not being done and the msg is interpreted as a client msg then i suspect it's a different problem we should address and hence not convinced this fix is correct.
> If the dest disconnects from you, you shouldn't be losing it as you still want it. If you have decided you don't want to know about it and also lose it, you shouldn't be sending it anything.

It is the way the problem was solved, i.e when we send a message, we keep just enough to resend to all possible target even if they are no longer in the peer_map and assume as we can that this will succeed.Using the peer_map to decide whether a connection is for a node is the reason we make a mistake because they are not related.Unless we want to re-design the re-send mechanism, there is little options here, currently 'is_node' is lying.

ustulation:
OK - probably good to have a discussion with what @dirvine wants here. If it speeds up the process, I'm happy to leave it as is and get this merged. Your call @jeanphilippeD 

I'm tagging this as approved in-case it's preventing someone from merging it

jeanphilippeD:
I'll merge this as soon as it is possible as it is currently blocking us from running soak tests.
We can revisit if we decide we need too, as if we find a better solution this will at least allow in the mean time for us to run tests and find more subtle failures.

PR #1864, opened 2019-10-23
serialize churn

### LINKED ISSUE #1860: Make churn event handling sequential
Here is my plan for handling mutliple add/remove at the same time. This borrows an idea from old Node ageing plan.

When doing a churn that result in SectionInfo/ElderInfo change (all at the moment), Keep that as flagged with any additional information needed for completing the churn (For example something like https://github.com/maidsafe/routing/compare/fleming...fizyk20:issue_1774#diff-bb7fddc1a592903612286af78e19ad6bR141 churn_in_progress info needed for BLS). We only reset this flag when the process is complete (i.e SectionInfo + Add/Remove Elder complete).

When that flag is on, OnlineEvent/Offline are moved to a queue. When the current churn is completed and flag is reset, we progress the next events. (This is very fast compared to the rate of Elder added/removed so we should not really have multiple elder changes). One option is to process all the events at once, which seem reasonable, since it is shared state all good nodes will create the votes in the same

jeanphilippeD:
Partial fix to allow work on DKG and also better handle Relocation as well.
Closes #1860 

Re-enable multiple_joining_nodes test.

Implement serialization of events:
in multiple_joining_nodes the 5 joining nodes are added and no nodes should fail.

Fix multiple issues related to dropping messages
 => backlog messages when not yet an elder
 => Forward routing messages received as Adult to correct destination.
 => Increase Signature accumulator time out and log when incomplete message is dropped.
 => Provide correct proof when sending to a Prefix authority (multiple sections).

Note: The ConnectionRequest should no longer be needed but can be better addressed as part of the follow up PR completing PeerMap removal.


src/chain/chain.rs
```diff
@@ -263,6 +276,7 @@ impl Chain {
                         .collect(),
                 };
 
+                self.process_churn = false;
```
madadam: Should we only flip this to `false` when the section info is for our own section?

jeanphilippeD: Yes, I'll push the updated branch if you are looking at it.

madadam: Sure. Or if you want me to wait with the review a bit, let me know.

jeanphilippeD:
> Sure. Or if you want me to wait with the review a bit, let me know.

Good to get early feedback. pushed the additional change I'm working with. still having issue with simultaneous_joining_nodes after a few 10s iterations.

src/chain/chain.rs
```diff
@@ -304,6 +317,20 @@ impl Chain {
             | AccumulatingEvent::Relocate(_) => (),
         }
 
+        let start_churn_event = match event {
+            AccumulatingEvent::Online(_) | AccumulatingEvent::Offline(_) => true,
```
madadam: Also `Relocate`.

src/chain/chain.rs
```diff
@@ -304,6 +317,20 @@ impl Chain {
...
+            _ => false,
+        };
+
+        if start_churn_event {
+            if self.process_churn {
+                self.state.churn_event_backlog.push_front(event);
+                return Ok(None);
+            } else {
+                self.process_churn = true;
```
madadam: One potential issue about this is that you expect the `Online` / `Offline` to always be followed by `SectionInfo`. But what about nodes that join but are not promoted to elders? Wouldn't we block any further churn events until someone else is promoted/demoted? This is not a problem in the current codebase, because we always promote everyone. But it won't stay that way for long.

madadam: Thinking about this problem, maybe it's not an issue. We would always know if we want to promote a new node immediately or not. If we decide to not promote then, we can clear the `progress_churn` flag explicitly (and process the next backlogged event, if any).

src/chain/shared_state.rs
```diff
@@ -177,14 +195,25 @@ impl SharedState {
...
+                log_or_panic!(
+                    LogLevel::Error,
+                    "{} - update_with_genesis_related_info different churn_event_backlog:\n{:?},\n{:?}",
+                    log_ident,
+                    self.churn_event_backlog,
+                    churn_event_backlog
+                );
+            }
         }
+        self.handled_genesis_event = true;
```
jeanphilippeD: Need to be set always, not just here, otherwise initial nodes will think they are not ready to handle backlog.

src/states/joining_peer.rs
```diff
@@ -234,11 +253,14 @@ impl Base for JoiningPeer {
...
+        p2p_node: P2pNode,
         _outbox: &mut dyn EventBox,
     ) -> Result<Transition, RoutingError> {
-        debug!("{} Unhandled direct message: {:?}", self, msg);
-
+        debug!(
+            "{} Unhandled direct message, adding to backlog: {:?}",
+            self, msg
+        );
+        self.direct_msg_backlog.push((p2p_node, msg));
```
madadam: Prevent duplicate `Relocate` message handling: either ignore `Relocate` here or add a check in `Elder::handle_relocate` to ignore the message if its `pub_id` is different from ours.

src/states/joining_peer.rs
```diff
@@ -43,12 +43,13 @@ pub struct JoiningPeer {
     network_service: NetworkService,
     routing_msg_filter: RoutingMessageFilter,
     msg_backlog: Vec<SignedRoutingMessage>,
```
madadam: I would rename this one to `routing_msg_backlog` now that we have `direct_msg_backlog` too, for explicitness.

src/states/joining_peer.rs
madadam: Hm, given that relocation is not yet implicit, this is not an issue yet. I'll handle it myself.

src/states/joining_peer.rs
```diff
@@ -135,6 +135,18 @@ impl JoiningPeer {
...
+                content:
+                    MessageContent::ConnectionRequest {
+                        conn_info, pub_id, ..
+                    },
+                src: Authority::Node(_),
+                dst: Authority::Node(_),
+            } => {
+                self.peer_map_mut().insert(pub_id, conn_info);
+                self.send_direct_message(&pub_id, DirectMessage::ConnectionResponse);
+                Ok(Transition::Stay)
```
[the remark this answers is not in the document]
jeanphilippeD: Because ConnectionRequest will be removed by @octol follow up PR, maybe best to avoid refactoring here.

jeanphilippeD:
Fixed review comment.

src/chain/chain.rs
```diff
@@ -252,6 +255,16 @@ impl Chain {
     /// If the event is a `EldersInfo` or `NeighbourInfo`, it also updates the corresponding
     /// containers.
     pub fn poll(&mut self) -> Result<Option<(AccumulatingEvent, EldersChange)>, RoutingError> {
+        if self.state.handled_genesis_event
+            && !self.process_churn
+            && self.state.change == PrefixChange::None
```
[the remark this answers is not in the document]
jeanphilippeD: I think so, before, we would never process an Online/Offline event if `self.state.change == PrefixChange::None` because of `is_valid_transition`.

src/states/elder/mod.rs
```diff
@@ -1137,6 +1154,35 @@ impl Base for Elder {
         &mut self.timer
     }
 
+    fn finish_handle_transition(&mut self, outbox: &mut dyn EventBox) -> Transition {
+        debug!("{} - State changed finish to Elder.", self);
```
madadam: Weird wording. What about "State change to Elder finished" or just "State changed to Elder"?

src/states/adult.rs
```diff
@@ -289,6 +283,34 @@ impl Base for Adult {
         &mut self.timer
     }
 
+    fn finish_handle_transition(&mut self, outbox: &mut dyn EventBox) -> Transition {
+        debug!("{} - State changed to Adult.", self);
+
+        for msg in self.routing_msg_backlog.drain(..).collect_vec() {
```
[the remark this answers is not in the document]
jeanphilippeD: Yes, but that add a clone for the content, while the drain do not, however, it might be better to use mem::replace.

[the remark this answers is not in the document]
madadam: Good point! It's fine then.

src/states/adult.rs
```diff
@@ -284,27 +287,23 @@ impl Base for Adult {
     }
 
     fn finish_handle_transition(&mut self, outbox: &mut dyn EventBox) -> Transition {
-        debug!("{} - State changed to Adult.", self);
+        debug!("{} - State changed to Adult finished.", self);
```
madadam: "changed" -> "change"

PR #1871, opened 2019-10-30
fix out of order neighbour info

### LINKED ISSUE #1870: NeighbourInfo being processed before the previous NeighbourInfo
So 2 problems:
- What we see here is the result of a NeighbourInfo being processed before the previous NeighbourInfo for that section was. As a result we are loosing a neigbour prefix.- The sub-problem is that we do not reach quorum for the previous NeighbourInfo, seemingly because the elder change at the time the message was received result in not enough node having received the message to get the 2/3 quorum.My plan:

Solution:
- Separate NeighbourInfo vote from SectionInfo as they are 2 different things and that cause confusion.
- Use a 1/3 quorum guaranteeing at least one honest node for Parsec event that are the result of received Section messages. => This will make it very difficult to not reach quorum unless many elder change at once.
- Ensure we only process NeigbourInfo quorum if it is the direct descendant of the neighbour we know.

jeanphilippeD:
Closes #1870 

The core of the solution is to ensure
 1- We do not process a NeighbourInfo before we processed its parent section.
 2 - We do not ignore NeighbourInfo because we voted for a follow up one (still needed for quorum).

Additionally, start isolating the NeighbourInfo Parsec event, as it is not the same as a SectionInfo.
Additional work was not required to fix this issue and can be done as needed or in other PRs.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -668,26 +668,6 @@ impl Chain {
         self.neighbour_elders_p2p().map(P2pNode::public_id)
     }
 
-    /// Returns `true` if we know the section with `elders_info`.
-    ///
-    /// If `check_signed` is `true`, also trust sections that we have signed but that haven't
-    /// accumulated yet.
-    #[cfg(feature = "mock_base")]
-    pub fn is_trusted(&self, elders_info: &EldersInfo, check_signed: bool) -> bool {
-        let is_proof = |si: &EldersInfo| si == elders_info || si.is_successor_of(elders_info);
-        let mut signed = self
-            .signed_events()
-            .filter_map(AccumulatingEvent::elders_info);
-        if check_signed && signed.any(is_proof) {
-            return true;
-        }
-        if elders_info.prefix().matches(self.our_id.name()) {
-            self.state.our_infos().any(is_proof)
-        } else {
-            self.neighbour_infos().any(is_proof)
-        }
-    }
-
     /// Return the keys we know
     pub fn get_their_keys_info(&self) -> impl Iterator<Item = (&Prefix<XorName>, &SectionKeyInfo)> {
         self.state.get_their_keys_info()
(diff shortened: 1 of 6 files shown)


src/chain/network_event.rs
```diff
@@ -78,6 +78,9 @@ pub enum AccumulatingEvent {
     NeighbourMerge(Digest256),
     SectionInfo(EldersInfo),
 
+    // Voted for received message with keys to we can update neighbour_info.
```
maqi: Not quite clear of what this comment means.
does it mean `Voted for received message with keys that we can update our neighbour_info to` ?

src/chain/chain.rs
```diff
@@ -753,7 +730,7 @@ impl Chain {
     fn should_skip_accumulator(&self, event: &NetworkEvent) -> bool {
         // FIXME: may also need to handle non SI votes to not get handled multiple times
         let si = match event.payload {
-            AccumulatingEvent::SectionInfo(ref si) => si,
+            AccumulatingEvent::SectionInfo(ref si) | AccumulatingEvent::NeighbourInfo(ref si) => si,
```
maqi: It seems the `AccumulatingEvent::NeighbourInfo` is always being handled as the same to `AccumulatingEvent::SectionInfo`.
Then why have this extra enum ?

jeanphilippeD: This was done as the first step to make them different. But as I was modifying the code I found fixing the route cause did not require it, so I did not progress it further. However, multiple place would be clearer if handled differently as mentioned in PR description. As well, it helps with logs.

src/chain/network_event.rs
jeanphilippeD: will fix

src/chain/chain.rs
maqi: ok.
I'd still prefer the `NeighbourInfo` to be in a separate PR.
However won't make this as a blocking point.

jeanphilippeD:
Fixed review comment and Rebase.

PR #1873, opened 2019-10-31
Fix additional failures with multi join

jeanphilippeD:
Fix additional seeds with mock_base and mock.
Refactor/simplify code in is_validation to support debugging.

### DIFF
--- a/tests/mock_network/mod.rs
+++ b/tests/mock_network/mod.rs
@@ -28,6 +28,7 @@ use routing::{
     mock::Network, Event, EventStream, NetworkConfig, NetworkParams, Prefix, XorName,
     XorTargetInterval,
 };
+use std::collections::BTreeMap;
 
 pub const LOWERED_ELDER_SIZE: usize = 3;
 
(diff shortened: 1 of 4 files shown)


tests/mock_network/mod.rs
```diff
@@ -65,6 +66,55 @@ fn nodes_with_prefix_mut<'a>(
...
+pub fn count_sections_members_if_split(nodes: &[TestNode]) -> BTreeMap<Prefix<XorName>, usize> {
+    let mut counts = BTreeMap::new();
+    for pfx in nodes.iter().map(node_prefix_if_split) {
+        *counts.entry(pfx).or_default() += 1;
+        let _ = counts.entry(pfx.sibling()).or_default();
+    }
+    counts
+}
+
+pub fn node_prefix_if_split(node: &TestNode) -> Prefix<XorName> {
```
maqi: this function seems better to be a helper function in node ?

tests/mock_network/mod.rs
```diff
@@ -65,6 +66,55 @@ fn nodes_with_prefix_mut<'a>(
         .filter(move |node| prefix.matches(&node.name()))
 }
 
+pub fn count_sections_members_if_split(nodes: &[TestNode]) -> BTreeMap<Prefix<XorName>, usize> {
+    let mut counts = BTreeMap::new();
+    for pfx in nodes.iter().map(node_prefix_if_split) {
+        *counts.entry(pfx).or_default() += 1;
+        let _ = counts.entry(pfx.sibling()).or_default();
```
maqi: why need to create an entry for sibling here?
we already used `entry().or_default` above

tests/mock_network/mod.rs
maqi: this function seems better to be a helper function in node ?

tests/mock_network/mod.rs
```diff
@@ -65,6 +66,55 @@ fn nodes_with_prefix_mut<'a>(
...
+    None
+}
+
+fn can_accept_node_without_split(
+    count_if_split_node: &BTreeMap<Prefix<XorName>, usize>,
+    min_split_size: usize,
+) -> bool {
+    count_if_split_node
+        .values()
+        .any(|count| *count < min_split_size * 2 - 1)
```
maqi: So this means `a node can be accepted without split when ANY of the input prefix contains less than 2*min_size-1 nodes` ?
I thought shall be `ALL` ? or we return someting like `Some(prefix)` ?

tests/mock_network/mod.rs
jeanphilippeD: Could be but it is really specialized, not clear there would be much use in other tests. so it does not seem reasonable to pollute that interface.

tests/mock_network/mod.rs
jeanphilippeD: Because it failed the test :-)
The problem is if the first 3 nodes happen to be in the same sub prefix the map only contain one prefix. with this it contain one sub prefix with 3 and the other sub-prefix with 0.

tests/mock_network/mod.rs
jeanphilippeD: If any are `*count < min_split_size * 2 - 1`, then at least one sub prefix can accept more nodes.
if none do, then there is no node we can create that could be accepted so we avoid the infinite loop.

tests/mock_network/mod.rs
maqi: > Because it failed the test ...  the other sub-prefix with 0

hmm...  this means the map has to be `prefix symmetric`, right?
I think the test failure is due to not catering the `non-symmetric` case. 
i.e. not generating new nodes evently.

I won't insisit to cartering for it, but better leave a comment here to remind us why this is required (and maybe need to cover it later on) ?

jeanphilippeD: This is because of how we use the map, if the map does not cover the full address space we have a problem. The reason we have a problem is when we create a new node, we look for which sub-prefix it is in and if there is space. If there is no sub-prefix for it, then it cannot be added. (In the degenerate case I mentioned, 3 node in same sub-prefix 0, you can never add any node to sub-prefix 1, and it never split).

jeanphilippeD:
Fixed review comment: Added a comment and used `unwrap!` to make more explicit the assumption on the map.

PR #1876, opened 2019-11-01
Review comment fixes for #1868

octol:
Fixes for the review comments in https://github.com/maidsafe/routing/pull/1868

That PR was merged before review comments were addressed as they were minor and blocking fixes to tests.

### DIFF
--- a/src/id.rs
+++ b/src/id.rs
@@ -239,6 +239,11 @@ impl P2pNode {
         }
     }
 
+    /// Creates a `ConnectionInfo` from the `P2pNode` instance.
+    pub fn into_connection_info(self) -> ConnectionInfo {
+        self.connection_info
+    }
+
     /// Returns the `PublicId`.
     pub fn public_id(&self) -> &PublicId {
         &self.public_id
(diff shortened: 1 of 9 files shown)


src/id.rs
```diff
@@ -257,7 +261,7 @@ impl P2pNode {
 
 impl Debug for P2pNode {
     fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
-        write!(formatter, "PublicId(name: {})", self.public_id.name())
+        write!(formatter, "{:?}", self.public_id)
```
madadam: This impl makes the debug output of `P2pNode` indistinguishable from the one of `PublicId`. So I would at least do something like:
```rust
write!(formatter, "P2pNode(name: {})", self.public_id.name())
```
but maybe better to also include the conn info:
```rust
write!(formatter, "P2pNode({} at {})", self.public_id.name(), self.conn_info.peer_addr)
```
or something like that.

octol: Yep good point, addressed in latest update

PR #1879, opened 2019-11-05
Replace member container in `EldersInfo` and fix order trait impl

octol:
- Replace member container set with map in `EldersInfo`.
- Implement order traits correctly for `P2pNode`.
- Improve performance by not unnecessarily clone and collect into sets and maps in `EldersInfo` methods. The `churn` tests show a ~10% improvement for me locally on average.

These things were pointed out during the review of #1868

### DIFF
--- a/src/chain/bls_emu.rs
+++ b/src/chain/bls_emu.rs
@@ -163,7 +163,7 @@ mod test {
             .enumerate()
             .map(|(index, full_id)| {
                 (
-                    *full_id.public_id(),
+                    *full_id.public_id().name(),
                     P2pNode::new(
                         *full_id.public_id(),
                         ConnectionInfo {
--- a/src/chain/bls_emu.rs
+++ b/src/chain/bls_emu.rs
@@ -162,12 +162,15 @@ mod test {
             .iter()
             .enumerate()
             .map(|(index, full_id)| {
-                P2pNode::new(
+                (
                     *full_id.public_id(),
-                    ConnectionInfo {
-                        peer_addr: ([127, 0, 0, 1], (index + 9000) as u16).into(),
-                        peer_cert_der: vec![],
-                    },
+                    P2pNode::new(
+                        *full_id.public_id(),
+                        ConnectionInfo {
+                            peer_addr: ([127, 0, 0, 1], (index + 9000) as u16).into(),
+                            peer_cert_der: vec![],
+                        },
+                    ),
                 )
             })
             .collect();
--- a/src/chain/bls_emu.rs
+++ b/src/chain/bls_emu.rs
@@ -70,7 +70,7 @@ impl PublicKeyShare {
 
 impl PublicKeySet {
     pub fn from_elders_info(elders_info: EldersInfo) -> Self {
-        let threshold = elders_info.members().len() * THRESHOLD_NUMERATOR / THRESHOLD_DENOMINATOR;
+        let threshold = elders_info.len() * THRESHOLD_NUMERATOR / THRESHOLD_DENOMINATOR;
         Self {
             threshold,
             elders_info,
@@ -87,7 +87,7 @@ impl PublicKeySet {
     {
         let sigs: BTreeMap<_, _> = shares
             .into_iter()
-            .filter(|(pk, _ss)| self.elders_info.members().contains(&pk.0))
+            .filter(|(pk, _ss)| self.elders_info.is_member(&pk.0))
             .map(|(pk, ss)| (pk.0, *ss))
             .collect();
         // In the BLS scheme, more than `threshold` valid signatures are needed to obtain a
@@ -114,8 +114,7 @@ impl PublicKey {
         sig.sigs
             .iter()
             .filter(|&(pk, ss)| {
-                self.0.elders_info.members().contains(pk)
-                    && PublicKeyShare(*pk).verify(ss, msg.as_ref())
+                self.0.elders_info.is_member(pk) && PublicKeyShare(*pk).verify(ss, msg.as_ref())
             })
             .count()
             > self.0.threshold
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -229,9 +229,9 @@ impl Chain {
         match event {
             AccumulatingEvent::SectionInfo(ref info)
             | AccumulatingEvent::NeighbourInfo(ref info) => {
-                let old_neighbours: BTreeSet<_> = self.neighbour_elders_p2p().cloned().collect();
+                let old_neighbours: BTreeSet<_> = self.neighbour_elder_nodes().cloned().collect();
                 self.add_elders_info(info.clone(), proofs)?;
-                let new_neighbours: BTreeSet<_> = self.neighbour_elders_p2p().cloned().collect();
+                let new_neighbours: BTreeSet<_> = self.neighbour_elder_nodes().cloned().collect();
 
                 if let Some((ref cached_info, _)) = self.state.split_cache {
                     if cached_info == info {
@@ -504,18 +504,18 @@ impl Chain {
             .get_member_connection_info(&pub_id)
             .ok_or(RoutingError::PeerNotFound(pub_id))?;
 
-        let mut elders_p2p = self.state.new_info.p2p_members().clone();
-        let _ = elders_p2p.insert(P2pNode::new(pub_id, connection_info.clone()));
+        let mut elder_nodes = self.state.new_info.p2p_members().clone();
+        let _ = elder_nodes.insert(pub_id, P2pNode::new(pub_id, connection_info.clone()));
 
         // TODO: the split decision should be based on the number of all members, not just elders.
-        if self.should_split(&elders_p2p)? {
-            let (our_info, other_info) = self.split_self(elders_p2p.clone())?;
+        if self.should_split(&elder_nodes)? {
+            let (our_info, other_info) = self.split_self(elder_nodes.clone())?;
             self.state.change = PrefixChange::Splitting;
             return Ok(vec![our_info, other_info]);
         }
 
         self.state.new_info = EldersInfo::new(
-            elders_p2p,
+            elder_nodes,
             *self.state.new_info.prefix(),
             Some(&self.state.new_info),
         )?;
@@ -529,11 +529,7 @@ impl Chain {
         self.assert_no_prefix_change("remove elder");
 
         let mut elders = self.state.new_info.p2p_members().clone();
-        let connection_info = self
-            .get_member_connection_info(&pub_id)
-            .ok_or(RoutingError::PeerNotFound(pub_id))?;
-        let p2p_node = P2pNode::new(pub_id, connection_info.clone());
-        let _ = elders.remove(&p2p_node);
+        let _ = elders.remove(&pub_id);
 
         if self.our_id() == &pub_id {
             self.is_elder = false;
@@ -673,8 +669,8 @@ impl Chain {
     pub fn elders_p2p(&self) -> impl Iterator<Item = &P2pNode> {
         self.neighbour_infos()
             .chain(iter::once(self.state.our_info()))
-            .flat_map(EldersInfo::p2p_members)
-            .chain(self.state.new_info.p2p_members())
+            .flat_map(EldersInfo::member_nodes)
+            .chain(self.state.new_info.member_nodes())
     }
 
     /// Returns a set of elders we should be connected to.
@@ -707,12 +703,12 @@ impl Chain {
 
     /// Returns elders from our own section according to the latest accumulated `SectionInfo`.
     pub fn our_elders(&self) -> impl Iterator<Item = &P2pNode> + ExactSizeIterator {
-        self.state.our_info().p2p_members().iter()
+        self.state.our_info().member_nodes()
     }
 
     /// Returns all neighbour elders.
-    pub fn neighbour_elders_p2p(&self) -> impl Iterator<Item = &P2pNode> {
-        self.neighbour_infos().flat_map(EldersInfo::p2p_members)
+    pub fn neighbour_elder_nodes(&self) -> impl Iterator<Item = &P2pNode> {
+        self.neighbour_infos().flat_map(EldersInfo::member_nodes)
     }
 
     /// Return the keys we know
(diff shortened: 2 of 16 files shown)


src/id.rs
```diff
@@ -284,10 +287,45 @@ impl PartialOrd for P2pNode {
 
 impl Ord for P2pNode {
     fn cmp(&self, other: &Self) -> Ordering {
-        self.public_id.cmp(&other.public_id)
+        let self_socket_addr = SocketAddrWrapper(self.connection_info.peer_addr);
```
jeanphilippeD: Not ideal, you still have a discrepency between Ord and Eq.
Either you implement Ord and Eq or you derive Ord and Eq.

If you want to go for the wrapper implementation, probably better to store that wrapper as oppose to constructing it on each comparison, also your wrapper may be better as ConnectionInfoWrapper since that is what you store.
Then you can Derive Ord,Eq,Hash on P2pNode and implement them on ConnectionInfoWrapper.

src/chain/chain.rs
```diff
@@ -291,26 +291,33 @@ impl Chain {
...
-                        .difference(&old_neighbours)
-                        .cloned()
-                        .collect(),
-                    removed: old_neighbours
-                        .difference(&new_neighbours)
-                        .cloned()
-                        .collect(),
-                };
+                // Create BTreeSet for the purpose of taking the difference.
+                // WIP: can we do better?
```
jeanphilippeD: Yes, not ideal

src/lib.rs
```diff
@@ -68,7 +68,7 @@
     unconditional_recursion,
     unknown_lints,
     unsafe_code,
-    unused,
+    //unused,
```
octol: Oops these should be removed

src/id.rs
```diff
@@ -279,51 +280,46 @@ impl Display for P2pNode {
...
-        compare_socket_addr(self.0, other.0)
+        let self_addr = deconstruct_socket_addr(&self.0.peer_addr);
+        let other_addr = deconstruct_socket_addr(&other.0.peer_addr);
+        (self_addr, &self.0.peer_cert_der).cmp(&(other_addr, &other.0.peer_cert_der))
     }
 }
 
-fn compare_socket_addr(left_addr: SocketAddr, right_addr: SocketAddr) -> Ordering {
-    let to_tuple = |socket_addr: SocketAddr| match socket_addr {
+fn deconstruct_socket_addr(socket_addr: &SocketAddr) -> (Ipv6Addr, u16, u32, u32) {
```
jeanphilippeD: If you include the &peer_cert_der to the return result, it will be much simpler to implement the Ord/Eq/Hash.

src/id.rs
```diff
@@ -279,51 +280,46 @@ impl Display for P2pNode {
     }
 }
 
-impl PartialOrd for P2pNode {
-    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        Some(self.cmp(other))
+#[derive(Clone, Serialize, Deserialize)]
```
jeanphilippeD: Wonder if to make it all consistent it would make sense to implement Serialize/Deserialize as well. Not sure.

src/chain/chain.rs
```diff
@@ -291,26 +291,33 @@ impl Chain {
...
+                let old_ids: BTreeSet<_> = old_neighbours.keys().cloned().collect();
+                let new_ids: BTreeSet<_> = new_neighbours.keys().cloned().collect();
+                let added_ids: BTreeSet<_> = new_ids.difference(&old_ids).cloned().collect();
+                let removed_ids: BTreeSet<_> = old_ids.difference(&new_ids).cloned().collect();
+
+                let added = new_neighbours
+                    .into_iter()
+                    .filter(|(pub_id, _)| added_ids.contains(pub_id))
+                    .collect();
```
jeanphilippeD: can remove new set and just do:
                 let added = new_neighbours
                    .into_iter()
                    .filter(|(pub_id, _)| !old_neighbours.contains_key(pub_id))
                    .collect();

octol: Ah good point, yes I'll try that! Thanks!

src/chain/chain.rs
octol: That block is now redone!

src/id.rs
octol: Agree, done!

src/chain/chain.rs
```diff
@@ -1484,9 +1473,9 @@ impl Chain {
 #[derive(Default)]
 pub struct EldersChange {
     // Peers that became elders.
-    pub added: BTreeSet<P2pNode>,
+    pub added: BTreeMap<PublicId, P2pNode>,
     // Peers that ceased to be elders.
-    pub removed: BTreeSet<P2pNode>,
+    pub removed: BTreeMap<PublicId, P2pNode>,
```
madadam: I don't think this change was necessary. We use this stuct only to update the neighbour connections (`Elder::update_neighbour_connections`) and we don't need to lookup by `PublicId` there.

octol: Yes with the new implementation of `Ord` of `P2pNode` we can probable go back to having `BTreeSet<P2pNode>`, without any wrappers even probably?

octol: Reverted to `BTreeSet<P2pNode>` for Elder updates in the latest revision

src/chain/elders_info.rs
```diff
@@ -80,25 +80,32 @@ impl EldersInfo {
...
         &self.members
     }
 
+    pub fn member_nodes(&self) -> impl Iterator<Item = &P2pNode> {
+        self.members.values()
+    }
+
     pub fn member_names(&self) -> BTreeSet<XorName> {
-        self.members.iter().map(P2pNode::name).copied().collect()
+        self.members.values().map(P2pNode::name).copied().collect()
```
jeanphilippeD: Need to fix performance issue there: (avoid creating a set all the time it is very expensive).
(Optionally also check that the count is enough before doing the filter if that also improve performance).

```
    /// Returns `true` if the proofs are from a quorum of this section.
    pub fn is_quorum(&self, proofs: &ProofSet) -> bool {
        // WIP: the call to members is probably to heavy?
        proofs
            .ids()
            .filter(|id| self.members().contains(id))
            .count()
            * QUORUM_DENOMINATOR
            > self.members.len() * QUORUM_NUMERATOR
    }
    /// Returns `true` if the proofs are from all members of this section.
    pub fn is_total_consensus(&self, proofs: &ProofSet) -> bool {
        proofs
            .ids()
            .filter(|id| self.members().contains(id))
            .count()
            == self.members.len()
    }
```

https://maidsafe.slack.com/archives/C03DZQTKU/p1573127337176200?thread_ts=1573063288.154600&cid=C03DZQTKU

src/chain/elders_info.rs
```diff
@@ -80,25 +80,32 @@ impl EldersInfo {
...
+            .members
+            .iter()
+            .chain(&other.members)
+            .map(|(pub_id, p2p_node)| (*pub_id, p2p_node.clone()))
+            .collect();
         Self::new(members, self.prefix.popped(), vec![self, other])
     }
 
     // WIP: remove me (or make less heavy)
     pub fn members(&self) -> BTreeSet<PublicId> {
```
jeanphilippeD: Probably this is problematic, at call site it looks cheap to call, it probably should not look like a straight accessor.
Maybe best to rename to something like `members_as_set`.

Also should it better be an iterator that we return? (it depends how often it is used and how).

src/chain/elders_info.rs
```diff
@@ -80,25 +80,32 @@ impl EldersInfo {
...
+    // WIP: rename to member_map?
+    pub fn p2p_members(&self) -> &BTreeMap<PublicId, P2pNode> {
         &self.members
     }
 
+    pub fn member_nodes(&self) -> impl Iterator<Item = &P2pNode> {
+        self.members.values()
+    }
+
     pub fn member_names(&self) -> BTreeSet<XorName> {
```
jeanphilippeD: Probably same here `member_names_as_set`.
Also should it better be an iterator that we return? (it depends how often it is used and how).

src/chain/elders_info.rs
```diff
@@ -29,7 +29,7 @@ use std::{
 #[derive(Default, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub struct EldersInfo {
     /// The complete list of the section's elders' IDs.
-    members: BTreeSet<P2pNode>,
+    members: BTreeMap<PublicId, P2pNode>,
```
jeanphilippeD: Looking at the follow up PR, Probably better to do that change here and use XorName as the key directly, this way you won't have to jump through many hoops in follow up PR to find P2pNode for a XorName.

src/chain/elders_info.rs
octol: >   Maybe best to rename to something like members_as_set.
Agree
> Also should it better be an iterator that we return? (it depends how often it is used and how).
The reason we return sets is mostly because we call `contains` on them. However yeah we can probably call `contains_key` on the underlying map directly instead. Maybe even get rid of the method all together

src/chain/elders_info.rs
octol: Agree, whenever we copy/clone to a new container it makes sense to indicate it in the name

src/chain/elders_info.rs
octol: Makes sense, I'll have a go at that

src/chain/chain.rs
```diff
@@ -1071,11 +1055,11 @@ impl Chain {
             .into_iter()
             .flat_map(|(_, si)| {
                 si.member_names()
-                    .into_iter()
                     .sorted_by(|name0, name1| name.cmp_distance(name0, name1))
             })
             .filter(|name| connected_peers.contains(&name))
             .take(count)
+            .cloned()
```
madadam: You can use `copied` here as `XorName` implements `Copy`. In general it's a good idea to use `copied` instead of `cloned` so that if we ever make `XorName` non-`Copy` for whatever reason, this code would stop compiling which would make us aware of a potential performance impact.

src/chain/elders_info.rs
```diff
@@ -29,7 +29,7 @@ use std::{
 #[derive(Default, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub struct EldersInfo {
     /// The complete list of the section's elders' IDs.
```
madadam: We might need to update this comment.

src/chain/elders_info.rs
```diff
@@ -120,22 +136,19 @@ impl EldersInfo {
...
     /// Returns `true` if the proofs are from a quorum of this section.
     pub fn is_quorum(&self, proofs: &ProofSet) -> bool {
-        // WIP: the call to members is probably to heavy?
-        proofs
-            .ids()
-            .filter(|id| self.members().contains(id))
-            .count()
-            * QUORUM_DENOMINATOR
-            > self.members.len() * QUORUM_NUMERATOR
+        if proofs.len() * QUORUM_DENOMINATOR <= self.len() * QUORUM_NUMERATOR {
```
madadam: I would add a comment here explaining why are we doing this check seemingly twice.

src/chain/elders_info.rs
```diff
@@ -120,22 +136,19 @@ impl EldersInfo {
...
     }
 
     /// Returns `true` if the proofs are from all members of this section.
     pub fn is_total_consensus(&self, proofs: &ProofSet) -> bool {
-        proofs
-            .ids()
-            .filter(|id| self.members().contains(id))
-            .count()
-            == self.members.len()
+        if proofs.len() < self.len() {
```
madadam: Also here.

src/chain/chain.rs
octol: Yep agree, makes sense!

src/id.rs
octol: To be honest I'm not entirely sure myself either, but my feeling is that we should just use the generated one

src/chain/elders_info.rs
octol: Dropped this commit as local tests didn't show any performance improvements

src/chain/elders_info.rs
octol: Same as above

src/chain/elders_info.rs
```diff
@@ -28,8 +28,8 @@ use std::{
 /// change, due to an elder being added or removed, or the section splitting or merging.
 #[derive(Default, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub struct EldersInfo {
-    /// The complete list of the section's elders' IDs.
-    members: BTreeSet<P2pNode>,
+    /// The section's complete set of elders as a map from it's name to a `P2pNode`.
```
madadam: "it's name" -> "its name"

madadam: But maybe better "their name"

src/chain/elders_info.rs
```diff
@@ -120,22 +136,13 @@ impl EldersInfo {
...
     }
 
     /// Returns `true` if the proofs are from all members of this section.
     pub fn is_total_consensus(&self, proofs: &ProofSet) -> bool {
-        proofs
-            .ids()
-            .filter(|id| self.members().contains(id))
-            .count()
-            == self.members.len()
+        proofs.ids().filter(|id| self.is_member(id)).count() == self.members.len()
```
jeanphilippeD: not consistent with is_quorum (ideally would use `self.len()` as well. (Also keep level of abstraction consistent which is good).

src/chain/bls_emu.rs
```diff
@@ -70,7 +70,7 @@ impl PublicKeyShare {
 
 impl PublicKeySet {
     pub fn from_elders_info(elders_info: EldersInfo) -> Self {
-        let threshold = elders_info.members().len() * THRESHOLD_NUMERATOR / THRESHOLD_DENOMINATOR;
+        let threshold = elders_info.len() * THRESHOLD_NUMERATOR / THRESHOLD_DENOMINATOR;
```
jeanphilippeD: General comment:
Could you Add in PR description that this also address a performance issue, and any number you have.

src/chain/elders_info.rs
octol: Fixed!

PR #1880, opened 2019-11-05
handle disconnection only via unresponsiveness track

### LINKED ISSUE #1875: Tackle the disconnect part of Remove ConnectionRequest/Response pattern and PeerMap work
This is a split off issue from the work of [PR 1865 Remove ConnectionRequest/Response pattern and PeerMap](https://github.com/maidsafe/routing/pull/1865) to tackle the disconnect part.

It also remove the last use of send_connection_request Allowing to remove ConnectionRequest, and prepping the work to remove ConnectionResponse.

So far we only have elder and want to deal with disconnection when using the Responsiveness Tracking.
When we will want to deal with Adult, probably a different mechanism will be required.

maqi:
This PR contains the work of handling disconnection via responsiveness tracking.
The main idea is that when a disconnection of peer is detected, we now don't need to do anything.
Just leave it to the `responsive tracking` to report that peer as unresponsive and remove it later on.

Closes #1875

### DIFF
--- a/src/states/adult.rs
+++ b/src/states/adult.rs
@@ -353,11 +353,6 @@ impl Base for Adult {
         Transition::Stay
     }
 
-    fn handle_peer_lost(&mut self, pub_id: PublicId, _: &mut dyn EventBox) -> Transition {
-        debug!("{} - Lost peer {}", self, pub_id);
-        Transition::Stay
-    }
-
     fn handle_direct_message(
         &mut self,
         msg: DirectMessage,
--- a/src/states/adult.rs
+++ b/src/states/adult.rs
@@ -589,6 +589,10 @@ impl Approved for Adult {
         Ok(())
     }
 
+    fn check_vote_status(&mut self) {
+        debug!("{} - Not Required CheckVotingStatus For Adult", self);
+    }
+
     fn handle_their_key_info_event(
         &mut self,
         _key_info: SectionKeyInfo,
--- a/src/states/elder/mod.rs
+++ b/src/states/elder/mod.rs
@@ -96,6 +96,8 @@ pub struct Elder {
     /// DKG cache
     dkg_cache: BTreeMap<BTreeSet<PublicId>, EldersInfo>,
     rng: MainRng,
+    #[cfg(not(feature = "mock_parsec"))]
+    unresponsive_nodes: BTreeMap<PublicId, usize>,
 }
 
 impl Elder {
@@ -252,6 +254,8 @@ impl Elder {
             pfx_is_successfully_polled: false,
             dkg_cache: Default::default(),
             rng: details.rng,
+            #[cfg(not(feature = "mock_parsec"))]
+            unresponsive_nodes: Default::default(),
         }
     }
 
@@ -1653,6 +1657,48 @@ impl Approved for Elder {
         self.disconnect(&pub_id);
 
         Ok(())
+    }
+
+    // Checking members vote status and vote to remove those non-resposive nodes.
+    // Mock parsec will fetch all peers' votes in one poll loop, hence doesn't need the logic of
+    // expecting extra accusations. And even may have the risk of not having enough accusations if
+    // switch to the same code as for real parsec.
+    #[cfg(feature = "mock_parsec")]
+    fn check_vote_status(&mut self) {
+        let unresponsive_nodes = self.chain.check_vote_status();
+        for pub_id in unresponsive_nodes {
+            self.vote_for_event(AccumulatingEvent::Offline(pub_id));
+        }
+    }
+
+    #[cfg(not(feature = "mock_parsec"))]
+    fn check_vote_status(&mut self) {
+        let unresponsive_nodes = self.chain.check_vote_status();
+        for (id, count) in mem::replace(&mut self.unresponsive_nodes, Default::default()) {
+            if unresponsive_nodes.contains(&id) {
+                let new_count = count.saturating_add(1);
+                info!(
+                    "{} detected {:?} as unresponsive for {} times when having {} members",
+                    self,
+                    id,
+                    new_count,
+                    self.chain.our_info().len(),
+                );
+                // The average chance that the missing peer will pick us as gossip recipient, so
+                // that we can know its votes (or we know its votes via others) is
+                // `1/number_of_peers`. So requiring `the node to be accused more times than there
+                // is elders in the section` is mainly just to ensure the full knowledge is reached
+                // before making the accuse.
+                if new_count > self.chain.our_info().len() {
+                    self.vote_for_event(AccumulatingEvent::Offline(id));
+                } else {
+                    let _ = self.unresponsive_nodes.insert(id, new_count);
+                }
+            }
+        }
+        for id in unresponsive_nodes {
+            let _ = self.unresponsive_nodes.entry(id).or_insert(1);
+        }
     }
 }
 
--- a/src/states/elder/mod.rs
+++ b/src/states/elder/mod.rs
@@ -1056,29 +1056,6 @@ impl Elder {
         Ok((targets.into_iter().cloned().collect(), dg_size))
     }
 
-    // Check whether we are connected to any elders. If this node loses all elder connections,
-    // it must be restarted.
-    fn check_elder_connections(&mut self, outbox: &mut dyn EventBox) -> bool {
-        if self
-            .peer_map()
-            .connected_ids()
-            .filter(|id| self.chain.our_id() != *id)
-            .any(|id| self.chain.is_peer_our_elder(id))
-        {
-            true
-        } else {
-            debug!("{} - Lost all elder connections.", self);
-
-            // Except network startup, restart in other cases.
-            if self.chain.our_info().version() > 0 {
-                outbox.send_event(Event::RestartRequired);
-                false
-            } else {
-                true
-            }
-        }
-    }
-
     fn check_signed_message_trust(&self, msg: &SignedRoutingMessage) -> Result<(), RoutingError> {
         if msg.check_trust(&self.chain) {
             Ok(())
@@ -1272,35 +1249,6 @@ impl Base for Elder {
         self.network_service
             .service_mut()
             .disconnect_from(conn_info.peer_addr);
-        Transition::Stay
-    }
-
-    fn handle_peer_lost(&mut self, pub_id: PublicId, outbox: &mut dyn EventBox) -> Transition {
-        debug!("{} - Lost peer {}", self, pub_id);
-
-        if !self.check_elder_connections(outbox) {
-            return Transition::Terminate;
-        }
-
-        if self.chain.is_peer_our_member(&pub_id) {
-            self.vote_for_event(AccumulatingEvent::Offline(pub_id));
-        }
-
-        if self.chain.is_peer_elder(&pub_id) {
-            debug!(
-                "{} - Sending connection request to {} due to lost peer.",
-                self, pub_id
-            );
-
-            let our_name = *self.name();
-            let _ = self.send_connection_request(
-                pub_id,
-                Authority::Node(our_name),
-                Authority::Node(*pub_id.name()),
-                outbox,
-            );
-        }
-
         Transition::Stay
     }
 
--- a/tests/mock_network/churn.rs
+++ b/tests/mock_network/churn.rs
@@ -9,15 +9,15 @@
 use super::{
     clear_relocation_overrides, count_sections, create_connected_nodes,
     create_connected_nodes_until_split, current_sections, gen_range, gen_range_except,
-    poll_and_resend, verify_invariant_for_all_nodes, TestNode,
+    poll_and_resend, verify_dropped_nodes, verify_invariant_for_all_nodes, TestNode,
 };
 use itertools::Itertools;
 use rand::Rng;
 use routing::{
     mock::Network,
     test_consts::{UNRESPONSIVE_THRESHOLD, UNRESPONSIVE_WINDOW},
-    Authority, Event, EventStream, NetworkConfig, NetworkParams, XorName, QUORUM_DENOMINATOR,
-    QUORUM_NUMERATOR,
+    Authority, Event, EventStream, NetworkConfig, NetworkParams, PublicId, XorName,
+    QUORUM_DENOMINATOR, QUORUM_NUMERATOR,
 };
 use std::{
     cmp,
@@ -35,7 +35,7 @@ fn drop_random_nodes<R: Rng>(
     rng: &mut R,
     nodes: &mut Vec<TestNode>,
     max_per_pfx: Option<usize>,
-) -> BTreeSet<XorName> {
+) -> BTreeSet<PublicId> {
     let mut dropped_nodes = BTreeSet::new();
     let elder_size = |node: &TestNode| unwrap!(node.inner.elder_size());
     let node_section_size = |node: &TestNode| {
@@ -75,8 +75,11 @@ fn drop_random_nodes<R: Rng>(
 
         *unwrap!(drop_count.get_mut(&pfx)) += 1;
         let dropped = nodes.remove(i);
-        assert!(dropped_nodes.insert(dropped.name()));
+        assert!(dropped_nodes.insert(dropped.id()));
     }
+
+    verify_dropped_nodes(rng, nodes, &dropped_nodes);
+
     dropped_nodes
 }
 
@@ -226,6 +229,7 @@ fn random_churn<R: Rng>(
     assert!(max_drop > 0);
     let dropped_nodes = drop_random_nodes(rng, nodes, Some(max_drop));
     warn!("Dropping nodes: {:?}", dropped_nodes);
+
     BTreeSet::new()
 }
 
--- a/tests/mock_network/drop.rs
+++ b/tests/mock_network/drop.rs
@@ -7,26 +7,29 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use super::{
-    create_connected_nodes, poll_all, poll_and_resend, verify_invariant_for_all_nodes, TestNode,
+    create_connected_nodes, poll_all, verify_dropped_nodes, verify_invariant_for_all_nodes,
+    TestNode,
 };
-use rand::Rng;
-use routing::{mock::Network, Event, EventStream, NetworkParams};
+use routing::{mock::Network, NetworkParams, PublicId};
+use std::{collections::BTreeSet, iter};
 
-// Drop node at index and verify its own section detected it.
-fn drop_node(nodes: &mut Vec<TestNode>, index: usize) {
+// Drop node at index and verify its still recognise as elder by other section members.
+fn drop_node(nodes: &mut Vec<TestNode>, index: usize) -> BTreeSet<PublicId> {
     let node = nodes.remove(index);
-    let name = node.name();
+    let id = node.id();
     let close_names = node.close_names();
 
     drop(node);
 
     // Using poll_all instead of poll_and_resend here to only let the other nodes realise the node
-    // got disconnected, but not make more progress.
+    // got disconnected, but not make more progress. We expect the node to remain elder until
+    // detected as unresponsive.
     let _ = poll_all(nodes);
 
     for node in nodes.iter_mut().filter(|n| close_names.contains(&n.name())) {
-        assert!(!node.inner.is_connected(name));
+        assert!(node.inner.is_peer_our_member(&id));
     }
+    iter::once(id).collect()
 }
 
 #[test]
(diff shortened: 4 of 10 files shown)


src/node.rs
```diff
@@ -467,6 +467,15 @@ impl Node {
             elder.trigger_relocation(node_id, destination_address)
```
jeanphilippeD: Not a refactor.

tests/mock_network/churn.rs
```diff
@@ -234,9 +237,38 @@ fn random_churn<R: Rng>(
...
+) {
+    for _ in 0..UNRESPONSIVE_WINDOW {
+        let event: Vec<_> = rng.gen_iter().take(100).collect();
+        nodes.iter_mut().for_each(|node| {
+            let _ = node
+                .inner
+                .elder_state_mut()
+                .map(|state| state.vote_for_user_event(event.clone()));
+        });
+        poll_and_resend(&mut nodes);
```
jeanphilippeD: Seem we want to only do `poll_and_resend` once.
i.e vote for all event we need to vote (i.e UNRESPONSIVE_WINDOW), then call poll_and_resend.
then we should assert that the dropped_node is gone because it should be, i.e panic otherwise, but not loop to try again?

maqi: hmm...  sending too many user data at once could cause other valid node to be accused as unresponsive, as the `votes` from other peers haven't arrived yet via polling.
That's why it has a `vote -> poll -> check` pattern here.
Maybe I could change it to send out a bunch of Threshold votes first, then vote & poll one by one?

jeanphilippeD: Did I miss something, but I thought that would not be an issue since we only track the vote that reached consensus? Why do you think you need to do one at a time? That is not how things will happen in real world, we will have a bunch of votes between each gossip event.

maqi: we only requires `supermajority` to reach consensus, and the distribution of messages are not random enough. Which means there is chance that some nodes' vote always got received AFTER consensus, and be accused of un-responsive.
It won't happen often, but may fail the soak test with certain chance.

src/node.rs
```diff
@@ -467,6 +467,15 @@ impl Node {
...
         }
     }
+
+    /// Check whether the peer is an elder in our section.
+    pub fn is_peer_our_elder(&self, pub_id: &PublicId) -> bool {
+        if let Some(elder) = self.elder_state() {
+            elder.is_peer_our_elder(pub_id)
+        } else {
+            false
+        }
```
octol: Looks fine, I kinda like the chained formulation but that's of course highly arbitrary ;)
```rust
self.elder_state().map(|elder| elder.is_peer_our_elder(pub_id)).unwrap_or_false()
```

tests/mock_network/churn.rs
```diff
@@ -234,9 +237,37 @@ fn random_churn<R: Rng>(
...
+// Sending out enough user data so that the dropped node can be detected via unresponsive tracking.
+fn verify_dropped_nodes<R: Rng>(
+    rng: &mut R,
+    mut nodes: &mut Vec<TestNode>,
+    dropped_nodes: &BTreeSet<PublicId>,
+) {
+    let is_still_elder = |node: &TestNode| {
+        dropped_nodes
+            .iter()
+            .any(|dropped_node| node.inner.is_peer_our_elder(dropped_node))
```
madadam: This is probably fine currently, but once promotion/demotion is implemented, we might in theory have a situation where a node that is supposed to be removed is only demoted and this test would still pass. Calling `is_peer_our_member` instead might make the tests more robust.

tests/mock_network/drop.rs
```diff
@@ -25,7 +24,7 @@ fn drop_node(nodes: &mut Vec<TestNode>, index: usize) {
     let _ = poll_all(nodes);
 
     for node in nodes.iter_mut().filter(|n| close_names.contains(&n.name())) {
-        assert!(!node.inner.is_connected(name));
+        assert!(node.inner.is_peer_our_elder(&id));
```
madadam: Shouldn't we assert that the peer is **not** our elder?

tests/mock_network/churn.rs
maqi: yeah, good suggestion, I will change to use `is_peer_our_member`

tests/mock_network/drop.rs
maqi: at this moment, the dropped node is not detected by others as unresponsive, as there is not enough events accumulated. So the node still remained as elder from other peers' perspective.

madadam: But then this check is useless because it would pass even if we didn't drop the node.

maqi: well, I think the purpose of the test is to confirm the disconnection won't cause an immediate state change?
I agree it will pass if we don't drop a node. However, I think it could expose an issue if we changed the disconnection handling logic to `act immediately` ?

madadam: I think we might need to rethink this test. Maybe do something like:

1. Create a network with a couple of nodes
2. Drop one node
3. Generate some traffic (user events) to let the other nodes detect the dropped node being unresponsive
4. Verify all the existing nodes detected the node being offline.

So basically the same think you do in the `verify_dropped_nodes` function in the churn tests. So maybe just extract that function into `mock_network/utils.rs` and then just call it here?

maqi: ok. will do that

maqi: ok. done.
However, this place is just a helper function, not a test.
so put the `verify_dropped_nodes` in the testing part.

tests/mock_network/drop.rs
```diff
@@ -41,35 +43,12 @@ fn node_drops() {
...
     );
     let mut nodes = create_connected_nodes(&network, elder_size + 2);
-    drop_node(&mut nodes, 0);
 
-    // Trigger poll_and_resend to allow remaining nodes to gossip and
-    // update their chain accordingly.
-    poll_and_resend(&mut nodes);
-    verify_invariant_for_all_nodes(&network, &mut nodes);
-}
+    let dropped_node = drop_node(&mut nodes, 0);
```
madadam: I think the `drop_node` function is now close to useless. What about removing it and doing just something like:
```rust
let dropped_nodes = iter::once(nodes.remove(0).id()).collect(); // collect the id of the removed node into a `BTreeSet`.
verify_dropped_nodes(&mut rng, &mut nodes, &dropped_nodes);
```

maqi: I still think we shall keep that to confirm the disconnection won't cause status change immediately.

madadam: Hm, OK. But maybe add a comment explaining why we are doing it? Something like we expect the node to remain elder until detected as unresponsive, or something like that.

maqi: pushed updated comment

src/states/elder/mod.rs
```diff
@@ -1737,6 +1689,40 @@ impl Approved for Elder {
...
         Ok(())
     }
+
+    // Checking members vote status and vote to remove those non-resposive nodes.
+    #[cfg(feature = "mock_parsec")]
+    fn check_voting_status(&mut self) {
+        let unresponsive_nodes = self.chain.check_vote_status();
+        for pub_id in unresponsive_nodes.iter() {
+            self.vote_for_event(AccumulatingEvent::Offline(*pub_id));
+        }
```
madadam: If dropped the `.iter()` you wouldn't need to dereference the `pub_id`.

src/states/elder/mod.rs
```diff
@@ -1737,6 +1689,40 @@ impl Approved for Elder {
 
         Ok(())
     }
+
+    // Checking members vote status and vote to remove those non-resposive nodes.
+    #[cfg(feature = "mock_parsec")]
+    fn check_voting_status(&mut self) {
```
madadam: For consistency, I'd either rename this to `check_vote_status` or rename `Chain::check_vote_status` to `check_voting_status`.

src/states/elder/mod.rs
```diff
@@ -1737,6 +1689,40 @@ impl Approved for Elder {
...
+            if unresponsive_nodes.contains(&id) {
+                let new_count = count.saturating_add(1);
+                info!(
+                    "{} detected {:?} as unresponsive for {} times when having {} members",
+                    self,
+                    id,
+                    new_count,
+                    self.chain.our_info().len(),
+                );
+                if new_count > self.chain.our_info().len() {
```
madadam: I'm not clear why we need the node to be accused more times than there is elders in the section. Maybe add a comment explaining that?

src/states/elder/mod.rs
```diff
@@ -1737,6 +1689,40 @@ impl Approved for Elder {
...
+    // Checking members vote status and vote to remove those non-resposive nodes.
+    #[cfg(feature = "mock_parsec")]
+    fn check_voting_status(&mut self) {
+        let unresponsive_nodes = self.chain.check_vote_status();
+        for pub_id in unresponsive_nodes.iter() {
+            self.vote_for_event(AccumulatingEvent::Offline(*pub_id));
+        }
+    }
+
+    #[cfg(not(feature = "mock_parsec"))]
```
madadam: Why can't we use this impl for `feature = "mock_parsec"` too?

maqi: When using mock_parsec, the polling will give all peers' votes in one polling loop.
Which means there will be less extra polling that will make the unresponsive nodes to be accused for enough times.

src/states/elder/mod.rs
maqi: The average chance that the missing peer will pick usas gossip recipient, so that we can know its votes (or we know its votes via others) is 1/number_of_peers.
So requiring `the node to be accused more times than there is elders in the section` is mainly just to ensure the full knowledge is reached before making the accuse.

madadam: OK, could you add it as comment there?

src/states/elder/mod.rs
madadam: So is this just an optimization to make `mock_parsec` tests faster?

src/states/elder/mod.rs
maqi: ok, will do

src/states/elder/mod.rs
maqi: no. more about avoiding the edge case that a disconnected node not being accused, due to our polling stopped too quick.

madadam: Can that happen though? The polling will only stop when there are no unpolled observations. So after the polling stops, everyone will have seen all the injected user-data votes and they will also see that the dropped node voted for none of them. So they should always accuse it. Or am I missing something?

maqi: if want to use the same code as to the mock_base, the diff of mock_parsec here is that all observations comes in within one polling loop. which means there is no `unpolled observations` left over and the polling could be stopped soon.
Meanwhile we are still expecting the dropped node to be accused for couple of more times.

madadam: Parsec poll happens only when receiving gossip. So right after you inject all those fake votes, `has_unpolled_observations` will return `true` for everyone, because everyone put some observations into their parsecs, but no one polled them yet. So when `poll_and_resend` is called, it will do at least one iteration, so at least some nodes will receive gossips. Because of how mock parsec is implemented, that would cause those nodes to reach consensus on all the votes (this would not be the case in real parsec). So that means they will also see that the dropped node did not vote and will accuse it. There might be some nodes that did not receive any gossip, but that means they did not poll their parsecs so they will still return `true` from `has_unpolled_observations`, which will trigger another iteration of polling. This will go on until everyone received at least one gossip which mean they also reached consensus on all the votes so they will all accuse the dropped node.

So I still don't see how in mock_parsec we can miss accusing the node.

maqi: The issue only happens for mock_base, not for mock_parsec.
mock_parsec is totally fine, that's why its function doesn't need extra logic handling (looking for extra accusations).

However, if want to use the same code as to mock_parsec, then the problem comes, because there probably won't be enough extra accusations.

madadam: >  there probably won't be enough extra accusations

Ah OK, I think I see the problem now.

madadam: So my reasoning for starting this thread is that I think it's worth to avoid conditional compilation as much as possible, because it can cause issues like tests passing with mock parsec but failing with real parsec (or vice-versa). I'm OK with doing it this way though, if there is no other way of it would be too much hassle. I have alternative idea though: in `verify_dropped_nodes`, we can use the `continue_if` predicate (by using `poll_and_resend_with_options`) to poll until the nodes are voted offline by everyone. A downside to that is that the tests might be a bit slower. 

Not sure what's best. What do the other reviewers think?

maqi: using `poll_and_resend_with_options` won't help here for mock_base, as the issue is there is valid node being accused.  :(

>  it can cause issues like tests passing with mock parsec but failing with real parsec (or vice-versa)

Well, the mock_parsec and real_parsec already has different behaviour here (one gives everything within one poll loop, the other gives things gradually). Hence we have to have different code to handle diff situation.  :(

madadam: Valid node being accused is solved by the multiple accusation thing you have in this PR. All I'm saying is to use that same logic for `mock_parsec` too, where the `poll_and_resend_with_options` could be used to prevent the polling to stop to early.

But you are right, there logic between real parsec and mock parsec is already different enough, so having a single implementation for this wouldn't achieve much. So I think I'm OK with this as long as the other reviewers are. @jeanphilippeD , @octol - what do you think guys?

maqi: could be a good idea to use `poll_and_resend_with_options `, will have a try

octol: Tricky one

> I have alternative idea though: in verify_dropped_nodes, we can use the continue_if predicate > (by using poll_and_resend_with_options) to poll until the nodes are voted offline by everyone. A > downside to that is that the tests might be a bit slower.

If it was my choice and it was feasible I'd opt for slightly slower and avoid the conditional compilation. If it works that is...

maqi: @madadam , tried with `poll_and_resend_with_options `, however it doesn't work. 
As `approved::parsec_poll` only got called when there is a parsec gossip, which only created when gossip_timeout got ticked. However, the advancing of clock only happens when `poll_all` or `is_busy` returns true within `poll_and_resend_with_options `. Unfortunately, they are both `false` in this case. Which means no matter how many times polled, there is no clock advancing , hence no gossip created.

I am not sure if it worth to change the behaviour of `poll_and_resend_with_options ` ?

madadam: Hm, interesting and good catch. 

> I am not sure if it worth to change the behaviour of poll_and_resend_with_options 

I think it is worth it, because without it it's not as useful. But doesn't have to be in this PR. Up to you.

maqi: changing the behaviour fails other test :(
I'd prefer leave it in a separate PR

madadam: OK, then let's leave it as is.

src/states/elder/mod.rs
```diff
@@ -1706,6 +1658,48 @@ impl Approved for Elder {
...
+            self.vote_for_event(AccumulatingEvent::Offline(pub_id));
+        }
+    }
+
+    #[cfg(not(feature = "mock_parsec"))]
+    fn check_vote_status(&mut self) {
+        let unresponsive_nodes = self.chain.check_vote_status();
+        for (id, count) in mem::replace(&mut self.unresponsive_nodes, Default::default()) {
+            if unresponsive_nodes.contains(&id) {
+                let new_count = count.saturating_add(1);
```
jeanphilippeD: So I can see how that works with some bulk group of vote happening, some nodes are behind and detected as unresponsive, and as time of, all there votes come in and they are no longer unresponsive.

What is not clear to me is how that work with continuous voting happening.
The first time, a slower node votes may not have been there yet, counted as unresponsive
The second time, their vote for the first batch has come, but the window as moved, and their vote for the second batch are not there yet, still counted as unresponsive.
The third time, window continue to move, and so there older votes coming still not make them look responsive.
etc.. eventually they are voted offline for the reason they are a bit behind (not much just enough for a sufficient length of time).

I'm thinking we can write a test for this that would fail for that code, and also represent more accurately the behaviour of the network (i.e continuous events voted).

maqi: > eventually they are voted offline for the reason they are a bit behind (not much just enough for a sufficient length of time).

It means it needs to be lagged for THRESHOLD number of consensused events behind. Which shall not happend, as its vote shall come in within Section_Member_Size parsec polls normally (same reasoning as at https://github.com/maidsafe/routing/pull/1880#discussion_r349081366 )

> I'm thinking we can write a test for this that would fail for that code

Not sure about this. The test sounds more likely that node to be `muted` actually  :(

jeanphilippeD: I do not think that is reasonable to say the node shall not lag by `UNRESPONSIVE_THRESHOLD` or even `UNRESPONSIVE_WINDOW` would not happen:

 example:

```
Network: 200 events per second require voting by all nodes.
All voters:      Vote 200 events -1sec> Vote 200 events -1sec> Vote 200 events -1sec> Vote 200 events -1sec> ...
Lagging by 2 second:     Nothing -1sec> Nothing         -1sec> Vote 200 events -1sec> Vote 200 events -1sec> ...
Consensus:
 Nothing
 -1sec> 200 events consensused for all voters, but lagging node: Lagging node unresponsive.
 -1sec> 200 events consensused for all voters, but lagging node: Lagging node unresponsive.
 -1sec> 200 events consensused for all voters, lagging node votes for first 200 events, all these are no longer in the window: Lagging node unresponsive.
 -1sec> 200 events consensused for all voters, lagging node votes for second 200 events, all these are no longer in the window: Lagging node unresponsive.
 -1sec> 200 events consensused for all voters, lagging node votes for third 200 events, all these are no longer in the window: Lagging node unresponsive.
...
 -> Lagging node voted offline.
```

madadam: I think if we expect 200 (or whatever) events per one gossip, then we would need to adjust the windows and threshold accordingly. Also, if a node lags that much then perhaps it's OK to consider it as unresponsive?

madadam: Alternatively, the window can be measured not in events but something else. Number of gossips? Or even actual time (seconds)? That could come with its own set of issues though...

jeanphilippeD: Well not handling lagging node may be a problem here if we want to use that mechanism to support upgrade where the node could possibly drop out for a bit, so is 2 second lagging too much?

Moreover, that 2 second is with the gossip event every second, something we would likely want to decrease to decrease parsec latency, as a result a much smaller lag may be enough. 

Maybe a fixed window is problematic if what we care about is that they voted for some event.
If what we are looking for is did they vote for enough event during the multiple gossip event we got, then maybe that window should reflect that.

But there are some things for consideration:
 - I am not clear that the way we are calling `self.check_vote_status();` make it shared state (i.e I think different nodes will call it after different event, I think it should be shared state if called from within the `while let Some(block) = self.parsec_map_mut().poll() {` block, I think, but then it is also not related to a gossip event being received).
 - (There may be consideration of malicious attacks by generating gossip, but that does not seem to be the case).
 - (a secondary consideration would be a window growing too big be problematic in term of performance.)


I'm sorry I'm not really proposing solutions here, I would like to, but I have yet to see a way that seem consistent with a flow of events being consensused. I think what you are saying @madadam may be correct in that maybe the window should be adjusted. But if the window is sufficient so we can handle these, then what is the purpose of the second windowing happening where we validate over multiple gossip event?

maqi: > Maybe a fixed window is problematic

could be. However to make all members using the same window size is another shared_state need to be exchanged. :(

>  called from within the `while let Some(block) = self.parsec_map_mut().poll() {` block

Previously it was. However it incur a performance regression. Maybe what we can do is: remain at the present calling place but with an extra condition that there shall be something polled from parsec this time.

> But if the window is sufficient so we can handle these, then what is the purpose of the second windowing happening where we validate over multiple gossip event

This is mainly due to the testing pattern: number of injected user-data depends on the window size. unless we change it to what Adam suggested that injecting user-data gradually.

jeanphilippeD: > Maybe what we can do is: remain at the present calling place but with an extra condition that there shall be something polled from parsec this time.

I am not clear how that solve the problem highlighted.

maqi: Same as to calling inside I think. just reduce multiple calls into one time call.

jeanphilippeD: With gossip
```
A10 <- B9
 |      |    
 |     B8
 |      |
A9 ->- B7 
``
B may see a block for A9, loop exit, a block for B7, loop exit, block for B8, loop exit, Block for B9, loop exit.
A may see a block for A9, loop exit, a block for B7, then B8, then B9, then A9, then A10 loop exit.

The reason is when getting gossip creating A10, Alice got all the other sync event that B saw one at a time.

File: src/states/adult.rs [hunk 12]
```diff
@@ -589,6 +589,10 @@ impl Approved for Adult {
         Ok(())
     }
 
+    fn check_vote_status(&mut self) {
+        debug!("{} - Not Required CheckVotingStatus For Adult", self);
```
jeanphilippeD: Why? won't it have a different view than the rest of the nodes when reaching elder status.

maqi: `check_vote_status` only incurs `casting vote of offline`, won't change anything in the `status tracker`.
The `tracker` is affected by add_expectation and add_vote, which is part of the `poll_event` call. Which allows the the adult having the same view as elders'.

maqi:
PR becomes obseleted and the function was already covered by recent refactoring work.

PR #1865, opened 2019-10-24
WIP: Remove ConnectionRequest/Response pattern and PeerMap

octol:
Work in progress / draft.

Removes the `ConnectionRequest/Response` pattern and replaces it by bundling `ConnectionInfo` with `PublicId` in a new `P2pNode` struct. The idea is that this will also allow us to get rid of `PeerMap`.

Update: split off into #1868, #1884, #1885 . Disconnects are handled in #1880.

### TODO
- [x] Add `P2pNode` struct and pass it around
- [x] Remove `PeerMap` (or radically simplify)
- [ ] Handle disconnects
- [x] Remove `ConnectionRequest/Response`
- [ ] Address all WIP comments

Part of #1855

PR #1884, opened 2019-11-06
Remove ConnectionRequest/Response - Part 2

octol:
This is the second part of #1865 and removes most calls to `PeerMap`.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -663,6 +663,28 @@ impl Chain {
             .map(|member_info| member_info.p2p_node.connection_info())
     }
 
+    /// Returns a section member `P2pNode`
+    pub fn get_member_p2p_node(&self, name: &XorName) -> Option<&P2pNode> {
+        self.state
+            .our_members
+            .get(name)
+            .map(|member_info| &member_info.p2p_node)
+    }
+
+    /// Returns a neighbour `P2pNode`
+    pub fn get_neighbour_p2p_node(&self, name: &XorName) -> Option<&P2pNode> {
+        self.state
+            .neighbour_infos
+            .iter()
+            .find(|(pfx, _)| pfx.matches(name))
+            .and_then(|(_, elders_info)| elders_info.member_map().get(name))
+    }
+
+    pub fn get_p2p_node(&self, name: &XorName) -> Option<&P2pNode> {
+        self.get_member_p2p_node(name)
+            .or_else(|| self.get_neighbour_p2p_node(name))
+    }
+
     /// Returns a set of elders we should be connected to.
     pub fn elders(&self) -> impl Iterator<Item = &P2pNode> {
         self.neighbour_infos()
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -671,6 +671,14 @@ impl Chain {
             .map(|member_info| &member_info.p2p_node)
     }
 
+    /// Returns an old section member `P2pNode`
+    fn get_post_split_sibling_member_p2p_node(&self, name: &XorName) -> Option<&P2pNode> {
+        self.state
+            .post_split_sibling_members
+            .get(name)
+            .map(|member_info| &member_info.p2p_node)
+    }
+
     pub fn get_our_info_p2p_node(&self, name: &XorName) -> Option<&P2pNode> {
         self.state.our_info().member_map().get(name)
     }
@@ -688,6 +696,7 @@ impl Chain {
         self.get_member_p2p_node(name)
             .or_else(|| self.get_our_info_p2p_node(name))
             .or_else(|| self.get_neighbour_p2p_node(name))
+            .or_else(|| self.get_post_split_sibling_member_p2p_node(name))
     }
 
     /// Returns a set of elders we should be connected to.
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -143,8 +143,8 @@ impl Chain {
     fn get_age_counters(&self) -> BTreeMap<PublicId, AgeCounter> {
         self.state
             .our_members
-            .iter()
-            .map(|(pub_id, member_info)| (*pub_id, member_info.age_counter))
+            .values()
+            .map(|member_info| (*member_info.p2p_node.public_id(), member_info.age_counter))
             .collect()
     }
 
@@ -331,8 +331,8 @@ impl Chain {
         let our_prefix = *self.state.our_prefix();
         let mut details_to_add = Vec::new();
 
-        for (pub_id, member_info) in self.state.our_joined_members_mut() {
-            if pub_id == trigger_node {
+        for (name, member_info) in self.state.our_joined_members_mut() {
+            if member_info.p2p_node.public_id() == trigger_node {
                 continue;
             }
 
@@ -340,18 +340,15 @@ impl Chain {
                 continue;
             }
 
-            let destination = compute_relocation_destination(
-                pub_id.name(),
-                trigger_node.name(),
-                &mut self.dev_params,
-            );
+            let destination =
+                compute_relocation_destination(name, trigger_node.name(), &mut self.dev_params);
             if our_prefix.matches(&destination) {
                 // Relocation destination inside the current section - ignoring.
                 continue;
             }
 
             details_to_add.push(RelocateDetails {
-                pub_id: *pub_id,
+                pub_id: *member_info.p2p_node.public_id(),
                 destination,
                 age: member_info.age() + 1,
             })
@@ -427,9 +424,7 @@ impl Chain {
     pub fn add_member(&mut self, p2p_node: P2pNode, age: u8) {
         self.assert_no_prefix_change("add member");
 
-        let pub_id = *p2p_node.public_id();
-
-        match self.state.our_members.entry(pub_id) {
+        match self.state.our_members.entry(*p2p_node.name()) {
             Entry::Occupied(mut entry) => {
                 if entry.get().state == MemberState::Left {
                     // Node rejoining
@@ -443,14 +438,14 @@ impl Chain {
                         LogLevel::Error,
                         "{} - Adding member that already exists: {}",
                         self,
-                        pub_id
+                        p2p_node,
                     );
                     return;
                 }
             }
             Entry::Vacant(entry) => {
                 // Node joining for the first time.
-                let _ = entry.insert(MemberInfo::new(age, p2p_node.connection_info().clone()));
+                let _ = entry.insert(MemberInfo::new(age, p2p_node.clone()));
             }
         }
 
@@ -466,7 +461,7 @@ impl Chain {
         if let Some(info) = self
             .state
             .our_members
-            .get_mut(&pub_id)
+            .get_mut(pub_id.name())
             .filter(|info| info.state == MemberState::Joined)
         {
             info.state = MemberState::Left;
@@ -655,7 +650,7 @@ impl Chain {
     pub fn is_peer_our_member(&self, pub_id: &PublicId) -> bool {
         self.state
             .our_members
-            .get(pub_id)
+            .get(pub_id.name())
             .map(|info| info.state == MemberState::Joined)
             .unwrap_or(false)
     }
@@ -664,8 +659,8 @@ impl Chain {
     pub fn get_member_connection_info(&self, pub_id: &PublicId) -> Option<&ConnectionInfo> {
         self.state
             .our_members
-            .get(&pub_id)
-            .map(|member_info| &member_info.connection_info)
+            .get(pub_id.name())
+            .map(|member_info| member_info.p2p_node.connection_info())
     }
 
     /// Returns a set of elders we should be connected to.
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -1187,25 +1187,6 @@ impl Chain {
             .collect_vec()
     }
 
-    /// Returns whether the table contains the given `name`.
-    fn has(&self, name: &XorName) -> bool {
-        self.get_section_legacy(name)
-            .map_or(false, |section| section.contains(name))
-    }
-
-    /// Returns the section matching the given `name`, if present.
-    /// Includes our own name in the case that our prefix matches `name`.
-    fn get_section_legacy(&self, name: &XorName) -> Option<BTreeSet<XorName>> {
-        if self.our_prefix().matches(name) {
-            return Some(self.our_info().member_names().copied().collect());
-        }
-        self.state
-            .neighbour_infos
-            .iter()
-            .find(|&(ref pfx, _)| pfx.matches(name))
-            .map(|(_, ref info)| info.member_names().copied().collect())
-    }
-
     /// Returns the `count` closest entries to `name` in the routing table, including our own name,
     /// sorted by ascending distance to `name`. If we are not close, returns `None`.
     pub fn closest_names(
@@ -1280,15 +1261,7 @@ impl Chain {
     ///     - if our name *is* the destination, returns an empty set; otherwise
     ///     - if the destination name is an entry in the routing table, returns it; otherwise
     ///     - returns the `N/3` closest members of the RT to the target
-    pub fn targets(
-        &self,
-        dst: &Authority<XorName>,
-        connected_peers: &[&XorName],
-    ) -> Result<(Vec<XorName>, usize), Error> {
-        // FIXME: only filtering for now to match RT.
-        // should confirm if needed esp after msg_relay changes.
-        let is_connected = |target_name: &XorName| connected_peers.contains(&target_name);
-
+    pub fn targets(&self, dst: &Authority<XorName>) -> Result<(Vec<&P2pNode>, usize), Error> {
         let candidates = |target_name: &XorName| {
             let filtered_sections =
                 self.closest_sections_info(*target_name)
@@ -1297,7 +1270,7 @@ impl Chain {
                         (
                             prefix,
                             members.len(),
-                            members.member_names().collect::<Vec<_>>(),
+                            members.member_nodes().collect::<Vec<_>>(),
                         )
                     });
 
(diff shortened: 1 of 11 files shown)


src/states/elder/mod.rs
```diff
@@ -950,29 +954,42 @@ impl Elder {
...
+            // WIP: neet to get targets without using the peer_map (get_targets uses peer_map
+            // internally)
+            let (targets, dg_size) = self.get_targets(signed_msg.routing_message())?;
+            (
+                targets
+                    .into_iter()
+                    .filter_map(|public_id| {
+                        self.peer_map
+                            .get_connection_info(&public_id)
+                            .map(|conn_info| P2pNode::new(public_id, conn_info.clone()))
```
jeanphilippeD: A bit strange too

src/chain/chain.rs
```diff
@@ -628,6 +628,49 @@ impl Chain {
...
+            .find(|(pub_id, _)| pub_id.name() == name)
+            .map(|(_, member_info)| member_info.connection_info.clone())
+    }
+
+    /// Returns the `P2pNode` for a member of our section.
+    pub fn get_member_p2p_node_by_id(&self, pub_id: &PublicId) -> Option<P2pNode> {
+        self.state
+            .our_members
+            .get(&pub_id)
+            .map(|member_info| P2pNode::new(*pub_id, member_info.connection_info.clone()))
```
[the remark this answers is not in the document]
octol: No `our_members` stores a `MemberInfo`

src/chain/chain.rs
```diff
@@ -628,6 +628,49 @@ impl Chain {
...
+    }
+
+    /// Returns a section member `P2pNode`
+    // WIP: by changing to Map<XorName, P2pNode> we could make this faster
+    pub fn get_member_p2p_node(&self, name: &XorName) -> Option<P2pNode> {
+        self.state
+            .our_members
+            .iter()
+            .find(|(pub_id, _)| pub_id.name() == name)
+            .map(|(pub_id, member_info)| P2pNode::new(*pub_id, member_info.connection_info.clone()))
```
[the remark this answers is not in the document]
octol: See previous reply

src/chain/chain.rs
```diff
@@ -628,6 +628,49 @@ impl Chain {
...
+    }
+
+    /// Returns a neighbour `P2pNode`
+    // WIP: by changing to Map<XorName, P2pNode> we could make this faster
```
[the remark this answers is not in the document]
octol: Agree. Change done in previous PR

src/chain/chain.rs
```diff
@@ -628,6 +628,49 @@ impl Chain {
...
+    pub fn get_neighbour_p2p_node(&self, name: &XorName) -> Option<P2pNode> {
+        self.neighbour_infos()
+            .flat_map(|elders_info| elders_info.member_map().iter())
+            .find(|(pub_id, _)| pub_id.name() == name)
+            .map(|(_, p2p_node)| p2p_node.clone())
```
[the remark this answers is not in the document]
octol: Agree. With the changes in the preceding PR this will now be replaced by a lookup

src/states/common/approved.rs
```diff
@@ -139,19 +143,23 @@ pub trait Approved: Base {
...
                 if recipients.is_empty() {
                     // Parsec hasn't caught up with the event of us joining yet.
                     return;
                 }
 
-                recipients.retain(|pub_id| self.peer_map().has(pub_id));
-                if recipients.is_empty() {
+                let p2p_recipients: Vec<_> = recipients
+                    .into_iter()
+                    .filter_map(|pub_id| self.chain().get_member_p2p_node_by_id(&pub_id))
```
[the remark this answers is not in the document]
octol: Do you mean we could want to send parsec gossip to nodes that are not in our or our neighbour section infos?

src/chain/chain.rs
```diff
@@ -628,6 +628,49 @@ impl Chain {
             .map(|member_info| &member_info.connection_info)
     }
 
+    /// Returns the `ConnectionInfo` for a `XorName` in our section.
+    // WIP: this is probably unnecessary slow
+    #[cfg(feature = "mock_base")]
+    pub fn get_member_name_connection_info(&self, name: &XorName) -> Option<ConnectionInfo> {
+        self.state
```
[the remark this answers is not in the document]
octol: Agree, I'll update the map to take a `XorName` as key and `MembersInfo` to also contain a `P2pNode` instead of a `ConnectionInfo`. That way we can to straight lookups and return references

src/chain/chain.rs
```diff
@@ -628,6 +628,49 @@ impl Chain {
...
+            .our_members
+            .iter()
+            .find(|(pub_id, _)| pub_id.name() == name)
+            .map(|(_, member_info)| member_info.connection_info.clone())
```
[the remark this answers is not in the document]
octol: Agree, see above

src/chain/chain.rs
```diff
@@ -628,6 +628,49 @@ impl Chain {
...
+        self.state
+            .our_members
+            .iter()
+            .find(|(pub_id, _)| pub_id.name() == name)
+            .map(|(pub_id, member_info)| P2pNode::new(*pub_id, member_info.connection_info.clone()))
+    }
+
+    /// Returns a neighbour `P2pNode`
+    // WIP: by changing to Map<XorName, P2pNode> we could make this faster
+    pub fn get_neighbour_p2p_node(&self, name: &XorName) -> Option<P2pNode> {
```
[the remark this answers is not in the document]
octol: Yep agree

src/states/common/approved.rs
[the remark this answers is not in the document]
jeanphilippeD: We only send parsec gossip between elders of the same section, or member of our own Section that are about to become elder.

I think the strangeness was coming from getting all this info and then throwing it away.
Not sure, It still seem the most sensible way of expressing it.

src/states/elder/mod.rs
```diff
@@ -965,29 +969,42 @@ impl Elder {
...
-        let (target_pub_ids, dg_size) = if let Some(target) = single_target {
-            (vec![*target], 1)
+        let (target_p2p_nodes, dg_size) = if let Some(target) = single_target {
+            (vec![target.clone()], 1)
         } else {
-            self.get_targets(signed_msg.routing_message())?
+            // WIP: neet to get targets without using the peer_map (get_targets uses peer_map
+            // internally)
+            let (targets, dg_size) = self.get_targets(signed_msg.routing_message())?;
+            (
```
jeanphilippeD: I'm not sure about that WIP.
It seems get_tragets use peer_map to find connected, also we are still using get_targets, so not clear what we are doing here.

Question, why would we not get rid of is_connected from within get_targets + chain::targets?
All nodes that can be targets are elders now, so it seem we find them in our or our neighbour'elder then we don't have to check for is_connected.

Additionally, you can also replace the return type from PublicId to P2pNodes. (though you could keep the return type as is in this PR, and use self.chain.get_p2p_node. so we do not use peer_map here.

octol: Yes that's right, we do so in part 3. The reason that commit is in part 3 and not part 2 is that I've not yet investigated the test failures I get from it

jeanphilippeD: Ok, that seem weird I did not expect a test failure here, can you point me to the commit you do it?

src/states/common/approved.rs
```diff
@@ -141,19 +144,27 @@ pub trait Approved: Base {
...
                 }
 
-                recipients.retain(|pub_id| self.peer_map().has(pub_id));
-                if recipients.is_empty() {
+                let p2p_recipients: Vec<_> = recipients
+                    .into_iter()
+                    .filter_map(|pub_id| {
+                        self.peer_map()
+                            .get_connection_info(pub_id)
+                            .map(|conn_info| P2pNode::new(*pub_id, conn_info.clone()))
```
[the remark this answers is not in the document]
octol: Correct, we make this change in part 3, still it causes test failures for the split test I've not yet managed to figure out. By the looks of it we need to handle disconnects for this to work as the only difference I can spot is that we end up sending to recipients that we've lost connection to. It's the `relocate_without_split` that causes problems.

src/states/elder/mod.rs
octol: https://github.com/maidsafe/routing/pull/1885/commits/bf3a29ce4453b3359e91f76dfc52edcd52f4cca6

jeanphilippeD: Thanks, I'll have a look tomorrow.

octol: I had a look at it and with a little tweaking I got the tests to pass

jeanphilippeD: Great :-)

src/states/elder/mod.rs
```diff
@@ -1050,34 +1054,19 @@ impl Elder {
     fn get_targets(
         &self,
         routing_msg: &RoutingMessage,
-    ) -> Result<(Vec<PublicId>, usize), RoutingError> {
-        // TODO: even if having chain reply based on connected_state,
-        // we remove self in targets info and can do same by not
-        // chaining us to conn_peer list here?
-        let conn_peers = self.connected_peers();
+    ) -> Result<(Vec<P2pNode>, usize), RoutingError> {
+        let conn_peers: Vec<_> = self.chain.elders().map(P2pNode::name).collect();
```
jeanphilippeD: Do we still need conn_peers? can we remove it from targets?

src/states/elder/mod.rs
```diff
@@ -1050,34 +1054,19 @@ impl Elder {
...
         let (targets, dg_size) = self.chain.targets(&routing_msg.dst, &conn_peers)?;
         Ok((
             targets
                 .into_iter()
-                .filter_map(|name| self.peer_map.get_id(&name))
-                .copied()
+                .filter_map(|name| self.chain.get_p2p_node(&name))
```
jeanphilippeD: Would it make sense for chain.targets to return P2pNodes? (Might event be possible to return `&P2pNode`

src/states/elder/mod.rs
jeanphilippeD: Also I see you dedup the other ones, would duplicate cause issue here?

src/chain/chain.rs
```diff
@@ -695,24 +711,38 @@ impl Chain {
...
+    }
+
+    /// Returns the elders for a neighbour section.
+    /// Returns None if the `Prefix` provided wasn't our own section or a neigbour.
+    pub fn get_section_elders(
+        &self,
+        names: &Prefix<XorName>,
+    ) -> Option<&BTreeMap<XorName, P2pNode>> {
+        if self.our_prefix() == names {
+            Some(self.our_info().member_map())
```
octol: Is this correct? Should it be from `our_info` or new_info`?

jeanphilippeD: our_info seem to make more sense.
new_info was essentially a way to handle remove and add concurrently, it should probably be removed, hopefully sometime soon.

src/states/adult.rs
```diff
@@ -394,16 +392,23 @@ impl Base for Adult {
 
         // We should only be connected to our own Elders - send to all of them
         // Need to collect IDs first so that self is not borrowed via the iterator
-        let target_ids: Vec<_> = self.peer_map.connected_ids().cloned().collect();
+        //
+        // WIP: this is probably out of date? How else do we know which our section members are?
+        let target_nodes = self
```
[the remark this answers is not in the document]
madadam: > Question, which routing messages are we sending as an adult? (I think currently we backlog them all).

It seems only `ConnectionRequest`, but with these changes we might not need to anymore? If not, we can probably remove this whole function from here and from `Base` and keep it only in `Elder`.

src/states/common/approved.rs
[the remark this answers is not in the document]
madadam: Alternatively, we can use `P2pNode` as `parsec::PublicId`, but we would have to come up with the corresponding `SecretId` somehow.

src/states/common/base.rs
```diff
@@ -512,24 +512,18 @@ pub fn from_network_bytes(data: NetworkBytes) -> Result<Message, RoutingError> {
...
-impl MessageRecipient for PublicId {
-    fn resolve<'a>(&'a self, peer_map: &'a PeerMap) -> Option<&'a ConnectionInfo> {
-        peer_map.get_connection_info(self)
-    }
-}
-
-impl MessageRecipient for XorName {
-    fn resolve<'a>(&'a self, peer_map: &'a PeerMap) -> Option<&'a ConnectionInfo> {
-        peer_map.get_connection_info(self)
+impl MessageRecipient for ConnectionInfo {
```
madadam: I think we can probably remove this trait and go back to sending messages via `ConnectionInfo` only. Unless we want to keep the convenience of being able to send using both with `ConnectionInfo` and `P2pNode` which I'm not sure is worth the complexity.

src/states/joining_peer.rs
```diff
@@ -145,8 +145,9 @@ impl JoiningPeer {
                 src: Authority::Node(_),
                 dst: Authority::Node(_),
             } => {
-                self.peer_map_mut().insert(pub_id, conn_info);
-                self.send_direct_message(&pub_id, DirectMessage::ConnectionResponse);
+                self.peer_map_mut().insert(pub_id, conn_info.clone());
+                let p2p_node = P2pNode::new(pub_id, conn_info);
+                self.send_direct_message(&p2p_node, DirectMessage::ConnectionResponse);
```
madadam: Do you need to construct the `P2pNode`? `send_direct_message` should work with raw `ConnetionInfo` too.

src/chain/chain.rs
```diff
@@ -691,6 +691,7 @@ impl Chain {
     }
 
     /// Returns a set of elders we should be connected to.
+    // WIP: should we remove potential duplicates?
```
madadam: We should also not include `new_info.member_nodes` here, but that can maybe be done in other PR.

src/states/common/base.rs
octol: Good point. As these things are further simplified in Part 3 PR I'll see if it makes sense to just remote it in that PR

src/states/joining_peer.rs
octol: Good catch, I'll just pass the connection info

src/chain/chain.rs
octol: Yep I agree. Ok if I add that change to the Part 3 PR so as to not invalidate the soak testing?

madadam: Sure. It might event be done as a completely separate PR. Up to you.

octol: Thanks. In that case I'll raise a separate PR I think :)

src/chain/shared_state.rs
```diff
@@ -40,7 +40,9 @@ pub struct SharedState {
     /// This is not a `BTreeSet` as it is ordered according to the sequence of pushes into it.
     pub our_infos: NonEmptyList<EldersInfo>,
     /// Info about all members of our section - elders, adults and infants.
-    pub our_members: BTreeMap<PublicId, MemberInfo>,
+    pub our_members: BTreeMap<XorName, MemberInfo>,
+    /// members that we had before last split.
+    pub our_old_members: BTreeMap<XorName, MemberInfo>,
```
madadam: Maybe better call this `sibling_members` ?

src/states/adult.rs
```diff
@@ -416,16 +415,23 @@ impl Base for Adult {
 
         // We should only be connected to our own Elders - send to all of them
         // Need to collect IDs first so that self is not borrowed via the iterator
-        let target_ids: Vec<_> = self.peer_map.connected_ids().cloned().collect();
+        //
+        // WIP: this is probably out of date? How else do we know which our section members are?
```
madadam: Is this still WIP?

src/states/common/approved.rs
```diff
@@ -400,8 +415,10 @@ pub trait Approved: Base {
             self, their_pub_id
         );
 
-        self.peer_map_mut().insert(their_pub_id, their_conn_info);
-        self.send_direct_message(&their_pub_id, DirectMessage::ConnectionResponse);
+        self.peer_map_mut()
+            .insert(their_pub_id, their_conn_info.clone());
+        let their_p2p_node = P2pNode::new(their_pub_id, their_conn_info);
+        self.send_direct_message(&their_p2p_node, DirectMessage::ConnectionResponse);
```
madadam: Again, no need to create the `P2pNode`. Just pass the `ConnectionInfo` directly.

src/states/adult.rs
octol: Not for this PR, WIP is for Part 3 and I'll try just removing the whole function as per your other comment

src/states/common/approved.rs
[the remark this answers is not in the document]
jeanphilippeD: @octol should we addres that comment?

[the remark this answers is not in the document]
jeanphilippeD: Ok, lets leave it for part 3.

src/chain/shared_state.rs
octol: I'll update to `pre_split_sibling_members`

madadam: Well, technically they are `post_split` siblings  as they weren't siblings before the split. Naming is hard :)

madadam: So I think just `sibling_members` should be enough. Don't think it's ambiguous as that term is used only when referring to splits.

jeanphilippeD: They are only the sibling member at the time of the split, after that the actual sibling members will change, and we keep just a snapshot.

src/states/common/approved.rs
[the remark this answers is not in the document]
octol: Part of #1898

src/states/common/base.rs
octol: Removed in https://github.com/maidsafe/routing/pull/1898

src/chain/chain.rs
octol: Removed in https://github.com/maidsafe/routing/pull/1898

PR #1885, opened 2019-11-06
Remove ConnectionRequest/Response - Part 3

octol:
Work in progress.

This is the third part and final part of #1865, and currently depends on disconnection handling #1880.
Immediate follow up from #1914 and #1908.

Removes all traces of `ConnectionRequest`/`Response`.


madadam:
Superseded by #2033 .

PR #1886, opened 2019-11-06
DKG on top of parsec reset

### LINKED ISSUE #1774: start DKG when membership change (1-4 days)
Could be Online -> DKG (DKG complete) -> Add (need to have stored info) -> SectionInfo
   Maybe in case of split: Online -> DKG (DKG complete) -> (No need for Add) SectionInfo
   => vote for DkgStart in `handle_online_event`.
   => Only vote for Add or remove when DkgResult comes back out of parsec as oppose to in `handle_online_event`.
   => Does not cater for split.

jeanphilippeD:
This PR start DKG when Online/Offline is consensused, and trigger the SectionInfo vote when the DKG complete.

This also re-enable mock_base testing with real Parsec, as newly added node get the gossip from the parsec instance they were added from, like when they were added as an elder before.

Unit tests are updated.
multiple_joining_nodes is disabled for mock_base because of timeout better solved after other works.

Closes #1774

### DIFF
--- a/src/mock/parsec/mod.rs
+++ b/src/mock/parsec/mod.rs
@@ -120,6 +120,14 @@ where
         Ok(())
     }
 
+    pub fn vote_for_as(&mut self, observation: Observation<T, S::PublicId>, vote_id: &S) {
+        state::with(self.section_hash, |state| {
+            let holder =
+                ObservationHolder::new(observation, vote_id.public_id(), self.consensus_mode);
+            state.vote(vote_id, holder);
+        });
+    }
+
     pub fn gossip_recipients(&self) -> impl Iterator<Item = &S::PublicId> {
         let iter = if self.peer_list.contains(self.our_id.public_id()) {
             Some(
--- a/src/states/elder/mod.rs
+++ b/src/states/elder/mod.rs
@@ -31,7 +31,7 @@ use crate::{
         SignedRoutingMessage,
     },
     outbox::EventBox,
-    parsec::{self, ParsecMap},
+    parsec::{self, DkgResultWrapper, ParsecMap},
     pause::PausedState,
     peer_map::PeerMap,
     routing_message_filter::RoutingMessageFilter,
@@ -1529,6 +1529,15 @@ impl Approved for Elder {
         info!("{} - handle Offline: {}.", self, pub_id);
         self.remove_member(pub_id, DisconnectTime::Now, outbox)?;
 
+        Ok(())
+    }
+
+    fn handle_dkg_result_event(
+        &mut self,
+        _participants: &BTreeSet<PublicId>,
+        _dkg_result: &DkgResultWrapper,
+    ) -> Result<(), RoutingError> {
+        // TODO
         Ok(())
     }
 
--- a/src/states/elder/mod.rs
+++ b/src/states/elder/mod.rs
@@ -102,6 +102,8 @@ pub struct Elder {
     pfx_is_successfully_polled: bool,
     /// Peers we will disconnect from in the future.
     delayed_disconnects: HashMap<u64, PublicId>,
+    /// DKG cache
+    dkg_cache: BTreeMap<BTreeSet<PublicId>, EldersInfo>,
 }
 
 impl Elder {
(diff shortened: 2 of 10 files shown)


tests/mock_network/mod.rs
```diff
@@ -245,7 +245,12 @@ fn node_joins_in_front() {
     verify_invariant_for_all_nodes(&network, &mut nodes);
 }
 
+// Only run for mock parsec, as with DKG Joining node times waiting for NodeApproval.
+// Elder go on to process it and take it on as an Elder.
+// This would not be an issue if Joining did not time out, or if elder processed them quicker.
```
madadam: There is a way to disable the join timeouts using `poll_and_resend_with_options(nodes, PollOptions::default().fire_join_timeout(false))`. Wouldn't that help?

src/mock/parsec/mod.rs
```diff
@@ -120,6 +120,14 @@ where
         Ok(())
     }
 
+    pub fn vote_for_as(&mut self, observation: Observation<T, S::PublicId>, vote_id: &S) {
```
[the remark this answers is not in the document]
jeanphilippeD: Why would that be a problem?

src/states/elder/mod.rs
```diff
@@ -102,6 +102,8 @@ pub struct Elder {
     pfx_is_successfully_polled: bool,
     /// Peers we will disconnect from in the future.
     delayed_disconnects: HashMap<u64, PublicId>,
+    /// DKG cache
+    dkg_cache: BTreeMap<BTreeSet<PublicId>, EldersInfo>,
```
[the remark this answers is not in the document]
jeanphilippeD: I don't know, it means that we have to have 2 special cases, while here we push that info in the cache as needed and get it from it when ready. Seem simpler?

> Also, shouldn't this be in SharedState?
I don't think so. I don't think we should have a DKG running when we create a new Parsec. (This might be something we need to be careful about in Prune, but it would not seem sensible to initiate pruning if we are already having a DKG on to rest Parsec).

src/mock/parsec/mod.rs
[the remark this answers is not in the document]
madadam: The mock network tests are supposed to simulate real network and you are not supposed to vote as someone else. On the other hand, you also don't have other peoples full ids there, so there is little chance for this to be misused? So probably OK to leave it as is. Up to you,.

tests/mock_network/mod.rs
jeanphilippeD: Not really, what fire_join_timeout do is move the clock forward if nothing more is left to do.
What happens here is, with gossip every 1 second, Parsec takes to many seconds to reach the different stage of consensus (i.e it has to add 5 nodes consensusing DKG + SectionInfo, and that takes too many round of gossip then the join node time out).

src/states/elder/mod.rs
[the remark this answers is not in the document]
madadam: > Seem simpler?

I uses more memory and It breaks the [Single Source of Truth](https://en.wikipedia.org/wiki/Single_source_of_truth) rule, but maybe it's something we can live with for now?

tests/mock_network/mod.rs
jeanphilippeD: Sorry I lost the reply in github.
fire_join_timeout move the fake_clock forward when nothing is left to do.
This is not what happens here, we gossip every 1 second, and parsec takes to many round of gossip to reach consensus on the 5 nodes we are adding for DKG + SectionInfo, while parsec still gossip we pass the Join timeout.

Option includes increasing Join timeout OR decreasing Parsec gossip time OR avoid doing so many DKG/SectionInfo before adding the last adult.

madadam: I see. Isn't this going to be a problem in the real network then? Should we make the join timeout longer?

jeanphilippeD: I think the solution will come from promotion/demotion and "avoid doing so many DKG/SectionInfo before adding the last adult." i.e add all adults in the queue first before promoting any elder.

madadam: Ninja'd :)

PR #1887, opened 2019-11-07
Perf fix experiments

jeanphilippeD:
Here are the things that popped out when looking at profiling and some quick fixes to prove if they made a difference.

The simplest one is:
 - clear NodeApproval serialized data 14s -> 12s (11.5)

The fairly simple and biggest one is:
- Fix is_quorum creating many BTreeSet: 12s -> 8s
(Found using RUSTFLAGS="-C inline-threshold=25 -g" to avoid inlining and get clearer result in profiling).

HashSet made no difference was confused by too much inline.
No check for consensus is not ideal, maybe there are better way to do that.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -23,7 +23,7 @@ use itertools::Itertools;
 use log::LogLevel;
 use std::cmp::Ordering;
 use std::{
-    collections::{BTreeMap, BTreeSet},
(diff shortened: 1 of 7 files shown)


jeanphilippeD:
Most fixes are now merged and the not checking for genesis info is more of an information than something we want to used.

Closing now.

PR #1888, opened 2019-11-07
fix/elder: handle the case of bulk parsec poll

maqi:
When executing tests, parsec polls observations in a bulk. This
will cause issue to the unresponsive track, making valid peer to
be considered as unresponsive.
This commit handle the case by having extra conditions of carrying
out vote status check, when a bulk of parsec poll happens.


maqi:
closed as the approach shall be replaced by a more proper one.

PR #1890, opened 2019-11-12
Access to the connection info of elders

### LINKED ISSUE #1889: Expose API to gather `ConnectionInfo`s for known elders in a section
`routing::Node` should expose public API to gather the connection information for all known elders within a section. This is required for supporting RPC in Vaults to send the section conn. info to bootstrapped clients.

ustulation:
Expose connection information of the elders in our section to the
user library. Vauls for example will use this to give it to a client
wanting to connect to the network (and thus via its section).

Closes #1889

### DIFF
--- a/src/node.rs
+++ b/src/node.rs
@@ -11,7 +11,7 @@ use crate::{
     chain::NetworkParams,
     error::{InterfaceError, RoutingError},
     event_stream::{EventStepper, EventStream},
-    id::{FullId, PublicId},
+    id::{FullId, P2pNode, PublicId},
     outbox::EventBox,
     pause::PausedState,
     quic_p2p::{OurType, Token},
@@ -26,7 +26,7 @@ use std::{net::SocketAddr, sync::mpsc};
 
 #[cfg(feature = "mock_base")]
 use {
-    crate::{chain::SectionProofChain, id::P2pNode, utils::XorTargetInterval, Chain, Prefix},
+    crate::{chain::SectionProofChain, utils::XorTargetInterval, Chain, Prefix},
     std::{
         collections::{BTreeMap, BTreeSet},
         fmt::{self, Display, Formatter},
@@ -183,6 +183,11 @@ impl Node {
     /// to the given one.
     pub fn close_group(&self, name: XorName, count: usize) -> Option<Vec<XorName>> {
         self.machine.current().close_group(name, count)
+    }
+
+    /// Returns the connection information of all the current section elders.
+    pub fn our_elders_info(&self) -> Option<impl Iterator<Item = &P2pNode>> {
+        self.machine.current().our_elders()
     }
 
     /// Returns the `PublicId` of this node.
--- a/src/state_machine.rs
+++ b/src/state_machine.rs
@@ -131,6 +131,16 @@ impl State {
         )
     }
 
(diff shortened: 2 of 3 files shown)


src/state_machine.rs
```diff
@@ -131,6 +131,14 @@ impl State {
         )
     }
 
+    pub fn our_elders<'a>(&'a self) -> Option<Box<dyn Iterator<Item = &P2pNode> + 'a>> {
```
jeanphilippeD: I think the pattenr that is used when not all state provide the interface is to have an explicit match, which read as well or better when we have only 1 or 2 state that match. That avoid having to add a new function to the `Base`, just so we can call it. (Also, it is consistent with Terminated which is always special cased.


```
    pub fn our_elders(&self) -> Option<Iterator<Item = &P2pNode>> {
        match *self {
            State::Elder(ref state) => Some(state.our_elders()),
            State::Adult(_) | State::BootstrappingPeer(_) | State::JoiningPeer(_) | State::Terminated => None,
        }
    }
```

src/states/common/base.rs
```diff
@@ -44,6 +44,17 @@ pub trait Base: Display {
...
     fn send_routing_message(&mut self, routing_msg: RoutingMessage) -> Result<(), RoutingError>;
 
+    // TODO: Due to https://stackoverflow.com/a/39490692 it's currently not possible do it without
+    // Box'ing. It's not a big price though as the collection is still only referenced and the
+    // internals are not cloned - just that the iterator returned is moved to the heap. The
+    // iterator still references the elements in the original collection.
+    //
+    // Once the feature lands in stable Rust, the Box'ing can be removed and impl Trait used for
+    // Trait functions too.
+    fn our_elders<'a>(&'a self) -> Option<Box<dyn Iterator<Item = &P2pNode> + 'a>> {
```
jeanphilippeD: Not needed, if dispatched to the right state from Node.

src/states/elder/mod.rs
```diff
@@ -1164,6 +1164,10 @@ impl Base for Elder {
         &self.full_id
     }
 
+    fn our_elders<'a>(&'a self) -> Option<Box<dyn Iterator<Item = &P2pNode> + 'a>> {
+        Some(Box::new(self.chain.our_elders()))
```
jeanphilippeD: I don't think we need to box if function not defined in Base.

src/node.rs
```diff
@@ -184,6 +184,11 @@ impl Node {
         self.machine.current().close_group(name, count)
     }
 
+    /// Returns the connection information of all the current section elders.
+    pub fn our_elders_info<'a>(&'a self) -> Option<Box<dyn Iterator<Item = &P2pNode> + 'a>> {
```
jeanphilippeD: One more thing, should we have a small test for this?

madadam: Question: why return `Option<Iterator>` instead of just returning empty iterator in the case we are not elder?

ustulation: As discussed, an empty iterator could be misguiding and indicate that there are no elders vs `Option::None` for and `Option<impl Iterator>` would mean that it's an invalid request or not possible to create this iterator. So leaving it as `Option` for now.

jeanphilippeD: No test then? it is probably ok for what is an accessor I guess.

PR #1892, opened 2019-11-13
Test improvements

madadam:
This PR improves the test framework:

1. RNG seed can now be specified using `SEED` environment variable. Example:
    ```
    SEED="[1,2,3,4]" cargo test --release --features=mock_base aggressive_churn
    ```
2. Logs are now enabled automatically when the `RUST_LOG` env variable is set, instead of having to manually insert `log::init` call into the tests. Works only in mock-network tests. Example:
    ```
    RUST_LOG=routing=trace cargo test ...
    ```
3. The current seed is printed on test failure but also when the test is interrupted with Ctrl+C or by sending `SIGTERM` signal to the process (that is, killing it). Note: to make this work, routing now hijacks the Ctrl+C handler if `mock_base` is enabled. Currently this happens inside the `Network` struct which is part of mock quic-p2p. That means this can interfere with vault's Ctrl+C handler if vault uses routing with `mock_base` enabled and also uses routing's mock quick-p2p (which I think is the plan). If this turns out to be an issue, we might want to put this feature behind a separate feature flag or somehow make sure it is only used inside routing itself and not by other crates.
4. `maidsafe_utilities::SeededRng` is replaced by `TestRng` which is simpler. It doesn't have thread-local accessor (no `thread_rng`), so the rng instance must now be passed down explicitly to everywhere it is needed. This is a deliberate decision to make the code more explicit and thus easier to understand.
5. There is now a `MainRng` type alias which evaluates to `OsRng` in production and `TestRng` in tests, so no generics or trait objects are necessary anymore.


Cargo.toml
```diff
@@ -14,11 +14,13 @@ edition = "2018"
 bytes = "~0.4.12"
 config_file_handler = "~0.11.0"
 crossbeam-channel = "~0.3.9"
+ctrlc = { version = "~3", optional = true, features = ["termination"] }
```
jeanphilippeD: Is that not a dev dependency?

Cargo.toml
```diff
@@ -14,11 +14,13 @@ edition = "2018"
...
 fake_clock = "~0.3.0"
 fxhash = "~0.2.1"
 hex = "~0.2.0"
 hex_fmt = "~0.1.0"
 itertools = "~0.6.1"
+lazy_static = { version = "~1", optional = true }
```
jeanphilippeD: same?

Cargo.toml
madadam: No, because the code that depends on this lives in `src` and is not exclusively behind `cfg(test)`.

madadam: I do have an idea how to make it possible though. Instead of adding this seed printing functionality to `mock::quic_p2p::Network`, we would create a new type, say `TestEnv` which would live in `tests/mock_network_utils.rs` and which would contain the mock-quic-p2p `Network` as well as the seed printing stuff. Then we would use this `TestEnv` everywhere we currently use `Network` in the mock network tests. Might be better to do it in a followup PR (if at all), to keep this one small and focused (as much as possible).

src/rng.rs
```diff
@@ -0,0 +1,374 @@
...
+            collections::HashMap,
+            process,
+            sync::{Mutex, Once},
+            thread::{self, ThreadId},
+        };
+
+        static HANDLER_SET: Once = Once::new();
+
+        lazy_static! {
+            static ref ACTIVE_SEEDS: Mutex<HashMap<ThreadId, (String, Seed)>> =
```
jeanphilippeD: Why is there a HashMap? I thought before we only had a single seed running, is there a reason for that?

src/rng.rs
```diff
@@ -0,0 +1,374 @@
...
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), rand_crypto::Error> {
+        self.0.fill_bytes(dest);
+        Ok(())
+    }
+}
+
+impl<R: CryptoRng> rand_crypto::CryptoRng for RngCompat<R> {}
+
+// Rng implementation used in production. Uses `OsRng` for maximum cryptographic security.
+#[cfg(not(any(test, feature = "mock_base")))]
```
jeanphilippeD: any reason to not use it for test as well if mock_base is not enabled?

jeanphilippeD: I guess lack of reproductibility?

src/rng.rs
madadam: There is no single global seed anymore and there can be multiple tests running in parallel. Also, this functionality is only to support interruptions (Ctrl+C, etc...).

src/rng.rs
madadam: I guess I can do that, but what would we gain by it?

madadam: That would apply to unit tests only and there we should not depend on rngs anyway. But in case we do for some reason, at least it gives us reproducibility, like you said.

jeanphilippeD: I was wondering whether we would want test using real random(checking we did not completely messed that) , but I guess unreproductible tests are not that useful.

madadam:
> Change looks good.
> Quite more involved than I expected, so maybe wait until part 2 of the conn info work is merged before merging?

Yup, sounds good.

PR #1893, opened 2019-11-14
fix/chain: extend cached keys to pass churn test

jeanphilippeD:
Fix aggressive_churn Some([2134840958, 10390792, 1455924021, 536582862]).
AckMessage sent to ourselves was received when one node already purged
the relevant self key from cached keys.

Test:
Verify seed pass.
soak test.


octol:
Looks good to me! 20 is still a pretty small number so I don't see any downsides to upping the size

PR #1896, opened 2019-11-15
Carry Online votes over parsec resets

madadam:


jeanphilippeD:
Merged with https://github.com/maidsafe/routing/pull/1897

PR #1898, opened 2019-11-18
Move PeerMap to NetworkService

octol:
- Move `PeerMap` to `NetworkService` as a prepatory step for Part 3 (#1885) in the connection request/response refactoring. The intention (in Part 3) is to strip down `PeerMap` to only keep track of client `SocketAddr` and mapping certificates to nodes.
Any notion of "connectedness" is set to be deprecated.
- Remove `peer_map` call in `send_parsec_gossip`
- Fix a few minor issues pointed out in the review of #1884

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -695,12 +695,10 @@ impl Chain {
     }
 
     /// Returns a set of elders we should be connected to.
-    // WIP: should we remove potential duplicates?
     pub fn elders(&self) -> impl Iterator<Item = &P2pNode> {
         self.neighbour_infos()
             .chain(iter::once(self.state.our_info()))
             .flat_map(EldersInfo::member_nodes)
-            .chain(self.state.new_info.member_nodes())
     }
 
     /// Checks if given `PublicId` is an elder in our section or one of our neighbour sections.
--- a/src/states/adult.rs
+++ b/src/states/adult.rs
@@ -51,7 +51,6 @@ pub struct AdultDetails {
     pub gen_pfx_info: GenesisPfxInfo,
     pub routing_msg_backlog: Vec<SignedRoutingMessage>,
     pub direct_msg_backlog: Vec<(P2pNode, DirectMessage)>,
-    pub peer_map: PeerMap,
     pub routing_msg_filter: RoutingMessageFilter,
     pub timer: Timer,
     pub network_cfg: NetworkParams,
@@ -69,7 +68,6 @@ pub struct Adult {
     routing_msg_backlog: Vec<SignedRoutingMessage>,
     direct_msg_backlog: Vec<(P2pNode, DirectMessage)>,
     parsec_map: ParsecMap,
-    peer_map: PeerMap,
     parsec_timer_token: u64,
     routing_msg_filter: RoutingMessageFilter,
     timer: Timer,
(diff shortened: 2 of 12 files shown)


src/chain/chain.rs
```diff
@@ -662,6 +662,12 @@ impl Chain {
             .map(|member_info| member_info.p2p_node.connection_info())
     }
 
+    /// Returns the `P2pNode` for a member of our section.
+    pub fn get_member_p2p_node_by_id(&self, pub_id: &PublicId) -> Option<&P2pNode> {
+        self.get_member_p2p_node(pub_id.name())
+            .filter(|p2p_node| p2p_node.public_id() == pub_id)
```
jeanphilippeD: Is the filter necessary?
And if not, not sure it really help compared to `self.get_member_p2p_node(pub_id.name())`

src/states/adult.rs
```diff
@@ -435,6 +432,8 @@ impl Base for Adult {
         // Need to collect IDs first so that self is not borrowed via the iterator
         //
         // WIP: this is probably out of date? How else do we know which our section members are?
+        // WIP: once ConnectionRequest/ConnectionResponse is removed this whole function can
```
jeanphilippeD: Not sure about that, maybe.
However, when Elder are demoted to Adults, they should forward RoutingMessages to Elders.

src/states/common/base.rs
```diff
@@ -380,32 +375,20 @@ pub trait Base: Display {
...
-                "{} Less than dg_size valid targets! dg_size = {}; targets = {:?}; valid targets = [{:?}]; msg = {:?}",
-                self,
-                dg_size,
-                dst_targets,
-                dst_targets
-                    .iter()
-                    .filter(|dst| dst.resolve().is_some())
-                    .format(", "),
-                message
+                "{} Less than dg_size valid targets! dg_size = {}; targets = {:?}; msg = {:?}",
```
jeanphilippeD: Wondering if this warn should be log_or_panic now. Seem that it should never happen.

src/chain/chain.rs
octol: Yeah good question, is it possible to happen? That the duplication is inconsistent?

src/states/elder/mod.rs
```diff
@@ -1414,9 +1419,9 @@ impl Base for Elder {
                     signed_msg.routing_message(),
                     target
                 );
-                let p2p_node = p2p_node.clone();
+                let conn_info = p2p_node.connection_info().clone();
```
maqi: is this clone really necessary?

octol: Otherwise the mutable borrow of `self` in `send_direct_message` conflicts with the immutable borrow of `self` when getting the reference to the `p2p_node` above

src/states/adult.rs
```diff
@@ -307,11 +304,11 @@ impl Base for Adult {
     }
 
     fn peer_map(&self) -> &PeerMap {
-        &self.peer_map
+        &self.network_service().peer_map
     }
 
     fn peer_map_mut(&mut self) -> &mut PeerMap {
-        &mut self.peer_map
+        &mut self.network_service_mut().peer_map
```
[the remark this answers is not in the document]
octol: Yep sounds like a good idea, I'll add that to the follow up PR

PR #1899, opened 2019-11-18
Relocate request

madadam:
This PR implement the relocate request feature - that is - a section can request a node from another section to be relocated into it. This is a high-level overview of how the feature works:

1. Relocate request must be invoked explicitly by the upper layer. Autonomous relocate requests are not implemented yet (it will be a separate task)
2. Triggering relocate request starts by first voting for it in parsec
3. When it accumulates, we calculate a recipient of the request (a random, but common `XorName`) 
4. We send the request to that name as a `PrefixSection` -> `Section` message. We also store the recipient name in the shared state.
5. When the recipient section receives it, it votes to respond to it
6. When that vote accumulates, it decides whether it can fulfill the request by checking the number of nodes in the section. If it has more than `ELDER_SIZE`, it accepts it, otherwise it denies it.

## Accept

7. The recipient section votes to relocate its youngest node into the requester section.
8. That node relocates
9. When the requester section accumulates the `Online` vote, it resets the relocate request recipient name from shared state. The process now ends.

## Deny

7. The recipient section sends a deny message back to the requester section
8. When the requester section receives the message, it votes again, putting the previous recipient name into the vote to make it unique
9. When the vote accumulates, new recipient name is calculated by rehashing the previous name.
10. Go to point 4

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -17,7 +17,7 @@ use crate::{
     error::RoutingError,
     id::{P2pNode, PublicId},
     parsec::{DkgResult, DkgResultWrapper},
-    relocation::{self, RelocateDetails},
+    relocation::{self, RelocateDetails, RelocateRequestRecipientError},
     routing_table::{Authority, Error},
     utils::LogIdent,
     BlsPublicKeySet, BlsSecretKeyShare, BlsSignature, ConnectionInfo, Prefix, XorName, Xorable,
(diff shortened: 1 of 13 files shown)


jeanphilippeD:
> 7. The recipient section votes to relocate its youngest node into the requester section.

Not sure that is ideal, if the section is in need of nodes, it probably would be good to avoid having all requested section sending infants, or even any infant. Probably better to relocate a random adult?

dirvine:
https://github.com/maidsafe/routing/issues/2106

Issue #1891, opened 2019-11-13
Sign new real BLS Key with old one


jeanphilippeD:
Whenever we get a DkgResult: (1-3 days)
   Create `RealSectionKeyInfo` and sign it with signature share from the key we are holding.
   (Need to store secret key if not done before - Yes needed).

PR #1900, opened 2019-11-19
Start signing next key with previous BLS key

jeanphilippeD:
Partially address #1891 

Use DkgResult and BLS keys to sign the next key.
This is incomplete in the sense that we are not yet signing the SectionKeyInfo as we should, but provide the initial setup to continue the work of removing bls_emu

### DIFF
--- a/src/chain/network_event.rs
+++ b/src/chain/network_event.rs
@@ -11,11 +11,20 @@ use crate::{
     crypto::Digest256,
     id::{FullId, P2pNode, PublicId},
     parsec,
+    //parsec::DkgResult,
     relocation::RelocateDetails,
     routing_table::Prefix,
-    BlsPublicKeyShare, BlsSignature, BlsSignatureShare, RoutingError, XorName,
+    BlsPublicKeyShare,
+    BlsSignature,
+    BlsSignatureShare,
+    RealBlsPublicKeyShare,
+    RealBlsSecretKeyShare,
+    RealBlsSignatureShare,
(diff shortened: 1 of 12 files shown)


src/parsec.rs
```diff
@@ -293,6 +294,17 @@ impl ParsecMap {
     }
 }
 
+// Generate a DkgResult that can be used for the first node
+pub fn generate_first_dkg_result(rng: &mut MainRng) -> DkgResult {
+    let threshold = 0;
+    let index = 0;
+    let rng = rng::new_from(rng);
```
madadam: It's not necessary to create new rng here. Just use the one passed to this function.

jeanphilippeD:
Rebased only

jeanphilippeD:
Fixed review comment

jeanphilippeD:
Fix other review comment.

src/chain/network_event.rs
```diff
@@ -61,6 +70,29 @@ impl EventSigPayload {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
+pub struct RealBlsEventSigPayload {
```
octol: I guess "Real" will be removed once the work is complete?

jeanphilippeD: Yes, it is just a bridge so some pieces can co-exist while we are solving the missing pieces.

PR #1903, opened 2019-11-20
Sign RealSectionKeyInfo with BLS key

### LINKED ISSUE #1891: Sign new real BLS Key with old one

jeanphilippeD:
Closes #1891 

Sign RealSectionKeyInfo, this also force the code & test to provide the BlsPublicKey where needed.
This is an essential part of being able to remove bls_emu.

Note:
- `self.vote_for_section_info(merged_info)?;` can no longer be called as it require a BlsPublicKey for the new section, and we do not have it. Merge is broken already so comment this out. The whole merge code may be removed when needed.
-`our_section_next_bls_keys_sig` was not doing the right thing, as that signature need to be provided by all nodes whether they were participants or not, they will have the BlsPublicKey and need to sign it.

### DIFF
--- a/src/chain/chain_accumulator.rs
+++ b/src/chain/chain_accumulator.rs
@@ -259,16 +259,19 @@ pub struct RemainingEvents {
 
 #[cfg(test)]
 mod test {
-    use super::super::{EldersInfo, SectionKeyInfo};
+    use super::super::{EldersInfo, RealBlsEventSigPayload, RealSectionKeyInfo};
     use super::*;
     use crate::{
         id::FullId,
+        parsec::generate_bls_threshold_secret_key,
         rng::{self, MainRng},
-        BlsPublicKeyShare,
+        BlsPublicKeyShare, RealBlsPublicKeySet,
     };
     use parsec::SecretId;
     use std::iter;
     use unwrap::unwrap;
+
+    const TEST_DATA_FOR_SIGN: [u8; 1] = [1u8];
 
     struct TestData {
         pub our_id: PublicId,
@@ -293,18 +296,32 @@ mod test {
         ))
     }
 
-    fn random_section_info_sig_payload(rng: &mut MainRng) -> EventSigPayload {
+    fn random_section_info_sig_payload(
+        rng: &mut MainRng,
+    ) -> (EventSigPayload, RealBlsEventSigPayload, RealBlsPublicKeySet) {
+        let participants = 2;
+        let first_secret_key_index = 0;
+        let bls_keys = generate_bls_threshold_secret_key(rng, participants);
+        let bls_secret_key_share = bls_keys.secret_key_share(first_secret_key_index);
         let (id, first_proof) = random_ids_and_proof(rng);
-        EventSigPayload {
-            pub_key_share: BlsPublicKeyShare(*id.public_id()),
-            sig_share: first_proof.sig,
-        }
+
+        (
+            EventSigPayload {
+                pub_key_share: BlsPublicKeyShare(*id.public_id()),
+                sig_share: first_proof.sig,
+            },
+            RealBlsEventSigPayload {
+                pub_key_share: bls_secret_key_share.public_key_share(),
+                sig_share: bls_secret_key_share.sign(&TEST_DATA_FOR_SIGN),
+            },
+            bls_keys.public_keys(),
+        )
     }
 
     fn random_ids_and_proof(rng: &mut MainRng) -> (FullId, Proof) {
         let id = FullId::gen(rng);
         let pub_id = *id.public_id();
-        let sig = id.sign_detached(&[1]);
+        let sig = id.sign_detached(&TEST_DATA_FOR_SIGN);
 
         (id, Proof { pub_id, sig })
     }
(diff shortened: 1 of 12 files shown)


src/chain/chain_accumulator.rs
```diff
@@ -293,12 +294,24 @@ mod test {
         ))
     }
 
-    fn random_section_info_sig_payload(rng: &mut MainRng) -> EventSigPayload {
+    fn random_section_info_sig_payload(
+        rng: &mut MainRng,
+    ) -> (EventSigPayload, RealBlsEventSigPayload, RealBlsPublicKeySet) {
```
maqi: maybe worth to make this triple as a struct?

src/chain/chain_accumulator.rs
```diff
@@ -293,12 +294,24 @@ mod test {
...
+        let bls_keys = generate_bls_threshold_secret_key(rng, 2);
```
maqi: why choosing 2 ?
and better make it as a const ?

src/chain/chain_accumulator.rs
```diff
@@ -293,12 +294,24 @@ mod test {
...
+        let bls_secret_key_share = bls_keys.secret_key_share(0);
```
maqi: key_share always indexed from 0 ? 
shall there be a comment at least explain why?

src/chain/chain_accumulator.rs
```diff
@@ -293,12 +294,24 @@ mod test {
...
-        }
+
+        (
+            EventSigPayload {
+                pub_key_share: BlsPublicKeyShare(*id.public_id()),
+                sig_share: first_proof.sig,
+            },
+            RealBlsEventSigPayload {
+                pub_key_share: bls_secret_key_share.public_key_share(),
+                sig_share: bls_secret_key_share.sign(&[1]),
```
maqi: is `[1]` just a random user-data that selected?
better to be a const as well ?

src/parsec.rs
```diff
@@ -296,14 +311,25 @@ impl ParsecMap {
...
     DkgResult::new(secret_key_set.public_keys(), Some(secret_key_share))
 }
 
+/// Generate a BLS SecretKeySet for the given number of participants.
+/// Used for generating first node, or for test.
+pub fn generate_bls_threshold_secret_key(
+    rng: &mut MainRng,
+    participants: usize,
+) -> RealBlsSecretKeySet {
+    let threshold = participants.saturating_sub(1) / 3;
```
maqi: better have `3` as a const ?

src/states/elder/mod.rs
```diff
@@ -342,8 +342,9 @@ impl Elder {
                 debug!("{} Failed to send Merge: {:?}.", self, err);
             }
         }
-        if let Some(merged_info) = self.chain.try_merge()? {
-            self.vote_for_signed_event(merged_info)?;
+        if let Some(_merged_info) = self.chain.try_merge()? {
+            // Cannot vote for this new section info as we do not have a BlsPublicKey available.
+            // self.vote_for_section_info(merged_info)?;
```
maqi: I thought the `merge` logic shall be removed?

src/chain/chain_accumulator.rs
jeanphilippeD: Only a triple because we have a fake and real, will become back a pair when there is only the real one. (i.e bls_emu gone)

src/chain/chain_accumulator.rs
jeanphilippeD: pretty arbitrary, allow to have 2 different keys. Will make it a named local

src/chain/chain_accumulator.rs
jeanphilippeD: I don't think it make sense to comment here, it is the secret_key_share api.
I can make it a named constant as well.

src/states/elder/mod.rs
jeanphilippeD: Other PR.

src/parsec.rs
jeanphilippeD: Went for comment instead seemed more useful.

jeanphilippeD:
Fixed review comments and clippy.

src/parsec.rs
```diff
@@ -296,14 +311,27 @@ impl ParsecMap {
...
 
+/// Generate a BLS SecretKeySet for the given number of participants.
+/// Used for generating first node, or for test.
+pub fn generate_bls_threshold_secret_key(
+    rng: &mut MainRng,
+    participants: usize,
+) -> RealBlsSecretKeySet {
+    // The BLS scheme will require more than `participants / 3`
+    // shares in order to construct a full key or signature.
+    let threshold = participants.saturating_sub(1) / 3;
```
madadam: Why do yo subtract 1?

jeanphilippeD: It is from Parsec.
```
/// Threshold to use for running DKG
pub fn dkg_threshold(participants_count: usize) -> usize {
    participants_count.saturating_sub(1) / 3
}
```

I tried to think about it, I think the reason we substract 1, is because for the example of 3, we need 2*t + 1 answer from the DKG to complete, that would mean 3 if threshold was 1, but then that would mean all answered and that cannot be the case.

For the example of 4, 2t+1 = 3, t = 1 invalid, which may not have answered.
So I think for number divisible by 3, the would be to big. (Sorry, a bit hand wavy, I think I and @fizyk20 had thought through it a bit more when we did the DKG in parsec, I guess the key_gen doc in parsec is probably the best source of info).

madadam: OK, makes sense.

PR #1905, opened 2019-11-20
Handle SectionInfo and NeighbourInfo separately + remove merge logic

### LINKED ISSUE #1904: Handle own section info separately from neighbour info + remove merge logic
They are both currently handled in one function (`handle_section_info_event`) even though they are quite different, which makes the code unnecessarily complex and confusing.

Also remove the last remaining bits of the merging logic which is non-functional anyway and only clutters the code.

madadam:
1. Clean up the last remaining bits of merge logic 
2. Separate the handling of `SectionInfo` and `NeighbourInfo` events, for clarity.

Closes #1904


src/states/elder/mod.rs
```diff
@@ -1582,53 +1586,35 @@ impl Approved for Elder {
...
                 self,
                 old_pfx,
                 elders_info.prefix()
             );
-        } else if self_sec_update {
+        } else {
             self.reset_parsec()?;
         }
 
-        self.update_neighbour_connections(neighbour_change, outbox);
```
jeanphilippeD: I am not 100% sure we can remove this `update_neighbour_connections`, because it does establish connections with other elders as well, but it is unlikely to be an issue.

src/states/elder/mod.rs
```diff
@@ -1637,6 +1623,25 @@ impl Approved for Elder {
...
+        &mut self,
+        elders_info: EldersInfo,
+        old_pfx: Prefix<XorName>,
+        neighbour_change: EldersChange,
+        outbox: &mut dyn EventBox,
+    ) -> Result<(), RoutingError> {
+        info!("{} - handle NeighbourInfo: {:?}.", self, elders_info);
+
+        if elders_info.prefix().is_extension_of(&old_pfx) {
+            self.finalise_split(outbox)?;
```
jeanphilippeD: We really do not want this happening, finalize split should only occur from a handle_section_info_event.
In practice this would never happen because we need to reach consensus on SectionInfo to send the NeighbourInfo, so they will always consensused after.

You can just delete that if and its content.

PR #1906, opened 2019-11-21
Remove bls_emu and use real BLS

### LINKED ISSUE #1907: Use Real BLS
Replace content of `SectionKeyInfo` with `RealSectionKeyInfo` and see what breaks.  Fix things.

jeanphilippeD:
Closes #1907 

Remove bls_emu: tweak quorum test as threshold in bls_emu was off by 1 in some cases (did not do the saturating_sub like Parsec and mock Parsec.

Some things still need some work (see commit comments), but it provide a reasonable baseline for further work.


src/chain/chain_accumulator.rs
```diff
@@ -233,12 +233,32 @@ impl AccumulatingProof {
...
+            .collect();
+
+        trace!(
+            "combine_signatures {:?} -1- {:?} -2- {:?} -3- {:?}",
+            positions,
+            fr_and_shares,
+            pk_set,
+            elder_info
+        );
+        pk_set.combine_signatures(fr_and_shares.into_iter()).ok()
```
madadam: What about
```rust
let shares = elder_info.member_ids().enumerate().filter_map(|(index, pub_id)| {
    self.sig_shares.get(pub_id).map(|sig_payload| (index, sig_payload.sig_share))
});
pk_set.combine_signatures(shares).ok()
```
?

jeanphilippeD: Yes :-)

src/messages/mod.rs
```diff
@@ -174,14 +174,20 @@ impl SignedRoutingMessage {
...
         content: RoutingMessage,
-        full_id: &FullId,
+        full_id: &BlsSecretKeyShare,
         pk_set: BlsPublicKeySet,
         proof: SectionProofChain,
     ) -> Result<SignedRoutingMessage> {
         let mut signatures = BTreeMap::new();
-        let pk_share = BlsPublicKeyShare(*full_id.public_id());
+        let pk_share = full_id.public_key_share();
+        // WIP: Get more efficient.
```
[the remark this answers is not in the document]
jeanphilippeD: SecretKeyShare::from_mut does modify the given Fr.
I'm thinking we probably want to stick with indices, though we might want to store the index or indices somewhere, but I think this is more for a follow up PR.

src/messages/mod.rs
```diff
@@ -173,14 +173,20 @@ impl SignedRoutingMessage {
...
+        let position = (0..100)
```
madadam: Wouldn't `(0..elder_size)` be enough? You would have to pass `elder_size` as argument though...

jeanphilippeD:
Fixed clippy.

I'm looking at how to address the other comment, but the right way to do it (i.e I think keep the index with the secret key share is bigger than I would want for this PR).

@madadam is it ok to fix the not yet fixed comment in another PR so we can merge this since it has gone through soak test?

src/chain/shared_state.rs
```diff
@@ -647,6 +652,17 @@ impl SectionProofChain {
...
     }
 }
+impl Debug for SectionProofChain {
+    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
+        write!(
+            formatter,
+            "SectionProofChain {{ genesis_key_info: {:?}, blocks: {:?}, validate: {} }}",
+            self.genesis_key_info,
+            self.blocks,
+            self.validate(),
```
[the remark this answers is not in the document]
jeanphilippeD: Fixed with https://github.com/maidsafe/routing/pull/1913

src/messages/mod.rs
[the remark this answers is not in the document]
jeanphilippeD: Fixed with https://github.com/maidsafe/routing/pull/1913

src/messages/mod.rs
jeanphilippeD: Fixed with https://github.com/maidsafe/routing/pull/1913

PR #1908, opened 2019-11-21
Strip down PeerMap

octol:
Strip down peer_map to the minimum we need it to do.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -29,6 +29,7 @@ use std::{
     collections::{btree_map::Entry, BTreeMap, BTreeSet},
     fmt::{self, Debug, Display, Formatter},
     iter, mem,
+    net::SocketAddr,
 };
 
 #[cfg(feature = "mock_base")]
@@ -692,6 +693,11 @@ impl Chain {
         self.neighbour_infos()
             .chain(iter::once(self.state.our_info()))
             .flat_map(EldersInfo::member_nodes)
+    }
+
+    pub fn find_p2p_node_from_addr(&self, socket_addr: &SocketAddr) -> Option<&P2pNode> {
+        self.elders()
+            .find(|p2p_node| p2p_node.peer_addr() == socket_addr)
     }
 
     /// Checks if given `PublicId` is an elder in our section or one of our neighbour sections.
--- a/src/node.rs
+++ b/src/node.rs
@@ -399,12 +399,14 @@ impl Node {
             .unwrap_or_default()
     }
 
-    /// Returns a set of elders we should be connected to.
+    /// Returns the elders in our and neighbouring sections.
     pub fn elders(&self) -> impl Iterator<Item = &PublicId> {
-        self.chain()
-            .into_iter()
-            .flat_map(Chain::elders)
-            .map(P2pNode::public_id)
+        self.elder_nodes().map(P2pNode::public_id)
+    }
+
+    /// Returns the elders in our and neighbouring sections.
+    pub fn elder_nodes(&self) -> impl Iterator<Item = &P2pNode> {
+        self.chain().into_iter().flat_map(Chain::elders)
     }
 
     /// Returns whether the given `PublicId` is a member of our section.
@@ -469,8 +471,8 @@ impl Node {
     }
 
     /// Indicates if this node has the connection info to the given peer.
-    pub fn is_connected<N: AsRef<XorName>>(&self, name: N) -> bool {
-        self.machine.current().is_connected(name)
+    pub fn is_connected(&self, socket_addr: &SocketAddr) -> bool {
+        self.machine.current().is_connected(socket_addr)
     }
 
     /// Provide a SectionProofChain that proves the given signature to the section with a given
--- a/src/peer_map.rs
+++ b/src/peer_map.rs
@@ -6,24 +6,13 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::{id::PublicId, xor_name::XorName, ConnectionInfo};
+use crate::ConnectionInfo;
 use fxhash::{FxHashMap as HashMap, FxHashSet as HashSet};
-use std::{collections::hash_map::Entry, net::SocketAddr};
+use std::net::SocketAddr;
 
-/// This structure holds the bi-directional association between peers public id and their network
-/// connection info. This association can be create in two ways:
-/// 1. When both pieces of information (public id and connection info) are obtained at the same
-///    time, call `insert`. This happens when a third party (other members of the section) sends
-///    them to us.
-/// 2. Otherwise its a two step process: first, when the connection to the peer is established at
-///    the network layer, call `connect`. Then when their public id is received, call `identify`.
-///    This happens when the peer connects to us and then sends us a message which contains their
-///    public id.
 #[derive(Default)]
 pub struct PeerMap {
-    forward: HashMap<XorName, ConnectionInfo>,
-    reverse: HashMap<SocketAddr, HashSet<PublicId>>,
-    pending: HashMap<SocketAddr, PendingConnection>,
+    connections: HashMap<SocketAddr, ConnectionInfo>,
     clients: HashSet<SocketAddr>,
 }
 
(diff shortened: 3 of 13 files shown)


src/chain/chain.rs
```diff
@@ -694,6 +695,11 @@ impl Chain {
             .flat_map(EldersInfo::member_nodes)
     }
 
+    pub fn get_p2p_node_from_addr(&self, socket_addr: &SocketAddr) -> Option<&P2pNode> {
```
jeanphilippeD: can you name it 'find_...` get suggest fast access.

src/node.rs
```diff
@@ -407,6 +407,11 @@ impl Node {
             .map(P2pNode::public_id)
     }
 
+    /// Returns a set of elders we should be connected to.
```
jeanphilippeD: `/// Returns a set of elders we can communicate with`

At the end we won't necessarily be connected, so probably best to not add a comment that will soon be misleading.

src/peer_map.rs
```diff
@@ -33,115 +22,33 @@ impl PeerMap {
...
-            if entry.get().is_empty() {
-                let _ = entry.remove();
-                return Some(conn_info);
-            }
-        }
-
-        None
+    // Marks the connection as severed at the network layer.
+    pub fn disconnect(&mut self, socket_addr: SocketAddr) -> Option<ConnectionInfo> {
+        self.connections.remove(&socket_addr)
```
jeanphilippeD: I like these short functions :-)

src/states/adult.rs
```diff
@@ -530,7 +531,16 @@ impl Approved for Adult {
         let _ = self.chain.poll_relocation();
 
         self.remove_elder(pub_id, outbox)?;
-        self.disconnect(&pub_id);
+
+        if let Some(p2p_node) = self.chain.get_p2p_node(pub_id.name()) {
+            let p2p_node = p2p_node.clone();
+            self.disconnect(&p2p_node.peer_addr());
+        } else {
+            info!(
```
jeanphilippeD: this is a log_or_panic, that should never happen.

src/states/adult.rs
```diff
@@ -589,7 +599,16 @@ impl Approved for Adult {
         info!("{} - handle Relocate: {:?}.", self, details);
         self.chain.remove_member(&details.pub_id);
         self.remove_elder(details.pub_id, outbox)?;
-        self.disconnect(&details.pub_id);
+
+        if let Some(p2p_node) = self.chain.get_p2p_node(details.pub_id.name()) {
+            let p2p_node = p2p_node.clone();
+            self.disconnect(&p2p_node.peer_addr());
+        } else {
+            info!(
```
jeanphilippeD: Same, log_or_panic

src/states/bootstrapping_peer.rs
```diff
@@ -137,7 +137,7 @@ impl BootstrappingPeer {
         };
 
         self.send_direct_message(&dst, DirectMessage::BootstrapRequest(destination));
-        self.peer_map_mut().connect(dst);
+        self.peer_map_mut().connect(dst.peer_addr, dst);
```
jeanphilippeD: might as well only have a single param and do that within the `connect` function?

src/states/common/base.rs
```diff
@@ -241,19 +245,8 @@ pub trait Base: Display {
...
-            let other_transition = self.handle_peer_lost(pub_id, outbox);
-
-            if let Transition::Stay = transition {
-                transition = other_transition
-            }
-        }
-
-        transition
+        let _ = self.peer_map_mut().disconnect(peer_addr);
+        self.handle_peer_lost(peer_addr, outbox)
```
jeanphilippeD: That is interesting, why this change?

src/states/elder/mod.rs
```diff
@@ -1059,11 +1060,23 @@ impl Elder {
...
+                .unwrap_or(false)
+        };
         if self
             .peer_map()
-            .connected_ids()
-            .filter(|id| self.chain.our_id() != *id)
-            .any(|id| self.chain.is_peer_our_elder(id))
+            .connected_peers()
+            .filter(|conn_info| &our_connection_info != *conn_info)
+            .any(|conn_info| connection_info_belongs_to_our_elder(conn_info))
```
jeanphilippeD: Seem a bit more complicated than I would expect.

```
       let our_name = self.chain.our_name();
        if self
            .chain
            .our_info()
            .member_map()
            .iter()
            .filter(|(name, _)| name != our_name)
            .any(|(_, p2p_node)| !self.peer_map().has(p2p_node.peer_addr()))  {
            return true;
        } ...

```

src/states/elder/mod.rs
```diff
@@ -1538,7 +1558,15 @@ impl Approved for Elder {
         }
 
         self.remove_elder(pub_id, outbox)?;
-        self.disconnect(&pub_id);
+        if let Some(p2p_node) = self.chain.get_p2p_node(pub_id.name()) {
+            let p2p_node = p2p_node.clone();
+            self.disconnect(&p2p_node.peer_addr());
+        } else {
+            info!(
```
jeanphilippeD: log or panic

src/states/elder/mod.rs
```diff
@@ -1702,7 +1730,15 @@ impl Approved for Elder {
 
         self.chain.remove_member(&pub_id);
         self.remove_elder(pub_id, outbox)?;
-        self.disconnect(&pub_id);
+        if let Some(p2p_node) = self.chain.get_p2p_node(pub_id.name()) {
+            let p2p_node = p2p_node.clone();
+            self.disconnect(&p2p_node.peer_addr());
+        } else {
+            info!(
```
jeanphilippeD: log or paninc?

src/chain/chain.rs
octol: Yes good point!

src/states/bootstrapping_peer.rs
octol: Yep you're right, that would be simpler (and more correct)

src/states/elder/mod.rs
octol: Good point, I'll update!

src/states/bootstrapping_peer.rs
octol: We don't have the 1-to-many mapping anymore as we did when the key was `PublicId`

src/states/common/base.rs
octol: We don't have the 1-to-many mapping as when the key was `PublicId`

jeanphilippeD: Could you be a bit more verbose, I still don't understand why it is ok to call handle_peer_lost here.

octol: Hmm not sure I understand, we called it before as well?

jeanphilippeD: Good point, I just completely misread the original loop.

src/states/common/approved.rs
```diff
@@ -413,13 +415,26 @@ pub trait Approved: Base {
...
+        self.peer_map_mut().connect(their_conn_info.clone());
         self.send_direct_message(&their_conn_info, DirectMessage::ConnectionResponse);
 
         Ok(())
     }
 
+    fn disconnect_by_id_lookup(&mut self, pub_id: &PublicId) {
+        if let Some(node) = self.chain().get_p2p_node(pub_id.name()) {
+            let node = node.clone();
+            self.disconnect(&node.peer_addr());
```
jeanphilippeD: clone or copy the peer_addr instead of the full node since that is what you use.

src/states/elder/mod.rs
```diff
@@ -629,14 +629,14 @@ impl Elder {
...
-        if !self.peer_map().has(&pub_id) {
+        if !self.peer_map().has(p2p_node.peer_addr()) {
             trace!(
                 "{} - Not yet connected to {} - use p2p_node.",
                 self,
                 p2p_node
             );
             self.peer_map_mut()
-                .insert(pub_id, p2p_node.connection_info().clone());
+                .connect(p2p_node.connection_info().clone());
```
jeanphilippeD: OK, this might be ok for now.
Ideally I'd think these would be gone (i.e only update the peer map on quic_p2p event or disconnect call), does it fail if we remove the following?
I'm thinking we should get a `ConnectedTo` event from quic_p2p when sending that?

```
             self.peer_map_mut()
                .connect(p2p_node.connection_info().clone()); 
```

Same apply for all `connect` calls before a `send_direct_message`

src/states/joining_peer.rs
```diff
@@ -147,12 +147,14 @@ impl JoiningPeer {
             RoutingMessage {
                 content:
                     MessageContent::ConnectionRequest {
-                        conn_info, pub_id, ..
+                        conn_info,
+                        pub_id: _pub_id,
```
jeanphilippeD: you can just remove the pub_id field.

src/peer_map.rs
```diff
@@ -6,24 +6,13 @@
...
-/// 2. Otherwise its a two step process: first, when the connection to the peer is established at
-///    the network layer, call `connect`. Then when their public id is received, call `identify`.
-///    This happens when the peer connects to us and then sends us a message which contains their
-///    public id.
 #[derive(Default)]
 pub struct PeerMap {
-    forward: HashMap<XorName, ConnectionInfo>,
-    reverse: HashMap<SocketAddr, HashSet<PublicId>>,
-    pending: HashMap<SocketAddr, PendingConnection>,
+    connections: HashMap<SocketAddr, ConnectionInfo>,
```
madadam: Now the socket address is stored twice - in the key and in the value. Could we change this to `HashMap<SocketAddr, Certificate>` ?

src/states/common/approved.rs
```diff
@@ -362,13 +362,15 @@ pub trait Approved: Base {
...
         }
 
-        if self.peer_map().has(&their_pub_id) {
-            trace!(
-                "{} - Not sending connection request to {} - already connected.",
-                self,
-                their_pub_id
-            );
-            return Ok(());
+        if let Some(p2p_node) = self.chain().get_p2p_node(their_pub_id.name()) {
```
madadam: Why do we need to send the connection request if we have the `P2pNode`?

src/states/common/approved.rs
```diff
@@ -413,13 +415,26 @@ pub trait Approved: Base {
...
-            .insert(their_pub_id, their_conn_info.clone());
+        self.peer_map_mut().connect(their_conn_info.clone());
         self.send_direct_message(&their_conn_info, DirectMessage::ConnectionResponse);
 
         Ok(())
     }
 
+    fn disconnect_by_id_lookup(&mut self, pub_id: &PublicId) {
+        if let Some(node) = self.chain().get_p2p_node(pub_id.name()) {
+            let node = node.clone();
```
madadam: Instead of cloning the whole `P2pNode`, you can just copy the peer address.

src/states/elder/mod.rs
octol: Removing those `connect` calls does still seem to pass the tests

src/states/elder/mod.rs
```diff
@@ -363,8 +364,7 @@ impl Elder {
 
     fn establish_connection(&mut self, node: P2pNode) {
         self.send_direct_message(node.connection_info(), DirectMessage::ConnectionResponse);
-        self.peer_map_mut()
-            .insert(*node.public_id(), node.into_connection_info());
+        self.peer_map_mut().connect(node.connection_info().clone())
```
madadam: Do we still need to do this? We already have the full connection info. I think it should be sufficient to call `peer_map.connect` only in `handle_connected_to`. This also applies to the other places where we do this. Similarly, we should probably only do `peer_map.disconnect` in `handle_connection_failure`.

src/peer_map.rs
octol: This was indeed discussed during planning. The argument for `ConnectionInfo` was that you can return a reference and just use that directly. I think both ways has merit. What do you think @jeanphilippeD ?

jeanphilippeD: We had a chat before, this is simpler and the gain is very small, that avoid having to rebuild the ConnectionInfo everytime we get it.

src/states/elder/mod.rs
octol: Agree, I've started removing the `connect` calls to see if it still works

jeanphilippeD:
> Looks mostly OK, just not sure about those explicit peer_map.connect and peer_map.disconnect calls. I don't think we need them anywhere expect handle_connected_to and handle_connection_failure.

@madadam agree with connect. Was not sure for disconnect, do we get an event from quic_p2p when we initiate the disconnect ourselves?

src/peer_map.rs
madadam: OK, fair enough. I think at some point in the future we should change the way we store the certificate from `Vec<u8>` to `Rc<[u8]>` (or `Arc<[u8]>`) to make them cheap to clone. Then it wouldn't matter to return them by value instead of reference so we would be able to avoid this duplication.

src/states/elder/mod.rs
octol: What about `connect` in `send_bootstrap_request`? I think we still need it there

madadam: I'm not sure. Do we? I think it's the same thing. When we receive a response, that response will be preceded by `ConnectedTo` event from quic-p2p and we would `peer_map.connect` there.

src/states/common/approved.rs
octol: If we have their `P2pNode` we are connected, but are we also connected at the network layer?

madadam:
> do we get an event from quic_p2p when we initiate the disconnect ourselves?

We don't. We need to disconnect explicitly then. But I think `Base::disconnect` does that already.

src/states/common/approved.rs
madadam: Not necessarily, but the whole idea behind this work is that it doesn't matter. We make no difference between knowing someones connection info and actually being connected. I think.

src/states/elder/mod.rs
octol: All the tests failed when I tried removing that one, just thought I'd check before started investigating the failures

madadam: Hm, OK. Then maybe put it back. There might be something subtle going on in `BootstrappingPeer`. We can always revisit it later.

src/states/common/approved.rs
jeanphilippeD: (That is once it is finished, but that specific piece of work does not change the model of all node we know are actually connected with quic_p2p).

PR #1910, opened 2019-11-21
Remove new_info from SharedState

madadam:


src/chain/chain.rs
```diff
@@ -925,7 +925,7 @@ impl Chain {
             AccumulatingEvent::SectionInfo(elders_info, _)
             | AccumulatingEvent::NeighbourInfo(elders_info) => {
                 if elders_info.prefix().is_compatible(self.our_prefix())
-                    && elders_info.version() > self.state.new_info.version()
+                    && elders_info.version() > self.state.our_info().version() + 1
```
maqi: shall we use saturating_add at least?

jeanphilippeD: Why? If we are one away from overfow our version we have other problems, don't we?

maqi: Yes. however the problem of overflow is more serious I think?

jeanphilippeD: Is it?
we will also try to increase the version on next member change, how are we going to handle it.
Also why `saturating_add`, how is that more correct, aren't we going to panic otherwise, which is probably what we should do here?

madadam: I think the point is that it will never happen. For example, assuming one churn event per second, it would take 500 billion years to reach that. If safe network is still around, it would probably switch to `u128` versions :)

maqi: there is only option of panic/wrapping_add/saturating_add. And I was only considering that saturating_add seems more suitable than wrapping_add.

And I will take the point that it shall never happen. So good with leave as it is.  :)

PR #1914, opened 2019-11-22
Only connect PeerMap on ConnectedTo

octol:


src/states/bootstrapping_peer.rs
```diff
@@ -268,14 +265,6 @@ impl Base for BootstrappingPeer {
         Transition::Terminate
     }
 
-    fn handle_connected_to(
```
jeanphilippeD: might be worth also removing it from base.rs and call the peer_map connect directly from handle_network_event since it would match the level of abstraction for that function which do so for client.

octol: I agree the mismatch in the level of abstraction stands out, note though that `handle_connected_to` is used in `elder/tests.rs` so the logic would presumably be have to duplicated there then?

jeanphilippeD: ok, probably best not too for now

PR #1915, opened 2019-11-25
API for vaults

### LINKED ISSUE #1912: Bootstrap response from adults should be a rebootstrap variant
Currently, as seen [here](https://github.com/maidsafe/routing/blob/f2b2fdc100582c828e8091156539ae4663c1f52a/src/states/adult.rs#L245) the adults always respond with a bootstrap error response to the joining nodes. The problem with this is that the bootstrap nodes that a joining node knows about might have been an elder at some point but are adults now and getting a rebootstrap variant of the bootstrap response is better than getting an error response (in worst case if all the nodes it knows are now adults in the network).

If adults reply with rebootstrap variant containing the connection info of the section elders then it's a more useful behaviour for the joining nodes who can now contact the section elder(s) and bootstrap procedure continues as normal (either a join or further rebootstrap responses).

Current this logic would be unused in routing so good to have a test case to confirm everything works as expected.

ustulation:
Fixes #1912

### DIFF
--- a/src/state_machine.rs
+++ b/src/state_machine.rs
@@ -153,6 +153,17 @@ impl State {
         }
     }
 
+    pub fn closest_elders_to<'a>(
+        &'a self,
+        name: &XorName,
+    ) -> Option<Box<dyn Iterator<Item = &P2pNode> + 'a>> {
+        match *self {
+            State::Elder(ref state) => Some(Box::new(state.closest_elders_to(name))),
+            State::Adult(ref state) => Some(Box::new(state.closest_elders_to(name))),
+            State::BootstrappingPeer(_) | State::JoiningPeer(_) | State::Terminated => None,
+        }
+    }
+
     pub fn our_connection_info(&mut self) -> Result<ConnectionInfo, RoutingError> {
         state_dispatch!(
             self,
--- a/src/state_machine.rs
+++ b/src/state_machine.rs
@@ -9,7 +9,7 @@
 use crate::{
     action::Action,
     chain::{EldersInfo, GenesisPfxInfo},
-    error::RoutingError,
+    error::{InterfaceError, RoutingError},
     id::{P2pNode, PublicId},
     network_service::NetworkBuilder,
     outbox::EventBox,
@@ -140,6 +140,16 @@ impl State {
             | State::JoiningPeer(_)
             | State::Adult(_)
             | State::Terminated => None,
+        }
+    }
+
+    pub fn matches_our_prefix(&self, name: &XorName) -> Result<bool, RoutingError> {
+        match *self {
+            State::Elder(ref state) => Ok(state.our_prefix().matches(name)),
+            State::Adult(ref state) => Ok(state.our_prefix().matches(name)),
+            State::BootstrappingPeer(_) | State::JoiningPeer(_) | State::Terminated => {
+                Err((InterfaceError::InvalidState).into())
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/states/adult.rs
```diff
@@ -115,13 +115,17 @@ impl Adult {
         Ok(node)
     }
 
+    pub fn closest_elders_to(&self, name: &XorName) -> impl Iterator<Item = &P2pNode> {
+        self.chain.closest_section_info(*name).1.member_nodes()
```
jeanphilippeD: Is that correct?
Probably needs test for this.

I do not think this is correct, the chain as created will still have information as it was at the point of the last Elder change (though it may change). I think this needs to be explicit and return our current elders.

src/states/adult.rs
```diff
@@ -115,13 +115,17 @@ impl Adult {
         Ok(node)
     }
 
+    pub fn closest_elders_to(&self, name: &XorName) -> impl Iterator<Item = &P2pNode> {
```
jeanphilippeD: Maybe the name is not quite accurate enough, it does not suggest that we then need to ask these for the closest elders to? (Could it be something like `closest_known_elders_to`? does not seem ideal).

src/states/elder/mod.rs
```diff
@@ -840,13 +841,13 @@ impl Elder {
             return Transition::Stay;
         }
 
-        let closest_section = self.chain.closest_section(&details.content().destination).0;
-        let conn_infos: Vec<_> = self
+        let closest_section_elders = self
```
jeanphilippeD: Why not use `closest_elders_to` here too?

src/state_machine.rs
```diff
@@ -143,6 +143,27 @@ impl State {
...
+            State::BootstrappingPeer(_) | State::JoiningPeer(_) | State::Terminated => {
+                Err((InterfaceError::InvalidState).into())
+            }
+        }
+    }
+
+    pub fn closest_elders_to<'a>(
+        &'a self,
+        name: &XorName,
+    ) -> Option<Box<dyn Iterator<Item = &P2pNode> + 'a>> {
```
jeanphilippeD: Would Result with InvalidState makes more sense here?

src/state_machine.rs
```diff
@@ -143,6 +143,27 @@ impl State {
...
+            State::Adult(ref state) => Ok(state.our_prefix().matches(name)),
+            State::BootstrappingPeer(_) | State::JoiningPeer(_) | State::Terminated => {
+                Err((InterfaceError::InvalidState).into())
+            }
+        }
+    }
+
+    pub fn closest_elders_to<'a>(
+        &'a self,
+        name: &XorName,
```
jeanphilippeD: XorName is copy. so it is ok to pass by value.
and if you do, I think you can drop explicit lifetimes.

Not sure it is better?

ustulation: `XorName` has nothing to do with the lifetimes here. It's the return type that must be valid for the duration of `self` as it's a ref from something somewhere within `self`.

src/states/adult.rs
ustulation: I didn't understand this point. Shouldn't we be getting info only from the chain as that's the only view we trust ?

jeanphilippeD: My point is we should probably call `self.chain.our_info().member_nodes()`

src/state_machine.rs
jeanphilippeD: Ok, I thought lifetime to self were implicit if no other arguments were a reference, but maybe not in this context.

PR #1919, opened 2019-11-26
Make sure messages are trusted by all nodes during ongoing split

### LINKED ISSUE #1810: Node in lagging part of a split may get unexpected untrusted message
The design of AckMessage requires that nodes are always able to validate trust on reception to avoid having to cache them.

Soak testing showed a very rare failure (30'000 churn tests).
Seed reproduced against fleming at 4310c9cea08587c63850b19ee8c6355dab41f0d7

**Summary of investigation in soak test failure:**
Reveal a bug in our AckMessage/check trust mechanism when split occurs.
Should be relatively straight forward to mitigate.

**Actions to fix it**
- Ensure no message is untrusted: ?Send the proof needed by both side of a split when we detect a split, until we get consensus of AckMessage from both side of the split?
- Ensure If we receive an AckMessage before we split and that is for the other side of the split, we do not process it (?add target prefix in AckMessage?).
- Add regression test: better test case that would trigger that bug much much faster.

**Detail of investigation, extract from log attached.**

What happens 10 split, 100(v24), 101 is not quite there yet.
100 send an AckMessage for 001 at version (001)v24(10:29:01.820723063), and 001 get consensus on it, so nodes that see that consensus will only send proof from (001)v24 and above for message to 100.
001 send an AckMessage for 100 at version (100)v25(10:29:01.820786819).

Elder(a6f083..(10)) receives it. It still has not seen the split, and thinks it is the authority for it, and realize it does not trust it (only trust to (001)v23).
=> This is a break of contract, nodes should be able to assume that all messages send by valid nodes will reach them with needed trust.
==> This could be addressed by ensuring both side of a new split Acked our version before pruning proof we send to either side.
==> (Pb: Could stop prunning for a while if one side of split does not communicate, not an issue, happen rarely and only result in some additional proof for a while).
=> How will that message be processed by the `Elder(a6f083..(10))` if it had enough proof and so trust it as it should.
==> Could that message be confused, should it be disabiguated with the destination prefix
==> Probably: negligable cost since message happen rarely (on churn), so saving a few bytes for the destination prefix is not necessary.

[issue_ack_message_split.txt](https://github.com/maidsafe/routing/files/3648150/issue_ack_message_split.txt)

madadam:
Consider this situation:

A section `P` splits into `P0` and `P1`. Say `P0` already updates some other section `R` about their knowledge. Then `R` sends a message to `P0`. `P1` might not have fully caught itself to the split so it still thinks its prefix is `P` and so it considers itself in authority of that message. But if the message is proven only for `P0`, then `P1` would not trust it.

The fix is to construct the proof so that it is still trusted by both `P0` and `P1` until `P1` also realizes the split and updates `R` about their knowledge.

Closes #1810

Fixes failure of `aggressive_churn` with `features=mock` and seed `[3801065637, 2093151754, 3731311097, 4215731170]`.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -731,22 +731,10 @@ impl Chain {
             && self.is_new(elders_info)
     }
 
-    /// Returns the index of the public key in our_history that will be trusted by the target
-    /// Authority
-    fn proving_index(&self, target: &Authority<XorName>) -> u64 {
-        self.state
-            .their_knowledge
-            .iter()
-            .filter(|(prefix, _)| target.is_compatible(prefix))
-            .map(|(_, index)| *index)
-            .min()
-            .unwrap_or(0)
-    }
-
     /// Provide a SectionProofChain that proves the given signature to the section with a given
     /// prefix
     pub fn prove(&self, target: &Authority<XorName>) -> SectionProofChain {
-        let first_index = self.proving_index(target);
+        let first_index = self.state.proving_index(target);
         self.state.our_history.slice_from(first_index as usize)
     }
 
--- a/src/chain/shared_state.rs
+++ b/src/chain/shared_state.rs
@@ -11,8 +11,8 @@ use super::{
     MemberState, MIN_AGE_COUNTER,
 };
 use crate::{
-    error::RoutingError, id::PublicId, relocation::RelocateDetails, utils::LogIdent, BlsPublicKey,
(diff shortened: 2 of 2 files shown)


madadam:
> Looks good, I assume it fixes the given seed (Is it worth putting seed in commit or PR comment?)

It does. I added the seed into the description.

PR #1920, opened 2019-11-26
Improve log for common failures

jeanphilippeD:
Error when reaching MAX_POLL are common and the error message and logs do not give enough information. Improve log to give the basic starting point needed to investigate failure.

### DIFF
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -331,8 +331,11 @@ impl SignedRoutingMessage {
                 } else {
                     log_or_panic!(
                         LogLevel::Error,
-                        "Combining signatures failed on {:?}!",
-                        self
+                        "Combining signatures failed on {:?}! Part Shares: {:?}, Part Set: {:?}, Partial: {:?}",
+                        self,
+                        partial.shares,
+                        partial.pk_set,
+                        partial,
                     );
                 }
             }
--- a/src/mock/parsec/mod.rs
+++ b/src/mock/parsec/mod.rs
@@ -221,9 +221,30 @@ where
 
     pub fn has_unpolled_observations(&self) -> bool {
         state::with::<T, S::PublicId, _, _>(self.section_hash, |state| {
-            state.has_unconsensused_observations(&self.peer_list)
+            state
+                .unconsensused_observations_for_peers(&self.peer_list)
+                .next()
+                .is_some()
                 || state.get_block(self.first_unconsensused).is_some()
         }) || self.our_unpolled_observations().next().is_some()
+    }
+
+    pub fn unpolled_observations_string(&self) -> String {
+        use itertools::Itertools;
+        let value = state::with::<T, S::PublicId, _, _>(self.section_hash, |state| {
+            format!(
+                "unconsensused_observations_for_peers: {:?}, first_unconsensused block: {:?}",
+                state
+                    .unconsensused_observations_for_peers(&self.peer_list)
+                    .format(", "),
+                state.get_block(self.first_unconsensused),
+            )
+        });
+        format!(
+            "{}, our_unpolled_observations: {:?}",
+            value,
+            self.our_unpolled_observations().format(", "),
+        )
     }
 
     pub fn our_unpolled_observations(&self) -> impl Iterator<Item = &Observation<T, S::PublicId>> {
--- a/src/parsec.rs
+++ b/src/parsec.rs
@@ -244,6 +244,22 @@ impl ParsecMap {
             .flatten()
     }
 
+    #[cfg(feature = "mock_parsec")]
+    pub fn unpolled_observations_string(&self) -> String {
+        let parsec = if let Some(parsec) = self.map.values().last() {
+            parsec
+        } else {
+            return String::new();
+        };
+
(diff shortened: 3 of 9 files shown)


jeanphilippeD:
Rebased

src/parsec.rs
```diff
@@ -244,6 +244,22 @@ impl ParsecMap {
...
+        let parsec = if let Some(parsec) = self.map.values().last() {
+            parsec
+        } else {
+            return String::new();
+        };
+
+        parsec.unpolled_observations_string()
+    }
+
+    #[cfg(all(not(feature = "mock_parsec"), feature = "mock_base"))]
```
maqi: will `#[cfg(not(feature = "mock_parsec"))]` be enough?

src/mock/parsec/mod.rs
```diff
@@ -221,11 +221,31 @@ where
...
+                .next()
+                .is_some()
                 || state.get_block(self.first_unconsensused).is_some()
         }) || self.our_unpolled_observations().next().is_some()
     }
 
+    pub fn unpolled_observations_string(&self) -> String {
+        let value = state::with::<T, S::PublicId, _, _>(self.section_hash, |state| {
+            format!(
+                "unconsensused_observations_for_peers: {:?}, first_unconsensused block: {:?}",
```
maqi: is the info of the `peers` needed?

madadam: I think that is useful, but not sure we need the "first unconsensused block". Isn't that going to be included among the unconsensused observations anyway?

tests/mock_network/utils.rs
```diff
@@ -295,7 +295,22 @@ pub fn poll_and_resend_with_options(nodes: &mut [TestNode], mut options: PollOpt
...
+    }
+
+    let first_node_busy = nodes.iter().find(|node| node_busy(node));
+    let node_name = first_node_busy.map(|node| node.name());
+    let unpolled_string = first_node_busy.map_or(String::new(), |node| {
+        node.inner.unpolled_observations_string()
+    });
+    panic!(
+        "poll_and_resend has been called {} times. first busy: {:?} : {}",
+        MAX_POLL_CALLS, node_name, unpolled_string
```
madadam: ...same here.

src/messages/mod.rs
```diff
@@ -331,8 +331,11 @@ impl SignedRoutingMessage {
                 } else {
                     log_or_panic!(
                         LogLevel::Error,
-                        "Combining signatures failed on {:?}!",
-                        self
+                        "Combining signatures failed on {:?}! Part Shares: {:?}, Part Set: {:?}, Partial: {:?}",
+                        self,
+                        partial.shares,
+                        partial.pk_set,
+                        partial,
```
madadam: Why do we need to output the shares and the pk_set separately when we also output the whole `partial` thing anyway?

src/mock/parsec/mod.rs
jeanphilippeD: No, one is the first unconsused blok by any peer, while `our_unpolled_observations` are only the one we voted for ourselves.

src/parsec.rs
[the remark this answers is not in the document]
jeanphilippeD: Maybe, not sure, if we actually need it for debugging real parsec test issue we can implement it then, so I'm not clear we really need an issue now.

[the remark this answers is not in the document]
madadam: I've seen the `poll_and_resend has been called 1000 times` error with real parsec too, so I'm pretty sure it would be useful.

src/mock/parsec/mod.rs
madadam: My question was about whether we need `first_unconsensused_block`, not whether we need `out_unpolled_observations`. The later one is definitely useful, but the former I think is already included in `unconsensused_observations_for_peers`, so might be redundant.

jeanphilippeD: Is it, then why did we have that condition, was it redundant?
```
             state.has_unconsensused_observations(&self.peer_list)
                || state.get_block(self.first_unconsensused).is_some() 
```

jeanphilippeD: Is it not a consensused observation if it is a block, just happen to not have been polled yet?

madadam: Hm, good point. OK, let's leave it in.

madadam: Yeah, I think `state.get_block(self.first_unconsensused)` is a block that already reached consensus but this node hasn't realized it yet. `has_unconsensused_observations` is for observations that haven't reached consensus yet. So it's fine as is.

src/messages/mod.rs
jeanphilippeD: Because the whole partial omit these as irrelevant (which is true in most cases, but not here).

src/parsec.rs
jeanphilippeD: Because that function is only use for test and so only available for mock base (might have been clippy complaining).

jeanphilippeD:
Fixed review comments

PR #1921, opened 2019-11-27
Only combine valid sigs for accumulated events

### LINKED ISSUE #1917: Check the 1/3 of signatures are all valid and use only the valid subset
Before calling combine, check the 1/3 of signatures are all valid and use only the valid subset + test.

Part of #1779

octol:
Resolves #1917 

TODO: 
- [x] Verify it works
- [ ] Tests


jeanphilippeD:
Superseeded by https://github.com/maidsafe/routing/pull/1965

PR #1927, opened 2019-11-29
feat\routing: export mock quic-p2p for vault use

maqi:
This contains the work of export mock quick-p2p for vault use

### DIFF
--- a/src/mock/quic_p2p/mod.rs
+++ b/src/mock/quic_p2p/mod.rs
@@ -23,7 +23,9 @@ use std::{
     iter,
     net::{IpAddr, SocketAddr},
     rc::Rc,
+    str::FromStr,
 };
+use structopt::StructOpt;
 
 /// Builder for `QuickP2p`.
 pub struct Builder {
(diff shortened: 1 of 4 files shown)


src/mock/quic_p2p/mod.rs
```diff
@@ -132,11 +134,18 @@ impl QuicP2p {
 }
 
 /// Configuration for `QuicP2p`.
-#[derive(Default, Clone)]
+#[derive(Debug, Default, Clone, Eq, PartialEq, Serialize, Deserialize, StructOpt)]
 pub struct Config {
     /// Hard-coded contacts.
+    #[structopt(
```
jeanphilippeD: is that because the real quic_p2p has this?

maqi: just because the vault code heavily use it.  :(

octol: Yep I ran into the same conundrum. I wonder if it's possible to wrap these structs to avoid structopt in routing? I suspect structop might cause a bit of build time overhead?

PR #1931, opened 2019-12-02
forward hop messages as an Adult

jeanphilippeD:
Adults will be long lived with node ageing:
 - Pass Hop messages to our Elders as we may have been demoted.
 - Process signature accumulation and pass messages to our elders as we may have been demoted.

Note:
We are still keeping messages in our backlog so we can process them our_selves, we may have to vote for something as an elder. We will likely have to clean this up further to avoid stale messages being processed and fail signature check.


jeanphilippeD:
Rebased so now a stand alone PR

jeanphilippeD:
Rebased and fixed review comment.

PR #1934, opened 2019-12-03
Fix relocation overrides

madadam:
This PR fixes a bug in the relocation override implementation which could cause a failure in the following scenario:

1. A relocation override is set for nodes in a section
2. A node which is not elder gets relocated from that section, using (and consuming) the relocation override
3. Because that node was not an elder, it does not cause new `SectionInfo` to be voted
4. Another node joins the section and is immediately promoted to elder
5. As it is catching up to the parsec history, it encounters the event that caused the above relocation
6. It processes the relocation, but as it joined later, the relocation overrides are not set on it, so it uses different relocation destination
7. And thus the new node gets out of sync with the rest of the section

The fix is to maintain a map of `original_relocation_destination` to `overriden_relocation_destination` so nodes will always use the same overridden destination no matter when the run the calculation.

Also, this PR refactors the relocation overrides by moving all the logic into a separate `RelocationOverrides` type (backed by a thread-local storage), thus avoiding polluting the production code with test-only code.

### DIFF
--- a/src/relocation.rs
+++ b/src/relocation.rs
@@ -225,7 +225,11 @@ mod overrides {
     }
 
     struct OverrideInfo {
+        // Name that will be used as the next relocation destination.
         next: XorName,
+        // Map of original relocation destinations to the overridden ones. As this map is shared
+        // among all nodes in the network, this assures that every node will pick the same
+        // destination name for a given relocated node no matter when the calculation is performed.
         used: HashMap<XorName, XorName>,
     }
 
--- a/src/relocation.rs
+++ b/src/relocation.rs
@@ -13,11 +13,15 @@ use crate::{
     crypto::{self, signing::Signature},
     error::RoutingError,
     id::{FullId, PublicId},
+    routing_table::Prefix,
     xor_name::{XorName, XOR_NAME_LEN},
     BlsSignature,
 };
 use maidsafe_utilities::serialisation::serialise;
 use std::fmt;
+
+#[cfg(feature = "mock_base")]
+pub use self::overrides::Overrides;
 
 /// Details of a relocation: which node to relocate, where to relocate it to and what age it should
 /// get once relocated.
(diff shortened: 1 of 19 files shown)


tests/mock_network/mod.rs
```diff
@@ -467,19 +451,16 @@ fn simultaneous_joining_nodes_three_section_with_one_ready_to_split() {
         SimultaneousJoiningNode {
             dst_section_prefix: small_prefix,
             src_section_prefix: small_prefix,
-            dst_relocation_interval_prefix: Some(small_prefix_to_add),
```
jeanphilippeD: Does this test still does what it was suppose to do?

madadam: I would say it does what it was doing before this PR, but that might not necessary be what we want. This is because the relocation_interval override wasn't actually used (only the relocation destination).

tests/mock_network/node_ageing.rs
```diff
@@ -256,9 +258,6 @@ fn section_churn(
     assert!(min_section_size < max_section_size);
 
     let mut rng = network.new_rng();
-    let next_relocation_dst = unwrap!(nodes_with_prefix(nodes, prefix).next())
```
jeanphilippeD: How does that bit now work without that?

tests/mock_network/utils.rs
```diff
@@ -510,7 +489,6 @@ pub fn add_connected_nodes_until_split(
         | Event::SectionSplit(..) => (),
         event => panic!("Got unexpected event: {:?}", event),
     });
-    clear_relocation_overrides(nodes);
```
jeanphilippeD: So I think I missed how this work because we use to set the relocation before doing thing and then reset them, and we do not need to clear anymore?

tests/mock_network/node_ageing.rs
madadam: The overrides are now shared, so they would apply to any new node too.

src/relocation.rs
```diff
@@ -122,10 +126,127 @@ impl RelocatePayload {
...
+    use std::{
+        cell::{Cell, RefCell},
+        collections::HashMap,
+    };
+
+    /// Mechanism for overriding relocation destinations. Useful for tests.
+    pub struct Overrides;
+
+    impl Overrides {
+        /// Create new instance of relocation overrides. There can be only one per thread.
```
maqi: `shall be only one instance per thread`?
same to the below

tests/mock_network/mod.rs
jeanphilippeD: Then maybe we broke the test before, because I think this test relied on the node going to the right sub section (like the other relocate during split maybe). Maybe could add an issue to fix it after (with the other test). Maybe the solution is that during test, when a joning node look for its new FullId, if mock_base it will look at some override and generate a full id in that prefix?

src/relocation.rs
```diff
@@ -122,10 +126,127 @@ impl RelocatePayload {
...
+    impl Drop for Overrides {
+        fn drop(&mut self) {
+            self.clear();
+            GUARD.with(|guard| guard.set(false));
+        }
+    }
+
+    struct OverrideInfo {
+        next: XorName,
+        used: HashMap<XorName, XorName>,
```
maqi: better have a comment explains what the key/value are for the `used` map?

tests/mock_network/utils.rs
madadam: I this this was remnant from the time where relocation was part of bootstrapping. It hasn't been necessary for some time.

tests/mock_network/mod.rs
madadam: Yeah, I think we should revise this test to make sure it's doing what we want it to do. But my point was that it wasn't this PR that changed it.

tests/mock_network/node_ageing.rs
```diff
@@ -273,10 +272,6 @@ fn section_churn(
         match churn {
             Churn::Add => {
                 add_node_to_prefix(network, nodes, prefix);
-                unwrap!(nodes.last_mut())
-                    .inner
-                    .set_next_relocation_dst(next_relocation_dst);
```
maqi: do we need to setup relocation here?

src/relocation.rs
madadam: I think "can" implies that it is actually impossible to create more than one instance (because it will panic otherwise), while "shall" to me sounds like you are just advised to not create more than one.

tests/mock_network/node_ageing.rs
```diff
@@ -134,15 +136,11 @@ fn relocate_causing_split() {
...
-    let mut nodes = create_connected_nodes_until_split_with_options(
-        &network,
-        vec![1, 1],
-        ChurnOptions {
-            suppress_relocation: true,
-        },
-    );
-    let oldest_age_counter = oldest_age_counter_after_only_adds(&nodes);
+    let mut nodes = create_connected_nodes_until_split(&network, vec![1, 1]);
+    let oldest_age_counter = node_age_counter(&nodes, 0);
```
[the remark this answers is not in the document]
madadam: We always assumed that `nodes[0]` is the oldest node. We are just more explicit about it now.

tests/mock_network/node_ageing.rs
madadam: See this discussion: https://github.com/maidsafe/routing/pull/1934/files/0a65a186f5434b47b7e2126ccae27b570c38dacb#r353183679

tests/mock_network/mod.rs
madadam: Created the issue: https://github.com/maidsafe/routing/issues/1935

PR #1936, opened 2019-12-03
Support pause/resume also for adults

madadam:
Also improve the pause/resume tests. There are now two: one for simple, no-split case and one for when the split happens in between the pause and resume.

### DIFF
--- a/src/state_machine.rs
+++ b/src/state_machine.rs
@@ -379,15 +379,14 @@ impl StateMachine {
     }
 
     pub fn pause(self) -> Result<PausedState, RoutingError> {
-        // TODO: should we allow pausing from other states too?
-        match self.state {
-            State::Elder(state) => {
-                let mut state = state.pause()?;
-                state.network_rx = Some(self.network_rx);
-                Ok(state)
-            }
-            _ => Err(RoutingError::InvalidStateForOperation),
-        }
+        let mut paused_state = match self.state {
+            State::Elder(state) => state.pause()?,
+            State::Adult(state) => state.pause()?,
+            _ => return Err(RoutingError::InvalidStateForOperation),
+        };
+
+        paused_state.network_rx = Some(self.network_rx);
+        Ok(paused_state)
     }
 
     pub fn resume(mut state: PausedState) -> (mpmc::Sender<Action>, Self) {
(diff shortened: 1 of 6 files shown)


src/state_machine.rs
```diff
@@ -407,6 +413,8 @@ impl StateMachine {
             events: Vec::new(),
         };
 
+        info!("{} - Resume", machine.current());
```
maqi: shall this info more about the `state before resume` ?

src/state_machine.rs
```diff
@@ -379,23 +379,29 @@ impl StateMachine {
...
-                let mut state = state.pause()?;
-                state.network_rx = Some(self.network_rx);
-                Ok(state)
-            }
-            _ => Err(RoutingError::InvalidStateForOperation),
-        }
+        info!("{} - Pause", self.current());
+
+        let mut paused_state = match self.state {
+            State::Elder(state) => state.pause()?,
```
maqi: can we have something like `state.pause(self.network_rx)` so that we can avoid assigning it separately?

tests/mock_network/mod.rs
```diff
@@ -572,13 +572,37 @@ fn carry_out_parsec_pruning() {
...
+    let prefix =
+        add_connected_nodes_until_one_away_from_split(&network, &mut nodes, &[Prefix::default()])
+            [0];
 
+    let new_node_id = FullId::within_range(&mut network.new_rng(), &prefix.range_inclusive());
+    node_pause_and_resume(network, nodes, new_node_id)
+}
+
+// Pause a random node, then add new node with the given id, then resume the paused node and verify
+// everything still works as expected.
```
maqi: so the expected behaviour of this resumed node is that it will be given a new id?

src/state_machine.rs
jeanphilippeD: (Not sure that would be better as it would require the state to know about that detail.)

maqi: it has to, as there is assignment later on immediately ?

jeanphilippeD: neither elder nor adults do

maqi: https://github.com/maidsafe/routing/pull/1936/files#diff-e79e9ddf41b90c06a8194192f55e31deR390 ?

madadam: Yeah, I wouldn't worry too much about details like this. This pause/resume mechanics is not finished yet and there are likely going to be changes, both the the internal details and possibly to the public API as well. There is issue raised for it: https://github.com/maidsafe/routing/issues/1937

src/state_machine.rs
madadam: It's just so we can see in the logs when resume happened. I didn't need it to be more detailed. When we actually need that, we can always add it.

tests/mock_network/mod.rs
madadam: No. What the test does is it pauses a random node, then adds new node (different one), then resumes the paused node. So the `new_node_id` argument is for the newly added node, not the paused node.

maqi: ah, ok. missed that

PR #1939, opened 2019-12-04
Improve `Connected` event handling (backports from 1929)

madadam:
Couple of commits dealing with the handling of `Event::Connected`, backported from #1929 .

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -57,6 +57,15 @@ pub enum ClientEvent {
     },
 }
 
+/// An Event raised as node complete joining
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum ConnectEvent {
+    /// Node first joining the network
+    First,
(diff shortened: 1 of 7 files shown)


src/event.rs
```diff
@@ -86,7 +95,7 @@ pub enum Event {
     /// Our own section has been split, resulting in the included `Prefix` for our new section.
     SectionSplit(Prefix<XorName>),
     /// The client has successfully connected to a proxy node on the network.
-    Connected,
+    Connected(ConnectEvent),
```
maqi: it is referred as `connect_type` in the later on Debug impl .
Also, `ConnectType` seems reflects the usage more accurately?

madadam: My preference would be to have it as separate events: `Event::Connected` and `Event::Relocated`. We discussed it with @jeanphilippeD  and decided it's something that can be addressed later.

PR #1940, opened 2019-12-05
resend Join bootstrap response if outdated elder version

jeanphilippeD:
Refactor code to allow post - processing events
Refactor code to have EldersInfo in BootstrapResponse::Join
Refactor code to have EldersInfo version in JoinRequest
Ensure new name match the same prefix has relocation destination even if split occur
JoiningPeer update and resend JoinRequest if receive BootstrapResponse::Join with newer info (could be the first responder was lagging, or the section updated).
Resend BootstrapResponse::Join if churn, or if receive JoinRequest that could have been for us in a previous section.

Commit comments have more details.

May address issue #1861 

Test:
Verify the fix address the issue found soak testing PR #1929

### DIFF
--- a/src/states/elder/mod.rs
+++ b/src/states/elder/mod.rs
@@ -62,6 +62,16 @@ use crate::messages::Message;
 const TICK_TIMEOUT: Duration = Duration::from_secs(15);
 const GOSSIP_TIMEOUT: Duration = Duration::from_secs(2);
 
+struct CompleteParsecReset {
+    /// The new genesis prefix info.
+    pub gen_pfx_info: GenesisPfxInfo,
+    /// The cached events that should be revoted. Not shared state: only the ones we voted for.
+    /// Also contains our votes that never reached consensus.
+    pub to_vote_again: BTreeSet<NetworkEvent>,
+    /// The events to process and not re-insert.
+    pub event_to_send: Option<Event>,
+}
+
 pub struct ElderDetails {
     pub chain: Chain,
     pub network_service: NetworkService,
@@ -381,11 +391,14 @@ impl Elder {
         self.send_direct_message(node.connection_info(), DirectMessage::ConnectionResponse);
     }
 
-    fn append_our_unpolled_observations(
-        &mut self,
-        mut reset_data: ParsecResetData,
-    ) -> ParsecResetData {
-        reset_data.cached_events.extend(
+    fn complete_parsec_reset_data(&mut self, reset_data: ParsecResetData) -> CompleteParsecReset {
+        let ParsecResetData {
+            gen_pfx_info,
+            mut cached_events,
+            completed_events,
+        } = reset_data;
+
+        cached_events.extend(
             self.parsec_map
                 .our_unpolled_observations()
                 .filter_map(|obs| match obs {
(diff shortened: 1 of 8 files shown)


src/states/elder/mod.rs
```diff
@@ -458,10 +481,59 @@ impl Elder {
...
+            | evt @ AccumulatingEvent::User(_) => {
+                log_or_panic!(LogLevel::Error, "unexpected event {:?}", evt);
+            }
+            AccumulatingEvent::Online(payload) => {
+                self.resend_bootstrap_response_join(&payload.p2p_node);
+            }
+        });
+    }
+
+    // Resend the response with ours our our sibling info in case of split.
```
madadam: ...with our sibling info...

src/states/elder/mod.rs
```diff
@@ -458,10 +481,59 @@ impl Elder {
...
+            | evt @ AccumulatingEvent::Relocate(_)
+            | evt @ AccumulatingEvent::SectionInfo(_, _)
+            | evt @ AccumulatingEvent::NeighbourInfo(_)
+            | evt @ AccumulatingEvent::TheirKeyInfo(_)
+            | evt @ AccumulatingEvent::SendAckMessage(_)
+            | evt @ AccumulatingEvent::User(_) => {
+                log_or_panic!(LogLevel::Error, "unexpected event {:?}", evt);
+            }
+            AccumulatingEvent::Online(payload) => {
+                self.resend_bootstrap_response_join(&payload.p2p_node);
```
madadam: Shouldn't this only happen on split and not on other parsec reset? Otherwise would we be unnecessarily resending the `BootstrapResponse`?

src/states/joining_peer.rs
```diff
@@ -259,6 +259,20 @@ impl Base for JoiningPeer {
         _outbox: &mut dyn EventBox,
     ) -> Result<Transition, RoutingError> {
         match msg {
+            DirectMessage::BootstrapResponse(BootstrapResponse::Join(info)) => {
+                if info.version() > self.elders_info.version() {
+                    if info.prefix().matches(self.full_id().public_id().name()) {
```
madadam: You can just do `...matches(self.name())`.

src/states/elder/mod.rs
jeanphilippeD: Any time churn happens, but not on parsec reset yes, I may have gotten that wrong

madadam: Why does it need to happen on non-split churn?

src/states/elder/mod.rs
```diff
@@ -62,6 +62,16 @@ use crate::messages::Message;
 const TICK_TIMEOUT: Duration = Duration::from_secs(15);
 const GOSSIP_TIMEOUT: Duration = Duration::from_secs(2);
 
+struct CompleteParsecReset {
+    /// The new genesis prefix info.
+    pub gen_pfx_info: GenesisPfxInfo,
+    /// The cached events that should be revoted. Not shared state: only the ones we voted for.
+    /// Also contains our votes that never reached consensus.
+    pub to_vote_again: BTreeSet<NetworkEvent>,
+    /// The events to process and not re-insert.
```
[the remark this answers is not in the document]
jeanphilippeD: Fixed with later commit.

src/states/elder/mod.rs
jeanphilippeD: ...with ours or our sibling's info... ?

src/states/elder/mod.rs
madadam: Ah, is it because some elders might get demoted and they would not re-vote it and so it would not accumulate?

jeanphilippeD: Yes

src/states/elder/mod.rs
madadam: Yes :+1:

jeanphilippeD:
Address small review comments

PR #1942, opened 2019-12-09
backport increase age even for infant from #1929

jeanphilippeD:
Backport a number of fixes from #1929 that stand alone.
The main item is always increase age even for infant as that was leading to stall in test otherwise. 
A number of related fixes are also included as well as the fixes necessary for soak test to pass.

Note:
The relocation_in_progress, churn_in_progress and split_in_progress could probably be improved but is left outside of the scope of that PR.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -64,6 +64,8 @@ pub struct Chain {
     event_cache: BTreeSet<NetworkEvent>,
     /// Marker indicating we are processing churn event
     churn_in_progress: bool,
+    /// Marker indicating we are processing a relocation.
+    relocation_in_progress: bool,
     /// The new dkg key to use when SectionInfo completes. For lookup, use the XorName of the
     /// first member in DKG participants and new ElderInfo. We only store 2 items during split, and
     /// then members are disjoint. We are working around not having access to the prefix for the
@@ -141,6 +143,7 @@ impl Chain {
             chain_accumulator: Default::default(),
             event_cache: Default::default(),
             churn_in_progress: false,
+            relocation_in_progress: false,
             new_section_bls_keys: Default::default(),
         }
     }
@@ -238,6 +241,7 @@ impl Chain {
     pub fn poll(&mut self) -> Result<Option<AccumulatedEvent>, RoutingError> {
         if self.state.handled_genesis_event
             && !self.churn_in_progress
+            && !self.relocation_in_progress
             && !self.state.split_in_progress
         {
             if let Some(event) = self.state.churn_event_backlog.pop_back() {
@@ -297,6 +301,7 @@ impl Chain {
                 self.update_their_knowledge(ack_payload.src_prefix, ack_payload.ack_version);
             }
             AccumulatingEvent::Relocate(_) => {
+                self.relocation_in_progress = false;
                 let signature = self.combine_signatures(proofs);
                 return Ok(Some(AccumulatedEvent::new(event).with_signature(signature)));
             }
@@ -313,7 +318,7 @@ impl Chain {
             _ => false,
         };
 
-        if start_churn_event && self.churn_in_progress {
+        if start_churn_event && (self.churn_in_progress || self.relocation_in_progress) {
             trace!(
                 "{} churn backlog {:?}, Other: {:?}",
                 self,
@@ -390,6 +395,12 @@ impl Chain {
 
     /// Returns the details of the next scheduled relocation to be voted for, if any.
     pub fn poll_relocation(&mut self) -> Option<RelocateDetails> {
+        assert!(
+            !self.relocation_in_progress,
+            "Only one relocation at a time {}",
+            self
+        );
+
         // Delay relocation until all backlogged churn events have been handled and no
         // additional churn is in progress.
         if self.churn_in_progress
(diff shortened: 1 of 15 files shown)


tests/mock_network/node_ageing.rs
```diff
@@ -245,6 +228,24 @@ fn remove_node_from_prefix(nodes: &mut Vec<TestNode>, prefix: &Prefix<XorName>)
...
+// the interval that allow demoting/relocating a node at each step.
+fn section_churn_allowing_relocate(
+    count: usize,
+    network: &Network,
+    nodes: &mut Vec<TestNode>,
+    prefix: &Prefix<XorName>,
+) {
+    // Keep the section size such that relocations can happen but splits can't.
+    // We need NETWORK_PARAMS.elder_size + 1 excluding relocating node for it to be demoted.
+    let min_size = (NETWORK_PARAMS.elder_size + 1) + 1;
```
madadam: Minor: why not `NETWORK_PARAMS.elder_size + 2` ?

src/chain/chain.rs
```diff
@@ -657,6 +657,7 @@ impl Chain {
 
     pub fn find_p2p_node_from_addr(&self, socket_addr: &SocketAddr) -> Option<&P2pNode> {
         self.elders()
+            .chain(self.our_joined_members())
```
madadam: Why do we need both `elders` and `our_joined_members` instead of just `our_joined_members` ?

src/node.rs
```diff
@@ -424,6 +424,11 @@ impl Node {
         self.chain().into_iter().flat_map(Chain::elders)
     }
 
+    /// Returns the members in our section.
+    pub fn our_joined_member_nodes(&self) -> impl Iterator<Item = &P2pNode> {
```
madadam: Minor: keep consistent naming. Either rename this to `our_joined_members` or rename `Chain::our_joined_members` to `Chain::our_joined_member_nodes`.

src/states/elder/mod.rs
```diff
@@ -356,6 +356,7 @@ impl Elder {
             let our_member_connections: HashSet<_> = self
```
madadam: This variable should be renamed as it's not just our members anymore.

src/chain/chain.rs
```diff
@@ -64,6 +64,8 @@ pub struct Chain {
     event_cache: BTreeSet<NetworkEvent>,
     /// Marker indicating we are processing churn event
     churn_in_progress: bool,
+    /// Marker indicating we are processing a relocation.
+    relocation_in_progress: bool,
```
madadam: Why do we need another flag instead of using `churn_in_progress` ?

src/chain/chain.rs
jeanphilippeD: elders also contain neighbour elders

tests/mock_network/node_ageing.rs
jeanphilippeD: it seemed more explicit, we want NETWORK_PARAMS.elder_size + 1  plus 1 relocating node.
But I can change it if that does not make sense

src/chain/chain.rs
madadam: Right. But shouldn't it then be `self.neighbour_elder_nodes().chain(self.our_joined_members())`. Otherwise you are including our elders twice.

src/chain/chain.rs
jeanphilippeD: Well they mean different thing, so maybe they could be made into an enum, but I would not want to overload the churn_in_progress bool.

churn_in_progress really means, EldersInfo is changeing, while relocation_in_progress EldersInfo will not change

tests/mock_network/node_ageing.rs
madadam: Nah, it's fine.

src/chain/chain.rs
jeanphilippeD: As mentionned in PR comment, probably want to clean it up, but maybe not in this PR?

src/chain/chain.rs
jeanphilippeD: yes, make sense.

jeanphilippeD:
Fixed soak test failure when revoting occurs.

jeanphilippeD:
Additional fix to ensure Adult process relocation the same as elders

src/node.rs
jeanphilippeD: Changed the function purpose to be known_nodes

jeanphilippeD:
Addressed review comments, and soak test seem happy for now (force pushed the review comment fixes)

src/node.rs
```diff
@@ -424,6 +424,11 @@ impl Node {
         self.chain().into_iter().flat_map(Chain::elders)
     }
 
+    /// Returns the members in our section and elders we know.
+    pub fn known_nodes(&self) -> impl Iterator<Item = &P2pNode> {
+        self.chain().into_iter().flat_map(Chain::our_joined_members)
```
madadam: Shouldn't this return `Chain::known_nodes` ?

jeanphilippeD: Yes, I messed something up with the review fixes and tests are failing as well

jeanphilippeD:
Force push fixing issues that were introduced
=> We want to include relocating nodes, 
=> known_nodes call correct function

Local soak test seem happy

jeanphilippeD:
Address soak test failure by backporting an additional commit + fix clippy unused

jeanphilippeD:
Added @madadam patch to avoid relocating on initial setup.

PR #1943, opened 2019-12-11
clean up poll & poll relocation

jeanphilippeD:
Clean up chain polling
Clean up poll relocation to be as part of chain polling

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -251,28 +251,57 @@ impl Chain {
                     event,
                     self.state.churn_event_backlog
                 );
-                return Ok(Some(AccumulatedEvent::new(event)));
+                return Ok(Some(event));
             }
         }
 
-        let (event, proofs) = {
-            let opt_event = self
-                .chain_accumulator
-                .incomplete_events()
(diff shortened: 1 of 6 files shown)


src/chain/chain.rs
```diff
@@ -238,41 +238,50 @@ impl Chain {
...
-            if let Some(event) = self.state.churn_event_backlog.pop_back() {
-                trace!(
-                    "{} churn backlog poll {:?}, Others: {:?}",
-                    self,
-                    event,
-                    self.state.churn_event_backlog
-                );
-                return Ok(Some(AccumulatedEvent::new(event)));
-            }
+    pub fn poll_accumulated(&mut self) -> Result<Option<PollAccumulated>, RoutingError> {
```
madadam: Maybe this can be renamed back to `poll` as there should be no confusion with all the other `poll_*` functions being private now?

jeanphilippeD: `poll` is just not that searchable as there are a bunch of thing doing poll, so I'm thinking a bit more explicit name is probably easier to follow, and clearer to read?

madadam: Ok,

PR #1944, opened 2019-12-11
fix/routing: simplify promotion/demotion and reduce number

jeanphilippeD:
We checked for promotion demotion everytime a churn event occured.
If we have some in the backlog, we risked changing multiple times in
quick sucession.

The new mechanism has multiple benefit:
 - Only update elders once if multiple node in churn backlog.
 - Clear interaction between the different polling (events, relocation,
   backloged churn, and Elder churn).
 - Avoid duplication that could break shared state in adult.
 - Shorten the flow since promote Demote does not happen as a result
   of an other event completing.

Test:
clippy + soak test

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -66,6 +66,8 @@ pub struct Chain {
     churn_in_progress: bool,
     /// Marker indicating we are processing a relocation.
     relocation_in_progress: bool,
+    /// Marker indicating that elders may need to change,
+    members_changed: bool,
     /// The new dkg key to use when SectionInfo completes. For lookup, use the XorName of the
     /// first member in DKG participants and new ElderInfo. We only store 2 items during split, and
     /// then members are disjoint. We are working around not having access to the prefix for the
@@ -144,6 +146,7 @@ impl Chain {
             event_cache: Default::default(),
             churn_in_progress: false,
             relocation_in_progress: false,
+            members_changed: false,
             new_section_bls_keys: Default::default(),
         }
     }
@@ -247,6 +250,10 @@ impl Chain {
             return Ok(Some(PollAccumulated::RelocateDetails(details)));
         }
 
+        if let Some(new_infos) = self.promote_and_demote_elders()? {
+            return Ok(Some(PollAccumulated::PromoteDemoteElders(new_infos)));
+        }
+
         let (event, proofs) = match self.poll_chain_accumulator() {
             None => return Ok(None),
             Some((event, proofs)) => (event, proofs),
@@ -326,11 +333,7 @@ impl Chain {
     }
 
     pub fn poll_churn_event_backlog(&mut self) -> Option<AccumulatedEvent> {
-        if self.state.handled_genesis_event
-            && !self.churn_in_progress
-            && !self.relocation_in_progress
-            && !self.state.split_in_progress
-        {
+        if self.can_poll_churn() {
             if let Some(event) = self.state.churn_event_backlog.pop_back() {
                 trace!(
                     "{} churn backlog poll {:?}, Others: {:?}",
@@ -354,7 +357,7 @@ impl Chain {
             _ => false,
         };
 
-        if start_churn_event && (self.churn_in_progress || self.relocation_in_progress) {
+        if start_churn_event && !self.can_poll_churn() {
             trace!(
                 "{} churn backlog {:?}, Other: {:?}",
                 self,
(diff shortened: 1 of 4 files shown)


src/chain/chain.rs
```diff
@@ -540,15 +548,19 @@ impl Chain {
...
     /// Returns a set of EldersInfos to vote for.
-    pub fn promote_and_demote_elders(&mut self) -> Result<Vec<EldersInfo>, RoutingError> {
-        self.assert_no_prefix_change("promote and demote");
+    fn promote_and_demote_elders(&mut self) -> Result<Option<Vec<EldersInfo>>, RoutingError> {
+        if !self.members_changed || !self.can_poll_churn() {
+            // Nothing changed that could impact elder set, or we cannot process it yet.
+            return Ok(None);
+        }
 
         // TODO: the split decision should be based on the number of all members, not just elders.
```
madadam: Not part of this PR, but this comment is outdated and can be deleted.

src/chain/chain.rs
```diff
@@ -540,15 +548,19 @@ impl Chain {
 
     /// Generate a new section info based on the current set of members.
     /// Returns a set of EldersInfos to vote for.
-    pub fn promote_and_demote_elders(&mut self) -> Result<Vec<EldersInfo>, RoutingError> {
-        self.assert_no_prefix_change("promote and demote");
+    fn promote_and_demote_elders(&mut self) -> Result<Option<Vec<EldersInfo>>, RoutingError> {
+        if !self.members_changed || !self.can_poll_churn() {
+            // Nothing changed that could impact elder set, or we cannot process it yet.
+            return Ok(None);
```
madadam: Why not just returning an empty vec?

madadam: Or maybe change the return type to something like:
```rust
enum NewEldersInfos {
    None,
    One(EldersInfo),
    Two(EldersInfo),
}
```
(but with better names)

jeanphilippeD: The empty vec make the call site/intent less clear (poll_accumulated), so in the end I converted it to an option.

The One, Tow would also make the call site (futher up) less clear as instead of looping through all the given info we'd have to have specific code for either case.

src/chain/chain.rs
```diff
@@ -66,6 +66,8 @@ pub struct Chain {
     churn_in_progress: bool,
     /// Marker indicating we are processing a relocation.
     relocation_in_progress: bool,
+    /// Marker indicating that elders may need to change,
```
maqi: can the comment be more explicit, say `Marker indicating elders may need to change, as there is membership change`

src/chain/chain.rs
madadam: You can `impl IntoIterator` for the `NewEldersinfos` thing if you want to be really fancy :) But I'm OK with it as is for now. We can always give it a refactoring pass some time later if we want.

src/chain/chain.rs
jeanphilippeD: It may be because of anything, that would change elders, in particular, if a member become Relocating. so I'm not sure it make sense to be more specific? (It just happen that both membership change and age increase happen on the same events)

src/chain/chain.rs
```diff
@@ -354,7 +357,7 @@ impl Chain {
             _ => false,
         };
 
-        if start_churn_event && (self.churn_in_progress || self.relocation_in_progress) {
+        if start_churn_event && !self.can_poll_churn() {
```
maqi: just to confirm that the `can_poll_churn` function will check two additional flag `handled_genesis_event` and `split_in_progress` from state.
Is this ok?

src/chain/chain.rs
```diff
@@ -433,11 +436,7 @@ impl Chain {
     fn poll_relocation(&mut self) -> Option<RelocateDetails> {
         // Delay relocation until all backlogged churn events have been handled and no
         // additional churn is in progress. Only allow one relocation at a time.
-        if self.relocation_in_progress
-            || self.churn_in_progress
-            || !self.state.churn_event_backlog.is_empty()
-            || !self.state.handled_genesis_event
-        {
+        if !self.can_poll_churn() || !self.state.churn_event_backlog.is_empty() {
```
maqi: just to confirm that the `can_poll_churn` function will check additional flag `split_in_progress` from state.
Is this ok?

src/chain/chain.rs
jeanphilippeD: Yes, it should be the same condition that allow an Online/Offline event to be coming out of that function

src/chain/chain.rs
jeanphilippeD: Yes, really split_in_progress will only be true if churn_in_progress is true also, so does not make much difference.
This set of flags can be cleaned up / improved later.

PR #1946, opened 2019-12-12
feat/TestRng: impl RngCore trait for TestRng

maqi:


src/rng.rs
```diff
@@ -137,6 +137,24 @@ mod test {
...
+        fn next_u64(&mut self) -> u64 {
+            self.0.next_u64()
+        }
+
+        fn fill_bytes(&mut self, dest: &mut [u8]) {
+            self.0.fill_bytes(dest)
+        }
+
+        fn try_fill_bytes(&mut self, _dest: &mut [u8]) -> Result<(), rand_crypto::Error> {
+            Ok(())
```
jeanphilippeD: Probably need to call `self.0.fill_bytes(dest)` before returning Ok, this should not have stub non functional implementation.

dirvine: Does `self.0.fill_bytes(dest)` not return the correct Result? I mean should that not be the last line with no `;` ? @maqi

madadam: `fill_bytes` returns `()`, but we need to return `Result`.

maqi: We can do something like `Ok(self.0.fill_bytes(dest))`
but I think it's bit mis-leading?

jeanphilippeD: Yes, it would be misleading.

PR #1947, opened 2019-12-12
Resend neighbour info we vote for

jeanphilippeD:
Backport some item written in #1929 

Also clean up the new NeighbourInfo resend mechanism so we can eventually use it for all section messages that must result in a consensus, and simplify the resend condition to always as it should only rarely happen and message filter will help.

Note: This is needed on all membership change as one of the new elder may not have received the message and voted for it otherwise (thus malicious elders could stall that event).

### DIFF
--- a/src/states/elder/mod.rs
+++ b/src/states/elder/mod.rs
@@ -26,7 +26,7 @@ use crate::{
     id::{FullId, P2pNode, PublicId},
     messages::{
         BootstrapResponse, DirectMessage, HopMessage, JoinRequest, MessageContent, RoutingMessage,
-        SignedRoutingMessage,
+        SecurityMetadata, SignedRoutingMessage,
     },
     outbox::EventBox,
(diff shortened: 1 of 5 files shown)


src/states/elder/mod.rs
```diff
@@ -542,33 +550,16 @@ impl Elder {
...
-                match self.send_signed_message(&msg) {
-                    Ok(()) => trace!("{} - Resend {:?}", self, msg.routing_message()),
-                    Err(error) => debug!(
-                        "{} - Failed to resend {:?}: {:?}",
-                        self,
-                        msg.routing_message(),
-                        error
-                    ),
-                }
+    fn resend_neighbour_infos(&mut self, _old_pfx: Prefix<XorName>) {
```
madadam: This is no longer used just for `NeighbourInfo` (or at least it doesn't have to be), so this method should be renamed. What about `resend_pending_voted_messages` ? Also the comment should be updated.

jeanphilippeD:
Fixed review comment.

PR #1948, opened 2019-12-16
Perform the elders connection check only when we lose elder

madadam:
In some situation we might not be connected to any of our elders yet and if we then lose some other connection, then check would incorrectly trigger node termination.


jeanphilippeD:
kick appveyor

PR #1949, opened 2019-12-16
Reenable elder unit tests

madadam:
This is a followup to #1929 . Only the last commit is new.


jeanphilippeD:
Will apply to original PR as we really need it for that PR to be complete.

madadam:
This is now part of #1929 .

Issue #1950, opened 2019-12-17
Clean up NodeAdded/Lost


jeanphilippeD:
NodeAdded/Lost with #1929 really mean Adult promotion/demotion.
From consumer both tests and vault, we probably want Adults/Infant notification as well, and also Clean up the name of NodeAdded/NodeLost to reflect new meaning.

cc @madadam

PR #1951, opened 2019-12-17
Clean member change

jeanphilippeD:
Avoid code duplication when member change between adults and Elders.
Remove comment about NodeAdded/NodeLost now tracked by https://github.com/maidsafe/routing/issues/1950

Also remove dead code now tracked by https://github.com/maidsafe/routing/issues/1953

### DIFF
--- a/src/states/common/approved.rs
+++ b/src/states/common/approved.rs
@@ -60,6 +60,14 @@ pub trait Approved: Base {
         outbox: &mut dyn EventBox,
     ) -> Result<(), RoutingError>;
 
+    /// Handle a member relocated.
+    fn handle_member_relocated(
+        &mut self,
+        payload: RelocateDetails,
+        signature: BlsSignature,
+        outbox: &mut dyn EventBox,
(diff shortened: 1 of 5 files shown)


jeanphilippeD:
Add commit to remove dead code tracked by #1953

src/states/common/approved.rs
```diff
@@ -437,6 +437,45 @@ pub trait Approved: Base {
...
+
+        Ok(())
+    }
+
+    fn handle_offline_event(
+        &mut self,
+        pub_id: PublicId,
+        outbox: &mut dyn EventBox,
+    ) -> Result<(), RoutingError> {
+        if !self.chain().can_remove_member(&pub_id) {
```
maqi: just to confirm, will an Adult node pass this check and carries out the block of actions below?

jeanphilippeD: yes, it only depends on the chain. note that this code existed in both adults.rs and elder/mod.rs before, so it should not be a behaviour change.

PR #1954, opened 2019-12-17
Improve churn tests

madadam:
This PR improves and refactors the churn tests.

- Remove code duplication and increase code reuse
- Extract parameters to allow more configurability of the churn tests
- Make the churn tests more strict 
- Increase the number of iterations to make it more likely that interesting cases are hit

### DIFF
--- a/tests/mock_network/churn.rs
+++ b/tests/mock_network/churn.rs
@@ -491,35 +491,33 @@ fn progress_and_verify<R: Rng>(
         warn!("Dropping {:?}", dropped_names);
     }
 
-    let (expectations, relocation_map) = match message_schedule {
+    let expectations = match message_schedule {
         MessageSchedule::AfterChurn => {
-            poll_and_resend(nodes);
-            let added_names = check_added_indices(nodes, added_indices);
-            log_churn_outcome(&added_names, &dropped_names);
-
+            poll_after_churn(nodes, added_indices, dropped_names);
             let expectations = setup_expectations(rng, nodes, network.elder_size());
             poll_and_resend(nodes);
-
-            (expectations, BTreeMap::default())
+            expectations
         }
         MessageSchedule::DuringChurn => {
             let expectations = setup_expectations(rng, nodes, network.elder_size());
-            let relocation_map = RelocationMapBuilder::new(&nodes);
-
-            poll_and_resend(nodes);
-            let added_names = check_added_indices(nodes, added_indices);
-            log_churn_outcome(&added_names, &dropped_names);
-
-            (expectations, relocation_map.build(&nodes))
+            poll_after_churn(nodes, added_indices, dropped_names);
+            expectations
         }
     };
 
-    expectations.verify(nodes, &relocation_map);
+    expectations.verify(nodes);
     verify_invariant_for_all_nodes(network, nodes);
     shuffle_nodes(rng, nodes);
 }
 
-fn log_churn_outcome(added_names: &BTreeSet<XorName>, dropped_names: &BTreeSet<XorName>) {
+fn poll_after_churn(
+    nodes: &mut [TestNode],
+    added_indices: BTreeSet<usize>,
+    dropped_names: BTreeSet<XorName>,
+) {
+    poll_and_resend(nodes);
+    let added_names = check_added_indices(nodes, added_indices);
+
     if !added_names.is_empty() {
         if !dropped_names.is_empty() {
             warn!(
@@ -542,15 +540,23 @@ struct MessageKey {
 }
 
 /// A set of expectations: Which nodes, groups and sections are supposed to receive a request.
-#[derive(Default)]
 struct Expectations {
     /// The Put requests expected to be received.
     messages: HashSet<MessageKey>,
     /// The section or section members of receiving groups or sections, at the time of sending.
     sections: HashMap<Authority<XorName>, HashSet<XorName>>,
+    relocation_map_builder: RelocationMapBuilder,
 }
 
 impl Expectations {
+    fn new(nodes: &[TestNode]) -> Self {
+        Self {
+            messages: HashSet::new(),
+            sections: HashMap::new(),
+            relocation_map_builder: RelocationMapBuilder::new(nodes),
+        }
+    }
+
     /// Sends a request using the nodes specified by `src`, and adds the expectation. Panics if not
     /// enough nodes sent a section message, or if an individual sending node could not be found.
     fn send_and_expect(
@@ -605,7 +611,9 @@ impl Expectations {
     }
 
     /// Verifies that all sent messages have been received by the appropriate nodes.
-    fn verify(mut self, nodes: &mut [TestNode], new_to_old_map: &BTreeMap<XorName, XorName>) {
+    fn verify(mut self, nodes: &mut [TestNode]) {
+        let new_to_old_map = self.relocation_map_builder.build(nodes);
+
         // The minimum of the section lengths when sending and now. If a churn event happened, both
         // cases are valid: that the message was received before or after that. The number of
         // recipients thus only needs to reach a quorum for the minimum number of node at one point.
@@ -715,7 +723,7 @@ fn setup_expectations<R: Rng>(
     // this makes sure we have two different sections if there exists more than one
     let auth_s1 = Authority::Section(!section_name);
 
-    let mut expectations = Expectations::default();
+    let mut expectations = Expectations::new(nodes);
 
     // Test messages from a node to itself, another node, a group and a section...
     expectations.send_and_expect(&content, auth_n0, auth_n0, nodes, elder_size);
--- a/tests/mock_network/churn.rs
+++ b/tests/mock_network/churn.rs
@@ -173,10 +173,7 @@ fn add_nodes<R: Rng>(rng: &mut R, network: &Network, nodes: &mut Vec<TestNode>) 
 
 /// Checks if the given indices have been accepted to the network.
 /// Returns the names of added nodes.
-fn check_added_indices(
-    nodes: &mut Vec<TestNode>,
-    new_indices: BTreeSet<usize>,
-) -> BTreeSet<XorName> {
+fn check_added_indices(nodes: &mut [TestNode], new_indices: BTreeSet<usize>) -> BTreeSet<XorName> {
     let mut added = BTreeSet::new();
     let mut failed = Vec::new();
 
@@ -204,7 +201,7 @@ fn check_added_indices(
 }
 
 // Shuffle nodes excluding the first node
-fn shuffle_nodes<R: Rng>(rng: &mut R, nodes: &mut Vec<TestNode>) {
+fn shuffle_nodes<R: Rng>(rng: &mut R, nodes: &mut [TestNode]) {
     rng.shuffle(&mut nodes[1..]);
 }
 
@@ -444,12 +441,6 @@ fn setup_expectations<R: Rng>(
     expectations
 }
 
-fn send_and_receive<R: Rng>(rng: &mut R, nodes: &mut [TestNode], elder_size: usize) {
-    let expectations = setup_expectations(rng, nodes, elder_size);
-    poll_and_resend(nodes);
-    expectations.verify(nodes, &Default::default());
-}
-
 // Helper to build a map of new names to old names.
 struct RelocationMapBuilder {
     initial_names: Vec<XorName>,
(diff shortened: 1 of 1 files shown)


tests/mock_network/churn.rs
```diff
@@ -20,18 +20,11 @@ use routing::{
 };
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 
-/// Randomly removes some nodes, but <1/3 from each section and never node 0.
-/// Never trigger merge: never remove enough nodes to drop to `elder_size`.
-/// max_per_pfx: limits dropping to the specified count per pfx. It would also
-/// skip prefixes randomly allowing sections to split if this is executed in the same
-/// iteration as `add_nodes_and_poll`.
+/// Randomly removes some nodes, but <1/3 from each section. Never remove enough nodes to drop to
```
jeanphilippeD: Is that 1/3 ok? I think we drop all in one go, 1/3 or all nodes in a section may be including all elders?

madadam: I think the function accounts for it. Maybe the comment needs to be more precise?

tests/mock_network/churn.rs
```diff
@@ -571,30 +577,34 @@ fn messages_during_churn() {
...
+        let (added_indices, dropped_names) = random_churn(
             &mut rng,
             &network,
             &mut nodes,
             churn_probability,
             min_section_num,
             max_section_num,
         );
+
+        if !dropped_names.is_empty() {
```
jeanphilippeD: I would remove the if and do it always, makes it easier to read logs, and simplify test code a tiny bit.
Same for Added nodes.

tests/mock_network/churn.rs
```diff
@@ -571,30 +577,34 @@ fn messages_during_churn() {
...
+        let added_names = check_added_indices(&mut nodes, added_indices);
         if !added_names.is_empty() {
             warn!("Added nodes: {:?}", added_names);
         }
 
-        shuffle_nodes(&mut rng, &mut nodes);
-
         expectations.verify(&mut nodes, &relocation_map);
         verify_invariant_for_all_nodes(&network, &mut nodes);
+        shuffle_nodes(&mut rng, &mut nodes);
```
jeanphilippeD: Why did the shuffle move from before verify to after?

madadam: No strong reason, I just wanted to have the verification bits together. Don't think it changes anything.

tests/mock_network/churn.rs
```diff
@@ -419,6 +410,70 @@ impl RelocationMapBuilder {
...
+        }
+        MessageSchedule::DuringChurn => {
+            let expectations = setup_expectations(rng, nodes, network.elder_size());
+            let relocation_map = RelocationMapBuilder::new(&nodes);
+
+            poll_and_resend(nodes);
+            let added_names = check_added_indices(nodes, added_indices);
+            log_churn_outcome(&added_names, &dropped_names);
+
+            (expectations, relocation_map.build(&nodes))
```
jeanphilippeD: would be nice to avoid the duplication, maybe worth putting that in a sub function `complete_churn` or better named.

```
             poll_and_resend(nodes);
            let added_names = check_added_indices(nodes, added_indices);
            log_churn_outcome(&added_names, &dropped_names); 
```

madadam: That's already done in later commits.

tests/mock_network/churn.rs
```diff
@@ -419,6 +410,70 @@ impl RelocationMapBuilder {
...
+    expectations.verify(nodes, &relocation_map);
+    verify_invariant_for_all_nodes(network, nodes);
+    shuffle_nodes(rng, nodes);
+}
+
+fn log_churn_outcome(added_names: &BTreeSet<XorName>, dropped_names: &BTreeSet<XorName>) {
+    if !added_names.is_empty() {
+        if !dropped_names.is_empty() {
+            warn!(
+                "Simultaneously added {:?} and dropped {:?}",
```
jeanphilippeD: I am not sure this is so helpful, much easier to follow similar sequence in logs.
I find it much easier to look for Dropped line followed by`Added`line, or added and dropped on same line but using same pattern.

so that seem that 

```
fn log_churn_outcome(added_names: &BTreeSet<XorName>, dropped_names: &BTreeSet<XorName>) {
   warn!("Added {:?}, dropped {:?}", added_names, dropped_names);
}
```

Would be simpler and more effective for me?

tests/mock_network/churn.rs
```diff
@@ -492,15 +490,23 @@ struct MessageKey {
 }
 
 /// A set of expectations: Which nodes, groups and sections are supposed to receive a request.
-#[derive(Default)]
 struct Expectations {
     /// The Put requests expected to be received.
     messages: HashSet<MessageKey>,
     /// The section or section members of receiving groups or sections, at the time of sending.
     sections: HashMap<Authority<XorName>, HashSet<XorName>>,
+    relocation_map_builder: RelocationMapBuilder,
```
jeanphilippeD: comment?

tests/mock_network/churn.rs
jeanphilippeD: I cannot really see it, but if you can tell me where and update the comment that would do.

tests/mock_network/churn.rs
```diff
@@ -18,20 +18,256 @@ use routing::{
...
+    // starts with just the root section with `elder_size` nodes.
+    initial_prefix_lens: Vec<usize>,
+    // When are messages sent during each iteration.
+    message_schedule: MessageSchedule,
+
+    // The grow phase lasts until the number of sections reaches this number.
+    grow_target_section_num: usize,
+    // Maximum number of iterations for the churn phase.
+    churn_max_iterations: usize,
+    // Probability that churn occurs for each iteration of the churn phase.
```
maqi: better give an explaination how this factor being used to affect the churn rate?

tests/mock_network/churn.rs
```diff
@@ -50,7 +286,7 @@ fn drop_random_nodes<R: Rng>(
         .collect();
     let mut drop_count: BTreeMap<_, _> = sections.keys().map(|pfx| (*pfx, 0)).collect();
     loop {
-        let i = gen_range(rng, 1, nodes.len());
+        let i = gen_range(rng, 0, nodes.len());
```
maqi: removing the first node could be dangerous.
there were many failures related to it previously

tests/mock_network/churn.rs
```diff
@@ -164,71 +383,101 @@ fn check_added_indices(
...
+
+fn poll_after_churn(
+    nodes: &mut [TestNode],
+    added_indices: BTreeSet<usize>,
+    dropped_names: BTreeSet<XorName>,
+) {
+    poll_and_resend(nodes);
+    let added_names = check_added_indices(nodes, added_indices);
+
+    if !added_names.is_empty() {
```
maqi: will
``` rust
match (added.names.is_empty(), dropped_names.is_empty()) {
    (true, true) =>,
    (true, false) =>,
    (false, true) =>,
    (false, false) =>,
}
```
be more readable ?

tests/mock_network/churn.rs
madadam: Isn't the existing comment enough? What would you suggest to add?

tests/mock_network/churn.rs
madadam: We should be able to remove even the first node, so if it causes issues it's a bug that should be addressed rather than avoided. Also the first node isn't that special anymore. The only difference from normal node is that it accepts bootstrap requests even when the section size is less than `elder_size` while other nodes don't. I'm not even sure why we even have that distinction...

tests/mock_network/churn.rs
```diff
@@ -165,20 +393,10 @@ fn add_nodes<R: Rng>(
             .create();
         if let Some(&pfx) = prefixes.iter().find(|pfx| pfx.matches(&node.name())) {
             assert!(prefixes.remove(&pfx));
-            if skip_some_prefixes && !rng.gen_weighted_bool(prefixes.len() as u32) {
```
jeanphilippeD: Wonder if this still make sense i.e not adding nodes to all prefixes? can be re-added in later PR.

tests/mock_network/churn.rs
```diff
@@ -18,7 +18,254 @@ use routing::{
...
     QUORUM_DENOMINATOR, QUORUM_NUMERATOR,
 };
-use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
+use std::{
+    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
+    usize,
+};
+
+#[test]
+fn aggressive_churn() {
```
jeanphilippeD: Wondering if it would make sense to have a test not sending any messages. for another Pr though.

tests/mock_network/churn.rs
maqi: specify the value shall be 0.0 - 1.0, and an example of 0.8 means 80% ?

tests/mock_network/churn.rs
madadam: Sure, the idea is that we can now have all kinds of tests by just using different parameters. Not sending any messages isn't currently supported though, but should be easy to add.

tests/mock_network/churn.rs
madadam: Hm, I think I removed this accidentally. I'll add it back in a followup PR, as you are suggesting.

tests/mock_network/churn.rs
dirvine: > I'm not even sure why we even have that distinction
I think that is just a hangover from the differentiating first node. We should not have that any more. Now any node should allow bostrap requests. It's muhc simpler now.

tests/mock_network/churn.rs
```diff
@@ -219,71 +456,91 @@ fn check_added_indices(
...
+    added_indices: BTreeSet<usize>,
+    dropped_names: BTreeSet<XorName>,
+) {
+    trace!(
+        "Adding {{{:?}}}, dropping {:?}",
+        added_indices
+            .iter()
+            .map(|index| nodes[*index].name())
+            .format(", "),
+        dropped_names
```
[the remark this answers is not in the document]
jeanphilippeD: `_dropped_names` is only for unused. even if it is for logging it is used.
I'm not having a particular problem with dropped_named being passed here so the log can be the desired shape looks fine to me in context.

tests/mock_network/churn.rs
```diff
@@ -219,71 +456,91 @@ fn check_added_indices(
...
 
-    // Use elder_size rather than section size to prevent collapsing any groups.
-    let max_drop = (unwrap!(nodes[0].inner.elder_size()) - 1)
-        * (QUORUM_DENOMINATOR - QUORUM_NUMERATOR)
-        / QUORUM_DENOMINATOR;
-    assert!(max_drop > 0);
-    let dropped_nodes = drop_random_nodes(rng, nodes, Some(max_drop));
-    warn!("Dropping nodes: {:?}", dropped_nodes);
-    BTreeSet::new()
+    warn!("Added {:?}, dropped {:?}", added_names, dropped_names);
```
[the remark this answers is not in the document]
jeanphilippeD: there are benefit in putting these in higher log level as well since they are useful to understand test tests even if we want to only get warning or errors.

tests/mock_network/churn.rs
```diff
@@ -449,7 +718,11 @@ impl Expectations {
...
 
-fn send_and_receive<R: Rng>(rng: &mut R, nodes: &mut [TestNode], elder_size: usize) {
+fn setup_expectations<R: Rng>(
+    rng: &mut R,
+    nodes: &mut [TestNode],
+    elder_size: usize,
+) -> Expectations {
     // Create random content and pick random sending and receiving nodes.
     let content: Vec<_> = rng.gen_iter().take(100).collect();
     let index0 = gen_elder_index(rng, nodes);
```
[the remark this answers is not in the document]
jeanphilippeD: So it should be the current elder to which we gave the task of sending that message that send it.
So yes, nodes that will become elder should not be sending that message.

tests/mock_network/churn.rs
```diff
@@ -481,237 +754,26 @@ fn send_and_receive<R: Rng>(rng: &mut R, nodes: &mut [TestNode], elder_size: usi
...
-        expectations.send_and_expect(&content, auth_s0, auth_n0, &mut nodes, elder_size);
+impl RelocationMapBuilder {
+    fn new(nodes: &[TestNode]) -> Self {
+        let initial_names = nodes.iter().map(|node| node.name()).collect();
+        Self { initial_names }
+    }
 
-        poll_and_resend(&mut nodes);
-        let new_to_old_map: BTreeMap<XorName, XorName> = nodes
+    fn build(self, nodes: &[TestNode]) -> BTreeMap<XorName, XorName> {
```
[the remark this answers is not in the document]
jeanphilippeD: Could be, not sure, it is already in the struct description though.

tests/mock_network/churn.rs
```diff
@@ -136,47 +391,33 @@ fn count_nodes_by_section(nodes: &[TestNode]) -> HashMap<Prefix<XorName>, Sectio
...
-
-    if !added_nodes.is_empty() {
-        warn!(
-            "    adding {{{}}}",
-            added_nodes.iter().map(|node| node.name()).format(", ")
-        );
+        added_nodes.push(node);
     }
 
     let mut min_index = 1;
```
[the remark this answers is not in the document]
jeanphilippeD: Not sure.

jeanphilippeD:
Superseded by https://github.com/maidsafe/routing/pull/1983

jeanphilippeD:
Superseded by https://github.com/maidsafe/routing/pull/1983

tests/mock_network/churn.rs
jeanphilippeD: Addressed by https://github.com/maidsafe/routing/pull/1985

PR #1957, opened 2019-12-18
bincode/error/refactor

dirvine:
Routing/refactor

Alter deliver group to n+1 / 3
Remove split buffer + extra method that copied this value
remove maidsafe_utilities::serialisation and use bincode directly
remove quic_error and replace with err-derive
use derive-more for simple From derives (mostly for error right now)
actually impl Error for our error types (used err-derive to simplify this)
refactor code + tests to cope with this.

Still a lot to do on the error story though. No point in passing ? everywhere, we need to recover where possible.
Also refactored some methods/functions that relied on bool for passing info


src/message_filter.rs
```diff
@@ -88,7 +87,10 @@ impl<Message: Hash> MessageFilter<Message> {
             .front()
             .map_or(false, |&(_, ref t)| *t <= now)
         {
-            let (hash_code, _) = unwrap!(self.timeout_queue.pop_front());
+            let (hash_code, _) = self
+                .timeout_queue
+                .pop_front()
+                .expect("failed to read expired message");
```
jeanphilippeD: `unwrap!(self.timeout_queue.pop_front(), "failed to read expired message");` ?

PR #1960, opened 2019-12-18
Re-enable drops in churn tests

madadam:
This PR fixes an accidental disabling of node drops in the churn tests. That uncovered some issues which this PR also fixes.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -526,8 +526,8 @@ impl Chain {
         }
     }
 
-    /// Remove a member from our section.
-    pub fn remove_member(&mut self, pub_id: &PublicId) {
+    /// Remove a member from our section. Returns the state of the member before the removal.
+    pub fn remove_member(&mut self, pub_id: &PublicId) -> MemberState {
         self.assert_no_prefix_change("remove member");
         self.members_changed = true;
 
@@ -538,10 +538,12 @@ impl Chain {
             // TODO: Probably should actually remove them
             .filter(|info| info.state != MemberState::Left)
         {
+            let member_state = info.state;
             info.state = MemberState::Left;
             self.state
                 .relocate_queue
                 .retain(|details| &details.pub_id != pub_id);
+            member_state
         } else {
             log_or_panic!(
                 LogLevel::Error,
@@ -549,6 +551,8 @@ impl Chain {
                 self,
                 pub_id
             );
+
+            MemberState::Left
         }
     }
 
@@ -894,10 +898,20 @@ impl Chain {
             && self.is_new(elders_info)
     }
 
-    /// Provide a SectionProofChain that proves the given signature to the section with a given
-    /// prefix
-    pub fn prove(&self, target: &Authority<XorName>) -> SectionProofChain {
-        let first_index = self.state.proving_index(target);
+    /// Provide a SectionProofChain that proves the given signature to the given destination
+    /// authority.
+    /// If `node_knowledge_override` is `Some`, it is used when calculating proof for
+    /// `Authority::Node` instead of the stored knowledge. Has no effect for other authority types.
+    pub fn prove(
+        &self,
+        target: &Authority<XorName>,
+        node_knowledge_override: Option<u64>,
+    ) -> SectionProofChain {
+        let first_index = match (target, node_knowledge_override) {
+            (Authority::Node(_), Some(knowledge)) => knowledge,
+            _ => self.state.proving_index(target),
+        };
+
         self.state.our_history.slice_from(first_index as usize)
     }
 
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -661,6 +661,12 @@ impl Chain {
         self.state.split_in_progress
     }
 
+    /// Returns whether a membership change is in progress (a node leaving, joining or being
+    /// relocated).
+    pub fn membership_change_in_progress(&self) -> bool {
+        self.churn_in_progress || self.relocation_in_progress
+    }
+
     /// Neighbour infos signed by our section
     pub fn neighbour_infos(&self) -> impl Iterator<Item = &EldersInfo> {
         self.state.neighbour_infos.values()
@@ -901,11 +907,6 @@ impl Chain {
     pub fn check_vote_status(&mut self) -> BTreeSet<PublicId> {
         let members = self.our_info().member_ids();
         self.chain_accumulator.check_vote_status(members)
-    }
-
-    /// Returns whether a churning is in progress.
-    pub fn is_churn_in_progress(&self) -> bool {
-        self.churn_in_progress
     }
 
     /// Returns `true` if the given `NetworkEvent` is already accumulated and can be skipped.
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -246,12 +246,14 @@ impl Chain {
             return Ok(Some(PollAccumulated::AccumulatedEvent(event)));
         }
 
-        if let Some(details) = self.poll_relocation() {
-            return Ok(Some(PollAccumulated::RelocateDetails(details)));
-        }
-
+        // Note: it's important that `promote_and_demote_elders` happens before `poll_relocation`,
+        // otherwise we might relocate a node that we still need.
         if let Some(new_infos) = self.promote_and_demote_elders()? {
             return Ok(Some(PollAccumulated::PromoteDemoteElders(new_infos)));
+        }
+
+        if let Some(details) = self.poll_relocation() {
+            return Ok(Some(PollAccumulated::RelocateDetails(details)));
         }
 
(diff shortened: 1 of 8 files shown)


tests/mock_network/churn.rs
```diff
@@ -20,65 +20,105 @@ use routing::{
...
-/// iteration as `add_nodes_and_poll`.
+/// Randomly removes some nodes.
+///
+/// Limits the number of nodes simultaneously dropped from a section such that the section still
+/// remains functional (capable of reaching consensus), also accounting for the fact that any
+/// dropped node might trigger relocation of other nodes. This limit is currently possibly too
+/// conservative and might change in the future, but it still allows removing at least one node per
+/// section.
+///
+/// If `max_per_pfx` is set, never drops more than that number of nodes per section.
```
maqi: `max_per_pfx` sounds we shall `never drops the number of nodes per section below that number` ?

tests/mock_network/churn.rs
```diff
@@ -20,65 +20,105 @@ use routing::{
...
+
+        if node.inner.is_elder() {
+            // Don't drop elder if a non-elder is already scheduled for drop.
+            if section.dropped_other_count > 0 {
+                continue;
+            }
+
+            section.dropped_elder_count += 1;
+        } else {
+            section.dropped_other_count += 1;
```
maqi: just to confirm, so we can drop multiple non-elder members, right?

madadam: Yes, that's the idea. So far it looks like it works, but let's wait for the soak test results.

src/states/common/approved.rs
```diff
@@ -505,8 +505,13 @@ pub trait Approved: Base {
             info!("{} - ignore Relocate: {:?} - not a member", self, details);
         } else {
             info!("{} - handle Relocate: {:?}.", self, details);
-            self.chain_mut().remove_member(&details.pub_id);
-            self.handle_member_relocated(details, signature, outbox)?;
+
+            // Note: it is important that `handle_member_relocated` is called before the member is
+            // actually removed. This is because the member info of the relocated node might still
+            // be needed by `handle_member_relocated`.
+            let pub_id = details.pub_id;
```
jeanphilippeD: I am not really sure I understand, is it because of `get_member_connection_info`?
Ohh it is because of the `node_knowledge` from the Relocating state used by chain::prove.

That implicit dependency seem less than ideal.
Should we have a new Option to pass to the chain::prove function, and get it out of self.chain_mut().remove_member(&pub_id); and pass it down to handle_member_relocated?

Or get the proof in that function and pass it down to handle_member_relocated (still think prove should use optional instead of using that state, that just seem to confusing)?

tests/mock_network/churn.rs
```diff
@@ -20,65 +20,107 @@ use routing::{
...
-        let pfx = nodes[i].our_prefix();
-        let (initial_size, max_drop) = sections[&pfx];
-        if drop_count.is_empty() {
-            break;
-        } else if drop_count.get(&pfx).is_none() {
+
+    // 10% probability that a node will be dropped.
+    let drop_probability = 0.1;
+
+    let mut sections = HashMap::<_, SectionInfo>::new();
```
jeanphilippeD: Maybe worth a sub function named since I had to look in depth to understand what was going on.
maybe:
`fn count_elders_and_others_nodes_by_prefix() -> SectionCountInfo`

tests/mock_network/churn.rs
```diff
@@ -20,65 +20,107 @@ use routing::{
...
     max_per_pfx: Option<usize>,
 ) -> BTreeSet<XorName> {
-    let mut dropped_nodes = BTreeSet::new();
-    let elder_size = |node: &TestNode| unwrap!(node.inner.elder_size());
-    let node_section_size = |node: &TestNode| {
-        node.inner
-            .section_elders(unwrap!(node.inner.our_prefix(), "{}", node.inner))
-            .len()
+    #[derive(Default)]
+    struct SectionInfo {
```
jeanphilippeD: maybe SectionCountInfo as SectionInfo is used in other context.

tests/mock_network/churn.rs
```diff
@@ -20,65 +20,107 @@ use routing::{
...
-        let early_terminate = max_per_pfx.map_or(false, |n| {
-            drop_count[&pfx] >= n || rng.gen_weighted_bool(drop_count.keys().len() as u32)
-        });
-        let normal_terminate =
-            ((drop_count[&pfx] + 1) * 3 >= initial_size) || (drop_count[&pfx] >= max_drop);
-        if early_terminate || normal_terminate {
-            let _ = drop_count.remove(&pfx);
+        let section = unwrap!(sections.get_mut(node.our_prefix()));
+        let elder_size = unwrap!(node.inner.elder_size());
+        let remaining = section.initial_elder_count + section.initial_other_count
```
jeanphilippeD: A function of SectionCountInfo?

tests/mock_network/churn.rs
```diff
@@ -20,65 +20,107 @@ use routing::{
...
+            - section.dropped_other_count
+            - section.dropped_elder_count;
+        let dropped = section.dropped_elder_count + section.dropped_other_count;
```
jeanphilippeD: same?

src/states/elder/mod.rs
```diff
@@ -1697,14 +1697,15 @@ impl Approved for Elder {
             );
             return Ok(());
         }
-        if self.chain.is_churn_in_progress() {
-            debug!(
-                "{} - Trying to carry out parsec pruning during churning.",
+        if self.chain.membership_change_in_progress() {
```
jeanphilippeD: I don't think this is correct, the reason we can ignore prune when elder member change is a new parsec is created at the end. not the case for relocation.

So probably what need to happen is relocation either work over pruned parsec, or prune still work.

src/chain/chain.rs
```diff
@@ -246,14 +246,16 @@ impl Chain {
             return Ok(Some(PollAccumulated::AccumulatedEvent(event)));
         }
 
-        if let Some(details) = self.poll_relocation() {
-            return Ok(Some(PollAccumulated::RelocateDetails(details)));
-        }
-
+        // Note: it's important that `promote_and_demote_elders` happens before `poll_relocation`,
+        // otherwise we might relocate a node that we still need.
```
jeanphilippeD: yes, that actually make more sense.

tests/mock_network/churn.rs
madadam: It's "max drops per prefix". I'll rename the var to make it more obvious.

src/chain/chain.rs
madadam: Should I also make this order explicit somehow?

jeanphilippeD: The comment seem to be explicit enough to me?

madadam: I'm thinking we are making the order explicit (statically enforced) here: https://github.com/maidsafe/routing/pull/1960#discussion_r359428920, so we might as well do it here too. Or do you think it's not worth it?

jeanphilippeD: The other one had an actual data dependency that was implicit, this one does not.
I'm not sure that would be worth trying to add something here. (Also the other was across multiple functions)

src/states/elder/mod.rs
jeanphilippeD: Considering the tests are passing, would it be worth raising an issue for this and merge this PR before you are on holiday? @madadam

madadam: OK, will do. But I might still be able to raise it as a separate PR before my leave.

madadam: https://github.com/maidsafe/routing/issues/1963

src/chain/chain.rs
```diff
@@ -894,10 +898,20 @@ impl Chain {
...
+    /// If `node_knowledge_override` is `Some`, it is used when calculating proof for
+    /// `Authority::Node` instead of the stored knowledge. Has no effect for other authority types.
+    pub fn prove(
+        &self,
+        target: &Authority<XorName>,
+        node_knowledge_override: Option<u64>,
+    ) -> SectionProofChain {
+        let first_index = match (target, node_knowledge_override) {
+            (Authority::Node(_), Some(knowledge)) => knowledge,
+            _ => self.state.proving_index(target),
```
jeanphilippeD: Should we remove this if in proving_index now?

```
    pub fn proving_index(&self, target: &Authority<XorName>) -> u64 {
        if let Authority::Node(name) = target {
            if let Some(info) = self.our_members.get(name) {
                if let MemberState::Relocating { node_knowledge } = info.state {
                    // Relocating node, or adult nodes need to be special cased:
                    // Use the proving index we last know for them.
                    return node_knowledge;
                }
            }
        }

PR #1965, opened 2019-12-19
Only combine valid sigs for accumulated events

### LINKED ISSUE #1917: Check the 1/3 of signatures are all valid and use only the valid subset

jeanphilippeD:
We need to create a valid signature. With an accumulated event, we will
always have threshold + 1 valid signature share. We may have invalid
shares as well that we need to ignore when combining signature to get
a valid signature.

Resolves #1917 
Superseed https://github.com/maidsafe/routing/pull/1921

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -24,6 +24,8 @@ use crate::{
 };
 use itertools::Itertools;
 use log::LogLevel;
+use maidsafe_utilities::serialisation::serialise;
+use serde::Serialize;
 use std::cmp::Ordering;
 use std::{
     collections::{btree_map::Entry, BTreeMap, BTreeSet},
@@ -318,9 +320,12 @@ impl Chain {
             AccumulatingEvent::AckMessage(ref ack_payload) => {
                 self.update_their_knowledge(ack_payload.src_prefix, ack_payload.ack_version);
             }
-            AccumulatingEvent::Relocate(_) => {
+            AccumulatingEvent::Relocate(ref relocate_details) => {
                 self.relocation_in_progress = false;
-                let signature = self.combine_signatures(proofs);
+                let signature = Some(
+                    self.check_and_combine_signatures(relocate_details, proofs)
+                        .ok_or(RoutingError::InvalidRelocation)?,
+                );
                 return Ok(Some(AccumulatedEvent::new(event).with_signature(signature)));
             }
             AccumulatingEvent::Online(_)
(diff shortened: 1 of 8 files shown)


src/chain/chain.rs
```diff
@@ -1173,15 +1178,45 @@ impl Chain {
...
+    pub fn check_and_combine_signatures<S: Serialize + Debug>(
+        &self,
+        signed_payload: &S,
+        proofs: AccumulatingProof,
+    ) -> Option<BlsSignature> {
+        let signed_bytes = serialise(signed_payload)
+            .map_err(|err| {
+                log_or_panic!(
+                    LogLevel::Error,
+                    "{} Failed to serialise accumulated event: {:?} for {:?}",
```
madadam: Should this be something like "Failed to serialise payload", since the function is generic and there is no requirement the payload actually is `AccumulatedEvent`?

src/chain/chain.rs
```diff
@@ -1173,15 +1178,45 @@ impl Chain {
...
+        proofs
+            .check_and_combine_signatures(
+                self.our_info(),
+                self.our_section_bls_keys(),
+                &signed_bytes,
+            )
+            .or_else(|| {
+                log_or_panic!(
+                    LogLevel::Error,
+                    "{} Failed to combine signatures for accumulated event: {:?}",
```
madadam: Same here.

src/error.rs
```diff
@@ -61,6 +61,8 @@ pub enum RoutingError {
     UntrustedMessage,
     #[error(display = "A new SectionInfo is invalid.")]
     InvalidNewSectionInfo,
+    #[error(display = "A Relocation is invalid.")]
+    InvalidRelocation,
```
madadam: Should this be more precise, i.e., `InvalidRelocationSignature` or something like that? Alternatively, why not just use the `InvalidSignature` variant?

jeanphilippeD: This error is similar to `InvalidNewSectionInfo`. Thee problem is if the signature is not correct, the whole Relocation will fail since we will ignore it. So I'm thinking it make sense as it is?

src/chain/chain.rs
jeanphilippeD: While the payload is somewhat generic, it is only for accumulated event.
This is re-enforced by the other parameter being an AccumulatingProof. The assumption that it is not ok to not be able to combine signature is because it is an accumulated event. So I don't think it make sense to update the log statement.

what do you think?

src/error.rs
madadam: OK.

src/chain/chain.rs
madadam: Shouldn't then the function take only `AccumulatedEvent`?

jeanphilippeD: We have already destructured. Also, it is not the AccumulatedEvent but the RelocateDetails or ElderInfo that we serialize to sign

madadam: OK. I mean it feels a bit wrong, but it's a very minor issue, so I'll accept it.

PR #1966, opened 2019-12-19
Remove routing_table

madadam:
This PR removes the routing_table module and moves the bits that we still need to more appropriate places. It also extracts the routing-table-like logic from `Chain` to a separate module so it's easy to unit-test it. The original routing_table tests are copied over to this module, but are commented out. They could be used as a guide when writing the new unit tests. The unit tests themselves are still TODO (which is why this is only draft PR).


jeanphilippeD:
Mostly superseeded by https://github.com/maidsafe/routing/pull/1972 except commits extracting `Router` as I was not 100% sure.

PR #1968, opened 2019-12-20
Remove SPLIT_BUFFER and safe_section size indirection.

dirvine:
SPLIT_BUFFER has no meaning or purpose. When the network is going to split it will, there is no concept of buffer here. 
Also, the indirection of safe_section_size min_section_size and min_split_size has been simplified. When the RT is removed then it will be safe_section_size only

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -37,10 +37,6 @@ use std::{
 #[cfg(feature = "mock_base")]
 use crate::crypto::Digest256;
 
-/// Amount added to `min_section_size` when deciding whether a bucket split can happen. This helps
-/// protect against rapid splitting and merging in the face of moderate churn.
-const SPLIT_BUFFER: usize = 1;
-
 /// Returns the delivery group size based on the section size `n`
 pub fn delivery_group_size(n: usize) -> usize {
     // this is an integer that is ≥ n/3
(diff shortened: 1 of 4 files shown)


src/node.rs
```diff
@@ -469,7 +469,7 @@ impl Node {
     ///
     /// Only if we have a chain (meaning we are elders) we will process this API
     pub fn min_split_size(&self) -> Option<usize> {
-        self.chain().map(|chain| chain.min_split_size())
+        self.chain().map(|chain| chain.safe_section_size())
```
jeanphilippeD: Did you mean to update the function name to here min_split_size - > safe_section_size?

dirvine: Ah yes, the min _split _size was a wrapper around safe_section_size + SPLIT_BUFFER, so I removed that with the split buffer.  Later we can remove min_seciotn_size and just use safe_section_size or vice versa as we have 2 functions doing the same thing really.

dirvine:
Yes @maqi  these can be all the same now. I was suggesting we do that later, but happy to add to this PR is you guys think it is best placed o do it now.

src/node.rs
jeanphilippeD: What I mean is that it is now a forwarding function called min_split_size that calls safe_section_size which is confusing, so this function should be renamed safe_section_size

dirvine: Ah yes, the RT min_section_size is forwarding, but will go away with the RT removal. We could change it here, but there is already a PR in play to remove RT.

src/chain/chain.rs
```diff
@@ -1251,8 +1251,8 @@ impl Chain {
             });
 
         // If either of the two new sections will not contain enough entries, return `false`.
-        let min_split_size = self.safe_section_size();
-        Ok(our_new_size >= min_split_size && sibling_new_size >= min_split_size)
+        let safe_section_size = self.safe_section_size();
```
jeanphilippeD: commit comment seem to have lines that should not be there?

PR #1971, opened 2019-12-23
Rand upgrade - WIP preview 

dirvine:


src/parsec.rs
```diff
@@ -387,7 +387,7 @@ fn create(rng: &mut MainRng, full_id: FullId, gen_pfx_info: &GenesisPfxInfo) ->
             &gen_pfx_info.first_info.member_ids().copied().collect(),
             gen_pfx_info.first_state_serialized.clone(),
             ConsensusMode::Single,
-            Box::new(rng::new_from(rng)),
+            Box::new(rng),
```
jeanphilippeD: I'm thinking we still want to create a new rng here?
What happen, are we trying to box a mut reference?

src/xor_name.rs
```diff
@@ -262,15 +263,17 @@ impl AsRef<XorName> for XorName {
...
-    use rand;
+    use rand::Rng;
     use std::cmp::Ordering;
     use unwrap::unwrap;
 
     #[test]
-    fn serialisation_xor_name() {
-        let obj_before: XorName = rand::random();
-        let data = unwrap!(serialise(&obj_before));
+    fn serialisation_xor_name() -> Result<(), RoutingError> {
```
jeanphilippeD: Still not clear that it is a better way of doing things, so far it seem to be less useful than the original.

src/xor_name.rs
```diff
@@ -298,28 +301,31 @@ mod tests {
 
     #[test]
     fn xor_name_equal_assertion() {
-        let type1: XorName = rand::random();
+        let mut rng = rand::thread_rng();
+        let type1: XorName = rng.gen::<XorName>();
         let type1_clone = type1;
-        let type2: XorName = rand::random();
+        let type2: XorName = rng.gen::<XorName>();
```
jeanphilippeD: do we need to declare the type twice? won't `rng.gen();` work?

src/rng.rs
```diff
@@ -213,12 +210,8 @@ mod test {
...
-    impl Rand for Seed {
-        fn rand<R: Rng>(rng: &mut R) -> Self {
-            // Note: the `wrapping_add` trick is a workaround for what seems to be a weakness in
-            // `XorShiftRng`. Without it we would sometimes end up with multiple rngs producing
-            // identical values.
-            // The idea is taken from: https://github.com/maidsafe/maidsafe_utilities/blob/24dfcbc6ee07a14bf64f3bc573f68cea01e06862/src/seeded_rng.rs#L92
+    impl Distribution<Seed> for Standard {
+        fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> Seed {
             Self([
                 rng.next_u32().wrapping_add(rng.next_u32()),
```
jeanphilippeD: If the comment goes, shouldn't the trick go too?
But since we are still using XorShift, shouldn't the comment stay?

src/rng.rs
```diff
@@ -48,15 +43,18 @@ impl<R: CryptoRng> rand_crypto::CryptoRng for RngCompat<R> {}
 // Rng implementation used in production. Uses `OsRng` for maximum cryptographic security.
 #[cfg(not(any(test, feature = "mock_base")))]
 mod implementation {
-    pub use rand::OsRng as MainRng;
+    // Parsec has a very old rand dependecy. This osRng is required to interact wiht that, until
+    // parsec upgrades
```
jeanphilippeD: Why is that?

jeanphilippeD: Also these `RngCompat` types are pretty trivial, so you can you one for now that implement I guess Rng for the right rand crate version. and then call `Box::new(ParsecRngCompat(rng::new_from(rng)))`

src/parsec.rs
dirvine: Ah ye, `OsRng` now (since 0.7) does not return a `Result`. On first use it constructs itself. So the use is a bit simpler now.
I was a bit torn with rand and all it's badly breaking changes with versions. Dalek is now 0.7, threshold_crypt 0.6.5 and parsec is 0.6.5 IIRC. They do not interoperate well at all. Then was thinking why force rand via libs and the only reason is deterministic everything (it's not though). However, I am not sure it matters a node has the exact same name between runs as the name does not matter as long as it fits a section. Same with threshold_crypto (it allows for rand ignoring by having a public random call). Again I am not sure the exact same keyshares matters there, we must assume that crate works. 
So leaves parsec and we can control that, but again do we need to from routing?

src/xor_name.rs
dirvine: Yes, it's still in the air IMO. Without backtrace it is perhaps less useful. I am still pondering this one and we should have a deeper chat. It does not belong in this PR though (I will remove it). 
It seems the more complex the test the more reason to use result based returns with accurate errors. Smaller unit test, perhaps should just panic/assert (preferred) Certainly unfinished discussion here I recon.

src/xor_name.rs
dirvine: I think this test is perhaps useless. It seems to only test `.clone` works and generating 2 random XorNames will be guaranteed different (which it won't). Personally I would remove this test.

src/rng.rs
dirvine: Ah yes it should, I agree, good catch

src/rng.rs
dirvine: I think this is something I do need to do, I fought this error for many hours this weekend. It seems newer versions of OsRng impl traits the old one does not have. So ends up in a cycle of messiness.

jeanphilippeD:
For reference: here is the branch that can be used to "solve" with hacks the compile error (need rebase now, but should be fairly small) https://github.com/jeanphilippeD/routing/tree/rand_adaptor

PR #1973, opened 2019-12-24
fix/routing: avoid unwrap! in production

jeanphilippeD:


jeanphilippeD:
Add two commit to ensure unwrap cannot be used in production (and fixing one missed usage).

src/lib.rs
```diff
@@ -220,6 +220,9 @@ pub(crate) use self::{
 #[cfg(feature = "mock_crypto")]
 pub(crate) use self::mock::crypto;
 
+#[cfg(any(test, feature = "mock_base"))]
+pub(crate) use unwrap::unwrap;
```
maqi: better put a comment here to explain why we do so.
so that we don't `change it back` later on accidently.

jeanphilippeD: I'm thinking someone wanting to change it back would look at the history of why it is this way, or the reviewers would question it.

PR #1974, opened 2019-12-24
Use safe_vault GitHub action

jeanphilippeD:
Copy the same github action as safe_vault to run all our clippy/test target use caching and ease maintenance.


jeanphilippeD:
Rebased once the `PR` github action all passed.

jeanphilippeD:
Appveyor is having the same issue as before should not block this merge though.

PR #1980, opened 2019-12-30
chore/ci: fix RUSTFLAGS in GHA being overriden by scripts

m-cat:


m-cat:
if you guys want to keep the RUSTFLAGS in the scripts we can do something like `export RUSTFLAGS=RUSTFLAGS:"-C ..."`

PR #1983, opened 2020-01-02
Improve churn tests

jeanphilippeD:
Supersede https://github.com/maidsafe/routing/pull/1954

This PR improves and refactors the churn tests.

- Remove code duplication and increase code reuse
- Extract parameters to allow more configurability of the churn tests
- Make the churn tests more strict 
- Increase the number of iterations to make it more likely that interesting cases are hit
- JP: Avoid recursive split by keeping add balanced.

### DIFF
--- a/tests/mock_network/churn.rs
+++ b/tests/mock_network/churn.rs
@@ -137,14 +137,7 @@ fn count_nodes_by_section(nodes: &[TestNode]) -> HashMap<Prefix<XorName>, Sectio
 }
 
 /// Adds node per existing prefix using a random proxy. Returns new node indices.
-/// skip_some_prefixes: skip adding to prefixes randomly to allowing sections to merge
-/// when this is executed in the same iteration as `drop_random_nodes`.
-fn add_nodes<R: Rng>(
-    rng: &mut R,
-    network: &Network,
-    nodes: &mut Vec<TestNode>,
-    skip_some_prefixes: bool,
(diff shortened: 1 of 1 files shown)


tests/mock_network/churn.rs
```diff
@@ -136,47 +391,57 @@ fn count_nodes_by_section(nodes: &[TestNode]) -> HashMap<Prefix<XorName>, Sectio
...
+        } else {
+            *bit_0 += 1;
+        }
+    }
+
+    counts
+        .into_iter()
+        .map(|(prefix, (bit_0, bit_1))| {
+            let next_bit_with_fewer_members = bit_0 > bit_1;
+            prefix.pushed(next_bit_with_fewer_members)
```
maqi: the comment of the function says: an appended bit: `0` if `bit` is `false`, and `1` if `bit` is `true`
which means when bit_0 less than bit_1, a `0` will be appended.
is it expected, or I missed something?

jeanphilippeD: yes, we want to add new member in the sub prefix with smallest members to keep things balanced.

so if the count of member in bit_0 is less than the one in bit_1, we add new member to prefix with next bit being 0.

maqi: ah, right. thx for clarification.

jeanphilippeD:
JP: Approved commits by @madadam

PR #1984, opened 2020-01-03
fix/routing: handle recursive split and remove test work around.

jeanphilippeD:
add_connected_nodes_until_split add all the nodes of leaf prefixes, thus
triggering recursive prefix case. Ensure that this split is not delayed
until the next online/offline event.

It is now trivial to handle recursive split in code, so remove the work
arround that was adding more nodes than needed, leaving additional split
yet to occur.

Fix the relocation suppression when adding node to section so it handle
prefix for which only a parent exists.

Test:
Run soak test + clippy.
Verify messages_during_churn [2481487885, 755787713, 2183041345, 3896795195]
does not fail, and still split properly (no relocation occur anymore).

### DIFF
--- a/src/relocation.rs
+++ b/src/relocation.rs
@@ -211,6 +211,16 @@ mod overrides {
             self.set(src_prefix, src_prefix.name())
         }
 
+        /// Suppress relocations from the given source prefix and its parent prefixes.
+        pub fn suppress_self_and_parents(&mut self, mut src_prefix: Prefix<XorName>) {
+            self.suppress(src_prefix);
+
+            while !src_prefix.is_empty() {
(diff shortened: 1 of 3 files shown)


src/relocation.rs
```diff
@@ -211,6 +211,16 @@ mod overrides {
             self.set(src_prefix, src_prefix.name())
         }
 
+        /// Suppress relocations from the given source prefix.
+        pub fn suppress_and_parents(&mut self, mut src_prefix: Prefix<XorName>) {
```
octol: The name is slightly confusing, what about `suppress_self_and_parents` or something similar?

jeanphilippeD:
Fixed review comment

PR #1985, opened 2020-01-03
fix/routing: treat bootstrapping to any node the same

jeanphilippeD:
Addressing comment https://github.com/maidsafe/routing/pull/1954#discussion_r361638168 from PR merged as https://github.com/maidsafe/routing/pull/1983.

Remove first node special case if not enough elders.

Test:
Update unit test
soak test + clippy

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -65,8 +65,3 @@ pub enum RoutingError {
     #[error(display = "An Elder DKG result is invalid.")]
     InvalidElderDkgResult,
 }
-
-#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
-pub enum BootstrapResponseError {
-    NotApproved,
-}
--- a/src/error.rs
+++ b/src/error.rs
@@ -69,5 +69,4 @@ pub enum RoutingError {
 #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
 pub enum BootstrapResponseError {
     NotApproved,
-    TooFewPeers,
 }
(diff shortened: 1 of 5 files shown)


src/error.rs
```diff
@@ -69,5 +69,4 @@ pub enum RoutingError {
 #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
 pub enum BootstrapResponseError {
```
maqi: probably no need to make this as `enum` anymore ? 
as you will only be `NotApproved`

src/states/elder/tests.rs
```diff
@@ -503,23 +503,25 @@ fn when_accumulate_offline_and_start_dkg_and_section_info_then_node_is_removed_f
...
 #[test]
-fn accept_previously_rejected_node_after_reaching_elder_size() {
+fn accept_node_before_and_after_reaching_elder_size() {
     // Set section size to one less than the desired number of the elders in a section. This makes
     // us reject any bootstrapping nodes.
     let mut elder_test = ElderUnderTest::new(ELDER_SIZE - 1);
     let node = JoiningNodeInfo::with_addr(&mut elder_test.rng, "198.51.100.0:5000");
 
-    // Bootstrap fails for insufficient section size.
+    // Bootstrap succeed even with to few elders.
```
maqi: `too` few

src/states/elder/tests.rs
```diff
@@ -503,23 +503,25 @@ fn when_accumulate_offline_and_start_dkg_and_section_info_then_node_is_removed_f
...
     elder_test.handle_bootstrap_request(*node.public_id(), node.connection_info());
-    assert!(!elder_test.is_connected(&node.connection_info().peer_addr));
+    assert!(elder_test.is_connected(&node.connection_info().peer_addr));
+
+    let node = JoiningNodeInfo::with_addr(&mut elder_test.rng, "198.51.100.1:5000");
```
maqi: this is actually a new node.
which means the node at address 198.51.100.0 could be dropped? (or at least indicates that as using the same variable name.)
maybe better use a different name?

jeanphilippeD:
Fixed review comments.

PR #1986, opened 2020-01-06
Don't fail if a user messages for non-elder is not delivered

madadam:
A recipient might be demoted before receiving the message and currently we raise the `MessageRecieved` event only as elders. This might change in the future.

### DIFF
--- a/tests/mock_network/churn.rs
+++ b/tests/mock_network/churn.rs
@@ -574,9 +574,9 @@ struct MessageKey {
     dst: Authority<XorName>,
 }
 
-/// A set of expectations: Which nodes, groups and sections are supposed to receive a request.
+/// A set of expectations: Which nodes, groups and sections are supposed to receive a message.
 struct Expectations {
-    /// The Put requests expected to be received.
+    /// The message expected to be received.
     messages: HashSet<MessageKey>,
     /// The section or section members of receiving groups or sections, at the time of sending.
     sections: HashMap<Authority<XorName>, HashSet<XorName>>,
@@ -594,7 +594,7 @@ impl Expectations {
         }
     }
 
-    /// Sends a request using the nodes specified by `src`, and adds the expectation. Panics if not
+    /// Sends a message using the nodes specified by `src`, and adds the expectation. Panics if not
     /// enough nodes sent a section message, or if an individual sending node could not be found.
     fn send_and_expect(
         &mut self,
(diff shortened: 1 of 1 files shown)


tests/mock_network/churn.rs
```diff
@@ -723,8 +723,21 @@ impl Expectations {
...
 
         for key in self.messages {
-            // All received messages for single nodes were removed: if any are left, they failed.
-            assert!(key.dst.is_multiple(), "Failed to receive request {:?}", key);
+            if let Authority::Node(curr_name) = key.dst {
+                let orig_name = new_to_old_map.get(&curr_name).unwrap_or(&curr_name);
+
+                if nodes
+                    .iter()
+                    .any(|node| node.inner.is_elder() && node.name() == *orig_name)
```
jeanphilippeD: This new code also seem to handle the case were a node was dropped.
How was it handled before, did elder destination not drop?

madadam: I think dropping happens before message sending, so we never sent a message to dropped node.

tests/mock_network/churn.rs
```diff
@@ -675,7 +675,7 @@ impl Expectations {
             })
             .collect();
         let mut section_msgs_received = HashMap::new(); // The count of received section messages.
-        for node in nodes {
+        for node in nodes.iter_mut() {
```
maqi: why we now requires a `mut` iterator?
looks like there is no change requiring that?

tests/mock_network/churn.rs
```diff
@@ -723,8 +723,21 @@ impl Expectations {
         }
 
         for key in self.messages {
-            // All received messages for single nodes were removed: if any are left, they failed.
-            assert!(key.dst.is_multiple(), "Failed to receive request {:?}", key);
+            if let Authority::Node(curr_name) = key.dst {
```
maqi: the test always only send message to `new name`s ?

tests/mock_network/churn.rs
madadam: [`try_next_ev`](https://github.com/maidsafe/routing/pull/1986/files#diff-99ea4a9327d9bac124927e60fdbc4e98R682) requires `mut`.

tests/mock_network/churn.rs
madadam: No, we send to old name, so this is buggy. Good catch!

tests/mock_network/churn.rs
maqi: but this is not a new change. `try_next_ev` was there previously.

madadam: No, but `nodes` is a mutable reference, so `for node in nodes` would consume it, but I need to use it again later.

maqi: Ah, Ok, I get it.

PR #1981, opened 2020-01-01
routing/api  Restricting visibility of private methods and types WIP

dirvine:
Initial commit. 

TBD
Refactor for all the mock features etc. Improve redability of the API and lib.rs

PR #1991, opened 2020-01-07
API cleanup

madadam:
Supersedes #1981 .

This PR reduces the exposed API surface, updated some doc comments and enables (and fixes) additional clippy lints.


src/lib.rs
```diff
@@ -181,6 +180,7 @@ pub use {
 
 #[cfg(feature = "mock_base")]
 pub use self::{
+    authority::Authority,
```
jeanphilippeD: Could we make `authority::Authority` pub(crate) if not mock_base so we can still have the useful names at root level?

src/chain/chain.rs
```diff
@@ -20,7 +20,7 @@ use crate::{
     parsec::{DkgResult, DkgResultWrapper},
     relocation::{self, RelocateDetails},
     utils::LogIdent,
-    BlsPublicKeySet, BlsSecretKeyShare, BlsSignature, ConnectionInfo, Prefix, XorName, Xorable,
```
jeanphilippeD: I thought we would need to pass at least signature back to vault, and possibly public key at some point?

madadam: There is no API currently that uses these. But we can always expose them later, when they are actually needed.

src/chain/chain.rs
```diff
@@ -1815,13 +1816,14 @@ mod tests {
...
         xor_space::{Prefix, XorName},
-        BlsSecretKeySet, ConnectionInfo,
+        ConnectionInfo,
     };
     use rand::Rng;
     use std::{
         collections::{BTreeMap, HashMap},
         str::FromStr,
     };
+    use threshold_crypto::SecretKeySet;
```
jeanphilippeD: this would make search and replace more difficult, probably best to stick with `bls::SecretKeySet`.

src/chain/chain_accumulator.rs
```diff
@@ -7,10 +7,14 @@
...
+use crate::id::PublicId;
 use log::LogLevel;
-use std::collections::{BTreeMap, BTreeSet, VecDeque};
-use std::{mem, rc::Rc};
+use std::{
+    collections::{BTreeMap, BTreeSet, VecDeque},
+    mem,
+    rc::Rc,
+};
+use threshold_crypto::{PublicKeySet, Signature};
```
jeanphilippeD: Same

src/chain/mod.rs
```diff
@@ -33,16 +33,19 @@ pub use self::{
...
 };
+use crate::PublicId;
 #[cfg(feature = "mock_base")]
-use crate::{error::RoutingError, id::P2pNode, BlsPublicKey, Prefix, XorName};
-use crate::{BlsPublicKeySet, PublicId};
+use crate::{error::RoutingError, id::P2pNode, Prefix, XorName};
 use std::collections::BTreeMap;
 use std::fmt::{self, Debug, Formatter};
+#[cfg(feature = "mock_base")]
+use threshold_crypto::PublicKey;
```
jeanphilippeD: same

src/chain/network_event.rs
```diff
@@ -27,6 +23,7 @@ use std::{
     collections::BTreeSet,
     fmt::{self, Debug, Formatter},
 };
+use threshold_crypto::{PublicKeyShare, SecretKeyShare, Signature, SignatureShare};
```
jeanphilippeD: same

src/chain/shared_state.rs
```diff
@@ -22,6 +22,7 @@ use std::{
     fmt::{self, Debug, Formatter},
     iter, mem,
 };
+use threshold_crypto::{PublicKey, PublicKeySet, Signature};
```
jeanphilippeD: same

src/lib.rs
```diff
@@ -157,24 +157,16 @@ use self::mock::quic_p2p;
...
+pub use self::{
+    chain::quorum_count,
+    error::{InterfaceError, RoutingError},
+    event::{ClientEvent, ConnectEvent, Event},
+    event_stream::EventStream,
+    id::{FullId, P2pNode, PublicId},
+    node::{Node, NodeBuilder},
+    pause::PausedState,
+    quic_p2p::{Config as NetworkConfig, NodeInfo as ConnectionInfo},
+    xor_space::{Prefix, XorName, XorNameFromHexError, Xorable, XOR_NAME_BITS, XOR_NAME_LEN},
```
jeanphilippeD: could we make these Bls* type pub(crate), similar question asked for Autority. probably apply to all changes so I won't repeat, or the threshold_cryptp -> bls name if that is the direction we prefer.

src/messages/mod.rs
```diff
@@ -27,6 +26,7 @@ use std::{
     fmt::{self, Debug, Formatter},
     mem,
 };
+use threshold_crypto::{PublicKeySet, Signature as BlsSignature, SignatureShare};
```
jeanphilippeD: same

src/parsec.rs
```diff
@@ -30,6 +29,7 @@ use std::{
     collections::{btree_map::Entry, BTreeMap},
     fmt, mem,
 };
+use threshold_crypto::SecretKeySet;
```
jeanphilippeD: same... same for the next ones.

src/lib.rs
madadam: So I think there are two ways to look at this:
1. There is some value in having the `lib.rs` file be clean and minimal so when a newcomer looks at it, they immediately see what the public API is without it being cluttered with internal stuff
2. On the other hand, there is also value in having some commonly used items re-exported at the top level, so they are easier to use.
Neither of these is particularly strong point and I don't really lean towards either of them, which is why I'll defer this to whatever you and/or @dirvine decide.

jeanphilippeD: One thing about point 2: if an item is re-exported for mock_base, it might be good to re-export it for crate otherwise so it can be used the same way in test and in crate. But whatever David decides.

madadam: The reason to re-export something for mock_base is mostly so that it can be used in the integration tests, not to make it convenient to use in unit tests.

src/lib.rs
```diff
@@ -110,6 +77,57 @@ extern crate serde_derive;
 #[macro_use]
 mod macros;
 
+// ############################################################################
+// Public API
+// ############################################################################
+pub use self::{
+    chain::quorum_count,
```
dirvine: quorum_count only used in a mock test, does not seem to be used in vaults (first API use).

src/lib.rs
```diff
@@ -110,6 +77,57 @@ extern crate serde_derive;
...
+    error::{InterfaceError, RoutingError},
+    event::{ClientEvent, ConnectEvent, Event},
+    event_stream::EventStream,
+    id::{FullId, P2pNode, PublicId},
+    node::{Node, NodeBuilder},
+    pause::PausedState,
+    quic_p2p::{Config as NetworkConfig, NodeInfo as ConnectionInfo},
+    xor_space::{Prefix, XorName, XorNameFromHexError, Xorable, XOR_NAME_BITS, XOR_NAME_LEN},
```
dirvine: seems only Prefix and XorName is required (in vault at least)

src/lib.rs
```diff
@@ -110,6 +77,57 @@ extern crate serde_derive;
 #[macro_use]
 mod macros;
 
+// ############################################################################
+// Public API
+// ############################################################################
+pub use self::{
+    chain::quorum_count,
+    error::{InterfaceError, RoutingError},
```
dirvine: InterFaceError is also unused I think, maybe good to look (next PR) at refactoring Error as well.

src/lib.rs
```diff
@@ -110,6 +77,57 @@ extern crate serde_derive;
...
+    event::{ClientEvent, ConnectEvent, Event},
```
dirvine: Not sure ConnectEvent is required in API as of yet either

src/lib.rs
madadam: OK, but I would maybe also expose `XOR_NAME_LEN`, which I think is common practice for types that publicly wrap a fixed-length array ([example](https://docs.rs/ed25519-dalek/1.0.0-pre.3/ed25519_dalek/constant.PUBLIC_KEY_LENGTH.html)).

src/lib.rs
dirvine: If possible I would go for clean lib.rs. However, we don't want to take it too far. So here I would be just ever so slightly in favour of this pub behind the feature flag as it is in the PR.

src/lib.rs
dirvine: Yes it's more compelling when you see more of these. I wonder is pub(crate) more idiomatic? I am on the fence, be good to get more feedback here.

src/lib.rs
madadam: `InterfaceError` is used in some of the public methods of `Node` ([e.g.](https://github.com/maidsafe/routing/blob/42760af38d173a623f86c37b1bdf5b6937618e06/src/node.rs#L245)), so I would keep it exposed for now.

src/lib.rs
dirvine: I would agree in this case as it fits a common pattern in rust :+1:  I was going for the keep as much private as possible /within reason but in his case I agree

src/lib.rs
madadam: It's part of `Event` which is public, so probably should be public too.

PR #1994, opened 2020-01-07
Re-expose mock quic-p2p

madadam:


madadam:
> In safe_vault, the following seem to be stuck compared to with the current lock file which pass. So there may be a bit more to it (though it does compile).
> 
> cargo test "$@" --release --features=mock_parsec --no-default-features

Hm, I'm getting a failure in routing tests when running with just `--features=mock_parsec`. Could be the same cause. Looking into it.

madadam:
Superseded by #1995 .

PR #1995, opened 2020-01-08
Revert some changes to allow safe_vault test to pass

jeanphilippeD:
Some of these should be re-applied together with fix with safe_vault if needed.
Take some commits from https://github.com/maidsafe/routing/pull/1994

Test:
Verify safe_vault test pass with small modification to cargo.toml

### DIFF
--- a/scripts/clippy
+++ b/scripts/clippy
@@ -4,4 +4,6 @@ set -x -e
 
 cargo clippy "$@" --all-targets
 cargo clippy "$@" --all-targets --features=mock_base
+cargo clippy "$@" --all-targets --features=mock_parsec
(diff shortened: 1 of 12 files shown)


scripts/clippy
```diff
@@ -4,4 +4,6 @@ set -x -e
 
 cargo clippy "$@" --all-targets
 cargo clippy "$@" --all-targets --features=mock_base
+cargo clippy "$@" --all-targets --features=mock_parsec
```
madadam: There is no `mock_parsec`

jeanphilippeD:
Fixed review comment

PR #1997, opened 2020-01-08
clean up timer to avoid work arround in safe_vault

jeanphilippeD:
Clean up timer to allow remove the incorrect call to `self.routing_node.borrow_mut().poll();` from safe_vault.

Test:
Verify test pass without that call using new routing.


jeanphilippeD:
Add reverted change as it now passed in safe_vault.

PR #1999, opened 2020-01-08
Extract mock quic p2p

madadam:


src/messages/direct.rs
```diff
@@ -25,7 +25,6 @@ use std::{
 };
 
 /// Direct message content.
-#[cfg_attr(feature = "mock_serialise", derive(Clone))]
 #[derive(Eq, PartialEq, Serialize, Deserialize)]
```
jeanphilippeD: need to remove that line since it is not necessary if mock_serialisation is off or gone.
Same for others

src/messages/mod.rs
```diff
@@ -30,7 +30,6 @@ use std::{
 /// Wrapper of all messages.
 ///
 /// This is the only type allowed to be sent / received on the network.
-#[cfg_attr(feature = "mock_serialise", derive(Clone))]
 #[derive(Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
```
jeanphilippeD: Same etc.

src/messages/direct.rs
jeanphilippeD: No i misread.

src/messages/mod.rs
jeanphilippeD: No I misread.

Cargo.toml
```diff
@@ -28,7 +28,7 @@ lazy_static = { version = "~1", optional = true }
 log = "~0.3.8"
 lru_time_cache = "~0.8.1"
 maidsafe_utilities = "~0.18.0"
-mock-quic-p2p = { git = "https://github.com/madadam/quic-p2p", branch="mock", optional = true }
```
jeanphilippeD: Could you update the commit that refer to madadam instead, probably best to avoid having commit refering to external repo.

PR #2000, opened 2020-01-08
fix/routing: pull parsec and remove rng compatibility for parsec

jeanphilippeD:


Cargo.toml
```diff
@@ -29,13 +29,12 @@ log = "~0.3.8"
 lru_time_cache = "~0.8.1"
 maidsafe_utilities = "~0.18.0"
 num-bigint = "~0.1.40"
-parsec = { git = "https://github.com/maidsafe/parsec", rev = "a6d8e85" }
+parsec = { git = "https://github.com/maidsafe/parsec", rev = "a3a1eb2" }
```
maqi: not related to this PR, maybe we shall consider using the `branch` arguement to avoid pointing to outdated parsec commits ?

jeanphilippeD: No, if you use a branch and a breaking change occur, you cannot develop.
Using rev allow to decide when to pull in the dependency. An alternative I guess would be to commit cargo.lock.

PR #2001, opened 2020-01-09
Trusted GenesisUpdate

### LINKED ISSUE #1945: Make sure `GenesisUpdate` message can be trusted
Currently we just apply the message without any verification. We should include any info necessary to verify it's integrity/trust when sending it and we should actually verify it when receiving. 

EDIT: Here is a plan to handle trust checking (copy-pasted from a slack thread):

```
Adam Cigánek  5 minutes ago
OK, so just to recap:
1. Modify the ParsecPoke mechanism to also include our current section knowledge.
2. Add a map to Elder to keep the non-elder member knowledge.
3. Use that map when creating proof for GenesisUpdate and Relocate messages.
4. Modify the signature accumulator to allow different length proofs (by taking the shortest common proof?).
5. Add trust check for non-elders
Does that sound right? If so, then I can start on that next week (it's my ODD tomorrow).

Jean-Philippe Dufraigne  3 minutes ago
 Modify the ParsecPoke mechanism to also include our current section knowledge.
I think we already do as we have a parsec version that should be sufficient I guess.
Alternatively have a different message to avoid overloading the meaning of Poke and leave us more option to optimize Poke as needed.

Jean-Philippe Dufraigne  2 minutes ago
Add a map to Elder to keep the non-elder member knowledge.
Yes, or chain, whichever is more appropriate for use when creating proof.

Jean-Philippe Dufraigne  1 minute ago
 Modify the signature accumulator to allow different length proofs (by taking the shortest common proof?).
I think it might already not check and just take the 1st proof it got. which is ok for our purpose, but maybe should be the longest proof to avoid malicious node creating too short proofs.

Jean-Philippe Dufraigne  1 minute ago
Otherwise seem like a plan to start on next week :slightly_smiling_face:

Adam Cigánek  < 1 minute ago
Yes, longest proof is probably more useful.
```

madadam:
Partial solves #1945 by implementing integrity check. We still need to check trust which will be handled in a followup PR.

Also includes minor cleanup/refactoring.

### DIFF
--- a/src/states/adult.rs
+++ b/src/states/adult.rs
@@ -21,7 +21,8 @@ use crate::{
     event::Event,
     id::{FullId, P2pNode, PublicId},
     messages::{
-        BootstrapResponse, DirectMessage, HopMessage, RoutingMessage, SignedRoutingMessage,
+        BootstrapResponse, DirectMessage, HopMessage, MessageContent, RoutingMessage,
+        SignedRoutingMessage,
     },
     network_service::NetworkService,
     outbox::EventBox,
@@ -409,7 +410,7 @@ impl Adult {
         &mut self,
         msg: SignedRoutingMessage,
         pub_id: PublicId,
-    ) -> Result<(), RoutingError> {
+    ) -> Result<Transition, RoutingError> {
         if !self.chain.is_peer_elder(&pub_id) {
             debug!(
                 "{} - Received message signature from not known elder (still use it) {}, {:?}",
@@ -418,9 +419,10 @@ impl Adult {
         }
 
         if let Some(signed_msg) = self.sig_accumulator.add_proof(msg) {
-            self.handle_signed_message(signed_msg)?;
+            self.handle_signed_message(signed_msg)
+        } else {
+            Ok(Transition::Stay)
         }
-        Ok(())
     }
 
     // If the message is for us, verify it then, handle the enclosed routing message and swarm it
@@ -428,7 +430,7 @@ impl Adult {
     fn handle_signed_message(
         &mut self,
         signed_msg: SignedRoutingMessage,
-    ) -> Result<(), RoutingError> {
+    ) -> Result<Transition, RoutingError> {
         if !self
             .routing_msg_filter
             .filter_incoming(signed_msg.routing_message())
@@ -439,7 +441,7 @@ impl Adult {
                 self,
                 signed_msg.routing_message()
             );
-            return Ok(());
+            return Ok(Transition::Stay);
         }
 
         self.handle_filtered_signed_message(signed_msg)
@@ -448,7 +450,7 @@ impl Adult {
     fn handle_filtered_signed_message(
         &mut self,
         signed_msg: SignedRoutingMessage,
-    ) -> Result<(), RoutingError> {
+    ) -> Result<Transition, RoutingError> {
         trace!(
             "{} - Handle signed message: {:?}",
             self,
@@ -457,11 +459,19 @@ impl Adult {
 
         if self.in_authority(&signed_msg.routing_message().dst) {
             self.check_signed_message_integrity(&signed_msg)?;
-            self.routing_msg_backlog.push(signed_msg.clone());
+
+            match &signed_msg.routing_message().content {
+                MessageContent::GenesisUpdate(info) => {
+                    return self.handle_genesis_update(info.clone())
+                }
+                _ => {
+                    self.routing_msg_backlog.push(signed_msg.clone());
+                }
+            }
         }
 
         self.send_signed_message_to_elders(&signed_msg)?;
-        Ok(())
+        Ok(Transition::Stay)
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/states/adult.rs
```diff
@@ -453,11 +455,22 @@ impl Adult {
 
         if self.in_authority(&signed_msg.routing_message().dst) {
             self.check_signed_message_integrity(&signed_msg)?;
-            self.routing_msg_backlog.push(signed_msg.clone());
+
+            let (msg, metadata) = signed_msg.into_parts();
+
+            match msg.content {
+                MessageContent::GenesisUpdate(info) => return self.handle_genesis_update(info),
```
jeanphilippeD: As you mentioned we probably need to validate trust here. was not an issue if only put in backlog because then we validated trust as elder.

jeanphilippeD: The problem is i'm not 100% sure it would be sound to do so as an adult as we don't ensure that all adults have updated trust info before sending it (like we do for elders). Maybe that should be a follow up PR/issue.

src/states/adult.rs
```diff
@@ -389,14 +388,9 @@ impl Adult {
             .map(|node| node.connection_info().clone())
             .collect();
 
-        let message = self.to_hop_message(signed_msg.clone())?;
+        let message = self.to_hop_message(signed_msg)?;
         self.send_message_to_targets(&targets, targets.len(), message);
 
-        // we've seen this message - don't handle it again if someone else sends it to us
-        let _ = self
```
jeanphilippeD: Why is that removed?

madadam: It's redundant because it's already done in `handle_signed_message`.

jeanphilippeD: Is that enough though? that create an imbalance with same code in Elder and I'm suspicious.
What if we accumulate a signature that trigger us to send a message as an adult, will we not forget to filter it?

madadam: We don't send any routing messages as adult right now. But even if we were, if we are the one sending, should it really go through *incoming* filter?

jeanphilippeD: We do through:

jeanphilippeD: We do through
```rust
    fn handle_message_signature(
        &mut self,
        msg: SignedRoutingMessage,
        pub_id: PublicId,
        if !self.chain.is_peer_elder(&pub_id) {
            debug!(
                "{} - Received message signature from not known elder (still use it) {}, {:?}",
                self, pub_id, msg
            );
        }

        if let Some(signed_msg) = self.sig_accumulator.add_proof(msg) {
            self.handle_signed_message(signed_msg)?;
        }
        Ok(())
    }
```

But then we already done the filtering so not a problem.

jeanphilippeD: > should it really go through incoming filter?

It is for Elders, and yes, so we don't process them if we receive the same message and resend them.

madadam: Then I think the incoming filter check should happen inside `handle_message_signature` where it seems more appropriate to me. Also for elders. On the other hand, this change is not really that relevant for this PR, so I'm now leaning towards reverting it. What do you think?

src/states/adult.rs
```diff
@@ -453,11 +449,22 @@ impl Adult {
...
         if self.in_authority(&signed_msg.routing_message().dst) {
             self.check_signed_message_integrity(&signed_msg)?;
-            self.routing_msg_backlog.push(signed_msg.clone());
+
+            let (msg, metadata) = signed_msg.into_parts();
+
+            match msg.content {
+                MessageContent::GenesisUpdate(info) => return self.handle_genesis_update(info),
+                _ => {
+                    let signed_msg = SignedRoutingMessage::from_parts(msg, metadata);
```
maqi: can we just use `signed_msg.clone()` above to avoid compose from parts again?
also, with this means we can reduce some code within this branch and remain as previous that call `self.send_signed_message_to_elders(&signed_msg)?;` once at the end ?

src/states/adult.rs
jeanphilippeD: Yes, I think that would make more sense as in a different PR that clean things up if we want it.

src/states/adult.rs
jeanphilippeD: seem like a good idea.

src/states/adult.rs
```diff
@@ -453,11 +454,19 @@ impl Adult {
...
         if self.in_authority(&signed_msg.routing_message().dst) {
             self.check_signed_message_integrity(&signed_msg)?;
-            self.routing_msg_backlog.push(signed_msg.clone());
+
+            match &signed_msg.routing_message().content {
+                MessageContent::GenesisUpdate(info) => {
+                    return self.handle_genesis_update(info.clone())
+                }
+                _ => {
+                    self.routing_msg_backlog.push(signed_msg.clone());
```
maqi: `self.routing_msg_backlog.push(signed_msg.clone())` will be called twice for this branch?
I think here we only need: `_ => (),` ?

PR #2002, opened 2020-01-09
fix/routing: avoid re-voting for completed event

jeanphilippeD:
Fix aggressive_churn: [1221425541, 378878981, 107272102, 3204539950] at
dc2abe182b4eadcbd13f67c5e6c4dcb4f5ee59e8.

Address issue introduced with #1993: Ensure we filter out all completed
events.
Also improve logs for debuging.

Test:
Verify the seed is fixed.
soak test + clippy

### DIFF
--- a/src/states/elder/mod.rs
+++ b/src/states/elder/mod.rs
@@ -405,32 +405,30 @@ impl Elder {
     fn complete_parsec_reset_data(&mut self, reset_data: ParsecResetData) -> CompleteParsecReset {
         let ParsecResetData {
             gen_pfx_info,
-            mut cached_events,
+            cached_events,
             completed_events,
         } = reset_data;
 
-        cached_events.extend(
-            self.parsec_map
-                .our_unpolled_observations()
-                .filter_map(|obs| match obs {
-                    parsec::Observation::OpaquePayload(event) => {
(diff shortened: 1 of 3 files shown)


src/states/elder/mod.rs
```diff
@@ -405,32 +405,30 @@ impl Elder {
...
-                    | parsec::Observation::Add { .. }
-                    | parsec::Observation::Remove { .. }
-                    | parsec::Observation::Accusation { .. }
-                    | parsec::Observation::StartDkg(_)
-                    | parsec::Observation::DkgResult { .. }
-                    | parsec::Observation::DkgMessage(_) => None,
-                })
-                .cloned(),
-        );
+        let cached_events: BTreeSet<_> = cached_events
```
maqi: just wondering why switch from the previous `extend` to the current `chain` ?

normally I will prefer not to use something like `let cached_events: BTreeSet<_> = cached_events`

jeanphilippeD: because the filter apply to both, not sure what you are suggesting?

maqi: Ah, I got what you mean. 
just forgot my above comment then, my mis-reading of the code-block.

PR #2005, opened 2020-01-14
Enable GenesisUpdate trust check 

### LINKED ISSUE #1945: Make sure `GenesisUpdate` message can be trusted

madadam:
This PR implements trust checking of `GenesisUpdate` message by non-elders. To achieve that, it introduces a mechanism by which elders keep track of the knowledge the other members have about their section. This knowledge is then used to construct the proof. This PR additioanally improves the unit test framework and add a couple of tests covering the message trust feature.

Closes #1945


src/states/adult/mod.rs
```diff
@@ -552,7 +553,8 @@ impl Base for Adult {
 
     fn handle_timeout(&mut self, token: u64, _: &mut dyn EventBox) -> Transition {
         if self.parsec_timer_token == token {
-            self.send_parsec_poke();
+            // TODO: send this only when the knowledge changes, not periodically.
```
jeanphilippeD: We need it to poke regularly though. Only half way through the combined change though so I may be missing something, but is it really sensible to use the same message for Poke and send_member_knowledge?

src/states/common/approved.rs
```diff
@@ -212,7 +221,10 @@ pub trait Approved: Base {
                 }
 
                 let rand_index = self.rng().gen_range(0, p2p_recipients.len());
-                (version, p2p_recipients[rand_index].clone())
+                let recipient = p2p_recipients[rand_index].clone();
+                let version = self.get_parsec_version_for(recipient.public_id());
```
jeanphilippeD: No sure about that will come back to it.

src/states/adult/mod.rs
madadam: It's not implemented in this PR, but I was thinking it should be sufficient to send this message only when our parsec version knowledge changes. That happens once when we transition from `JoiningPeer` to `Adult` and then every time we receive `GenesisUpdate`. So maybe we should only send it then? I'm not 100% sure about it which is why I haven't implemented it in this PR.

jeanphilippeD: I agree, this is what should happen for this new message.
But how does that support what poke was doing. i.e why could we do that for poke, will the gossip flow properly even if it is sent only once, at a point the recipient elder may be in any state. If so that is good.

src/states/elder/mod.rs
```diff
@@ -1291,32 +1316,39 @@ impl Elder {
...
-        let proof = self.chain.prove(&msg.dst, None);
+        let proof = self.chain.prove(&msg.dst, node_knowledge_override);
         let pk_set = self.our_section_bls_keys().clone();
         let secret_key = self.chain.our_section_bls_secret_key_share()?;
 
         SignedRoutingMessage::new(msg, secret_key, pk_set, proof)
     }
+
+    // Starts tracking knowledge of recently demoted elders.
+    fn update_demoted_members_knowledge(&mut self, removed_elders: &BTreeSet<P2pNode>) {
```
jeanphilippeD: I do not think this is correct, the member may not have processed it.
I think the correct way would be to wait for a message that the demoted elder send, so the demoted elder should send that message too.

src/states/elder/mod.rs
```diff
@@ -1596,16 +1628,22 @@ impl Approved for Elder {
         payload: OnlinePayload,
         outbox: &mut dyn EventBox,
     ) -> Result<(), RoutingError> {
+        let _ = self
```
jeanphilippeD: Not sure about that.

jeanphilippeD: I think there is a bit of an issue in that we could get the Poke event before reaching that point. We probably should have taken it into account and not ignored it.

src/states/elder/mod.rs
```diff
@@ -1810,6 +1853,22 @@ impl Approved for Elder {
         }
         Ok(())
     }
+
+    fn before_handle_parsec_gossip(&mut self, parsec_version: u64, sender: &PublicId) {
+        if parsec_version >= self.parsec_map.last_version() {
+            // This member has caught up to the latest parsec version. We don't need to track its
+            // knowledge any more.
+            let _ = self.members_knowledge.remove(sender.name());
```
jeanphilippeD: Don't we still need to track it, what about when it is demoted?

src/messages/mod.rs
```diff
@@ -298,17 +311,59 @@ impl SignedRoutingMessage {
...
+            return Err(MergeSecurityMetadataError {
+                kind: PublicKeySetMismatch,
+                other_message: SignedRoutingMessage::from_parts(
+                    other.content,
+                    SecurityMetadata::Partial(other_md),
+                ),
+            });
+        }
+
+        // TODO: should we also check the proof chains are compatible?
```
jeanphilippeD: Considering that TODO, would it make sense to postpone the commit `Take the longest proof chain when accumulating message signatures` to a later PR that would address that too? Currently because there are no malicious node any chain will be trusted.

src/states/common/approved.rs
jeanphilippeD: it does not quite fit with the poke thing, but probably ok now that we update the version with the poke message so that is the version we will get, and I guess with the removal of member version for elder, we get the latest version always which is what we want (i.e only send response to older version not request, except to adults that poke us).

src/states/adult/mod.rs
madadam: I think it should work. The only problem I can think of is what happens if some of the `MemberKnowledge` messages get lost. Because they would be sent only once, the elders that didn't receive them wouldn't get their members knowledge updated. Not sure how big of a problem it would really be, because we can expect that at least some of the elders do receive the message and those elders would then send the correct version gossip to the node.

src/states/elder/mod.rs
madadam: > the member may not have processed it yet.
Should be fine, because that means the member is still in the Elder state so when we send it `GenesisUpdate`, they would ignore it. Although maybe they should instead backlog it (and process when adult)?

src/states/elder/mod.rs
madadam: This is currently OK because the `MemberKnowledge` messages are sent periodically. But If we want to change that then this can become an issue.

src/states/elder/mod.rs
madadam: We re-insert it when the member is demoted.

src/states/adult/mod.rs
jeanphilippeD: It is a direct message so it would not get lost (i.e we will retry to send it until it succeed).
However, in the case of a node becoming an elder, they need to keep sending poke until the elders decide that that Adult is about to become an Elder, right in the DKG, so if there is nothing done for that that could be a problem (though maybe it is not an issue, the new node would be a gossip recipient and receive the right version of parsec gossip probably). (Though there may still be a problem if Elders move forward and send new version of gossip where Adult would be an Elder, but the Adult has not caught up yet, in this case poke make sense.

It just seem we are not 100% clear what should be done here.

src/messages/mod.rs
madadam: Yes. Also I'm no longer sure just taking the longer proof is sufficient. Say we have two messages, both have proof of length 1, but one is at  version 7 and the other 8. The current code will prefer the one that is already in the accumulator, but I think we should take the lower of the two (not 100% sure about this though).

src/states/elder/mod.rs
jeanphilippeD: But then they would ignore a GenesisUpdate that they should not ignore which would be problematic (Also could be other messages like relocation).

What I am saying is I think this function break the logic behind the member_knowledge map (That is value in the map will always be trusted by the node).

madadam: > But then they would ignore a GenesisUpdate that they should not ignore

Yes, which is why I'm thinking we should backlog it.

src/states/elder/mod.rs
jeanphilippeD: I think it could well go all wrong:
 - Node is relocated to our section.
 - Node is Approved -> Approval is sent
 - Node send MemberKnowledge to all elders
 - Most have not yet approved it and ignore it as a result.
 - Node is promoted to Elder
 - now looking for Parsec last always, all further MemberKnowledge are ignored.
 - Node never receive the gossip from the version it expects and never see itself promoted.

src/states/elder/mod.rs
madadam: But I do take you other point. But it won't solve the problem of lagging elder receiving `GenesisUpdate`.

src/states/elder/mod.rs
jeanphilippeD: (ok, though probably with a version the node has not reached yet, but that is part of other comment thread).

src/states/elder/mod.rs
jeanphilippeD: Yes, but we would only need to backlog message that we trusted as oppose to backlog any given random messages. (i.e message we really need to take action on).

src/states/adult/mod.rs
madadam: OK. I'm thinking about putting the parsec poke thing back exactly as it was and adding the `MemberKnowledge` as a separate message. Then we can decide what to do with the poke later and address it in a separate PR. What do you think?

jeanphilippeD: Not sure, that discussion moved me a bit closer to thinking that these 2 messages should be as 1 as you did. still need to think about it a bit.

src/states/elder/mod.rs
madadam: Yeah, this could be an issue. Perhaps we need to keep the parsec poke thing periodic.

jeanphilippeD: This is a problem even if parsec Poke is periodic, it is related to the use of MemberKnowledge to reply to a parsec poke.

src/states/elder/mod.rs
madadam: OK. So if we don't do this and only update the `members_knowledge` container when we receive the `MemberKnowledge` message, and also if we backlog `GenesisUpdate`, but only after successful proof check, then I think it should work. The worst case I can think of right now is this:

1. A node is demoted
2. It's lagging behind, so we don't receive it's `MemberKnowledge` yet. That means there is no entry for it in the `members_knowledge` container.
3. We progress the section again and send `GenesisUpdate` to all non-elders
4. Because there is no entry for the lagging node, we use the longest proof we have
5. The lagging node still thinking it is elder receives the message. It should trust it, because the proof is sufficiently long
6. It backlogs it
7. It eventually catches up, transitions to adult and processes the backlogged `GenesisUpdate`

So the only issue is the proof might get unnecessarily long, but that is probably OK.

madadam: Alternative: on demote, instead of inserting the members_knowledge of the current version, we use the last version that reached the total consensus via the `AckMessage` event.

src/states/elder/mod.rs
madadam: No, with periodic parsec poke (or `MemberKnowledge`), the elders do eventually get updated and will send the correct version gossip to the promoted node. Or am I missing something?

src/states/elder/mod.rs
jeanphilippeD: Yes, both of these could be option, I'm a bit tired so I'll probably need to conclude tomorrow for definitive answer, but here is my thought.

I'm not clear why we need to remove the value from member_knowledge, it seem they should be kept up to date by sending a new message like all Adults do everytime as an elder we see Parsec change (i.e when we send GenesisUpdate). It just happen that when dealing with an elder we do not want to use that value (so we could check if it is an elder and send latest parsec), except in the case of Poke in which case the node explicitely tells us it still needs the old value. So I'm thinking the special poke case need to still exists, and in the non poke case, we really send latest parsec always (i.e the code is more like it was before).

Then regarding GenesisUpdate messages proofs (or other messages), we always use the member_knowledge, beause we always have it, and like for SMD it means every message we sent will always have enough proof to be trusted at the time we sent them.

But I'll need to think a bit more about it.

jeanphilippeD: >  So I'm thinking the special poke case need to still exists, and in the non poke case, we really send latest parsec always (i.e the code is more like it was before).

Or maybe not, and we use the value that was last given to us, i.e if we have not received the updated info, then the node is probably lagging and we send the old parsec in our gossip, but if we see it we send the new parsec, that should work for existing elders (lagging) as well as new elders (lagging or not).

src/states/elder/mod.rs
jeanphilippeD: I'm thinking if we consider a node an elder we ignore MemberKnowledge, so if that node is  lagging at every poke we keep on sending them the new parsec version while they need the old one to catch up.

src/states/elder/mod.rs
madadam: > I'm not clear why we need to remove the value from member_knowledge

We don't need that value once the node becomes elder, because for elder we prove messages based on the `their_knowledge` container. So the current logic I have there is to remove the value when the node is promoted and put it back when it is demoted. Also in the promotion case I don't remove it straight away because the node might be lagging. So I only remove it when I receive first up-to-date parsec gossip from the node.

jeanphilippeD: > We don't need that value once the node becomes elder

I think we are going at it from a different angle. I'm saying it seems simpler to continue to maintain that collection without differentiating between Elders and Adults. My reason is that it avoid the corner case that occur in transition from/to Elder. If at some place we do not need that value, we can choose to ignore it, even if we store it. (It is 7 nodes out of 100 so not much lost here).


I'm also thinking there are also different cases, i.e:
- When sending a section message to ourselves we use their-knowledge
- But when sending a message to one node (Like for example if we were sending a Relocation message), it would seem to make sense to use the members_knowledge (eventually), and the same machanism as GenesisUpdate message.


Additionally for:
```
     fn before_handle_parsec_gossip(&mut self, parsec_version: u64, sender: &PublicId) {
        if parsec_version >= self.parsec_map.last_version() { 
```

I am not sure that is even sufficient to provide the guarantee that we seem to assume here. i.e when participating in the DKG, we will send gossip, but we are not yet an elder, which mean we will not be in the group required to have voted for a change in their_knowledge.

madadam: OK, I think I get your point now. It makes sense and looks like it makes things simpler, so :+1:

src/states/adult/mod.rs
```diff
@@ -235,20 +235,18 @@ impl Adult {
...
-        self.routing_msg_backlog.push(msg)
+    ) -> Result<Transition, RoutingError> {
+        match &msg.routing_message().content {
+            MessageContent::GenesisUpdate(info) => {
+                self.check_signed_message_trust(&msg)?;
+                self.handle_genesis_update(info.clone())
+            }
+            _ => {
+                self.routing_msg_backlog.push(msg.clone());
+                self.send_signed_message_to_elders(msg)?;
```
jeanphilippeD: That does not seem right, dispatch_routing_message did not do the send_signed_message in either Elder or Adult. I don't think it should now.

src/messages/mod.rs
jeanphilippeD: Did you mean to remove that commit?

src/states/adult/mod.rs
madadam: Why is that a problem?

madadam: Ok, I see. It's a bit inconsistent. I'll change it.

jeanphilippeD: Level of abstraction, the level above does the resending, dispatch is to handle message we need to address. It makes is clearer in the calling function that resending is happening as needed. (Here we split it in 2, part of resending is done in calling function.

This will be even worth with work I'm looking at to avoid reserialization, it would force us to pass to the dispatch function what is needed for the resend.

src/states/common/approved.rs
jeanphilippeD: So I'm still not clear how that fit in this PR:
 -> If you get a MemberKnowledge, we are going to be in the `Some((v, p)) => (v, p),` branch.
 -> If you don't in what circumstance do you need to send not the latest Parsec gossip? Did it change with this PR?

madadam: Hm, right. I guess this only makes sense if we went with the non-periodic message approach. Will revert it.

src/messages/direct.rs
```diff
@@ -91,7 +106,11 @@ impl Debug for DirectMessage {
             ConnectionResponse => write!(formatter, "ConnectionResponse"),
             ParsecRequest(v, _) => write!(formatter, "ParsecRequest({}, _)", v),
             ParsecResponse(v, _) => write!(formatter, "ParsecResponse({}, _)", v),
-            ParsecPoke(v) => write!(formatter, "ParsecPoke({})", v),
+            MemberKnowledge(payload) => write!(
+                formatter,
+                "MemberKnowledge {{ elders_version: {}, parsec_version: {} }}",
```
jeanphilippeD: is it not what `write!(formatter, {:?}, payload)` would print?

src/messages/mod.rs
```diff
@@ -559,144 +554,97 @@ mod tests {
...
             pk_set,
-            dummy_proof
-        ));
-        assert_eq!(signed_msg.signatures().expect("no signatures").len(), 1);
-        assert!(!signed_msg.check_fully_signed());
-
-        // Add an invalid signature for IDs 1 added by the 3rd malicious node.
-        // Add a valid signature for IDs 1 and 2 and an invalid one for ID 3
-        // Add an invalid signature for ID 3 added by the same 3rd malicious node.
-        let bad_sig = bls_secret_key_share_3.sign(&[1]);
```
jeanphilippeD: Is this part of the test no longer relevant (i.e testing that adding bad signatures do not make the signature invalid if enough good ones are present?

madadam: There is a test that invalid signatures don't count, but not that they don't invalidate valid signatures. I'll add it.

src/states/adult/mod.rs
```diff
@@ -547,13 +499,25 @@ impl Base for Adult {
             }
         }
 
+        for msg in mem::replace(&mut self.routing_msg_backlog, Default::default()) {
+            if let Transition::Stay = &transition {
+                match self.handle_filtered_signed_message(msg) {
```
jeanphilippeD: calling handle_filtered_signed_message make the node resend the message. At this point in finish_handle_transition we should probably only apply what needs to be done for us.

jeanphilippeD: Also, why do we switch around the direct and hop messages?

jeanphilippeD: Is it to be consistent with Elder, may make sense.

src/states/adult/mod.rs
```diff
@@ -6,6 +6,9 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+#[cfg(all(test, feature = "mock_base"))]
```
jeanphilippeD: Why do we build it in mock_base case?

madadam: Those tests don't require mock parsec because they don't use parsec at all.

PR #2007, opened 2020-01-15
Routing name tidy up

dirvine:
Phase I of a name tidy up and also start of reordering if statements to give us short circuit plus more readability. 

This PR is an example, I hope of some work we will tackle across libraries as we can now see the network in operation.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -24,17 +24,17 @@ use std::{
 #[derive(Clone, Eq, PartialEq)]
 pub enum Client {
     /// Inform the user (library) that we are connected to a client
-    ConnectedTo {
+    Connected {
         /// Client's endpoint
         peer_addr: SocketAddr,
     },
     /// Inform the user (library) that we are disconnected from a client
-    ConnectionFailureTo {
+    ConnectionFailure {
         /// Client's endpoint
         peer_addr: SocketAddr,
     },
     /// Inform the user (library) that we have a new message from a client
-    NewMessageFrom {
+    NewMessage {
         /// Client's endpoint
         peer_addr: SocketAddr,
         /// Client's message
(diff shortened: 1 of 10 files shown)


src/event.rs
```diff
@@ -22,26 +22,26 @@ use std::{
...
+pub enum Client {
     /// Inform the user (library) that we are connected to a client
-    ConnectedToClient {
+    ConnectedTo {
         /// Client's endpoint
         peer_addr: SocketAddr,
     },
     /// Inform the user (library) that we are disconnected from a client
-    ConnectionFailureToClient {
+    ConnectionFailureTo {
```
madadam: This enum seems to mirror the `Event` enum from `quic-p2p` to some extent, so maybe we should use the same naming? So this variant should be just `ConnectionFailure`.

madadam: This enum seems to mirror the `Event` enum from `quic-p2p` to some extent, so maybe we should use the same naming? So this variant should be just `ConnectionFailure`.

src/event.rs
```diff
@@ -22,26 +22,26 @@ use std::{
...
         /// Client's endpoint
         peer_addr: SocketAddr,
     },
     /// Inform the user (library) that we have a new message from a client
-    NewMessageFromClient {
+    NewMessageFrom {
```
madadam: Just `NewMessage`

src/event.rs
```diff
@@ -62,7 +62,7 @@ pub enum ClientEvent {
 
 /// An Event raised as node complete joining
 #[derive(Debug, Clone, Eq, PartialEq)]
-pub enum ConnectEvent {
+pub enum Connect {
```
madadam: I would rename this to `Connected`, to match the variant it is used in.

src/event.rs
```diff
@@ -141,30 +141,30 @@ impl Debug for Event {
     }
 }
 
-impl Debug for ClientEvent {
+impl Debug for Client {
     fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
         match *self {
-            Self::ConnectedToClient { peer_addr } => {
+            Self::ConnectedTo { peer_addr } => {
                 write!(formatter, "ClientEvent::ConnectedToClient - {}", peer_addr)
```
madadam: We probably want to change these strings too, to match the variant names.

src/node.rs
```diff
@@ -41,47 +41,47 @@ use {
...
     full_id: Option<FullId>,
     network_cfg: NetworkParams,
 }
 
-impl NodeBuilder {
+impl Builder {
     /// Configures the node to start a new network instead of joining an existing one.
     pub fn first(self, first: bool) -> Self {
-        Self { first, ..self }
+        Builder { first, ..self }
```
madadam: Why not use `Self` to avoid repeating the name? There is also a clippy lint for that which I think is enabled?

madadam: > There is also a clippy lint

It's [use_self](https://rust-lang.github.io/rust-clippy/master/) but it's not enabled because it has some annoying false positives right now. I'd still prefer to use `Self` instead of repeating the type name, but as we don't enforce it via a lint, I don't insist on it either.

jeanphilippeD: it seem some of the previous PRs converted TypeName to Self, so that seem to be a move backward, any reason for it?

src/event.rs
dirvine: Yes, I did ponder that for a while. I agree it makes more sense now you point out it mirrors quic-p2p. I will change that.

src/node.rs
dirvine: Yes, Self is best here. I will change it. I don't think this PR reverted that though it just never changed it to "fix" the repeated typename.

Issue #2010, opened 2020-01-15
Clean up message relay to not deserialize full message on send to next hop.


jeanphilippeD:
- [x]  => Part 1, still deserialize, but not re-serialize by re-using the message bytes we got. (init est: 3-4 days)
- [ ]   => need to handle partial deserialization for knowing target. (init est: 3-4 days)
   => desirable but not strictly crucial - performance improvement.

PR #2012, opened 2020-01-15
Avoid unecessary serialize for Hop messages

### LINKED ISSUE #2011: Don't Filter too much (since it would impact performance unless we use real BLS keys)
(init est: 2-3 days)
  => Attack send bad sig first => message rejected forever.
  => Once trusted only, or filter include signature

*Moved from SMD board.

jeanphilippeD:
Partially addresses #2010 
Closes #2011 

Avoid throwing away the serialized data we get from quic_p2p, and re-use it for filtering and for re-sending the Hop message.

Building on this work, we should avoid de-serializing the full message when all we need is the destination if we are not part of the destination authority.

We can probably also futher clean up the code following this change, in particular it is not clear we still need the HopMessage type.

### DIFF
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -19,8 +19,10 @@ use crate::{
     error::{Result, RoutingError},
     id::{FullId, PublicId},
     location::Location,
+    states::common::to_network_bytes,
     xor_space::{Prefix, XorName},
 };
+use bytes::Bytes;
 use log::LogLevel;
 use maidsafe_utilities::serialisation::serialise;
 use std::{
@@ -57,6 +59,62 @@ impl HopMessage {
     /// Wrap `content` for transmission to the next hop and sign it.
     pub fn new(content: SignedRoutingMessage) -> Result<Self> {
         Ok(Self { content })
+    }
+}
+
+#[derive(Eq, PartialEq, Clone)]
+pub struct HopMessageWithSerializedMessage {
+    /// Wrapped signed message.
+    content: SignedRoutingMessage,
+    /// Serialized Message as received or sent to quic_p2p.
+    serialized_message: Bytes,
+}
+
+impl HopMessageWithSerializedMessage {
+    /// Serialize message and keep both SignedRoutingMessage and Bytes.
+    pub fn new(content: SignedRoutingMessage) -> Result<Self> {
+        let hop_msg = HopMessage::new(content)?;
+        let message = Message::Hop(hop_msg);
+        let serialized_message = to_network_bytes(&message)?;
+
+        let content = if let Message::Hop(hop_msg) = message {
+            hop_msg.content
+        } else {
+            unreachable!("Created as Hop can only match Hop.")
+        };
+
+        Ok(Self {
+            content,
+            serialized_message,
+        })
+    }
+
+    /// Create from part with serialized_message the bytes of the serialization
+    /// for `Message::Hop(HopMessage::new(content))`
+    pub fn new_with_serialized_message(
+        content: SignedRoutingMessage,
+        serialized_message: Bytes,
+    ) -> Self {
+        Self {
+            content,
+            serialized_message,
+        }
+    }
+
+    pub fn signed_routing_message(&self) -> &SignedRoutingMessage {
+        &self.content
+    }
+
+    pub fn into_signed_routing_message(self) -> SignedRoutingMessage {
+        self.content
+    }
+
+    pub fn serialized_message(&self) -> &Bytes {
+        &self.serialized_message
+    }
+
+    pub fn message_dst(&self) -> &Location<XorName> {
+        &self.content.routing_message().dst
     }
 }
 
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -45,16 +45,16 @@ pub enum Message {
 
 #[allow(clippy::large_enum_variant)]
 pub enum MessageWithBytes {
-    Hop(HopMessageWithSerializedMessage),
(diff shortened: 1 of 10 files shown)


src/messages/mod.rs
```diff
@@ -58,6 +60,67 @@ impl HopMessage {
     }
 }
 
+#[derive(Eq, PartialEq, Clone)]
+pub struct HopMessageWithSerializedMessage {
```
madadam: Suggestions to make the names shorter without losing information:
- `HopMessageWithSerializedMessage` -> `HopMessageWithBytes`
- `serialized_message` -> `bytes`
- `new_with_serialized_message` -> `new_with_bytes`

src/messages/mod.rs
```diff
@@ -58,6 +60,67 @@ impl HopMessage {
...
+            serialized_message,
+        })
+    }
+
+    /// Create from part with serialized_message the bytes of the serialization
+    /// for `Message::Hop(HopMessage::new(content))`
+    pub fn new_with_serialized_message(
+        content: SignedRoutingMessage,
+        serialized_message: Bytes,
+    ) -> Self {
```
madadam: The problem with this method is that there is no guarantee that the bytes actually match the content. Wonder if there is some non-convoluted way to statically enforce that. Maybe something like this:

```rust
enum MessageWithBytes {
    Hop(HopMessageWithBytes),
    Direct(SignedDirectMessage) // we don't need the bytes for this one
}

impl MessageWithBytes {
    fn from_bytes(bytes: &Bytes) -> Result<Self> {
        match from_network_bytes(bytes)? {
            Message::Hop(msg) => Self::Hop(HopMessageWithBytes {
                content: msg.content,
                bytes: bytes.clone()
            })
            Message::Direct(msg) => Self::Direct(msg),
        }
    }
}
```
?

src/messages/mod.rs
```diff
@@ -58,6 +60,67 @@ impl HopMessage {
...
+        self.content
+    }
+
+    pub fn routing_message(&self) -> &RoutingMessage {
+        self.content.routing_message()
+    }
+
+    pub fn serialized_message(&self) -> Bytes {
+        // Bytes is backed by Rc so clone is cheap and result in the same memory held.
+        self.serialized_message.clone()
```
madadam: Why not let the caller do the clone if they need to?

src/messages/mod.rs
jeanphilippeD: Yes, looks pretty good, just one thing, the bytes are not just the serialized SignedRoutingMessage, but the actual on the wire message which is SignedRoutingMessage wrapped in enum and hop message.

But if that does not seem to be confusing I'll go with the rename you suggest (not sure my names do a better job making that distinction to be honest).

src/messages/mod.rs
jeanphilippeD: Yes, maybe something like this will be better.
Yes there is no static enforcement, it is up to the caller to do the right thing, will see if i can do what you suggest.

src/messages/mod.rs
madadam: Hm, I see. Well, if you think that distinction is important to keep, them maybe keep the name.

src/messages/mod.rs
jeanphilippeD: Because they pretty much always do, and I thought putting the comment here made more sense than having clone where we call it. Not sure I'm right.

madadam: I feel like letting the caller do the clone is slightly more idiomatic due to being more explicit. Similar to how I think it's better to take argument by value instead of taking it by reference and cloning internally. But I don't feel too strongly about this, so I'll leave it up to you.

jeanphilippeD: Yes, the problem is normally with Rc, we would use `Rc::clone()` (clippy suggest it I think), as that highlight it is a different kind of clone, but Bytes does not have it, so I'll probably keep it.

madadam: Yes, but `Bytes` isn't alias for `Rc`, it just contains one (actually I think it is `Arc`), so calling `clone` on it is perfectly fine even for clippy.

jeanphilippeD:
Straight rebase, checking clippy + test pass at each stage. will now address comments.

src/messages/mod.rs
jeanphilippeD: Ok, ideally I think it would be best if there was a `bytes::clone()` to make things clear.
Trying to work a way that would address both our concerns.

Option A:
Add a function `bytes_clone` somewhere with explanation that it is cheap and use it at the caller site of serialized_message.

Option B:
Have 2 function `serialized_message` not cloning, and another one doing cloning but also somewhat clear that it is a different clone. I don't really have a good idea.


Would option A makes sense to you, or do you have a good name for option B?

madadam: What is wrong with just returning `&Bytes` from `serialized_message` and letting the caller call `clone` on it? Is it because it's not obvious that it is a cheap clone?

jeanphilippeD: yes

jeanphilippeD: maybe that is not as important as I make it to be, but we seem to be trying to avoid unnecessary clones generally.

madadam: Ok. I'm not sure this is that important. Maybe just add a comment explaining the clone is cheap?

jeanphilippeD: Ok, I'll do that at each call site then

jeanphilippeD:
another straight rebase review comment fix comming

jeanphilippeD:
Addressed review comments in 2 new commits

PR #2013, opened 2020-01-15
Authority -> Location

dirvine:
Renamed Authority to Location. This is the next phase of cleanup and does appear to make more
locial sense when reading the code. It is likely Authority will make a come back though, but as
an actual authority that we can query for validity. Desing considerations though on whether
Authority will be src dst or content. There is reasonable argument that Authority should be for
whole message.

### DIFF
--- a/src/action.rs
+++ b/src/action.rs
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::{authority::Authority, error::InterfaceError, xor_space::XorName};
+use crate::{error::InterfaceError, location::Location, xor_space::XorName};
 use bytes::Bytes;
 use hex_fmt::HexFmt;
 use quic_p2p::Token;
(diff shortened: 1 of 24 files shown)


src/error.rs
```diff
@@ -31,7 +31,7 @@ pub enum RoutingError {
     #[error(display = "Invalid State.")]
     Terminated,
     #[error(display = "Invalid requester or handler authorities.")]
```
madadam: This description should probably be changed to "Invalid requester or handler locations". Also maybe look for any other instances of "authorities" (plural) and rename them as well.

src/action.rs
```diff
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::{authority::Authority, error::InterfaceError, xor_space::XorName};
```
jeanphilippeD: commit titles could be a bit more helpful, than just the scope and type of change.
We were also following the leaf guideline https://github.com/autumnai/leaf/blob/master/CONTRIBUTING.md#git-commit-guidelines so all lower case type/scope:

`refactor/routing: renamed Authority to Location`
`docs/routing: updated documentation for Location struct`
`docs/routing: rename authorities -> locations in docs.`

But in the case of this change, might as well keep a single commit (squash and fix the single commit) will make it easier to rebase as well.

PR #2014, opened 2020-01-16
Simplify errors

madadam:
Merge `InterfaceError` and `RoutingError` into one type.


dirvine:
I think we must be due a semver release and get save_vault on that actually. This will change the API, so perhaps a good time to do that (version update)

jeanphilippeD:
We kept on breaking the API, and probably have not finished yet, safe_vault if ok with its cargo.lock, just need to update it when pulling the new routing (was updated recently).

PR #2015, opened 2020-01-16
Partial deserialize for hop messages

### LINKED ISSUE #2010: Clean up message relay to not deserialize full message on send to next hop.

jeanphilippeD:
Closes #2010 

- Partial de-serialization of Hop message (find it is `Hop`, get `dst`).
- Use tuple serialization for SigneRoutingMessage to ease partial de-serialization.
- Use hash for logging/tracing with a log at the beginning with the routing message and hash.

### DIFF
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -26,7 +26,7 @@ use crate::{
 use bincode::serialize;
 use bytes::Bytes;
 use log::LogLevel;
-use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::{
     collections::BTreeSet,
     fmt::{self, Debug, Formatter},
@@ -310,29 +310,8 @@ pub struct PartialSignedRoutingMessage {
 
 impl<'de> Deserialize<'de> for PartialSignedRoutingMessage {
     fn deserialize<D: Deserializer<'de>>(deserialiser: D) -> std::result::Result<Self, D::Error> {
-        struct PeekVisitor;
-
-        impl<'de> de::Visitor<'de> for PeekVisitor {
-            type Value = PartialSignedRoutingMessage;
-
-            fn expecting(&self, formatter: &mut Formatter) -> fmt::Result {
-                formatter.write_str("expecting to find SignedRoutingMessage and getting field of PeakedSignedRoutingMessageInfo")
-            }
-
-            fn visit_seq<V>(
-                self,
-                mut seq: V,
-            ) -> std::result::Result<PartialSignedRoutingMessage, V::Error>
-            where
-                V: de::SeqAccess<'de>,
-            {
-                let dst = seq
-                    .next_element()?
-                    .ok_or_else(|| de::Error::invalid_length(0, &self))?;
-                Ok(PartialSignedRoutingMessage { dst })
-            }
-        }
-        deserialiser.deserialize_tuple(1, PeekVisitor)
+        let dst: Location = Deserialize::deserialize(deserialiser)?;
+        Ok(PartialSignedRoutingMessage { dst })
     }
 }
 
@@ -760,18 +739,29 @@ mod tests {
         let full_id = FullId::gen(&mut rng);
         let msg = gen_message(&mut rng);
 
-        let signed_msg = unwrap!(SignedRoutingMessage::single_source(msg, &full_id,));
+        let signed_msg_org = unwrap!(SignedRoutingMessage::single_source(msg, &full_id,));
 
-        let msg = unwrap!(HopMessageWithBytes::new(signed_msg, &LogIdent::new("node")));
+        let msg = unwrap!(HopMessageWithBytes::new(
+            signed_msg_org.clone(),
+            &LogIdent::new("node")
+        ));
         let bytes = msg.full_message_bytes();
-        let partial_content = unwrap!(bincode::deserialize::<PartialMessage>(bytes));
-        let partial_msg_head = unwrap!(bincode::deserialize::<PartialMessage>(&bytes[0..40]));
+        let full_msg = unwrap!(from_network_bytes(bytes));
+        let partial_msg = unwrap!(partial_from_network_bytes(bytes));
+        let partial_msg_head = unwrap!(partial_from_network_bytes(&bytes.slice(0, 40)));
 
-        let expected = PartialMessage::Hop(PartialSignedRoutingMessage {
-            dst: *msg.message_dst(),
+        let expected_partial = PartialMessage::Hop(PartialSignedRoutingMessage {
+            dst: signed_msg_org.routing_message().dst,
         });
-        assert_eq!(partial_content, expected);
-        assert_eq!(partial_msg_head, expected);
+        let signed_msg = if let Message::Hop(signed_msg) = full_msg {
+            Some(signed_msg)
+        } else {
+            None
+        };
+
+        assert_eq!(partial_msg, expected_partial);
+        assert_eq!(partial_msg_head, expected_partial);
+        assert_eq!(signed_msg, Some(signed_msg_org))
     }
 
     fn make_proof_chain(pk_set: &bls::PublicKeySet) -> SectionProofChain {
@@ -788,10 +778,9 @@ mod tests {
     fn gen_message(rng: &mut MainRng) -> RoutingMessage {
         use rand::distributions::Standard;
 
-        let name = rng.gen();
         RoutingMessage {
-            src: Location::Section(name),
-            dst: Location::Section(name),
+            src: Location::Section(rng.gen()),
+            dst: Location::Section(rng.gen()),
             content: MessageContent::UserMessage(rng.sample_iter(Standard).take(6).collect()),
         }
     }
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -20,6 +20,7 @@ use crate::{
     id::{FullId, PublicId},
     location::Location,
     states::common::{from_network_bytes, to_network_bytes},
+    utils::LogIdent,
     xor_space::{Prefix, XorName},
 };
 use bincode::serialize;
@@ -71,7 +72,7 @@ pub struct HopMessageWithBytes {
 
 impl HopMessageWithBytes {
     /// Serialize message and keep both SignedRoutingMessage and Bytes.
-    pub fn new(content: SignedRoutingMessage) -> Result<Self> {
+    pub fn new(content: SignedRoutingMessage, log_ident: &LogIdent) -> Result<Self> {
         let message = Message::Hop(content);
         let full_message_bytes = to_network_bytes(&message)?;
 
@@ -81,7 +82,16 @@ impl HopMessageWithBytes {
             unreachable!("Created as Hop can only match Hop.")
         };
 
-        Ok(Self::new_from_parts(content, full_message_bytes))
+        let message = Self::new_from_parts(content, full_message_bytes);
+
+        trace!(
+            "{} Creating message hash({:?}) {:?}",
+            log_ident,
+            message.full_message_crypto_hash,
+            message.content.routing_message(),
+        );
+
+        Ok(message)
     }
 
     fn new_from_parts(content: SignedRoutingMessage, full_message_bytes: Bytes) -> Self {
@@ -100,10 +110,6 @@ impl HopMessageWithBytes {
 
     pub fn into_signed_routing_message(self) -> SignedRoutingMessage {
         self.content
-    }
-
-    pub fn routing_message(&self) -> &RoutingMessage {
-        self.content.routing_message()
     }
 
     pub fn full_message_bytes(&self) -> &Bytes {
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -19,7 +19,7 @@ use crate::{
     error::{Result, RoutingError},
     id::{FullId, PublicId},
     location::Location,
-    states::common::{from_network_bytes, peek_from_network_bytes, to_network_bytes},
+    states::common::{from_network_bytes, partial_from_network_bytes, to_network_bytes},
     utils::LogIdent,
     xor_space::{Prefix, XorName},
 };
@@ -47,11 +47,11 @@ pub enum Message {
 
 #[derive(Debug, Eq, PartialEq, Hash, Deserialize)]
 #[allow(clippy::large_enum_variant)]
-pub enum MessagePeek {
+pub enum PartialMessage {
     /// A message sent between two nodes directly
     Direct(SignedDirectMessage),
     /// A message sent across the network (in transit)
-    Hop(PeekedSignedRoutingMessageInfo),
+    Hop(PartialSignedRoutingMessage),
 }
 
 #[allow(clippy::large_enum_variant)]
@@ -61,19 +61,14 @@ pub enum MessageWithBytes {
 }
 
 impl MessageWithBytes {
-    pub fn peek_from_bytes(bytes: Bytes) -> Result<Self> {
-        match peek_from_network_bytes(&bytes)? {
-            MessagePeek::Hop(msg_peek) => Ok(Self::Hop(HopMessageWithBytes::new_from_parts(
-                None, msg_peek, bytes,
+    pub fn partial_from_bytes(bytes: Bytes) -> Result<Self> {
+        match partial_from_network_bytes(&bytes)? {
+            PartialMessage::Hop(msg_partial) => Ok(Self::Hop(HopMessageWithBytes::new_from_parts(
+                None,
+                msg_partial,
+                bytes,
             ))),
-            MessagePeek::Direct(msg) => Ok(Self::Direct(msg, bytes)),
-        }
-    }
-
-    pub fn full_hop_msg_from_bytes(bytes: &Bytes) -> Result<SignedRoutingMessage> {
-        match from_network_bytes(bytes)? {
-            Message::Hop(msg) => Ok(msg),
-            Message::Direct(_msg) => Err(RoutingError::InvalidMessage),
+            PartialMessage::Direct(msg) => Ok(Self::Direct(msg, bytes)),
         }
     }
 }
@@ -83,8 +78,8 @@ impl MessageWithBytes {
 pub struct HopMessageWithBytes {
     /// Wrapped signed message.
     full_content: Option<SignedRoutingMessage>,
-    /// Peeked SignedRoutingMessage infos
-    peek_msg: PeekedSignedRoutingMessageInfo,
+    /// Partial SignedRoutingMessage infos
+    partial_content: PartialSignedRoutingMessage,
     /// Serialized Message as received or sent to quic_p2p.
     full_message_bytes: Bytes,
     /// Crypto hash of the full message.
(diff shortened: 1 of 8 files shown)


src/messages/mod.rs
```diff
@@ -43,17 +45,35 @@ pub enum Message {
...
+        match peek_from_network_bytes(&bytes)? {
+            MessagePeek::Hop(msg_peek) => Ok(Self::Hop(HopMessageWithBytes::new_from_parts(
+                None, msg_peek, bytes,
+            ))),
+            MessagePeek::Direct(msg) => Ok(Self::Direct(msg, bytes)),
+        }
+    }
+
+    pub fn full_hop_msg_from_bytes(bytes: &Bytes) -> Result<SignedRoutingMessage> {
+        match from_network_bytes(bytes)? {
```
maqi: the `from_network_bytes` function could be used to handle a Direct message, but here we only expect it to handle hop message?

src/messages/mod.rs
```diff
@@ -71,39 +93,53 @@ pub struct HopMessageWithBytes {
...
     /// Serialize message and keep both SignedRoutingMessage and Bytes.
-    pub fn new(content: SignedRoutingMessage) -> Result<Self> {
-        let message = Message::Hop(content);
+    pub fn new(full_content: SignedRoutingMessage, log_ident: &LogIdent) -> Result<Self> {
+        let message = Message::Hop(full_content);
         let full_message_bytes = to_network_bytes(&message)?;
 
-        let content = if let Message::Hop(content) = message {
-            content
+        let full_content = if let Message::Hop(full_content) = message {
```
maqi: we just composed the message, and then de-compose it?
is it better than just clone the full_content ?

src/messages/mod.rs
```diff
@@ -71,39 +93,53 @@ pub struct HopMessageWithBytes {
...
         };
 
-        Ok(Self::new_from_parts(content, full_message_bytes))
-    }
-
-    fn new_from_parts(content: SignedRoutingMessage, full_message_bytes: Bytes) -> Self {
+        let peek_msg = PeekedSignedRoutingMessageInfo {
+            dst: full_content.routing_message().dst,
+        };
+        let message = Self::new_from_parts(Some(full_content), peek_msg, full_message_bytes);
```
maqi: the above message is `Message::Hop(full_content)` , and here is `new_from_parts` 
if they are different, will it be better to have a different name ?

src/messages/mod.rs
```diff
@@ -231,14 +275,77 @@ impl Debug for SecurityMetadata {
...
     /// A request or response type message.
     content: RoutingMessage,
     /// Optional metadata for verifying the sender
     security_metadata: SecurityMetadata,
 }
 
+impl Serialize for SignedRoutingMessage {
+    fn serialize<S: Serializer>(&self, serialiser: S) -> ::std::result::Result<S::Ok, S::Error> {
+        (
+            &self.content.dst,
```
maqi: as the src and dst are having the same format, just to confirm that this is the preferred order to use, right?

src/messages/mod.rs
```diff
@@ -231,14 +275,77 @@ impl Debug for SecurityMetadata {
...
+impl<'de> Deserialize<'de> for SignedRoutingMessage {
+    fn deserialize<D: Deserializer<'de>>(deserialiser: D) -> ::std::result::Result<Self, D::Error> {
+        let (dst, src, content, security_metadata): (
+            Location,
+            Location,
+            MessageContent,
+            SecurityMetadata,
+        ) = Deserialize::deserialize(deserialiser)?;
+        Ok(Self {
+            content: RoutingMessage { dst, src, content },
```
maqi: RoutingMessage is defined as ` {src, dst, content}` at https://github.com/maidsafe/routing/blob/fleming/src/messages/mod.rs#L477-L484 ?

src/states/elder/mod.rs
```diff
@@ -706,20 +706,20 @@ impl Elder {
...
     fn handle_filtered_signed_message(
         &mut self,
-        msg: HopMessageWithBytes,
+        mut msg: HopMessageWithBytes,
     ) -> Result<(), RoutingError> {
         trace!(
             "{} - Handle signed message: {:?}",
             self,
-            msg.routing_message()
+            msg.full_message_crypto_hash()
```
maqi: a hash of message won't help much with debugging, meanwhile the previous whole message will do as it contains searchable key words of network actions.
so here (and for some other trace places as well), we may still need the whole msg to be logged.
if cann't do that, maybe just remove the log ?

src/messages/mod.rs
jeanphilippeD: Yes, it just move parts around without cloning.

src/messages/mod.rs
jeanphilippeD: yes, it is a full *hop* message from bytes. The direct message was already deserialized on its own.

src/messages/mod.rs
jeanphilippeD: We need dst for partial deserialization, so we need dst to be first.

src/messages/mod.rs
jeanphilippeD: it does not matter, since these are named fields, but I can fix.

src/states/elder/mod.rs
jeanphilippeD: If you look at the commit, you'll see that when creating a HopeMessageWithBytes from a SignedRoutingMessage we log it, so that first instance of seeing the hash in your log file should contain the full message.

Do you think it won't be enough?

maqi: Ah, ok.
Yeah, that shall be enough.

jeanphilippeD:
Addressed review comments in new commit.

src/messages/mod.rs
```diff
@@ -71,39 +93,58 @@ pub struct HopMessageWithBytes {
...
+    pub fn new(full_content: SignedRoutingMessage, log_ident: &LogIdent) -> Result<Self> {
+        let hop_msg_result = {
+            let (full_content, full_message_bytes) = {
+                let full_message = Message::Hop(full_content);
+                let full_message_bytes = to_network_bytes(&full_message)?;
+
+                if let Message::Hop(full_content) = full_message {
+                    (full_content, full_message_bytes)
+                } else {
+                    unreachable!("Created as Hop can nly match Hop.")
```
maqi: typo, `only` ?

src/states/elder/mod.rs
madadam: If this ends up being a hassle it might be worth to write a small utility to pipe the logs through which resolves the hashes.

jeanphilippeD: I think it may actually end up quite useful, generally I see the message I care at the start, and then I want to search for it and often I'm missing bits, a hash may be easier in the end :-)

jeanphilippeD:
Fixed typo.

src/messages/mod.rs
```diff
@@ -63,7 +82,9 @@ impl MessageWithBytes {
 #[derive(Eq, PartialEq, Clone)]
 pub struct HopMessageWithBytes {
     /// Wrapped signed message.
-    content: SignedRoutingMessage,
+    full_content: Option<SignedRoutingMessage>,
+    /// Peeked SignedRoutingMessage infos
+    peek_msg: PeekedSignedRoutingMessageInfo,
```
madadam: It seems that `peek_msg` is only relevant when `full_content` is `None`. So to avoid redundant states, I would change this to a single enum:

```rust
enum HopMessageMode { // you can come up with better name
    Peeked(PeekedSignedRoutingMessageInfo),
    Full(SignedRoutingMessage),
}

struct HopMessageWithBytes {
    content: HopMessageMode,
    // ...
}
```

src/messages/mod.rs
```diff
@@ -44,17 +45,35 @@ pub enum Message {
...
+    pub fn peek_from_bytes(bytes: Bytes) -> Result<Self> {
+        match peek_from_network_bytes(&bytes)? {
+            MessagePeek::Hop(msg_peek) => Ok(Self::Hop(HopMessageWithBytes::new_from_parts(
+                None, msg_peek, bytes,
+            ))),
+            MessagePeek::Direct(msg) => Ok(Self::Direct(msg, bytes)),
+        }
+    }
+
+    pub fn full_hop_msg_from_bytes(bytes: &Bytes) -> Result<SignedRoutingMessage> {
```
madadam: Why is this not a normal method that takes `self`? i.e.:
```rust
pub fn full_hop_msg_from_bytes(&self) -> Result<SignedRoutingMessage> {
    match from_network_bytes(&self.bytes) {
        ..
    }
}
```

src/messages/mod.rs
```diff
@@ -237,14 +275,77 @@ impl Debug for SecurityMetadata {
...
 /// Wrapper around a routing message, signed by the originator of the message.
-#[derive(Eq, PartialEq, Clone, Hash, Serialize, Deserialize)]
+#[derive(Debug, Eq, PartialEq, Clone, Hash)]
+pub struct PeekedSignedRoutingMessageInfo {
+    /// Destination location
+    pub dst: Location,
+}
+
+impl<'de> Deserialize<'de> for PeekedSignedRoutingMessageInfo {
+    fn deserialize<D: Deserializer<'de>>(deserialiser: D) -> ::std::result::Result<Self, D::Error> {
```
madadam: You don't need to prefix stuff with `::` (also below)

src/states/common/base.rs
```diff
@@ -452,3 +452,7 @@ pub fn to_network_bytes(message: &Message) -> Result<Bytes, RoutingError> {
 pub fn from_network_bytes(data: &Bytes) -> Result<Message, RoutingError> {
     Ok(bincode::deserialize(&data[..])?)
 }
+
+pub fn peek_from_network_bytes(data: &Bytes) -> Result<MessagePeek, RoutingError> {
+    Ok(bincode::deserialize(&data[..])?)
+}
```
madadam: As we no longer have mock serialization, we can now change `from_network_bytes` and `to_network_bytes` to be generic. The we wouldn't need this function.

madadam: Or maybe even remove those function altogether and use `bincode` directly.

src/messages/mod.rs
```diff
@@ -72,44 +93,53 @@ pub struct HopMessageWithBytes {
...
         }
     }
 
-    pub fn signed_routing_message(&self) -> &SignedRoutingMessage {
-        &self.content
-    }
-
-    pub fn into_signed_routing_message(self) -> SignedRoutingMessage {
-        self.content
+    pub fn take_or_deserialize_signed_routing_message(&mut self) -> Result<SignedRoutingMessage> {
```
madadam: Is such verbose name necessary? Isn't the source of the message an implementation detail that the caller doesn't need to be concerned with? I.e., just call it `take_signed_routing_message`.

src/messages/mod.rs
madadam: Also it should probably be called `signed_routing_message_from_bytes` for consistency.

src/messages/mod.rs
```diff
@@ -237,14 +275,77 @@ impl Debug for SecurityMetadata {
...
 
 /// Wrapper around a routing message, signed by the originator of the message.
-#[derive(Eq, PartialEq, Clone, Hash, Serialize, Deserialize)]
+#[derive(Debug, Eq, PartialEq, Clone, Hash)]
+pub struct PeekedSignedRoutingMessageInfo {
+    /// Destination location
+    pub dst: Location,
+}
+
+impl<'de> Deserialize<'de> for PeekedSignedRoutingMessageInfo {
```
madadam: I wonder, can't we just do:
```rust
let peeked_message: PeekedSignedRoutingMessageInfo = deserialize(&bytes)?;
```
Or would that fail because of not consuming the bytes to the end?

src/messages/mod.rs
```diff
@@ -237,14 +275,77 @@ impl Debug for SecurityMetadata {
 }
 
 /// Wrapper around a routing message, signed by the originator of the message.
-#[derive(Eq, PartialEq, Clone, Hash, Serialize, Deserialize)]
+#[derive(Debug, Eq, PartialEq, Clone, Hash)]
+pub struct PeekedSignedRoutingMessageInfo {
```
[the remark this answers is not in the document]
madadam: Or `PartialSignedRoutingMessage`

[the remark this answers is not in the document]
madadam: Or explicitly `SignedRoutingMessageDst`

src/messages/mod.rs
jeanphilippeD: That is not really the case, essentially full_content is only there to be taken. it is a cached value.
So when we took that value, the PeekedSignedRoutingMessageInfo should still be there.

src/messages/mod.rs
jeanphilippeD: Then it would need to be a method of `HopMessageWithBytes`.
I wanted to keep this function and `peek_from_bytes` together, maybe not a good idea?

src/messages/mod.rs
madadam: Then when we take the message out, we can set `content` back to the peeked thing.

src/states/common/base.rs
jeanphilippeD: I'm not sure the generic option really helps.
I think it is not unreasonable to look at removing these function and use bincode directly (though I'm not sure it would make things more ergonomic/readable so I'd be incline to leave it until we decide it makes more sense?

src/messages/mod.rs
madadam: It's not very idiomatic. Associated functions usually return `Self` or something that contains `Self`. Also you only call this function from `deserialize_signed_routing_message`, so why not just do the deserialization inline there and not even have this function?

src/messages/mod.rs
jeanphilippeD: take imply a quick move, I would want to make sure it is clear what is happening.
Since it is used so few times is that an issue?

src/messages/mod.rs
jeanphilippeD: We deserialize a tupple, since we serialize a tuple.
But I'll look whether I can just deserialize to a single `Location` value.

Cannot deserialize(&bytes)?; because bytes contain an enum of either direct or hop

src/messages/mod.rs
madadam: Then maybe `extract_signed_routing_message`. Or leave it as is if you prefer. Not a big issue for me.

src/messages/mod.rs
```diff
@@ -237,14 +275,77 @@ impl Debug for SecurityMetadata {
...
+    }
+}
+
+impl<'de> Deserialize<'de> for SignedRoutingMessage {
+    fn deserialize<D: Deserializer<'de>>(deserialiser: D) -> ::std::result::Result<Self, D::Error> {
+        let (dst, src, content, security_metadata): (
+            Location,
+            Location,
+            MessageContent,
+            SecurityMetadata,
```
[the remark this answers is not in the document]
jeanphilippeD: It complained originaly but will try again.

src/messages/mod.rs
jeanphilippeD: Does that really make things clearer, we need to add more code? I guess I could rename full_content to `signed_message_cache`?

but it is possible to implement `take` on `HopMessageMode` if you prefer it this way.

src/messages/mod.rs
madadam: I think you should be able to serialize a `Message` but deserialize `PeekMessage`. Then you wouldn't need this custom deserialization logic. But I'm not sure it would really work because it would not consume the whole buffer. But might be worth a try.

jeanphilippeD: We already not consuming the full buffer, so not really an issue. as I said will try.

src/messages/mod.rs
madadam: Well the only thing that's somehow bothering me is we are duplicating information: when `full_content` is `Some`, then the dst location is stored twice. Not saying this is a major issue, but it feels bit dirty.

src/messages/mod.rs
jeanphilippeD: Well that worked, that was an unfortunate rabbit hole.

src/messages/mod.rs
```diff
@@ -630,6 +742,26 @@ mod tests {
...
+        let msg = unwrap!(HopMessageWithBytes::new(signed_msg, &LogIdent::new("node")));
+        let bytes = msg.full_message_bytes();
+        let peek_msg = unwrap!(bincode::deserialize::<MessagePeek>(bytes));
+        let peek_msg_head = unwrap!(bincode::deserialize::<MessagePeek>(&bytes[0..40]));
+
+        let expected = MessagePeek::Hop(PeekedSignedRoutingMessageInfo {
+            dst: *msg.message_dst(),
+        });
+        assert_eq!(peek_msg, expected);
+        assert_eq!(peek_msg_head, expected);
```
madadam: Also verify that the bytes can be deserialized into the full message too.

src/messages/mod.rs
jeanphilippeD: Seem to be an awful lot of complicated code, for something that is I just cached a big value, and otherwise I need the dst.

```rust
#[derive(Eq, PartialEq, Clone)]
enum HopMessageDeserializedContent {
    Peeked(PartialSignedRoutingMessage),
    Full(SignedRoutingMessage),
}

impl HopMessageDeserializedContent {
    fn take_full(&mut self) -> Option<SignedRoutingMessage> {
        let dst = *self.message_dst();
        let value = mem::replace(self, Self::Peeked(PartialSignedRoutingMessage{dst}));
        match value {
            Self::Peeked(_) => None,
            Self::Full(msg) => Some(msg),
        }
    }

    fn message_dst(&self) -> &Location {
        match self {
            Self::Peeked(msg) => &msg.dst,
            Self::Full(msg) => &msg.routing_message().dst,
        }
    }
}
```

jeanphilippeD: so i think i'll stick with the less verbose solution for now.

jeanphilippeD:
Addressed review comments.

src/messages/mod.rs
```diff
@@ -43,17 +45,28 @@ pub enum Message {
     Hop(SignedRoutingMessage),
 }
 
+#[derive(Debug, Eq, PartialEq, Hash, Deserialize)]
+#[allow(clippy::large_enum_variant)]
+pub enum MessagePeek {
```
madadam: Since you accepted the "Partial" naming scheme, maybe use it everywhere for consistency. So this can be `PartialMessage` ?

src/messages/mod.rs
```diff
@@ -62,7 +75,9 @@ impl MessageWithBytes {
 #[derive(Eq, PartialEq, Clone)]
 pub struct HopMessageWithBytes {
     /// Wrapped signed message.
-    content: SignedRoutingMessage,
+    full_content: Option<SignedRoutingMessage>,
+    /// Peeked SignedRoutingMessage infos
+    peek_msg: PartialSignedRoutingMessage,
```
madadam: ...and this can be `partial_content`.

src/messages/mod.rs
```diff
@@ -231,14 +276,51 @@ impl Debug for SecurityMetadata {
...
+    /// Destination location
+    pub dst: Location,
+}
+
+impl<'de> Deserialize<'de> for PartialSignedRoutingMessage {
+    fn deserialize<D: Deserializer<'de>>(deserialiser: D) -> std::result::Result<Self, D::Error> {
+        let dst: Location = Deserialize::deserialize(deserialiser)?;
+        Ok(PartialSignedRoutingMessage { dst })
+    }
+}
```
madadam: Is this manual impl necessary? Why not just `derive(Deserialize)` instead?

src/messages/mod.rs
```diff
@@ -231,14 +276,51 @@ impl Debug for SecurityMetadata {
...
+#[derive(Eq, PartialEq, Clone, Hash)]
 pub struct SignedRoutingMessage {
     /// A request or response type message.
     content: RoutingMessage,
     /// Optional metadata for verifying the sender
     security_metadata: SecurityMetadata,
 }
 
+impl Serialize for SignedRoutingMessage {
+    fn serialize<S: Serializer>(&self, serialiser: S) -> ::std::result::Result<S::Ok, S::Error> {
```
madadam: Another redundant `::`

src/messages/mod.rs
```diff
@@ -231,14 +276,51 @@ impl Debug for SecurityMetadata {
...
+            &self.content.src,
+            &self.content.content,
+            &self.security_metadata,
+        )
+            .serialize(serialiser)
+    }
+}
+
+impl<'de> Deserialize<'de> for SignedRoutingMessage {
+    fn deserialize<D: Deserializer<'de>>(deserialiser: D) -> ::std::result::Result<Self, D::Error> {
```
madadam: Also here

src/states/common/base.rs
madadam: OK, I'm fine with postponing this one.

src/messages/mod.rs
jeanphilippeD: I think it is, the struct derive is different, but I'll try.

jeanphilippeD: It does work, though it feels really wrong, we serialize the SigneRoutingMessage as a tuple it seem more reasonable to deserialize it as a tuple. Cannot have tuple of 1, so next best thing I though was as the sole member.

jeanphilippeD:
Addressed review comments also strengthen the test a bit.

src/messages/mod.rs
madadam: It works with bincode (and possibly some other formats) because it doesn't store any tags, headers, etc.. So a struct of one field is serialized the same as that field by itself. It doesn't work in general though. We are unlikely to switch to other serialization format any time soon so this should be fine. Best to have it covered by tests though, which you do so we should be good.

jeanphilippeD: do you mean I should still just use derive(Deserialize)?

madadam: Also, a tuple of N elements is serialized as just the first element, then the second, etc... There is no header saying how many elements there are and there are no separators between the elements. This is because those information can be inferred from the type you are deserializing into.

madadam: > do you mean I should still just use derive(Deserialize)?

Yes. It achieves the same effect with less (hand-written) code.

jeanphilippeD:
Pushed commit to address last review comment (forgot to do that).

src/messages/mod.rs
```diff
@@ -62,7 +77,9 @@ impl MessageWithBytes {
 #[derive(Eq, PartialEq, Clone)]
 pub struct HopMessageWithBytes {
     /// Wrapped signed message.
-    content: SignedRoutingMessage,
+    full_content: Option<SignedRoutingMessage>,
+    /// Peeked SignedRoutingMessage infos
```
maqi: the comment may need to be updated as the field name has been changed ?

jeanphilippeD:
push force rebased version (tested local 2300 churn 5000 others).
previous push force fixed review comment

PR #2018, opened 2020-01-20
Fix hop message handling

madadam:
This PR fixes two failures:

1. When we are recently joined node and we have a backlogged message that we received while still in `JoiningPeer` state, we must forward the message to our elders if we can't handle it ourselves instead of ignoring it. This prevents losing `NeighbourInfo` votes (among possibly other things)
2. When we are elder and we receive a message whose proof is too new for us, we must not fail but instead forward the message to the other elders. This is because we've already been demoted, we just haven't catched up to it yet.

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -870,17 +870,42 @@ impl Chain {
 
     /// Returns `true` if the `proof_chain` contains a key we have in `their_keys` and that key is
     /// for a prefix compatible with proof_chain prefix.
-    pub fn check_trust(&self, proof_chain: &SectionProofChain) -> bool {
+    pub fn check_trust(&self, proof_chain: &SectionProofChain) -> TrustStatus {
         let last_prefix = proof_chain.last_public_key_info().prefix();
-        let filtered_keys: BTreeSet<_> = self
+        let known_key_infos: BTreeSet<_> = self
             .state
             .get_their_keys_info()
             .filter(|&(pfx, _)| last_prefix.is_compatible(pfx))
             .map(|(_, info)| info)
             .collect();
-        proof_chain
+
+        if proof_chain
             .all_key_infos()
-            .any(|key_info| filtered_keys.contains(key_info))
+            .any(|proof_key_info| known_key_infos.contains(proof_key_info))
+        {
+            return TrustStatus::Trusted;
+        }
+
+        let max_known_version = match known_key_infos
+            .iter()
+            .map(|known_key_info| known_key_info.version())
+            .max()
+        {
+            Some(version) => version,
+            None => return TrustStatus::ProofInvalid,
+        };
+
+        let min_proof_version = proof_chain
+            .all_key_infos()
+            .map(|proof_key_info| proof_key_info.version())
+            .min()
+            .expect("empty proof");
+
+        if min_proof_version > max_known_version {
+            TrustStatus::ProofTooNew
+        } else {
+            TrustStatus::ProofInvalid
+        }
     }
 
     /// Returns `true` if the `EldersInfo` isn't known to us yet.
@@ -1796,6 +1821,27 @@ impl EldersChangeBuilder {
                 .difference(&new_neighbour)
                 .cloned()
                 .collect(),
+        }
+    }
+}
+
+// Result of a message trust check.
+#[derive(Debug)]
+pub enum TrustStatus {
+    // Message is trusted.
+    Trusted,
+    // Message is untrusted because the proof is invalid.
+    ProofInvalid,
+    // Message trust cannot be determined because the proof starts at version that is newer than
+    // our latest one.
+    ProofTooNew,
+}
+
+impl TrustStatus {
+    pub fn is_trusted(&self) -> bool {
+        match self {
+            Self::Trusted => true,
+            _ => false,
         }
     }
 }
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -8,7 +8,7 @@
 
 use super::{
     chain_accumulator::{AccumulatingProof, ChainAccumulator, InsertError},
-    shared_state::{SectionKeyInfo, SectionProofBlock, SharedState, SplitCache},
+    shared_state::{SectionKeyInfo, SectionProofBlock, SharedState, SplitCache, TrustStatus},
     AccumulatedEvent, AccumulatingEvent, AgeCounter, EldersChange, EldersInfo, GenesisPfxInfo,
     MemberInfo, MemberPersona, MemberState, NetworkEvent, NetworkParams, Proof, ProofSet,
     SectionProofChain,
@@ -868,44 +868,9 @@ impl Chain {
         self.state.get_their_keys_info()
     }
 
-    /// Returns `true` if the `proof_chain` contains a key we have in `their_keys` and that key is
-    /// for a prefix compatible with proof_chain prefix.
+    /// Check whether we trust the given proof chain.
     pub fn check_trust(&self, proof_chain: &SectionProofChain) -> TrustStatus {
-        let last_prefix = proof_chain.last_public_key_info().prefix();
-        let known_key_infos: BTreeSet<_> = self
-            .state
-            .get_their_keys_info()
-            .filter(|&(pfx, _)| last_prefix.is_compatible(pfx))
-            .map(|(_, info)| info)
-            .collect();
-
-        if proof_chain
-            .all_key_infos()
-            .any(|proof_key_info| known_key_infos.contains(proof_key_info))
-        {
-            return TrustStatus::Trusted;
-        }
-
-        let max_known_version = match known_key_infos
-            .iter()
-            .map(|known_key_info| known_key_info.version())
-            .max()
-        {
-            Some(version) => version,
-            None => return TrustStatus::ProofInvalid,
-        };
-
-        let min_proof_version = proof_chain
-            .all_key_infos()
-            .map(|proof_key_info| proof_key_info.version())
-            .min()
-            .expect("empty proof");
-
-        if min_proof_version > max_known_version {
-            TrustStatus::ProofTooNew
-        } else {
-            TrustStatus::ProofInvalid
-        }
+        proof_chain.check_trust(self.state.get_their_keys_info())
     }
 
     /// Returns `true` if the `EldersInfo` isn't known to us yet.
@@ -1821,27 +1786,6 @@ impl EldersChangeBuilder {
                 .difference(&new_neighbour)
                 .cloned()
                 .collect(),
-        }
-    }
-}
-
-// Result of a message trust check.
-#[derive(Debug)]
-pub enum TrustStatus {
-    // Message is trusted.
-    Trusted,
-    // Message is untrusted because the proof is invalid.
-    ProofInvalid,
-    // Message trust cannot be determined because the proof starts at version that is newer than
-    // our latest one.
-    ProofTooNew,
-}
-
-impl TrustStatus {
-    pub fn is_trusted(&self) -> bool {
-        match self {
-            Self::Trusted => true,
-            _ => false,
         }
     }
 }
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -8,7 +8,7 @@
 
 use super::{
     chain_accumulator::{AccumulatingProof, ChainAccumulator, InsertError},
-    shared_state::{SectionKeyInfo, SectionProofBlock, SharedState, SplitCache, TrustStatus},
+    shared_state::{SectionKeyInfo, SectionProofBlock, SharedState, SplitCache},
     AccumulatedEvent, AccumulatingEvent, AgeCounter, EldersChange, EldersInfo, GenesisPfxInfo,
     MemberInfo, MemberPersona, MemberState, NetworkEvent, NetworkParams, Proof, ProofSet,
     SectionProofChain,
@@ -866,11 +866,6 @@ impl Chain {
     /// Return the keys we know
     pub fn get_their_keys_info(&self) -> impl Iterator<Item = (&Prefix<XorName>, &SectionKeyInfo)> {
         self.state.get_their_keys_info()
-    }
-
-    /// Check whether we trust the given proof chain.
-    pub fn check_trust<'a>(&self, proof_chain: &'a SectionProofChain) -> TrustStatus<'a> {
-        proof_chain.check_trust(self.state.get_their_keys_info())
     }
 
     /// Returns `true` if the `EldersInfo` isn't known to us yet.
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -869,7 +869,7 @@ impl Chain {
     }
 
     /// Check whether we trust the given proof chain.
-    pub fn check_trust(&self, proof_chain: &SectionProofChain) -> TrustStatus {
+    pub fn check_trust<'a>(&self, proof_chain: &'a SectionProofChain) -> TrustStatus<'a> {
         proof_chain.check_trust(self.state.get_their_keys_info())
     }
 
--- a/src/chain/shared_state.rs
+++ b/src/chain/shared_state.rs
@@ -622,11 +622,7 @@ impl SectionProofChain {
             None => return TrustStatus::ProofInvalid,
         };
 
-        let min_proof_version = self
-            .all_key_infos()
-            .map(|proof_key_info| proof_key_info.version())
-            .min()
-            .expect("empty proof");
+        let min_proof_version = self.genesis_key_info.version();
 
         if min_proof_version > max_known_version {
             TrustStatus::ProofTooNew
--- a/src/chain/shared_state.rs
+++ b/src/chain/shared_state.rs
@@ -513,6 +513,14 @@ impl SectionProofBlock {
             false
         }
     }
+
+    pub fn prefix(&self) -> &Prefix<XorName> {
+        self.key_info.prefix()
+    }
+
+    pub fn version(&self) -> u64 {
+        self.key_info.version()
+    }
 }
 
 #[derive(Debug, Eq, PartialEq, Clone, Hash, Serialize, Deserialize)]
@@ -534,7 +542,9 @@ impl SectionProofChain {
     }
 
     pub fn push(&mut self, block: SectionProofBlock) {
-        self.blocks.push(block);
+        if self.validate_next_block(&block) {
+            self.blocks.push(block);
+        }
     }
 
     pub fn validate(&self) -> bool {
@@ -582,6 +592,43 @@ impl SectionProofChain {
             genesis_key_info,
             blocks,
         }
+    }
+
+    fn validate_next_block(&self, block: &SectionProofBlock) -> bool {
+        let last = self.last_public_key_info();
+        if block.version() != last.version() + 1 {
+            log_or_panic!(
+                LogLevel::Error,
+                "Invalid next block version - not successor (last: {}, next: {})",
+                last.version(),
+                block.version()
+            );
+            return false;
+        }
+
+        if !block.prefix().is_compatible(last.prefix()) {
+            log_or_panic!(
+                LogLevel::Error,
+                "Invalid next block prefix - not compatible (last: {:?}, next: {:?})",
+                last.prefix(),
+                block.prefix()
+            );
+            return false;
+        }
+
+        if block.prefix() != last.prefix()
+            && block.prefix().bit_count() != last.prefix().bit_count() + 1
+        {
+            log_or_panic!(
+                LogLevel::Error,
+                "Invalid next block prefix - differs by more than one bit (last: {:?}, next: {:?})",
+                last.prefix(),
+                block.prefix()
+            );
+            return false;
+        }
+
+        true
     }
 }
 
--- a/src/chain/shared_state.rs
+++ b/src/chain/shared_state.rs
@@ -18,7 +18,7 @@ use bincode::{deserialize, serialize};
 use itertools::Itertools;
 use log::LogLevel;
 use std::{
-    collections::{BTreeMap, VecDeque},
+    collections::{BTreeMap, BTreeSet, VecDeque},
     fmt::{self, Debug, Formatter},
     iter, mem,
 };
@@ -594,6 +594,47 @@ impl SectionProofChain {
         }
     }
 
+    // Verify this proof chain against the given key infos.
+    pub fn check_trust<'a, I>(&self, their_key_infos: I) -> TrustStatus
+    where
+        I: IntoIterator<Item = (&'a Prefix<XorName>, &'a SectionKeyInfo)>,
+    {
+        let last_prefix = self.last_public_key_info().prefix();
+        let known_key_infos: BTreeSet<_> = their_key_infos
+            .into_iter()
+            .filter(|&(pfx, _)| last_prefix.is_compatible(pfx))
+            .map(|(_, info)| info)
+            .collect();
+
+        if self
+            .all_key_infos()
+            .any(|proof_key_info| known_key_infos.contains(proof_key_info))
+        {
+            return TrustStatus::Trusted;
+        }
+
+        let max_known_version = match known_key_infos
+            .iter()
+            .map(|known_key_info| known_key_info.version())
+            .max()
+        {
+            Some(version) => version,
+            None => return TrustStatus::ProofInvalid,
+        };
+
+        let min_proof_version = self
+            .all_key_infos()
+            .map(|proof_key_info| proof_key_info.version())
+            .min()
+            .expect("empty proof");
+
+        if min_proof_version > max_known_version {
+            TrustStatus::ProofTooNew
+        } else {
+            TrustStatus::ProofInvalid
+        }
+    }
+
     fn validate_next_block(&self, block: &SectionProofBlock) -> bool {
         let last = self.last_public_key_info();
         if block.version() != last.version() + 1 {
@@ -670,6 +711,27 @@ impl SectionKeyInfo {
 
     pub fn serialise_for_signature(&self) -> Result<Vec<u8>, RoutingError> {
         Ok(serialize(&self)?)
+    }
+}
+
+// Result of a message trust check.
+#[derive(Debug)]
+pub enum TrustStatus {
+    // Message is trusted.
+    Trusted,
+    // Message is untrusted because the proof is invalid.
+    ProofInvalid,
+    // Message trust cannot be determined because the proof starts at version that is newer than
+    // our latest one.
+    ProofTooNew,
+}
+
+impl TrustStatus {
+    pub fn is_trusted(&self) -> bool {
+        match self {
+            Self::Trusted => true,
+            _ => false,
+        }
     }
 }
 
--- a/src/chain/shared_state.rs
+++ b/src/chain/shared_state.rs
@@ -502,10 +502,6 @@ impl SectionProofBlock {
         &self.key_info
     }
 
-    pub fn key(&self) -> &bls::PublicKey {
-        self.key_info.key()
-    }
-
     pub fn verify_with_pk(&self, pk: bls::PublicKey) -> bool {
         if let Ok(to_verify) = self.key_info.serialise_for_signature() {
             pk.verify(&self.sig, to_verify)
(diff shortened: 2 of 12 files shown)


src/states/adult/mod.rs
```diff
@@ -527,7 +497,15 @@ impl Base for Adult {
...
-                match self.handle_backlogged_filtered_signed_message(msg) {
+                let msg = match HopMessageWithBytes::new(msg) {
+                    Ok(msg) => msg,
+                    Err(err) => {
+                        error!("{} - Failed to make message {:?}", self, err);
+                        continue;
+                    }
+                };
+
+                match self.handle_filtered_signed_message(msg) {
```
jeanphilippeD: This will forward messages whose trust may be outdated.
I think the idea was that we did not need to do that, and probably don't want to do that.

Is it to handle the JoiningPeer needing to forward the message, it already has a set of elders so it could forward message there like Adult do, instead of on transition, that would probably be better.

But probably ideally we would be to ensure that only Adults are transistion to Elders. i.e something that would stop a node being promoted if we have not detected that it transitioned to Adult already?

src/chain/chain.rs
```diff
@@ -870,17 +871,44 @@ impl Chain {
...
+
+        if min_proof_version > max_known_version {
+            return Err(ProofError::TooNew);
+        }
+
+        if max_proof_version < min_known_version {
+            return Err(ProofError::TooOld);
+        }
+
+        Err(ProofError::Invalid)
```
jeanphilippeD: I am not sure the logic is quite right in case of split, I'm not 100% sure, but split case seem a bit difficult to work out.
Still thinking about it.

src/states/elder/mod.rs
```diff
@@ -717,16 +717,22 @@ impl Elder {
         let dst = msg.message_dst();
         let signed_msg = msg.signed_routing_message();
         if self.in_location(dst) {
-            self.check_signed_message_trust(signed_msg)?;
+            let status = self.check_signed_message_trust(signed_msg)?;
             self.check_signed_message_integrity(signed_msg)?;
```
maqi: what if the status return in above step is `Ok(TrustStatus::Unknown)`, but here failing the integrity check?

src/chain/chain.rs
```diff
@@ -870,17 +871,44 @@ impl Chain {
...
+            .minmax()
+            .into_option()
+            .ok_or(ProofError::Invalid)?;
+        let (min_proof_version, max_proof_version) = proof_chain
             .all_key_infos()
-            .any(|key_info| filtered_keys.contains(key_info))
+            .map(|proof_key_info| proof_key_info.version())
+            .minmax()
+            .into_option()
+            .expect("empty proof");
```
jeanphilippeD: Would it make sense to avoid the unecessary iterator and expect and use something like:
```
let (min_proof_version, max_proof_version) = (proof_chain.genesis_public_key_info().version(), proof_chain.last_public_key_info().version()); // the genesis function is missing but probably should be added.
```

src/chain/chain.rs
```diff
@@ -870,17 +871,44 @@ impl Chain {
...
+
+        if min_proof_version > max_known_version {
+            return Err(ProofError::TooNew);
+        }
+
+        if max_proof_version < min_known_version {
+            return Err(ProofError::TooOld);
```
maqi: Looks like this `TooOld` will be handled as the same to `Invalid` ?
Then why need to differentiate it?

src/states/adult/mod.rs
madadam: > Is it to handle the JoiningPeer needing to forward the message, it already has a set of elders so it could forward message there like Adult do, instead of on transition, that would probably be better.

This sounds like decent idea that can make things simpler. I'm going to give it a try.

> But probably ideally we would be to ensure that only Adults are transistion to Elders. i.e something that would stop a node being promoted if we have not detected that it transitioned to Adult already?

I'm not sure I understand this. We do transition to `Elders` only from `Adults`. Or am I missing something?

jeanphilippeD: > I'm not sure I understand this. We do transition to Elders only from Adults. Or am I missing something?

jeanphilippeD: > I'm not sure I understand this. We do transition to Elders only from Adults. Or am I missing something?

Yes, the nodes go through Joining -> Adult -> Elder, however, a section may promote to Elder a node that still think of itself as Joining. Is that not the source of the bug?

src/chain/chain.rs
```diff
@@ -1800,6 +1828,16 @@ impl EldersChangeBuilder {
     }
 }
 
+#[derive(Error, Debug)]
+pub enum ProofError {
```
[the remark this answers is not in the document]
madadam: I was thinking about it, but it seems to me that when you are about to create an enum where one of the variant is basically `Ok`, then it might be more idiomatic to use `Result` instead. But not 100% convinced about it.

> It seem we are trying to reduce the number of error,

I'm not convinced that reducing the number of error overall is a meaningful goal. I think having these internal small, locally-scoped errors can make the code actually simpler to follow.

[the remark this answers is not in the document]
jeanphilippeD: > I think having these internal small, locally-scoped errors can make the code actually simpler to follow.

Could be, not sure, I'm not 100% convinced by that example that it makes code simpler to follow.

src/states/elder/mod.rs
madadam: Then we would not handle the message which is the correct thing to do.

src/chain/chain.rs
jeanphilippeD: Could strengthen that expectation I guess. It seem to be a sensible expectation.

src/chain/chain.rs
madadam: Good point. It might possibly be useful to differentiate these when we start to handle malice, but currently it's probably unnecessary. Will remove it.

src/chain/chain.rs
madadam: I think we can actually do better. We can modify `SectionProofChain` so that expectation is statically guaranteed. Probably not in this PR though.

src/states/adult/mod.rs
madadam: Ah yes, got you now. Yes, it is the source of the bug, but I don't think it's buggy by itself because it's just the node being a bit laggy.

jeanphilippeD: Indeed, what could be considered buggy is that we decide a node is an Elder before checking they can do it (i.e are already Adults, primed and ready to transition when notified).

src/chain/chain.rs
```diff
@@ -870,17 +871,44 @@ impl Chain {
...
+        let (min_proof_version, max_proof_version) = proof_chain
             .all_key_infos()
-            .any(|key_info| filtered_keys.contains(key_info))
+            .map(|proof_key_info| proof_key_info.version())
+            .minmax()
+            .into_option()
+            .expect("empty proof");
+
+        if min_proof_version > max_known_version {
+            return Err(ProofError::TooNew);
```
jeanphilippeD: Could it make sense to be a bit more explicit in the name:
TrustGenesisToNew?

And maybe:

KeyAlreadyPurged?

src/chain/chain.rs
jeanphilippeD: The logic seem sound, but it is not immediatly clear. because we only have keys compatible with the last prefix in `known_key_infos` we cannot have things from the sibling.

src/states/elder/mod.rs
```diff
@@ -717,16 +717,22 @@ impl Elder {
...
-            self.update_our_knowledge(signed_msg);
 
             if dst.is_multiple() {
```
jeanphilippeD: I've been thinking fora while that this should be done before checking the trust or integrity, would it make sense to swap it now, to avoid having the result then check status later?

src/states/elder/mod.rs
```diff
@@ -1323,6 +1329,33 @@ impl Elder {
...
     }
+
+    // Checks whether the message is trusted.
+    fn check_signed_message_trust(
+        &self,
+        msg: &SignedRoutingMessage,
+    ) -> Result<TrustStatus, RoutingError> {
+        match msg.check_trust(&self.chain) {
+            Ok(()) => Ok(TrustStatus::Known),
+            Err(ProofError::TooNew) if msg.routing_message().dst.is_multiple() => {
```
jeanphilippeD: Why shoud it be only if `dst.is_multiple()`?

src/states/elder/mod.rs
```diff
@@ -1323,6 +1329,33 @@ impl Elder {
...
+                // Proof is too new which can only happen if we've been already demoted but are
+                // lagging behind (or the sender is faulty/malicious). We can't handle the message
+                // ourselves but the other elders likely can.
+                Ok(TrustStatus::Unknown)
```
jeanphilippeD: Why not return the same error `Err(RoutingError::UntrustedMessage)`?
The only difference we care about is probably the log or panic?

src/states/adult/mod.rs
madadam: But that would involve the adult to check some kind of acknowledgement to the elders before they proceed with the promotion. Do we really want that?

jeanphilippeD: Yes, indeed, we already have Adult sending Knowledge information/Poke so we could use that.
Whether we want to or not is open question, simplify some things complicate others.

jeanphilippeD: My thought though, is that it probably is something we want because it ensures we know what kind of node receive a message, thus avoiding having to push more and more things into the JoiningPeer type.

madadam: I think this is a potential can of worms. We can use the `members_knowledge` container to know which nodes have caught up and only promote those that have. But that container is not shared state which means we can't base voting on it. So we would have to make it shared state with all the complications that it entails.

jeanphilippeD: Yes, we could not use the `members_knowledge` container, we could vote for node as Adults when we get their message though, and update the MemberInfo for it when it reach consensus (i.e like we do for Online).

But yes it could have further problem, still incline to think it will be a better solution in the long run, but does not have to be in that PR if we think the alternative is simpler to live with for now.

madadam: OK. If we decide to do this, then I'd strongly prefer to do it in a separate PR. Anyway, back to the point:
> Is it to handle the JoiningPeer needing to forward the message, it already has a set of elders so it could forward message there like Adult do, instead of on transition, that would probably be better.

I tried this but it doesn't work. The problem is that when we are `JoiningPeer`, we don't yet have enough information to decide whether to handle a message or relay it. We can only do that when we become `Adult`. So the only sensible thing to do as `JoiningPeer` is to backlog it.At least until we come up with different approach than backlogging everything.

madadam: > This will forward messages whose trust may be outdated.

So to handle this correctly, I think we need different logic for handling the `JoiningPeer -> Adult` transition from the `Elder` -> `Adult` one.

jeanphilippeD: > OK. If we decide to do this, then I'd strongly prefer to do it in a separate PR. Anyway, back to the point:

It could be a separate PR coming first if we cannot address the issues with having JoiningNode getting messages too early (but lets see what we can do without needing it).

> The problem is that when we are JoiningPeer, we don't yet have enough information to decide whether to handle a message or relay it.

What do you mean, I thought we would be relaying it always since we are a JoiningPeer, and also push it in our backlog for processing?

madadam: Ah, so backlog AND relay. OK, let's try that.

jeanphilippeD: Yes, this is what Adults do too, i.e they backlog for processing (but not re-relay), and relay message immediately. It also avoid that message being delayed.

madadam: OK, that makes sense. However is still doesn't work. At least now if `Adult::handle_backlogged_filtered_signed_message` is implemented the way it currently is. Problem is we still need to backlog a message that is meant for us as elder, otherwise when we get promoted the message gets lost. Even if someone would then send the message to us again, it would get filtered out because we've already seen it (This might point to a deeper problem in that if we don't handle the message ourselves, we should probably not incoming-filter it). 

Also:
> This will forward messages whose trust may be outdated.

I don't think this is a problem, because we clear the backlog on genesis update. 

So I'm now thinking reverting to the original approach - that is - call `handle_filtered_signed_message` in both cases (new and backlogged). Note this is different from what we do as `Elder` but it makes sense because as `Elder` we always process the message, but as `Adult` we don't.

madadam: Also, I don't think it makes much sense to relay messages as `JoiningPeer` because we don't expect to get stuck in that state for long, unlike `Adult`.

jeanphilippeD: > OK, that makes sense. However is still doesn't work. At least now if Adult::handle_backlogged_filtered_signed_message is implemented the way it currently is.

Is it that we drop the message when we become Elder in `Adult::handle_backlogged_filtered_signed_message`, are we not at the `in_location`? why?

jeanphilippeD: > Also, I don't think it makes much sense to relay messages as JoiningPeer because we don't expect to get stuck in that state for long, unlike Adult.

I'm confused, I thought that was the bug, that JoiningPeer received message that Elder should have received, is it not?

src/states/elder/mod.rs
```diff
@@ -714,24 +714,35 @@ impl Elder {
...
             }
-            // if addressed to us, then we just queue it and return
-            self.routing_msg_queue
-                .push_back(msg.into_signed_routing_message());
-        } else if let Err(error) = self.send_signed_message(&msg) {
-            debug!("{} Failed to send {:?}: {:?}", self, signed_msg, error);
+        }
+
+        // Relay closer to the destination or broadcast to the rest of our section.
+        if let Err(error) = self.send_signed_message(&msg) {
```
jeanphilippeD: Wouldn't moving that before `if self.in_location(msg.message_dst())` simplify the code?

src/states/elder/mod.rs
```diff
@@ -714,24 +714,35 @@ impl Elder {
...
-                if let Err(error) = self.send_signed_message(&msg) {
-                    debug!("{} Failed to send {:?}: {:?}", self, signed_msg, error);
+            if handle {
+                self.update_our_knowledge(signed_msg);
+                self.routing_msg_queue.push_back(signed_msg.clone());
+            }
+
+            if !msg.message_dst().is_multiple() {
+                if !handle {
+                    // Message is only for us but we can't handle it yet. Backlog it.
```
jeanphilippeD: Why should it be if it is only for us?

madadam: This is to cover the case where we receive a message with dst = Node whose proof is too new. But I guess this cannot really happen, because we use the `members_knowledge` thing for messages sent to nodes. So probably best to treat this case as normal untrusted message and fail. I'll change it.

src/states/elder/mod.rs
madadam: No, because we don't always want to do this. For example if the message is untrusted or fails the integrity check.

jeanphilippeD: But why shouldn't we? it seem that essentially when receiving a message we want to first send it to everyone that needs receiving it regardless of integrity check/trust since that is what we do for all hops.

That is why I'm thinking having this arbitrary check before resending is probably not necessary, i.e we should not try to avoid resending to all the needed destination regardless of what the processing we do tells us (Also we should probably resend before doing any processing to avoid unnecessary latency).

src/chain/chain.rs
jeanphilippeD: Would it make sense to still use that expectation here though?

src/states/adult/mod.rs
madadam: Let me summarize this:

The problem we have is that a `JoiningPeer` receives a message meant for `Elder` because it's lagging. The simplest solution to deal with this and the one currently implemented in this PR is this:
1. Backlog the message as `JoiningPeer`
2. When transitioning into `Adult`, pull the messages from the backlog, process those we can and re-backlog the others
3. When transitioning into `Elder`, pull the messages from the backlog and process them all

The main difference from the current fleming branch is the re-backlogging of message on transition to `Adult` which we don't do in fleming. This is because in fleming we have two pathways for handling messages: one for received messages and one for messages from backlog. They differ mostly in what we do with unhandled messages - in the first one we backlog them, in the second one we ignore them. 

Notes about some of your suggestions:
- Relaying messages also as `JoiningPeer` - we can do this but it's not going to buy us much at the cost of complicating the `JoiningPeer` state. This is because we don't stay in that state for long and we do relay the messages when transitioning to `Adult`. Also this wouldn't be enough to fix the primary failure anyway, so I decided to not implement it in this PR.
- Not re-backlogging messages when transitioning to `Adult` - already explained that this is the source of the failure. The concern about message becoming stale this way is unfounded, because we purge the backlog on every `GenesisUpdate`.

src/chain/shared_state.rs
```diff
@@ -534,7 +542,9 @@ impl SectionProofChain {
     }
 
     pub fn push(&mut self, block: SectionProofBlock) {
-        self.blocks.push(block);
+        if self.validate_next_block(&block) {
+            self.blocks.push(block);
+        }
     }
 
     pub fn validate(&self) -> bool {
```
jeanphilippeD: Should we call `validate_next_block` from this function as well/instead?

src/states/elder/mod.rs
```diff
@@ -714,24 +714,30 @@ impl Elder {
...
-            self.update_our_knowledge(signed_msg);
-
-            if dst.is_multiple() {
-                // Broadcast to the rest of the section.
-                if let Err(error) = self.send_signed_message(&msg) {
-                    debug!("{} Failed to send {:?}: {:?}", self, signed_msg, error);
-                }
+        if msg.message_dst().is_multiple() {
+            // Relay closer to the destination or broadcast to the rest of our section.
+            if let Err(error) = self.send_signed_message(&msg) {
```
jeanphilippeD: does noy seem to handle when we are a hop toward the single node destination.

src/states/elder/mod.rs
```diff
@@ -714,24 +714,30 @@ impl Elder {
...
+                    self,
+                    msg.signed_routing_message(),
+                    error
+                );
+            }
+        }
+
+        if self.in_location(msg.message_dst()) {
+            let signed_msg = msg.signed_routing_message();
+            let handle = self.verify_signed_message(signed_msg).map_err(|error| {
```
jeanphilippeD: do we actually need the `handle`? i.e can we simplify this function and `verify_signed_message` to return Error or nothing? would that make a change in execution trace?

src/states/elder/mod.rs
madadam: Yes, I just realized this by seeing a whole bunch of failing tests :grimacing:

madadam:
> Looks reasonable, I guess i did not expect the change I asked in particular to return key to require so much adaptation, but it seem reasonable. Left a couple of comments.

If you take it only up to [this commit](https://github.com/maidsafe/routing/pull/2018/commits/ca80c01e5062c9b5067a876129ad8776146bb2a9) then I think it implements only what you wanted. But I think the subsequent commits make thinks cleaner.

src/states/elder/mod.rs
madadam: Hm, with the relaying now happening first I think so. Let me give it a try.

madadam: Or maybe not. We don't want to `log_or_panic` if the proof is too new, but do want to in the other cases.

jeanphilippeD: Yes, but that happens from within `verify_signed_message` and does not need to leak out.

madadam: No, it doesn't anymore. I took that out to be able to use the function in both `handle_filtered_signed_message` and `handle_backlogged_filtered_signed_message`.

src/chain/shared_state.rs
```diff
@@ -627,44 +626,58 @@ impl SectionProofChain {
...
-                "Invalid next block version - not successor (last: {}, next: {})",
-                last.version(),
-                block.version()
-            );
-            return false;
+    fn validate_next_block(
+        &self,
+        last: &SectionKeyInfo,
+        next: &SectionProofBlock,
+    ) -> Result<(), NextBlockError> {
```
jeanphilippeD: I'm not sure why you moved the log or panic outside of this function, it does not seem clearer, and we really should really log or panic for either of the callers.

madadam: > should really log or panic for either of the callers

I disagree. In `push`, it's we who are calling it so we are fully in control of what we pass in so panicking is appropriate (passing an invalid block is a contract violation), but in `verify` we are calling it on a value that we received from someone else which can contain anything so we need to handle errors gracefully. 

You can also argue that this panics only in tests where we control everything but there seems to be push against `log_or_panic` so I wanted to make the code a bit future-proof.

src/states/elder/mod.rs
```diff
@@ -714,7 +714,20 @@ impl Elder {
...
+                    msg.signed_routing_message(),
+                    error
+                );
+            }
+        }
+
+        if in_location {
             let signed_msg = msg.signed_routing_message();
             let handle = self.verify_signed_message(signed_msg).map_err(|error| {
                 self.log_verify_failure(signed_msg, &error);
```
jeanphilippeD: Why not put the log_verify_failure from within so we can simplify the caller?

jeanphilippeD: Ha, I see what you mean by https://github.com/maidsafe/routing/pull/2018/files/453a07da18a1bf103260ccd0bea47a90d123ce21..dd0d5de157935bbc246104c7616ef2a4329559f5#r368989410

src/chain/shared_state.rs
```diff
@@ -627,44 +626,58 @@ impl SectionProofChain {
...
-            && block.prefix().bit_count() != last.prefix().bit_count() + 1
-        {
-            log_or_panic!(
-                LogLevel::Error,
-                "Invalid next block prefix - differs by more than one bit (last: {:?}, next: {:?})",
-                last.prefix(),
-                block.prefix()
-            );
-            return false;
+        if !next.verify_with_pk(*last.key()) {
```
jeanphilippeD: Should we really do that on push?

src/chain/shared_state.rs
```diff
@@ -627,44 +626,58 @@ impl SectionProofChain {
...
+            return Err(NextBlockError::Signature);
         }
 
-        true
+        Ok(())
     }
 }
 
+#[derive(Error, Debug)]
+enum NextBlockError {
```
jeanphilippeD: Seem way too complicated for what we need, after all we essentially don't expect to see that message.

```
fn validate_next_block(...) -> bool;

     pub fn push(&mut self, block: SectionProofBlock) {
        if self.validate_next_block(self.last_public_key_info(), &block) {
            self.blocks.push(block);
        } else {
            log_or_panic!(LogLevel::Error, "Invalid next block {:?} -> {:?}", self.last_public_key_info(), block),
        }
    }

File: src/chain/shared_state.rs (hunk 16 above)
Comment (jeanphilippeD) [reply src/chain/shared_state.rs]: Not 100% sure, so I'll follow what you say.

File: src/chain/shared_state.rs (hunk 18 above)
Comment (madadam) [reply src/chain/shared_state.rs]: I think so, otherwise you could push a block with invalid signature.

File: src/chain/shared_state.rs (hunk 16 above)
Comment (madadam) [reply src/chain/shared_state.rs]: I moved the `log_or_panic` outside of the function so that we can use it for both use cases where the panic is appropriate only in one of them.

File: src/chain/shared_state.rs (hunk 19 above)
Comment (madadam) [reply src/chain/shared_state.rs]: Fair enough. Will simplify it.

File: src/chain/shared_state.rs (hunk 18 above)
Comment (jeanphilippeD) [reply src/chain/shared_state.rs]: Ok, not sure, but I can see how that can be justified. (should not happen often anyway.)

File: src/states/elder/mod.rs [hunk 20]
```diff
@@ -714,24 +714,31 @@ impl Elder {
...
+            let signed_msg = msg.signed_routing_message();
+            let handle = self.verify_signed_message(signed_msg).map_err(|error| {
+                self.log_verify_failure(signed_msg, &error);
+                error
+            })?;
+
+            if handle {
+                self.update_our_knowledge(signed_msg);
+                self.routing_msg_queue
+                    .push_back(msg.into_signed_routing_message());
```
maqi: what's the difference between `msg.signed_routing_message` (https://github.com/maidsafe/routing/pull/2018/files#diff-bb7fddc1a592903612286af78e19ad6bR731) and `msg.into_signed_routing_message` ?

src/states/joining_peer.rs
```diff
@@ -310,8 +310,15 @@ impl Base for JoiningPeer {
...
+
+        trace!(
+            "{} - Handle signed message: {:?}",
+            self,
+            msg.routing_message()
+        );
+
         if self.in_location(&msg.routing_message().dst) {
-            self.check_signed_message_integrity(&msg)?;
+            // TODO: verify the message
```
maqi: still need to left this TODO over?

jeanphilippeD:
@maqi @madadam would it make sense to fix these extra minor thing as a follow up PR so we can merge both PRs?

src/states/joining_peer.rs
madadam: We currently can't properly verify a message when we are `JoiningPeer`. This needs to be addressed separately (it's one of the remaining points in https://github.com/maidsafe/routing/issues/1835).

src/states/adult/mod.rs
```diff
@@ -426,34 +425,16 @@ impl Adult {
...
-            }
-        }
+    fn verify_signed_message(&self, msg: &SignedRoutingMessage) -> Result<(), RoutingError> {
+        let result = match msg.verify(self.chain.get_their_keys_info()) {
+            Ok(VerifyStatus::Full) => Ok(()),
+            Ok(VerifyStatus::ProofTooNew) => Err(RoutingError::UntrustedMessage),
+            Err(error) => Err(error),
+        };
+        result.map_err(|error| {
+            self.log_verify_failure(msg, &error);
```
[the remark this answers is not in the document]
madadam: No, because I want to log for both `Err` and `Ok(ProofTooNew)`.

src/messages/mod.rs
```diff
@@ -572,14 +578,17 @@ mod tests {
...
 
         let signed_msg_1 = unwrap!(SignedRoutingMessage::new(msg, &sk_share_1, pk_set, proof));
         signed_msg_0.add_signature_shares(signed_msg_1);
         assert!(signed_msg_0.check_fully_signed());
 
         signed_msg_0.combine_signatures();
-        unwrap!(signed_msg_0.check_integrity());
+        assert_eq!(
+            unwrap!(signed_msg_0.verify(&their_key_infos)),
+            VerifyStatus::Full
```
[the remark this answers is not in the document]
madadam: There isn't currently, but we should add it. Maybe in a followup PR?

src/messages/mod.rs
```diff
@@ -195,8 +209,18 @@ pub struct SingleSrcSecurityMetadata {
...
-    pub fn verify_sig(&self, bytes: &[u8]) -> bool {
-        self.public_id.verify(bytes, &self.signature)
+    pub fn verify(&self, content: &RoutingMessage) -> Result<VerifyStatus, RoutingError> {
+        if content.src.single_signing_name() != Some(self.public_id.name()) {
+            // Signature is not from the source node.
+            return Err(RoutingError::InvalidMessage);
+        }
+
+        let signed_bytes = serialize(content)?;
+        if !self.public_id.verify(&signed_bytes, &self.signature) {
```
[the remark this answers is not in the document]
madadam: I think it's a difference between the APIs of the `dalek` and `threshold-crypto` crates. Not much we can do about it.

src/states/elder/mod.rs
madadam: One returns reference the other value.

madadam: (one borrows, the other moves)

src/messages/mod.rs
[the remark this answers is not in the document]
maqi: Ah, ok then.

src/messages/mod.rs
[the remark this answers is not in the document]
maqi: Yeah, I am OK with have a follow up PR to contain a test for it.

src/states/adult/mod.rs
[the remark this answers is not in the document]
maqi: Ah, right, ok then.

PR #2019, opened 2020-01-20
clean up to support partial deserialization

jeanphilippeD:


src/messages/mod.rs
```diff
@@ -68,6 +67,8 @@ pub struct HopMessageWithBytes {
     content: SignedRoutingMessage,
     /// Serialized Message as received or sent to quic_p2p.
     full_message_bytes: Bytes,
+    /// Crypto hahs of the full message.
+    full_message_crypto_hash: Digest,
```
madadam: Minor, but do we need to include `crypto` in the name? Can we just assume that "hash" means "crypto hash:?

src/messages/mod.rs
```diff
@@ -68,6 +67,8 @@ pub struct HopMessageWithBytes {
     content: SignedRoutingMessage,
     /// Serialized Message as received or sent to quic_p2p.
     full_message_bytes: Bytes,
+    /// Crypto hahs of the full message.
```
madadam: Typo: "Crypto hash..."

src/messages/mod.rs
```diff
@@ -31,6 +31,8 @@ use std::{
     mem,
 };
 
+type Digest = [u8; 32];
```
madadam: This is already defined in the `crypto` module. It's called `Digest256` though.

src/messages/mod.rs
jeanphilippeD: not really, we have Hash derive a lot and the .hash function is not crypto. I'd feel a bit more comfortable keeping it.

jeanphilippeD:
Fixed review comments

PR #2020, opened 2020-01-20
replace maidsafe_utilities::serialisation with bincode

jeanphilippeD:


src/error.rs
```diff
@@ -49,3 +49,11 @@ pub enum RoutingError {
     #[error(display = "Error while trying to receive a message from a mpsc channel.")]
     MpscRecvError(mpsc::RecvError),
 }
+
+// TODO dirvine, we need to complete the error story in our code. Here I am unboxing and that will
```
dirvine: This part is one I would like @jeanphilippeD  and @madadam  to look at a bit closer if possible. It feels a bit dirty, but let's see.

src/parsec.rs
```diff
@@ -120,11 +119,12 @@ impl ParsecMap {
...
         // Increase the size before fetching the parsec to satisfy the borrow checker
-        self.count_size(
-            serialisation::serialised_size(&request),
-            msg_version,
-            log_ident,
-        );
+        let ser_size = if let Ok(size) = bincode::serialized_size(&request) {
+            size
+        } else {
+            return (None, false);
```
maqi: Just to confirm that this bool flag is to ensure no parsec poll to be undertaken in case of cannot fetch a request's serialized_size .

same to the below response case.

src/parsec.rs
```diff
@@ -153,11 +153,12 @@ impl ParsecMap {
         log_ident: &LogIdent,
     ) -> bool {
         // Increase the size before fetching the parsec to satisfy the borrow checker
-        self.count_size(
-            serialisation::serialised_size(&response),
-            msg_version,
-            log_ident,
-        );
+        let ser_size = if let Ok(size) = bincode::serialized_size::<Response>(&response) {
```
maqi: why need `::<Response>` here, but not requiring `::<Request>` above ?

src/error.rs
jeanphilippeD: Ok replacing the Error with `Box<ErrorKind>` is sufficient to remove that code.

src/parsec.rs
jeanphilippeD: Yes, essentially this is a hint that doing parsec_poll will yield no new result since nothing was added to the newest parsec.

jeanphilippeD:
Force pushed fixing review comments.

PR #2023, opened 2020-01-22
WIP Prototype: avoid serializing for signature verification.

jeanphilippeD:
Note:
 - Still use serialize_size which is not ideal, we should be able
to get that information when we are deserializing, but not clear how
to do it, maybe with custom reader?
 - Alternatively if we wanted to avoid deserializing in most cases, we
   could possibly have the serialized size as part of the message. maybe

Test:
clippy + test

### DIFF
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -154,6 +154,15 @@ impl HopMessageWithBytes {
         &self.partial_content.dst
     }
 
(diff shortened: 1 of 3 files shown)


dirvine:
Thanks for this, as things are moving on at some pace we will need to rework much of this one, but we will take all the still relevant parts. :+1:

PR #2024, opened 2020-01-22
Verify NodeApproval message as JoiningPeer

### LINKED ISSUE #2022: Make sure NodeApproval message can be trusted
That message is received by a node in `JoiningPeer` state and so its trust currently cannot be fully verified. Find a way to fix that.

madadam:
- If we are a newly joining node, we don't have any trusted keys so we cannot fully verify the message. Instead we only check that the proof and the message signature are valid.
- If we are relocating node, we carry the trusted keys from our previous section into the `JoiningPeer` state and use that to fully verify the message

Closes #2022 .

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -397,6 +397,12 @@ impl Chain {
         let relocating_state = self.state.create_relocating_state();
         let mut details_to_add = Vec::new();
 
+        struct PartialRelocateDetails {
+            pub_id: PublicId,
+            destination: XorName,
+            age: u8,
+        }
+
         for (name, member_info) in self.state.our_joined_members_mut() {
             if member_info.p2p_node.public_id() == trigger_node {
                 continue;
@@ -418,7 +424,7 @@ impl Chain {
             }
 
             member_info.state = relocating_state;
-            details_to_add.push(RelocateDetails {
+            details_to_add.push(PartialRelocateDetails {
                 pub_id: *member_info.p2p_node.public_id(),
                 destination,
                 age: member_info.age() + 1,
@@ -429,7 +435,17 @@ impl Chain {
 
         for details in details_to_add {
             trace!("{} - Change state to Relocating {}", self, details.pub_id);
-            self.state.relocate_queue.push_front(details)
+
+            let destination_key_info = self
+                .latest_compatible_their_key_info(&details.destination)
+                .clone();
+            let details = RelocateDetails {
+                pub_id: details.pub_id,
+                destination: details.destination,
+                destination_key_info,
+                age: details.age,
+            };
+            self.state.relocate_queue.push_front(details);
         }
     }
 
@@ -864,8 +880,18 @@ impl Chain {
     }
 
     /// Return the keys we know
-    pub fn get_their_keys_info(&self) -> impl Iterator<Item = (&Prefix<XorName>, &SectionKeyInfo)> {
-        self.state.get_their_keys_info()
+    pub fn get_their_key_infos(&self) -> impl Iterator<Item = (&Prefix<XorName>, &SectionKeyInfo)> {
+        self.state.get_their_key_infos()
+    }
+
+    /// Returns the latest key info whose prefix is compatible with the given name.
+    pub fn latest_compatible_their_key_info(&self, name: &XorName) -> &SectionKeyInfo {
+        self.state
+            .get_their_key_infos()
+            .filter(|(prefix, _)| prefix.matches(name))
+            .map(|(_, info)| info)
+            .max_by_key(|info| info.version())
+            .unwrap_or_else(|| self.state.our_history.first_key_info())
     }
 
     /// Returns `true` if the `EldersInfo` isn't known to us yet.
@@ -2013,6 +2039,9 @@ mod tests {
         let relocate_details = RelocateDetails {
             pub_id: *chain.our_id(),
             destination: *chain.our_id().name(),
+            destination_key_info: chain
+                .latest_compatible_their_key_info(chain.our_id().name())
+                .clone(),
             age: 0,
         };
         let elder_size = chain.our_info().len();
--- a/src/chain/network_event.rs
+++ b/src/chain/network_event.rs
@@ -80,8 +80,12 @@ impl EventSigPayload {
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub struct OnlinePayload {
+    // Identifier of the joining node.
     pub p2p_node: P2pNode,
+    // The age the node should have after joining.
     pub age: u8,
+    // The version of the destination section that the joining node knows, if any.
+    pub their_knowledge: Option<u64>,
 }
 
 /// Routing Network events
--- a/src/chain/shared_state.rs
+++ b/src/chain/shared_state.rs
@@ -73,7 +73,7 @@ impl SharedState {
     ) -> Self {
         let pk_info = SectionKeyInfo::from_elders_info(&elders_info, bls_keys.public_key());
         let our_history = SectionProofChain::from_genesis(pk_info);
-        let their_key_info = our_history.last_public_key_info();
+        let their_key_info = our_history.last_key_info();
         let their_keys = iter::once((*their_key_info.prefix(), their_key_info.clone())).collect();
 
         let our_members = elders_info
@@ -291,7 +291,7 @@ impl SharedState {
         self.our_history.push(proof_block);
         self.our_infos.push(elders_info);
 
-        let key_info = self.our_history.last_public_key_info().clone();
+        let key_info = self.our_history.last_key_info().clone();
         self.update_their_keys(&key_info);
     }
 
@@ -382,7 +382,7 @@ impl SharedState {
     }
 
     /// Returns the reference to their_keys and any recent keys we still hold.
-    pub fn get_their_keys_info(&self) -> impl Iterator<Item = (&Prefix<XorName>, &SectionKeyInfo)> {
+    pub fn get_their_key_infos(&self) -> impl Iterator<Item = (&Prefix<XorName>, &SectionKeyInfo)> {
         self.their_keys
             .iter()
             .chain(self.their_recent_keys.iter().map(|(p, k)| (p, k)))
@@ -559,7 +559,7 @@ impl SectionProofSlice {
         )
     }
 
-    pub fn last_new_public_key_info(&self) -> Option<&SectionKeyInfo> {
+    pub fn last_new_key_info(&self) -> Option<&SectionKeyInfo> {
         self.blocks.last().map(|block| block.key_info())
     }
 
@@ -647,11 +647,11 @@ impl SectionProofChain {
     }
 
     pub fn push(&mut self, block: SectionProofBlock) {
-        if !validate_next_block(self.last_public_key_info(), &block) {
+        if !validate_next_block(self.last_key_info(), &block) {
             log_or_panic!(
                 LogLevel::Error,
                 "Invalid next block: {:?} -> {:?}",
-                self.last_public_key_info(),
+                self.last_key_info(),
                 block
             );
             return;
@@ -673,7 +673,11 @@ impl SectionProofChain {
         true
     }
 
-    pub fn last_public_key_info(&self) -> &SectionKeyInfo {
+    pub fn first_key_info(&self) -> &SectionKeyInfo {
+        &self.genesis_key_info
+    }
+
+    pub fn last_key_info(&self) -> &SectionKeyInfo {
         self.blocks
             .last()
             .map(|block| block.key_info())
@@ -863,7 +867,7 @@ mod test {
         // Assert
         //
         let actual_keys = state
-            .get_their_keys_info()
+            .get_their_key_infos()
             .map(|(p, info)| {
                 (
                     *p,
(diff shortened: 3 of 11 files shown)


src/chain/shared_state.rs
```diff
@@ -596,22 +600,32 @@ impl SectionProofChain {
...
+        for (prefix, info) in their_key_infos {
+            their_key_infos_count += 1;
+
+            if last_prefix.is_compatible(prefix) {
+                let _ = known_key_infos.insert(info);
+            }
+        }
+
+        if their_key_infos_count == 0 {
+            return TrustStatus::TheirKeyInfosEmpty;
```
jeanphilippeD: That does not seem right, we should know whether we can trust stuff or not and not call this function if we cannot.

src/chain/shared_state.rs
```diff
@@ -694,14 +708,16 @@ impl SectionKeyInfo {
 
 // Result of a message trust check.
 #[derive(Debug)]
-pub enum TrustStatus<'a> {
-    // Message is trusted. Contains the latest section public key.
-    Trusted(&'a bls::PublicKey),
+pub enum TrustStatus {
+    // Message is trusted.
+    Trusted,
```
jeanphilippeD: Why do you no longer return that key, I need it with what I'm working on.

src/messages/mod.rs
```diff
@@ -222,18 +222,25 @@ impl FullSecurityMetadata {
...
+            TrustStatus::TheirKeyInfosEmpty => VerifyStatus::Partial,
             TrustStatus::ProofTooNew => return Ok(VerifyStatus::ProofTooNew),
             TrustStatus::ProofInvalid => return Err(RoutingError::UntrustedMessage),
         };
 
         let signed_bytes = serialize(content)?;
-        if !public_key.verify(&self.signature, &signed_bytes) {
+        if !self
+            .proof
+            .last_public_key_info()
```
jeanphilippeD: Yes, step backward for me as well.

src/chain/network_event.rs
```diff
@@ -80,8 +80,12 @@ impl EventSigPayload {
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub struct OnlinePayload {
+    // Identifier of the joining node.
     pub p2p_node: P2pNode,
+    // The age the node should have after joining.
     pub age: u8,
+    // The knowledge of the joining node about the destination section.
```
maqi: what kind of knowledge, according to a later on usage, it seems the version of the section?
why not just name the field as: dst_sec_version if so?

madadam: I agree that `knowledge` is a bit vague, but it's already established term - we use it in `their_knowledge` in `SharedState` and this field is basically the same thing (the node's knowledge of out section). We can maybe try to come up with better name but we should be consistent.

jeanphilippeD: Maybe could use `their_knowledge` here too?

src/states/joining_peer.rs
```diff
@@ -143,6 +147,11 @@ impl JoiningPeer {
             };
             let join_request = JoinRequest {
                 elders_version,
+                knowledge: self
+                    .section_key_info
+                    .as_ref()
+                    .map(|info| info.version())
+                    .unwrap_or(0),
```
madadam: Question to reviewers: when we are newly joining node, we are basically requesting a proof that starts at the root. However, we still can't properly verify such proof because we have no trusted keys. This means the message will contain unnecessary long proof for no real benefit. Is it worth optimizing this? For example changing the `knowledge` field to `Option<u64>` and setting it to `None` for newly joining nodes? So the target section could then use shorter proof to save same space?

src/chain/network_event.rs
madadam: Yeah, sounds reasonable. Will do.

src/states/joining_peer.rs
jeanphilippeD: Yes, Maybe a new node could have knowledge of valid keys. but right now they do not so probably better to have it optional.

Latter we could add a mean for them to know genesis, or maybe with each update we could add all the trusted key at that point, but for now, we should make it clear that we do not require proof.

src/messages/mod.rs
```diff
@@ -403,13 +421,29 @@ impl SignedRoutingMessage {
...
+                .map(|_| VerifyStatus::Full),
             SecurityMetadata::Full(security_metadata) => {
                 security_metadata.verify(&self.content, their_key_infos)
             }
         }
     }
 
+    /// Verifies that the proof and the signature of this message are valid, but does not check
+    /// the proof against known keys.
+    pub fn verify_partial(&self) -> Result<(), RoutingError> {
```
jeanphilippeD: I'm not sure why we want verify partial. Essentially with my changeset you may not have a public key, so it is not possible.

I'd say, maybe if you want to check the signature you could look if there is a block and check against it, but you would need to make sure they give at least one block for these new node otherwise that would amount to nothing.

But should they really do that? we still don't have any trust root so why trust the key they give us?

src/states/adult/mod.rs
```diff
@@ -154,6 +154,12 @@ impl Adult {
         conn_infos: Vec<ConnectionInfo>,
         details: SignedRelocateDetails,
     ) -> Result<State, RoutingError> {
+        let their_key_infos = self
```
jeanphilippeD: This does not seem right, Adults don't have their_key_info populated, it needs to come from SignedRelocateDetails.
Am I missing something?

jeanphilippeD: I mean it only has our last Elders key infos, not any other section.

src/messages/mod.rs
madadam: I guess checking at least that the proof is valid (each block is signed with the previous block) and the signature is valid is better than nothing? But maybe it isn't, because without trusted keys, a malicious agent can forge any proofs/signatures without much effort. So probably better to not verify the message at all for now?

src/states/adult/mod.rs
madadam: Doesn't it come from `GenesisUpdate`?

jeanphilippeD: I believe the shared state is empty for now, but also it would be out of date.

src/states/adult/mod.rs
```diff
@@ -426,15 +433,12 @@ impl Adult {
...
-            Ok(VerifyStatus::Full) => Ok(()),
-            Ok(VerifyStatus::ProofTooNew) => Err(RoutingError::UntrustedMessage),
-            Err(error) => Err(error),
-        };
-        result.map_err(|error| {
-            self.log_verify_failure(msg, &error);
-            error
-        })
+        msg.verify(self.chain.get_their_key_infos())
+            .and_then(VerifyStatus::require_full)
```
jeanphilippeD: Nice, it is small but I like that very much :-)

src/states/bootstrapping_peer.rs
```diff
@@ -100,6 +104,13 @@ impl BootstrappingPeer {
         relocate_payload: Option<RelocatePayload>,
         _outbox: &mut dyn EventBox,
     ) -> Result<State, RoutingError> {
+        let section_key_info = self
```
jeanphilippeD: As mentioned, we probably should already have the single SectionKeyInfo for our destination knowledge.

src/states/adult/mod.rs
madadam: I don't think it's empty: https://github.com/maidsafe/routing/blob/fleming/src/chain/chain.rs#L614. Also it doesn't matter it's outdated. It doesn't need to contain the latest keys. Any keys whose prefix is compatible with the target prefix will do.

jeanphilippeD:
```
        let trimmed_info = GenesisPfxInfo {
            first_info: self.gen_pfx_info.first_info.clone(),
            first_bls_keys: self.gen_pfx_info.first_bls_keys.clone(),
            first_state_serialized: Default::default(),
            first_ages: self.gen_pfx_info.first_ages.clone(),
            latest_info: self.chain.our_info().clone(),
            parsec_version: self.gen_pfx_info.parsec_version,
        };

        let src = Location::PrefixSection(*trimmed_info.first_info.prefix());
        let content = MessageContent::NodeApproval(trimmed_info);
```

We trimm it out for NodeApproval, I guess we missed doing that for GenesisUpdate, so new node will not have the info, but ones that have seen an elder change will. Not ideal.

jeanphilippeD: Would it make things easier if we did the work of making relocate like GenesisUpdate i.e don't have to sign it before it goes through Parsec? then we can just add whatever we need when we send the message.

jeanphilippeD: Also you are correct it does not matter that it is out of date since its history will always be kept by the destination even if we trim it eventually it should not matter for that case. (But Adults should really not have that knowledge to start with).

src/chain/chain.rs
```diff
@@ -864,8 +864,8 @@ impl Chain {
     }
 
     /// Return the keys we know
```
jeanphilippeD: Would it make things easier for this change if we did the work of making relocate like GenesisUpdate i.e don't have to sign it before it goes through Parsec? then we can just add whatever we need when we send the message.

src/states/adult/mod.rs
madadam: > Would it make things easier if we did the work of making relocate like GenesisUpdate i.e don't have to sign it before it goes through Parsec? then we can just add whatever we need when we send the message.

Yes I think that's the plan. Not sure it makes sense to do it in this PR though.

jeanphilippeD: yes, it does not really make sense in this PR, could be if we wanted, but if it make that PR easier could be done in a PR that we rebase 2024 ontop?

src/states/elder/mod.rs
```diff
@@ -1161,7 +1173,11 @@ impl Elder {
     // Constructs a signed message, finds the nodes responsible for accumulation, and either sends
     // these nodes a signature or tries to accumulate signatures for this message (on success, the
     // accumulator handles or forwards the message).
-    fn send_routing_message(&mut self, routing_msg: RoutingMessage) -> Result<(), RoutingError> {
```
maqi: the `node_knowledge_override` is only populated when handle_candidate_approval.
maybe worth to put a comment here to explain the usage?

madadam:
Addressed the review comments. The relocating node now gets only the root key from the source section in the `RelocateDetails`, carries it over to the `JoiningPeer` state and uses it to verify the `NodeApproval` message. The `NodeApproval` message is now explicitly proven with starting index 0, so the root key can verify it.

src/chain/chain.rs
```diff
@@ -395,6 +395,7 @@ impl Chain {
 
         let our_prefix = *self.state.our_prefix();
         let relocating_state = self.state.create_relocating_state();
+        let root_key_info = self.root_key_info().clone();
```
jeanphilippeD: > Addressed the review comments. The relocating node now gets only the root key from the source section in the RelocateDetails, carries it over to the JoiningPeer state and uses it to verify the NodeApproval message. The NodeApproval message is now explicitly proven with starting index 0, so the root key can verify it.

Ok, that is different. I guess that eliminate the need for communicating what key we trust to the section we join.
Is it really what we want in the end though?

1 - do I understand correctly that existing fleming code was already always giving the full proof to new joining node?
(If so that make sense to exploit it).
2 - do we plan to change this to something more like what you had originally, (i.e giving our_knowledge to the joining section so they can give us a proof, otherwise they give nothing) to consider that issue complete, or are we happy with it and would only address it if we have a performance issue?

src/messages/mod.rs
```diff
@@ -228,12 +228,22 @@ impl FullSecurityMetadata {
             TrustStatus::ProofInvalid => return Err(RoutingError::UntrustedMessage),
         };
 
+        self.verify_signature(content, public_key)?;
+        Ok(VerifyStatus::Full)
+    }
+
+    #[cfg_attr(feature = "mock_base", allow(clippy::trivially_copy_pass_by_ref))]
+    fn verify_signature(
```
jeanphilippeD: I guess we can revert that function since it is not use and does not seem to bring much.

src/messages/mod.rs
```diff
@@ -403,7 +413,9 @@ impl SignedRoutingMessage {
             SecurityMetadata::None | SecurityMetadata::Partial(_) => {
                 Err(RoutingError::FailedSignature)
             }
-            SecurityMetadata::Single(security_metadata) => security_metadata.verify(&self.content),
+            SecurityMetadata::Single(security_metadata) => security_metadata
```
jeanphilippeD: Why that change? how did singe security metadata change?

src/chain/chain.rs
madadam: > 1 - do I understand correctly that existing fleming code was already always giving the full proof to new joining node?

I don't think this was the case. In fleming I think the proof given is the same we would give to any other node that is already a member of our section. 

> 2 - do we plan to change this to something more like what you had originally

I'd say no, unless there are performance (or other) issues. I kind of like the simplicity of this approach.

src/messages/mod.rs
madadam: Remnants from the previous version. Will revert.

madadam:
Rebased and update to use latest known destination section key info instead of the root info.

PR #2026, opened 2020-01-22
Remove tusted key from Hop messages

### LINKED ISSUE #1752: Improve message size/performance (Omit BlsPublicKey) (Optional)
(Init estimate 3-4 days)

In the normal case, we expect the BlsPublicKey used to signed the message to be in our own `their_keys`.
In case of churn we expect either the genesis key in the `proof: SectionProofChain` to be in `their_keys`, or one of its proven descendants.

In all cases, the genesis key itself may be omitted from the `proof: SectionProofChain`, as either it is in `their_keys` or is uneeded. The only necessary piece of information is the Prefix+version that will allow looking up the specific key needed.

As a result, in the normal non churn case, the `proof: SectionProofChain` will only contain the minimal Prefix+version information needed.

jeanphilippeD:
Closes #1752 

We should only validate trust when we have the root of trust:
The root of trust bls::PublicKey is no longer part of the Hop message from a Section.
For most message the security_metadata will now only be a Prefix + version(u64).
To validate the message we must find the key we trust and either use it to validate a new
key to trust, or the message itself.

### DIFF
--- a/src/chain/shared_state.rs
+++ b/src/chain/shared_state.rs
@@ -18,7 +18,7 @@ use bincode::{deserialize, serialize};
 use itertools::Itertools;
 use log::LogLevel;
 use std::{
-    collections::{BTreeMap, BTreeSet, VecDeque},
+    collections::{BTreeMap, VecDeque},
     fmt::{self, Debug, Formatter},
     iter, mem,
 };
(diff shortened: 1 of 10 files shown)


src/chain/shared_state.rs
```diff
@@ -520,15 +520,21 @@ impl SectionProofBlock {
...
-pub struct SectionProofChain {
-    genesis_key_info: SectionKeyInfo,
+pub struct SliceProofChain {
+    /// The version of the section key to use as root of trust.
+    version: u64,
+    /// The prefix of section key to use as root of trust.
+    prefix: Prefix<XorName>,
+    /// chain of trust to the section, if empty use root of trust.
     blocks: Vec<SectionProofBlock>,
 }
```
madadam: I might be misunderstanding something, but doesn't this still contain the public keys in all but the first block? I though the idea was the public keys would be taken only from `their_key_infos` so they don't have to be included in the messages?

src/chain/shared_state.rs
```diff
@@ -520,15 +520,21 @@ impl SectionProofBlock {
...
 
 #[derive(Debug, Eq, PartialEq, Clone, Hash, Serialize, Deserialize)]
-pub struct SectionProofChain {
-    genesis_key_info: SectionKeyInfo,
+pub struct SliceProofChain {
```
madadam: Minor: this is a "slice of SectionProofChain", so I think `SectionProofChainSlice` or just `SectionProofSlice` would be more appropriate name.

src/chain/shared_state.rs
madadam: Ah wait. I think I get it now. We might not have the latest keys, so they still need to be included in the message so we can update `their_key_infos` with it. It's fine then.

jeanphilippeD: Well the new key we don't know have to be included otherwise how do we learn of them.
But mostly what this mean is the root of trust is not included, we have to have it to verify the message. What this mean is for most messages for which blocks is empty, there is no public key included with the message.

jeanphilippeD:
Address review comment.

src/chain/shared_state.rs
```diff
@@ -537,8 +543,111 @@ impl SectionProofChain {
...
+        self.blocks.last().map(|block| block.key_info())
+    }
+
+    fn last_trusted_key_info<'a>(
+        &'a self,
+        last_trusted: &'a SectionKeyInfo,
+    ) -> Option<&'a SectionKeyInfo> {
+        let block_offset = last_trusted.version().saturating_sub(self.version) as usize;
+
+        if block_offset == 0 {
```
maqi: basically, this block of code is saying:
we only consider two cases are valid:
  a, last_trusted equals to self
  b, the block at (block_offset - 1) contains the same key_info as last_trusted
 
right?

src/chain/shared_state.rs
```diff
@@ -520,15 +520,21 @@ impl SectionProofBlock {
 }
 
 #[derive(Debug, Eq, PartialEq, Clone, Hash, Serialize, Deserialize)]
-pub struct SectionProofChain {
-    genesis_key_info: SectionKeyInfo,
+pub struct SectionProofSlice {
+    /// The version of the section key to use as root of trust.
```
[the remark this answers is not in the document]
jeanphilippeD: Could do but I'm not sure that brings much, the other suggestion in particular https://github.com/maidsafe/routing/pull/2026#discussion_r370020346 require to create structs when just trying to access some info, so I'm not clear the trade offs are worth it.

src/chain/shared_state.rs
```diff
@@ -537,8 +543,111 @@ impl SectionProofChain {
         self.blocks.len()
     }
 
+    pub fn last_prefix_version(&self) -> (&Prefix<XorName>, u64) {
```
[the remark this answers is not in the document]
jeanphilippeD: see https://github.com/maidsafe/routing/pull/2026#discussion_r370041465

src/chain/shared_state.rs
```diff
@@ -537,8 +543,111 @@ impl SectionProofChain {
...
+        } else if let Some(block) = self.blocks.get(block_offset - 1) {
+            if block.key_info() != last_trusted {
+                return None;
+            }
+        } else {
+            // Root of trust not found
+            return None;
+        }
+
+        let mut current = last_trusted;
```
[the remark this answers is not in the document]
jeanphilippeD: yes, because block[3] is version 6. (i.e if self.version is 2, then block[0] is version 3, block[1] is version 4, block[2] is version 5, block[3] is version 6.

src/chain/shared_state.rs
jeanphilippeD: Yes.
Really there is only one case:
 -  the block at (block_offset - 1) contains the same key_info as last_trusted.
It is just that in the case of block_offset, we only have prefix and version for `block_offset - 1` and they are self.version, self.preffix.

src/chain/shared_state.rs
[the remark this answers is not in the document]
maqi: well, as to the suggestion at that comment, we could have shared_state & block have an accessor function, say `.trust()` to return this struct/type directly.

All this suggestion is just to make it explicit that now `(prefix, version)` is always being used together to compose a `trust`.

src/chain/shared_state.rs
maqi: ok. thx for the clarification

src/chain/shared_state.rs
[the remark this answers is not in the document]
jeanphilippeD: It really does not have too. This is not what compose trust. trust is the key we have, it is just what we use to find the key we trust. i.e what is trust is using the key we got from their_keys to either validate the SectionProofBlocks or the message itself.

So we would need a better name for this new proposed struct if we thought we wanted it.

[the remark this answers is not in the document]
maqi: Ok, then.

PR #2027, opened 2020-01-27
fix/routing: use same mechanism for Relocate as GenesisUpdate

jeanphilippeD:
With GenesisUpdate, we can now send a SignedRoutingMessage to one of
our members, by directly sending it the direct message MessageSignature.
The node accumulate it at reception and process it when enough signature
is received. This remove the need to sign the message before going
through Parsec.

Test:
soak + clippy

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -21,7 +21,7 @@ use crate::{
     relocation::{self, RelocateDetails},
     utils::LogIdent,
     xor_space::Xorable,
-    ConnectionInfo, Prefix, XorName,
+    Prefix, XorName,
 };
 use bincode::serialize;
 use itertools::Itertools;
@@ -310,13 +310,8 @@ impl Chain {
             AccumulatingEvent::AckMessage(ref ack_payload) => {
                 self.update_their_knowledge(ack_payload.src_prefix, ack_payload.ack_version);
             }
-            AccumulatingEvent::Relocate(ref relocate_details) => {
+            AccumulatingEvent::Relocate(_) => {
                 self.relocation_in_progress = false;
-                let signature = Some(
-                    self.check_and_combine_signatures(relocate_details, proofs)
-                        .ok_or(RoutingError::InvalidRelocation)?,
-                );
-                return Ok(Some(AccumulatedEvent::new(event).with_signature(signature)));
             }
             AccumulatingEvent::Online(_)
             | AccumulatingEvent::Offline(_)
@@ -706,14 +701,6 @@ impl Chain {
             .unwrap_or(false)
     }
 
-    /// Returns the `ConnectioInfo` for a member of our section.
-    pub fn get_member_connection_info(&self, pub_id: &PublicId) -> Option<&ConnectionInfo> {
-        self.state
-            .our_members
-            .get(pub_id.name())
-            .map(|member_info| member_info.p2p_node.connection_info())
-    }
-
     /// Returns a section member `P2pNode`
     pub fn get_member_p2p_node(&self, name: &XorName) -> Option<&P2pNode> {
         self.state
@@ -926,12 +913,19 @@ impl Chain {
         target: &Location,
         node_knowledge_override: Option<u64>,
     ) -> SectionProofSlice {
-        let first_index = match (target, node_knowledge_override) {
+        let first_index = self.knowledge_index(target, node_knowledge_override);
+        self.state.our_history.slice_from(first_index as usize)
+    }
+
+    /// Provide a start index of a SectionProofSlice that proves the given signature to the given
+    /// destination location.
+    /// If `node_knowledge_override` is `Some`, it is used when calculating proof for
+    /// `Location::Node` instead of the stored knowledge. Has no effect for other location types.
+    pub fn knowledge_index(&self, target: &Location, node_knowledge_override: Option<u64>) -> u64 {
+        match (target, node_knowledge_override) {
             (Location::Node(_), Some(knowledge)) => knowledge,
             _ => self.state.proving_index(target),
-        };
-
-        self.state.our_history.slice_from(first_index as usize)
+        }
     }
 
     /// Check which nodes are unresponsive.
@@ -1814,7 +1808,7 @@ impl EldersChangeBuilder {
 #[cfg(test)]
 mod tests {
     use super::{
-        super::{quorum_count, EldersInfo, EventSigPayload, GenesisPfxInfo, MIN_AGE_COUNTER},
+        super::{EldersInfo, GenesisPfxInfo, MIN_AGE_COUNTER},
         *,
     };
     use crate::{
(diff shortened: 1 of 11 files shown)


src/chain/chain.rs
```diff
@@ -926,12 +913,19 @@ impl Chain {
...
         node_knowledge_override: Option<u64>,
     ) -> SectionProofSlice {
-        let first_index = match (target, node_knowledge_override) {
+        let first_index = self.knowledge_index(target, node_knowledge_override);
+        self.state.our_history.slice_from(first_index as usize)
+    }
+
+    /// Provide a start index of a SectionProofSlice that proves the given signature to the given
+    /// destination location.
+    /// If `node_knowledge_override` is `Some`, it is used when calculating proof for
```
maqi: This second paragraph of the comment is bit duplicated with the one above function `prove` ?
maybe one of them can be removed?

src/chain/chain.rs
```diff
@@ -2015,82 +2009,4 @@ mod tests {
             check_infos_for_duplication(&chain);
         }
     }
-
-    #[test]
```
maqi: These removed tests are no longer necessary, right?

src/chain/network_event.rs
```diff
@@ -51,19 +50,6 @@ pub struct EventSigPayload {
...
-        let sig_share = key_share.sign(&serialize(&payload)?[..]);
-        let pub_key_share = key_share.public_key_share();
-
-        Ok(Self {
-            pub_key_share,
-            sig_share,
-        })
-    }
-
     pub fn new_for_section_key_info(
```
maqi: maybe `new_for_section_key_info` now can be renamed as `new` function as it becomes the only constructor?

src/messages/mod.rs
```diff
@@ -580,6 +581,8 @@ pub enum MessageContent {
     },
     /// Update sent to Adults and Infants by Elders
     GenesisUpdate(GenesisPfxInfo),
+    /// Send from a section to the node being relocated.
```
maqi: better to be clustered with the above `NodeApproval` ?
also, similiar to `NodeApproval`, better state it is sent from `Section X` to the node being relocated ?

src/states/adult/mod.rs
```diff
@@ -258,10 +261,10 @@ impl Adult {
 
         self.network_service_mut().remove_and_disconnect_all();
 
-        Transition::Relocate {
-            details,
+        Ok(Transition::Relocate {
+            details: signed_msg.clone(),
```
maqi: is this clone really required?

src/states/joining_peer.rs
```diff
@@ -191,17 +191,20 @@ impl JoiningPeer {
...
-            msg.verify(as_iter(key_info))
-                .and_then(VerifyStatus::require_full)
-                .map_err(|error| {
-                    self.log_verify_failure(msg, &error, as_iter(key_info));
-                    error
-                })
-        } else {
-            // We don't have the root key info so we can't verify the message.
-            Ok(())
+            if let Some(details) = payload.relocate_details() {
```
maqi: in case the payload is invalid, i.e. `relocate_details` returns `None`, this function returns with `Ok(())`, is it ok?

src/chain/chain.rs
jeanphilippeD: Yes, they were tested the relocate being signed by Parsec, but we no longer do that (since it is signed when sending the message).

src/chain/network_event.rs
jeanphilippeD: Could be but the name of the struct is more generally useful, so it seem sensible to have a named constructor. Not sure.

src/messages/mod.rs
jeanphilippeD: Will fix, the comment, but I'm not clear why the clustering would make more sense. In a way this message behave like GenesisUpdate, (in that it is accumulated at the single node destination), so it seem better to cluster it with it.

src/states/adult/mod.rs
jeanphilippeD: Not strictly, though it would make the code more complicated if it was not and it is a rare message.

src/states/joining_peer.rs
jeanphilippeD: Yes, essentially we have a bug if that ever return None, that mean we did not actually have a RelocateDetails in the SignedRoutingMessage, which mean we have nothing to validate trust against, and so we are like a new node here. (i.e the other JoinType).

src/chain/chain.rs
jeanphilippeD: I don't really see why, essentially the effect of node_knowledge_override is the same for both public functions.
Unless we want something like:

```rust
\\\ Provide a SectionProofSlice that proves the given signature for the given parameter forwarded to `knowledge_index`
pub fn prove(
```

Is that what we want? (@madadam could you chip in as well).

maqi: It's just `node_knowledge_override` is a passing down parameter for the `prove` function, so I think we can skip one of the comment.

src/chain/network_event.rs
maqi: right. Ok then.
up to you to decide change or not, won't make this as a blocking comment.  :)

src/messages/mod.rs
maqi: I just thought they are both being `bootstrap process` related, and comment will be similar, so suggested to cluster them.
up to you to decide.

jeanphilippeD: They are very different though, one is when you join a section, the other when you leave it.

src/states/joining_peer.rs
maqi: Yeah, maybe in that situation, we shall join as a new node? or terminate?

src/messages/mod.rs
maqi: but within a `relocate process`, it is `you leave a section, then to join another`

src/states/joining_peer.rs
jeanphilippeD: I could put a log or panic here, but that bring little more, by that point we already sent the Join requests, If the message was invalid, Elder will have decided `"{} - Ignoring relocation JoinRequest from {} - invalid payload."` and not replied.

src/messages/mod.rs
jeanphilippeD: There are many ways to group or not group these, can be from the point of view of elder, or relocating/joining node. I think that the fact that the message behave/are used in a similar fashion (GenesisUpdate/Relocate) is the more relevant fact.

src/states/joining_peer.rs
maqi: > If the message was invalid, Elder will have decided "{} - Ignoring relocation JoinRequest from {} - invalid payload." and not replied.
Then the joining node may only get a `join request time out` error and have no idea what the problem is, and keeps re-join using the same info ? hmm...   I think a bit more detailed failure reason shall be given?

jeanphilippeD: But then, the bug is worth than that, it means a group of valid Elders sent an invalid message.
Essentially the problem is the type SignedRoutingMessage does not embed all the type information. (i.e that as an Adult we would never have reached the Relocate state if the message was not a Relocate message. One alternative is to store both RelocateDetails and SignedRoutingMessage, thus keeping type information but now getting duplicate info. Another alternative is to have a SignedRelocateMessage that contiain a SignedRoutingMessage that had to contain RelocateDetails, I can see that as a reasonable approach, I'll look at adding a new commit that would remove that confusion.

jeanphilippeD:
Pushed new commit to address review comments.

jeanphilippeD:
Rebase to fix error with updated dependency.
Also fixed the new type SignedRelocateDetails so Deserialization does not allow to break the invariant (that would panic a node receiving it).

src/states/common/approved.rs
```diff
@@ -521,14 +498,14 @@ pub trait Approved: Base {
...
 
-    fn check_signed_relocation_details(&self, details: &SignedRelocateDetails) -> bool {
-        match details.verify(self.chain().get_their_key_infos()) {
-            Ok(()) => true,
-            Err(error) => {
-                self.log_verify_failure(details, &error, self.chain().get_their_key_infos());
-                false
-            }
-        }
+    fn check_signed_relocation_details(&self, msg: &SignedRoutingMessage) -> bool {
```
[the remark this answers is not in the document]
jeanphilippeD: can do, makes little difference, will fix.

jeanphilippeD:
Pushed force fixing review comment.

PR #2029, opened 2020-01-28
Message unification part 1

madadam:
This PR contains some refactoring, cleanup and code reorganization in preparation for the message unification itself which will happen in a separate PR.


src/messages/mod.rs
```diff
@@ -329,9 +329,9 @@ impl Debug for SignedRoutingMessage {
     }
 }
 
-impl Debug for MessageContent {
+impl Debug for RoutingVariant {
```
jeanphilippeD: That should have moved with MessageContent to the new file in previous commit.

src/messages/direct.rs
```diff
@@ -18,14 +18,15 @@ use crate::{
...
 use std::{
     fmt::{self, Debug, Formatter},
     hash::{Hash, Hasher},
-    mem,
+    ops::Deref,
 };
 
 /// Body of DirectVariant
-#[derive(Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Eq, PartialEq, Hash, Serialize, Deserialize)]
```
jeanphilippeD: not 100% sure that was a win,  remove fairly simple implementation, but traded off a less easy usage.

src/states/adult/mod.rs
```diff
@@ -586,6 +585,15 @@ impl Base for Adult {
                 self.direct_msg_backlog.push((p2p_node, msg));
                 Ok(Transition::Stay)
             }
+            msg => {
```
jeanphilippeD: This is not ideal, we probably want to keep the exhaustive list in these matches because we need to look at them everytime a variant is added or changed.

src/states/adult/mod.rs
```diff
@@ -411,9 +410,9 @@ impl Adult {
         if self.in_location(msg.message_dst()) {
             let signed_msg = msg.take_or_deserialize_signed_routing_message()?;
             match &signed_msg.routing_message().content {
-                RoutingVariant::GenesisUpdate(info) => {
+                Variant::GenesisUpdate(info) => {
                     self.verify_signed_message(&signed_msg)?;
-                    return self.handle_genesis_update(info.clone());
+                    return self.handle_genesis_update((**info).clone());
                 }
                 _ => {
```
jeanphilippeD: hence the reason to enumerate all, it is strange that for the direct message we choose to do something different (i.e not add to backlog).

src/states/elder/mod.rs
```diff
@@ -1559,6 +1553,14 @@ impl Base for Elder {
                 );
                 self.direct_msg_backlog.push((p2p_node, msg));
             }
+            msg => {
```
jeanphilippeD: Same

src/states/elder/mod.rs
```diff
@@ -1559,6 +1553,14 @@ impl Base for Elder {
...
+                trace!(
```
jeanphilippeD: The same decision was not made for `dispatch_routing_message` because match was not exhausitve, was it right?
(I guess it does not matter very much for this intermediate state because it cannot happen).

src/messages/mod.rs
```diff
@@ -347,6 +351,11 @@ impl VerifyStatus {
     }
 }
 
+pub enum QueuedMessage {
```
jeanphilippeD: Yes, that change seem to make quite a lot of sense to me.

src/messages/security_metadata.rs
```diff
@@ -16,31 +16,7 @@ use crate::{
...
-use std::{
-    collections::BTreeSet,
-    fmt::{self, Debug, Formatter},
-};
-
-/// Metadata needed for verification of the sender.
-/// Contain shares of the section signature before combining into a BLS signature
-/// and into a FullSecurityMetadata.
-#[derive(Eq, PartialEq, Clone, Hash, Serialize, Deserialize)]
-pub struct PartialSecurityMetadata {
```
jeanphilippeD: It probably make more sense to fold that commit with the previous one since we are adding AccumulatingMessage to functionally replace PartialSecurityMetadata, and that removal change is pretty small.

src/location.rs
```diff
@@ -6,77 +6,86 @@
...
+#[derive(Clone, Copy, Eq, PartialEq, Hash, Serialize, Deserialize, Debug)]
+pub enum DstLocation {
+    /// Destination is a single node with the given name.
+    Node(XorName),
+    /// Destination are the nodes of the section whose prefix matches the given name.
+    Section(XorName),
+    /// Destination are the nodes whose name is matched by the given prefix.
+    Prefix(Prefix<XorName>),
+    /// Destination is the node at the `ConnectionInfo` the message is directly sent to.
+    Direct,
```
dirvine: I think this should go as well. We will always have the connectioninfo to send the message to. So when it's a message to forward the receiver only needs to check dst == him or forward, but we cannot send a message to anyone we don't have the connection info for. So here we use Direct, but I think that itsef may be soemthing the `send` method gives to `quic-p2p` when it selects a target Feels like we perhaps mix this up a little here. Maybe the unification push I have atm but the actual wire format will need to include the `<SocketAddr>` of the target and we can ignore the Xorname there. So the routing message has no direct, but the `quic-p2p` message needs a SocketAddr and the message. Maybe that is the answer?

src/messages/security_metadata.rs
```diff
@@ -18,18 +18,20 @@ use bincode::serialize;
 use serde::{Deserialize, Serialize};
 use std::fmt::{self, Debug, Formatter};
 
-/// Metadata needed for verification of the sender.
 #[derive(Eq, PartialEq, Clone, Hash, Serialize, Deserialize)]
-pub struct SectionSecurityMetadata {
-    pub proof: SectionProofSlice,
-    pub signature: bls::Signature,
+#[allow(clippy::large_enum_variant)]
+pub enum SecurityMetadata {
```
jeanphilippeD: Yes, flattening this enum make sense since it is simpler now.

src/messages/accumulating.rs
```diff
@@ -104,14 +104,14 @@ impl AccumulatingMessage {
             }
         };
 
-        let security_metadata = SecurityMetadata::Section {
+        let src_authority = SrcAuthority::Section {
```
jeanphilippeD: Yes, the rename seem pretty reasonable.

src/messages/accumulating.rs
```diff
@@ -0,0 +1,299 @@
...
+            dst,
+            variant,
+            proof,
+            public_key_set,
+            signature_shares,
+        })
+    }
+
+    /// Add the signature shares of `other` into this message.
+    /// Note: currently no validation is performed that the messages have the same content.
```
dirvine: Perhaps a big FIXME here to be addressed soon and we can look t verifying the content and sig at the same time and do away with a bad_sigs container.

src/location.rs
```diff
@@ -6,77 +6,86 @@
...
-    Section(XorName),
-    /// A set of nodes with names sharing a common prefix - may span multiple `Section`s present in
-    /// the routing table or only a part of a `Section`
-    PrefixSection(Prefix<XorName>),
-    /// A single node
-    Node(XorName),
+/// Message source location.
+#[allow(clippy::large_enum_variant)]
+#[derive(Clone, Copy, Eq, PartialEq, Hash, Serialize, Deserialize, Debug)]
+pub enum SrcLocation {
```
dirvine: Is it worth having these as seperate files?

src/messages/accumulating.rs
```diff
@@ -0,0 +1,299 @@
+// Copyright 2020 MaidSafe.net limited.
```
dirvine: The file should be called accumulating_message perhaps?

src/messages/authority.rs
```diff
@@ -0,0 +1,101 @@
+// Copyright 2020 MaidSafe.net limited.
```
dirvine: Again perhaps rename file src_authority ?

tests/mock_network/messages.rs
```diff
@@ -90,14 +91,23 @@ fn send_and_receive() {
...
+        .filter(|n| n.inner.is_elder() && n.in_dst_location(&dst))
     {
         loop {
             match node.try_next_ev() {
-                Ok(Event::MessageReceived { content, src, dst }) => {
+                Ok(Event::MessageReceived { content, src, .. }) => {
                     request_received_count += 1;
                     if req_content == content {
-                        if let Err(err) = node.inner.send_message(dst, src, res_content.clone()) {
+                        let res_src = SrcLocation::PrefixSection(*node.our_prefix());
```
jeanphilippeD: This change look a bit out of place in this commit (Splitting Location into SrcLocation and DstLocation, why is it relevant  here. I'm also not clear what we are trying to do here.

src/location.rs
dirvine: I wonder (just question) is `Src` and `Dst` alone perhaps not enough for these names?

src/location.rs
```diff
@@ -11,29 +11,18 @@ use crate::xor_space::{Prefix, XorName};
...
-    /// A single section whose prefix matches the given name
-    Section(XorName),
-    /// A set of nodes with names sharing a common prefix - may span multiple `Section`s present in
-    /// the routing table or only a part of a `Section`
-    PrefixSection(Prefix<XorName>),
-    /// A single node
+    /// A single node with the given name.
     Node(XorName),
+    /// A single section with the given prefix.
+    Section(Prefix<XorName>),
```
jeanphilippeD: Not 100% sure since SrcLocation::Section means something diferent from DstLocation::Section, seem a bit strange.

src/location.rs
jeanphilippeD: SrcLocation is more explicit and does not really add noise when used in code. I'd prefer we keep the full thing.

src/location.rs
jeanphilippeD: Could be a follow up PR if that make sense, maybe?

src/messages/accumulating.rs
```diff
@@ -31,18 +34,22 @@ pub struct AccumulatingMessage {
...
-        content: RoutingMessage,
+        src: Prefix<XorName>,
+        dst: DstLocation,
+        variant: Variant,
         section_share: &SectionKeyShare,
         public_key_set: bls::PublicKeySet,
         proof: SectionProofSlice,
     ) -> Result<Self> {
-        let bytes = serialize(&content)?;
+        let bytes = serialize(&(&src, &dst, &variant))?;
```
jeanphilippeD: This split of one type into multiple does not seem quite right, especially as we serialize it like that with a specific order in multiple places. Would it make sense to still have a type, maybe one that serialize like it happens here?

src/messages/authority.rs
```diff
@@ -6,37 +6,65 @@
...
 
 impl SrcAuthority {
+    pub fn location(&self) -> SrcLocation {
+        match self {
+            Self::Node { public_id, .. } => SrcLocation::Node(*public_id),
+            Self::Section { prefix, .. } => SrcLocation::Section(*prefix),
+        }
+    }
+
+    pub fn as_node(&self) -> Result<&PublicId> {
```
jeanphilippeD: Not 100% sure about this interface and `as_section`, they seem easy to use incorrectly.

src/location.rs
jeanphilippeD: I did thought it made sense in one file when reviewing.

src/messages/direct.rs
madadam: The custom `Hash` impl becomes more annoying when we merge `DirectVariant` and `RoutingVariant` into one type. Anyway, ideally we would have `Hash` implemented for `parsec::{Request, Response}`, then none of this would be necessary.

src/chain/chain.rs
```diff
@@ -1457,7 +1457,7 @@ impl Chain {
                 }
                 self.candidates(target_name)?
             }
-            Location::PrefixSection(ref prefix) => {
+            DstLocation::Prefix(ref prefix) => {
```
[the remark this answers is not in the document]
madadam: I know you wanted to remove it but I would prefer to do it in a separate PR because it's more involved - required changes in the `NeighbourInfo` messages flow.

src/messages/direct.rs
jeanphilippeD: Make sense, it only require to add Hash to parsec NetworkEvent since we already implement it anyway in Routing.

src/location.rs
madadam: So `DstLocation::Direct` does several things:
1. It's for messages sent directly to a socket address we know, so no hops/relays take place
2. We don't know or don't care about the recipient's name
3. The destination is whoever actually receives the message (so `in_location(Direct)` is always `true`)

I added this variant to keep the "direct message" flows we currently have in order to keep the amount of changes in this PR somehow reasonable. I don't mind trying to addres this in a followup PR though.

src/messages/accumulating.rs
madadam: I'd say this is a low priority because in practice the signature accumulator calls this function only on messages that have identical hashes (and thus content). Might be still worth at least raising a github issue for it.

src/location.rs
madadam: Not sure, they are pretty small and related. Makes more sense to me to keep then in the same file.

src/messages/authority.rs
madadam: They simplify the message dispatch though. In what way could they be used incorrectly?

src/location.rs
madadam: Yeah, me too. It's bit more explicit without being that much more verbose.

tests/mock_network/messages.rs
madadam: Previously we would just flip the src and dst when sending the response, but we can't do that anymore as they are different types. Also in later commit they actually have different variants so a simple mapping of src <-> dst won't work.

jeanphilippeD: Wouldn't it make sense to do it in the later commit then where it is necessary?
It just seem odd since we never use PrefixSection as a source so far (I think) (in later commit we start using the Prefix as the source, but I think we make it behave differently).

madadam: > Wouldn't it make sense to do it in the later commit then where it is necessary

But it is necessary here. I can't just take `SrcLocation` and pass it to where `DstLocation` is expected (which is what we did previously because they were the same types). Or am I misunderstanding what you mean?

src/location.rs
madadam: Should I rename it to `Prefix` perhaps?

tests/mock_network/messages.rs
jeanphilippeD: I mean you can match the dst to a DstLocation::Section(name) => SrcLocation(name) no?

jeanphilippeD: Like you do for the res_dst.

src/location.rs
jeanphilippeD: Maybe, but we might loose the context that this is a Section?

jeanphilippeD: Maybe the fact that they contain different type but named the same is less confusing. (like you did).

madadam: So `PrefixSection` then? As it was?

jeanphilippeD: It was meaning a different thing and is not particularly clear. I'd say lets forget that comment for now, I'm not sure the "fix" we have are helping.

src/chain/chain.rs
```diff
@@ -910,7 +910,7 @@ impl Chain {
     /// `Location::Node` instead of the stored knowledge. Has no effect for other location types.
```
jeanphilippeD: Minor: comment : Location -> DstLocation

src/chain/chain.rs
```diff
@@ -1423,9 +1427,9 @@ impl Chain {
     ///     - if our name *is* the destination, returns an empty set; otherwise
     ///     - if the destination name is an entry in the routing table, returns it; otherwise
     ///     - returns the `N/3` closest members of the RT to the target
-    pub fn targets(&self, dst: &Location) -> Result<(Vec<&P2pNode>, usize), RoutingError> {
+    pub fn targets(&self, dst: &DstLocation) -> Result<(Vec<&P2pNode>, usize), RoutingError> {
```
jeanphilippeD: Might be worth doing a search and replace for `Location::Section`, `Location::PrefixSection` ...

src/messages/accumulating_message.rs
```diff
@@ -269,9 +269,13 @@ mod tests {
...
         RoutingMessage {
-            src: SrcLocation::Section(rng.gen()),
+            src: SrcLocation::Section(gen_prefix(rng)),
             dst: DstLocation::Section(rng.gen()),
             content: Variant::UserMessage(rng.sample_iter(Standard).take(6).collect()),
         }
     }
+
+    fn gen_prefix(rng: &mut MainRng) -> Prefix<XorName> {
+        Prefix::new(rng.gen_range(0, 4), rng.gen())
```
jeanphilippeD: Any reason for the prefix to not be fully random?

src/messages/with_bytes.rs
```diff
@@ -139,18 +139,21 @@ mod tests {
...
 
     #[test]
     fn serialise_and_partial_at_hop_message() {
         let mut rng = rng::new();
         let full_id = FullId::gen(&mut rng);
-        let msg = gen_message(&mut rng);
-
-        let signed_msg_org = unwrap!(SignedRoutingMessage::single_source(msg, &full_id,));
+        let msg = RoutingMessage {
+            src: SrcLocation::Node(*full_id.public_id().name()),
```
jeanphilippeD: Why the change from SrcLocation::Section to Node? Not sure it makes much difference but I wonder. (Particularly since all our hop messages have a Section as a source.

src/messages/accumulating_message.rs
madadam: Not really except to keep a potential log output short. Shouldn't affect anything in this test anyway.

src/messages/with_bytes.rs
madadam: I don't believe it makes any difference for this test, but I though that since the message is `single_source` it would be more realistic.

PR #2030, opened 2020-01-29
Message unification part 2

madadam:
Followup to #2029


src/routing_message_filter.rs
```diff
@@ -51,8 +52,13 @@ impl RoutingMessageFilter {
     }
 
     // Filter incoming `RoutingMessage`. Return whether this specific message has already been seen.
-    pub fn filter_incoming(&mut self, msg: &HopMessageWithBytes) -> FilteringResult {
-        let hash = msg.full_message_crypto_hash();
+    pub fn filter_incoming(&mut self, msg: &MessageWithBytes) -> FilteringResult {
+        // FIXME: Not filtering direct messages until we decide what to do with them.
```
jeanphilippeD: I don't think there is a need for a FIXME. It is a reasonable approach.

src/routing_message_filter.rs
```diff
@@ -67,10 +73,15 @@ impl RoutingMessageFilter {
     // Return `KnownMessage` also if hashing the message fails - that can be handled elsewhere.
     pub fn filter_outgoing(
         &mut self,
-        msg: &HopMessageWithBytes,
+        msg: &MessageWithBytes,
         pub_id: &PublicId,
     ) -> FilteringResult {
-        let hash = msg.full_message_crypto_hash();
+        // FIXME: Not filtering direct messages until we decide what to do with them.
```
jeanphilippeD: Same

src/states/adult/mod.rs
```diff
@@ -321,118 +322,55 @@ impl Adult {
...
-
-    fn verify_signed_message(&self, msg: &SignedRoutingMessage) -> Result<(), RoutingError> {
-        msg.verify(self.chain.get_their_key_infos())
-            .and_then(VerifyStatus::require_full)
-            .map_err(|error| {
-                self.log_verify_failure(msg, &error, self.chain.get_their_key_infos());
-                error
-            })
+        let msg = msg.take_or_deserialize_message()?;
+        if self.should_handle_message(&msg)? {
```
jeanphilippeD: That is not quite right, we should not `take_or_deserialize_message` if we do not have `should_handle_message` true, that defeat the purpose.

jeanphilippeD: Maybe an option would be to have a function like, that does the take_or_deserialize_message inside it (since you want to do the verify in it).

```
if let Some(msg) = self.should_handle_message(&msg)? {
     self.handle_message(None, msg, outbox)
}
```
Probably need a different name.

src/states/adult/mod.rs
```diff
@@ -539,69 +462,119 @@ impl Base for Adult {
...
                 Ok(Transition::Stay)
             }
-            msg => {
-                trace!(
-                    "{} Unhandled direct message {}, ignoring: {:?}",
-                    self,
-                    p2p_node.public_id(),
-                    msg,
-                );
+            _ => {
```
jeanphilippeD: Probably best to be exhaustive here with the diferent variants.

src/states/adult/mod.rs
```diff
@@ -539,69 +462,119 @@ impl Base for Adult {
...
+        // Only verify messages we know how to handle.
+        match msg.variant {
+            Variant::GenesisUpdate(_)
+            | Variant::Relocate(_)
+            | Variant::MessageSignature(_)
+            | Variant::ParsecRequest(..)
+            | Variant::ParsecResponse(..)
+            | Variant::BootstrapRequest(_)
+            | Variant::ConnectionResponse => (),
+            _ => return Ok(true),
```
jeanphilippeD: I'm not really clear why we don't need to verify the other messages.
Also would be good to be exaustive here, what are the message we will not verify.

src/states/adult/tests.rs
```diff
@@ -211,7 +218,7 @@ fn genesis_update_message_successful_trust_check() {
     let gen_pfx_info = adult_test.gen_pfx_info(1);
     let msg = adult_test.genesis_update_message(gen_pfx_info);
 
-    let _ = unwrap!(adult_test.adult.handle_signed_message(msg));
+    let _ = unwrap!(adult_test.adult.handle_message(None, msg, &mut ()));
```
jeanphilippeD: I'm a bit confused about why this test function and the next call different functions.

src/states/bootstrapping_peer.rs
```diff
@@ -279,61 +283,68 @@ impl Base for BootstrappingPeer {
...
+    fn handle_message(
         &mut self,
-        msg: Variant,
-        p2p_node: P2pNode,
+        sender: Option<ConnectionInfo>,
+        msg: Message,
         _: &mut dyn EventBox,
-    ) -> Result<Transition, RoutingError> {
+    ) -> Result<Transition> {
+        let response = match msg.variant {
```
jeanphilippeD: Why the changes in this function? It seem we are still doing mostly the same thing?

src/states/common/base.rs
```diff
@@ -268,29 +260,44 @@ pub trait Base: Display {
...
+            );
 
-                self.handle_direct_message(msg, P2pNode::new(public_id, connection_info), outbox)
-            }
+            return Ok(Transition::Stay);
+        }
+
+        if self.should_relay_message(msg.message_dst()) {
+            // Relay closer to the destination or broadcast to the rest of our section.
+            self.relay_message(sender.clone(), msg.clone())?;
```
jeanphilippeD: Why do we need the sender to relay_message?

src/states/common/base.rs
```diff
@@ -268,29 +260,44 @@ pub trait Base: Display {
...
-                        trace!(
-                            "{} - Received direct message from unconnected peer {}: {:?}",
-                            self,
-                            public_id,
-                            msg
-                        );
-                        return Ok(Transition::Stay);
-                    };
+    ) -> Result<Transition> {
+        let sender = self.peer_map().get_connection_info(&src_addr).cloned();
```
jeanphilippeD: Should that be done further down, where we need it?

src/states/common/base.rs
```diff
@@ -354,28 +361,30 @@ pub trait Base: Display {
...
+            Ok(message) => message,
+            Err(error) => {
+                error!("{} - Failed to create message: {:?}", self, error);
+                return;
+            }
         };
 
-        let message = match to_network_bytes(&message) {
-            Ok(bytes) => bytes,
+        let bytes = match serialize(&message) {
```
jeanphilippeD: Not 100% sure that removing `to_network_bytes`... makes things clearer. It was quite good to see that this was serialization of a specific object we send on the wire. If we want to do this change, maybe should be in its own PR, but I'm not yet convinced we want to.

src/states/common/base.rs
```diff
@@ -427,12 +434,13 @@ pub trait Base: Display {
             .disconnect_from(peer_addr);
     }
 
-    fn send_msg_to_client(&mut self, peer_addr: SocketAddr, msg: Bytes, token: Token) {
-        let client = Peer::Client { peer_addr };
-        self.network_service_mut()
-            .service_mut()
-            .send(client, msg, token);
-    }
+    fn filter_incoming_message(&mut self, msg: &MessageWithBytes) -> bool;
```
jeanphilippeD: Would it make sense to keep these at the top like the others?

src/states/elder/mod.rs
```diff
@@ -665,178 +656,151 @@ impl Elder {
...
+
+        Ok(Transition::Stay)
     }
 
-    // If the message is for us, verify it then, handle the enclosed routing message and swarm it
-    // to the rest of our section when destination is targeting multiple; if not, forward it.
-    fn handle_signed_message(&mut self, msg: HopMessageWithBytes) -> Result<(), RoutingError> {
-        if !self.routing_msg_filter.filter_incoming(&msg).is_new() {
+    fn handle_accumulated_message(&mut self, msg: Message) -> Result<()> {
+        // FIXME: this is almost the same as `Base::handle_new_deserialized_message` - find a way
```
jeanphilippeD: Might be worth doing as part of that PR now it is smaller.

jeanphilippeD: Wondering if `add_proof` should return a `MessageWithBytes`.

src/states/elder/mod.rs
```diff
@@ -665,178 +656,151 @@ impl Elder {
...
-                self.log_verify_failure(&signed_msg, &error, self.chain.get_their_key_infos());
-                error
-            })?;
-
-            if handle {
-                self.update_our_knowledge(&signed_msg);
-                self.routing_msg_queue.push_back(signed_msg);
-            }
+        let msg = msg.take_or_deserialize_message()?;
+        if self.should_handle_message(&msg)? {
```
jeanphilippeD: Same `take_or_deserialize_message` should only happen after or within should_handle_message whne we know we need it.

src/states/elder/mod.rs
```diff
@@ -665,178 +656,151 @@ impl Elder {
...
-        match (msg.content, msg.src, msg.dst) {
-            (
-                NeighbourInfo(elders_info),
-                src @ SrcLocation::Section(_),
-                dst @ DstLocation::PrefixSection(_),
-            ) => {
-                self.handle_neighbour_info(elders_info, src, dst, src_authority)?;
-                Ok(Transition::Stay)
+        match msg.variant {
+            Variant::NeighbourInfo(elders_info) => {
```
jeanphilippeD: I think this is not right, we still need to do that same match we did before.
i.e still want to check the sending authority and the destination are as expected.

src/states/elder/mod.rs
```diff
@@ -1269,32 +1214,23 @@ impl Elder {
...
-                return Some(result);
-            }
-        };
+    fn get_signature_targets(&self) -> Vec<XorName> {
+        let mut list = self
+            .chain
+            .our_elders()
+            .map(|p2p_node| p2p_node.name())
+            .copied()
+            .sorted_by(|lhs, rhs| self.our_prefix().name().cmp_distance(lhs, rhs));
```
jeanphilippeD: Won't it be always the same targets? (won't it always be the first elders (maybe)?) is that what happened before? is that what we want (I thought not).

jeanphilippeD: Should it not be the ones closest to the destination?

src/states/elder/mod.rs
```diff
@@ -1325,19 +1261,39 @@ impl Elder {
...
+                // lagging behind (or the sender is faulty/malicious). We can't handle the
+                // message ourselves but the other elders likely can.
+                Ok(false)
+            }
+            Ok(VerifyStatus::ProofTooNew) => Err(RoutingError::UntrustedMessage),
+            Err(error) => Err(error),
+        }
+    }
+
+    fn is_message_verified(&self, msg: &Message) -> bool {
```
jeanphilippeD: Not sure the name is sensible as it seem to suggest a cheap operation. (one already done).

src/states/elder/mod.rs
```diff
@@ -1413,14 +1355,11 @@ impl Base for Elder {
...
-        self.send_routing_message(
-            RoutingMessage {
-                src,
-                dst,
-                content: Variant::UserMessage(content),
-            },
-            None,
-        )
+        if let DstLocation::Direct = dst {
+            return Err(RoutingError::BadLocation);
```
jeanphilippeD: Why is this an error? if this error ever occur is it because of a bug?

src/states/joining_peer.rs
```diff
@@ -295,49 +261,47 @@ impl Base for JoiningPeer {
                     }
                 }
             }
+            Variant::NodeApproval(gen_info) => {
```
jeanphilippeD: Again probably want to check the the src type and destination types are as expected.

src/states/joining_peer.rs
```diff
@@ -295,49 +261,47 @@ impl Base for JoiningPeer {
...
             Variant::ConnectionResponse | Variant::BootstrapResponse(_) => (),
-            _ => {
-                debug!(
-                    "{} Unhandled direct message from {}, adding to backlog: {:?}",
-                    self,
-                    p2p_node.public_id(),
-                    msg
-                );
-                self.msg_backlog.push(QueuedMessage::Direct(p2p_node, msg));
+            _ if msg.dst.is_single() => {
```
jeanphilippeD: Should we be exhaustive?

tests/mock_network/churn.rs
```diff
@@ -707,13 +707,11 @@ impl Expectations {
                             *section_msgs_received.entry(key).or_insert(0usize) += 1;
                         }
                     } else {
+                        let dst_name = unwrap!(dst.as_node());
                         assert_eq!(
-                            orig_name,
-                            dst.name(),
+                            orig_name, *dst_name,
```
jeanphilippeD: Better to keep assert unique.
`assert_eq!( Ok(orig_name), dst_name, ...` instead of unwrap followed by assert.

src/states/adult/mod.rs
madadam: The idea is to verify only the messages we actually handle, because the others might have proof too new for us and we shouldn't fail on that. We are relaying them to our elders anyway and they will do the verification.

Also, this is only really necessary because we have the `log_or_panic` thing in `verify_message`.

madadam: But makes sense to be exhaustive.

src/states/common/base.rs
madadam: So we can backlog them (we store the sender in the backlog).

src/states/adult/mod.rs
jeanphilippeD: If that is the case then it seem that we should do that in should_handle? (though i see that is needed for backlog. It just does not seem reasonable for a verify function to not verify messages. Maybe that verify_message should always return true, and then call it again at the point were we handle it. I'm not sure, but that strange dependency does not seem solid.

src/states/common/base.rs
jeanphilippeD: Should it not be None for relay messages?

src/messages/accumulating.rs
```diff
@@ -31,18 +34,22 @@ pub struct AccumulatingMessage {
...
-        content: RoutingMessage,
+        src: Prefix<XorName>,
+        dst: DstLocation,
+        variant: Variant,
         section_share: &SectionKeyShare,
         public_key_set: bls::PublicKeySet,
         proof: SectionProofSlice,
     ) -> Result<Self> {
-        let bytes = serialize(&content)?;
+        let bytes = serialize(&(&src, &dst, &variant))?;
```
jeanphilippeD: Question: do we need the `src` in the thing that we signed, it is us after all and will be in the self validating SrcAuthority?

madadam: Not sure. Say we take it out - is it guarantee that the src prefix is genuine then?

jeanphilippeD: Well, the src prefix + version will be the thing that we use to find the key on receiving a message, so if it is not right then it will not validate? (otherwise it is the top one, and it was signed by previous keys).

src/messages/mod.rs
```diff
@@ -1,614 +1,103 @@
...
-            &self.content.content,
-            &self.security_metadata,
-        )
-            .serialize(serialiser)
-    }
-}
+impl Message {
+    /// Creates a message from single node.
+    pub(crate) fn single_src(src: &FullId, dst: DstLocation, variant: Variant) -> Result<Self> {
+        let serialized = serialize(&(&dst, &variant))?;
```
jeanphilippeD: However, contrarily to the Section bls signature, the PublicId is not self validating, so we probably should have that in the signature.

src/states/elder/mod.rs
madadam: I think the prefix in `SrcLocation::Section` will always be out prefix no? So taking only our section elders should be enough?

> Should it not be the ones closest to the destination?

That's not what we did previously, but maybe it makes sense to do, so we don't always pick the same nodes?

jeanphilippeD: > That's not what we did previously, but maybe it makes sense to do, so we don't always pick the same nodes?

That is true, and probably it is not for that PR, or at least a separate commit, it just seem strange that we would always use the same target for message we send.

src/states/elder/mod.rs
madadam: `DstLocation::Direct` is not valid in this case because `UserMessage` is a "hop" message. 

> if this error ever occur is it because of a bug?

Yes, but not in routing. This is called by the upper layer.

src/states/elder/mod.rs
jeanphilippeD: > So taking only our section elders should be enough?
Yes, that is correct, what was strange that we would always take the same of our elders.

src/states/elder/mod.rs
jeanphilippeD: So I see you've added `let _ = msg.src.as_section()?;..` But it is only for this, that comment applied probably to all messages. I'm still thinking matching on the tuple make sense?

jeanphilippeD: But I guess it makes exhaustive matching not possible. Not sure what is better.

src/states/common/base.rs
```diff
@@ -268,28 +266,50 @@ pub trait Base: Display {
...
+            );
 
-                self.handle_direct_message(msg, P2pNode::new(public_id, connection_info), outbox)
-            }
+            return Ok(Transition::Stay);
+        }
+
+        if self.should_relay_message(msg.message_dst()) {
+            // Relay closer to the destination or broadcast to the rest of our section.
+            self.relay_message(msg.clone())?;
```
jeanphilippeD: Why is that cloned?

src/states/adult/mod.rs
```diff
@@ -321,118 +322,52 @@ impl Adult {
...
-    fn handle_signed_message(
+    fn handle_accumulated_message(
         &mut self,
-        msg: HopMessageWithBytes,
-    ) -> Result<Transition, RoutingError> {
-        if !self.routing_msg_filter.filter_incoming(&msg).is_new() {
+        msg: MessageWithBytes,
+        outbox: &mut dyn EventBox,
+    ) -> Result<Transition> {
+        if !self.filter_incoming_message(&msg) {
```
jeanphilippeD: did you meant to remove the duplication and call the base function? i.e https://github.com/maidsafe/routing/pull/2030#discussion_r372332158

src/states/adult/mod.rs
```diff
@@ -539,63 +459,127 @@ impl Base for Adult {
...
-        msg: HopMessageWithBytes,
-        _outbox: &mut dyn EventBox,
-    ) -> Result<Transition, RoutingError> {
-        self.handle_signed_message(msg)
+    fn filter_incoming_message(&mut self, message: &MessageWithBytes) -> bool {
+        self.msg_filter.filter_incoming(message).is_new()
+    }
+
+    fn should_verify_message(&self, msg: &Message) -> bool {
+        // As Adult, we only verify the messages that we know how to process ourselves.
```
jeanphilippeD: Better but still not ideal as is, I'd be happier if we can make that clearer.
Would it make sense to split the function and do something like that?

```rust
         if let Some(msg) = self.preprocess_message(msg)? {
            if self.should_verify_message(msg) {
                 self.verify_message(&msg)?;
                 self.handle_message(None, msg, outbox)
            } else {
                 self.handle_unverified_message(&msg)
            }
        } else {
            Ok(Transition::Stay)
        } 
````

Comment (jeanphilippeD) [reply src/states/adult/mod.rs]: Initial comment was https://github.com/maidsafe/routing/pull/2030#discussion_r372307058

File: src/states/elder/mod.rs (hunk 14 above)
Comment (madadam) [reply src/states/elder/mod.rs]: > But it is only for this, that comment applied probably to all messages

But I do that for all messages. If you notice there are calls like `as_section`, `as_node`, etc... In every branch except `UserMessage` because that one can be sent with any src/dst. Not sure matching on the tuple would buy as much. And as you said, we would lose exhaustive match.

Comment (jeanphilippeD) [reply src/states/elder/mod.rs]: Indeed, missed that.

File: src/messages/mod.rs (hunk 22 above)
Comment (madadam) [reply src/messages/mod.rs]: Hm, that's not what we were doing before though. But probably makes sense. Just to be sure - it's only the encryption_key part of it that's not self-validating, right?

Comment (jeanphilippeD) [reply src/messages/mod.rs]: So would make sense in its own commit.

File: src/states/common/base.rs (hunk 23 above)
Comment (madadam) [reply src/states/common/base.rs]: We still use the message below. Or are you saying we should pass it by reference and let the implementations of `relay_message` clone it only when necessary? That would probably be better.

File: src/states/adult/mod.rs (hunk 25 above)
Comment (madadam) [reply src/states/adult/mod.rs]: Sounds reasonable.

File: src/states/common/base.rs (hunk 23 above)
Comment (jeanphilippeD) [reply src/states/common/base.rs]: > we should pass it by reference and let the implementations of relay_message clone it only when necessary

Yes. relay_message should not need to clone generally.

File: src/states/adult/mod.rs [hunk 26]
```diff
@@ -493,31 +428,16 @@ impl Base for Adult {
...
-
-                        self.handle_filtered_signed_message(msg)
-                    }
-                    QueuedMessage::Direct(sender, msg) => {
-                        self.handle_direct_message(msg, sender, outbox)
-                    }
-                };
-
-                match result {
+                match self.handle_message(sender, message, outbox) {
```
jeanphilippeD: This does not do the verify of the message that was done before this commit since it happened in `handle_filtered_signed_message`.

src/states/adult/mod.rs
```diff
@@ -539,63 +459,127 @@ impl Base for Adult {
...
-        use crate::messages::Variant::*;
-        match msg {
-            MessageSignature(msg) => self.handle_message_signature(*msg, *p2p_node.public_id()),
-            ParsecRequest(version, par_request) => {
-                self.handle_parsec_request(version, par_request, p2p_node, outbox)
+    ) -> Result<Transition> {
+        trace!("{} - Handle message {:?}", self, msg);
+
+        match msg.variant {
+            Variant::GenesisUpdate(info) => self.handle_genesis_update(*info),
```
jeanphilippeD: Should it check that the source was a section, and the destination itself?
i.e like for Elder messages we probably should check the type of source/destination is as expected for all messages.

src/states/joining_peer.rs
jeanphilippeD: I don't think the check happen here because we ignore the result of as_section / as_node.

madadam: Oops, you are right. Forgot the question marks :flushed:

jeanphilippeD: Maybe could at least specify the type of the _ in this case, that would avoid that issue.

src/states/elder/mod.rs
jeanphilippeD: That fix me is still there (slightly changed). Probably best to fix it with an additional commit.

src/states/joining_peer.rs
jeanphilippeD: Have you thought about the suggestion to be explicit with the type `let _: type = ...?;` maybe clippy would complain but that would make it explicit that we don't want  a Result<..>

src/states/elder/mod.rs
madadam: The FIXME is there because it's not addressed yet and likely won't be in this PR. To address it properly would require some changes which I think are out of scope for this PR.

jeanphilippeD: What change would it require?

it seem to me that the only difference is that we want to do something slightly different if we did the accumulation.
 1 - either pass a parameter to allow the different behavior.
 2 - is there a good reason for us doing different thing in this case?

But I'm happy for it to be in another PR.

madadam: > What change would it require?

The `handle_accumulated_message` function is also called from `send_routing_message` which 1) doesn't take `outbox` and 2) doesn't return `Transition`. That means that `handle_accumulate_message` also can't take/return those. But `handle_new_deserialized_message` does take/return them.

My preferred solution to this would be to get rid of the `outbox` argument altogether and use a member variable instead (that would simplify the code overall, not just in this place). That would still leave us with the `Transition` return type but I think we can ignore that here because `Elders` never transition on receiving a message. Anyway, I think that is something for a future PR.

jeanphilippeD: ok

jeanphilippeD:
@dirvine was happy with the initial version presented in #2029 and confirm this is good so will merge.

PR #2031, opened 2020-01-30
Remove EventStream

madadam:
Followup to #2030.

Also includes removal of unnecessary test hack + minor cleanup.


src/state_machine.rs
```diff
@@ -544,117 +525,23 @@ impl StateMachine {
...
-                received = true;
-                let queue = match &action {
-                    Action::HandleTimeout(_) => &mut timed_out_events,
-                    _ => &mut events,
-                };
-                queue.push(EventType::Action(Box::new(action)));
-            }
-        }
-
-        // Interleave timer events with routing or network events.
```
jeanphilippeD: Note that this interleaving no longer happen, I think, so it may reduce the test coverage a bit.

src/action.rs
```diff
@@ -25,6 +25,7 @@ use std::{
 ///    2. `Action::Terminate` indicates to `Core` that no new actions should be taken and all
 ///       pending events should be handled.
 ///       After completion `Core` will send `Event::Terminated`.
+#[allow(clippy::large_enum_variant)]
```
jeanphilippeD: Just one thing I meant to ask before approving: did you update the safe_vault code and run its integration tests?

src/state_machine.rs
madadam: Yeah I know. Not sure we really need it. Also, [the crossbeam-channel docs](https://docs.rs/crossbeam-channel/0.4.0/crossbeam_channel/struct.Select.html) say this:

> Select allows you to define a set of channel operations, wait until any one of them becomes ready, and finally execute it. If multiple operations are ready at the same time, a **random one among them is selected**.

So maybe we get that feature for free now?

madadam: Hm, but that can break determinism :thinking:

jeanphilippeD: Yes, I think you are correct, we do get it somewhat for free. the select is "random" I think in a deterministic simple way, but that is likely enough for our purpose.

madadam: OK, so you don't think it breaks test repeatability then?

jeanphilippeD: > m, but that can break determinism
No, no worries here is it a reproducible random if I remember correctly when I looked it up

jeanphilippeD:
````
pub fn shuffle<T>(v: &mut [T]) {
    let len = v.len();
    if len <= 1 {
        return;
    }

    thread_local! {
        static RNG: Cell<Wrapping<u32>> = Cell::new(Wrapping(1406868647));
    }
```
Yes, I think we are good.

madadam: Yeah, just looking at it. It uses hard-coded seed so we should be good.

src/action.rs
dirvine: That should really be a case of us using semver properly and the consumer to update properly.

madadam: Just ran the tests. All passed.

PR #2033, opened 2020-02-04
Remove peer map

madadam:
With the changes in https://github.com/maidsafe/quic-p2p/pull/90 it's no longer necessary to keep track of the mapping between socket addresses and peer infos. Thus it's possible to remove `PeerMap`.

Note: with the `PeerMap` removal, we lost the ability to detect the situation where we get disconnected from all our peers. Previosly we would trigger restart in that case, but after this PR such situation is no longer handled at all. We might possibly want to bring that check back in some form, but likely not based on connectedness but rather on responsiveness.


src/states/adult/mod.rs
```diff
@@ -253,7 +253,15 @@ impl Adult {
...
             .collect();
 
-        self.network_service_mut().remove_and_disconnect_all();
+        // Disconnect from everyone we know.
+        let addrs: Vec<_> = self
+            .chain
+            .known_nodes()
+            .map(|node| node.connection_info().peer_addr)
+            .collect();
+        for addr in addrs {
```
jeanphilippeD: That seem unfortunate as it does not disconnect from those we do not know.
Is that suggesting there is a missing API here from quick_p2p that we need if we want to not keep track of those we are connected?

src/states/bootstrapping_peer.rs
```diff
@@ -179,9 +179,10 @@ impl BootstrappingPeer {
     }
 
     fn reconnect_to_new_section(&mut self, new_conn_infos: Vec<ConnectionInfo>) {
-        self.network_service_mut().remove_and_disconnect_all();
+        for addr in mem::take(&mut self.pending_requests) {
```
jeanphilippeD: Same, maybe.

tests/mock_network/drop.rs
```diff
@@ -14,19 +14,11 @@ use routing::{event::Event, mock::Environment, NetworkParams};
...
-
-    drop(node);
+    let _ = nodes.remove(index);
 
     // Using poll_all instead of poll_and_resend here to only let the other nodes realise the node
     // got disconnected, but not make more progress.
     let _ = poll_all(nodes);
-
-    for node in nodes.iter_mut().filter(|n| close_names.contains(&n.name())) {
-        assert!(!node.inner.is_connected(&endpoint));
```
jeanphilippeD: Did you solve the detecting offline node then? because we only get an offline event if we tried to connect?
That would be great news.

src/states/elder/tests.rs
```diff
@@ -496,30 +457,6 @@ fn when_accumulate_offline_and_start_dkg_and_section_info_then_node_is_removed_f
...
-
-    // Add new section member to reach elder_size.
-    let new_info = elder_test.new_elders_info_with_candidate();
-    elder_test.accumulate_online(elder_test.candidate.clone());
-    elder_test.accumulate_start_dkg(&new_info);
-    elder_test.accumulate_section_info_if_vote(&new_info);
-
-    // Bootstrap succeeds.
-    elder_test.handle_bootstrap_request(*node1.public_id(), node1.connection_info());
-    assert!(elder_test.is_connected(&node1.connection_info().peer_addr));
```
jeanphilippeD: While this test used connected to verify a property, I don't think it is what it test, it test that a node can join.
I'd be incline to think we should still have this test, just using a different property to validate it is the case.

BTW, we don't really have responsiveness check for Adults, how do we detect they drop with this PR?

src/states/adult/mod.rs
madadam: Do we need to disconnect from those we don't know? But if so, then yeah, we would need support from quic-p2p. Either a method to disconnect from all, or a method to list all connections.

tests/mock_network/drop.rs
madadam: Not sure what you mean?

src/states/adult/mod.rs
jeanphilippeD: I think so, that was what was happening before I think? Yes, I agree.

tests/mock_network/drop.rs
jeanphilippeD: How does offline detection works now?for elders? for adults?

src/states/elder/tests.rs
madadam: > we should still have this test, just using a different property to validate it is the case.

Maybe we can check that the elder sent `BootstrapResponse` back? Also, we don't do that thing where we reject the request if we don't have enough members anymore, so that part of the test is not necessary.

> BTW, we don't really have responsiveness check for Adults, how do we detect they drop with this PR?

We don't do drop detection by responsiveness yet. We rely on the `ConnectionFailure` event from quic-p2p. This has not changed in this PR.

tests/mock_network/drop.rs
madadam: By receiving the `ConnectionFailure` event.

jeanphilippeD: So I guess we are still connected or trying to send messages then, because otherwise why would we get a connection failure? Will this happen on all Elders?

jeanphilippeD: I'm wondering, if adults are not connected to all elders currently, would they detect them dropping before an Elder change occur? Is there a missing test for that?

src/states/adult/mod.rs
madadam: We don't disconnect from unknown peers as elders either, so maybe this issue needs a more systematic approach? Maybe we can auto-disconnect from anyone who connects to us or sends us a message and who we don't recognize?

tests/mock_network/drop.rs
madadam: > I'm wondering, if adults are not connected to all elders currently, would they detect them dropping before an Elder change occur?

I don't think so, but why would adults need to be able to detect elder going offline? We currently don't do anything there: https://github.com/maidsafe/routing/blob/fleming/src/states/adult/mod.rs#L439-L442 

> Is there a missing test for that?

As adults don't do anything on peer drop I'm not sure what is there to test. Is there maybe an edge case you see?

PR #2038, opened 2020-02-06
Gone peers detection

### LINKED ISSUE #2037: Improve offline detection
Currently we detect nodes as offline when we receive the `ConnectionFailure` quic-p2p event from them. This is bad because a node can go offline only temporarily (due to random connection issues, node restart or other reasons) but we would still vote it off the section.

We need to come up with better solution. Some ideas:

- Use the `UnsentUserMessage` mechanism of quic-p2p: when a message fails to be sent to a particular peer more than some threshold number of times, we can flag that peer as being offline
- Or when we detect the connection failure, we can try to reconnect a couple of times and flag the peer as offline only if we fail all of them.
- Or something else.

Note: we cannot use non-responsiveness detection for this, because that applies only to elders (it is based on parsec participation) and we need a solution for all members.

madadam:
This PR improves the handling of peers that go offline. Instead of immediately voting them off the section, we give them chance to come back. This is to allow for temporary network connectivity issues, node restarts, and so on.

- When the peer gets disconnected or when messages sent to them fail more than allowed number of times, we consider them as "lost"
- When peer is "lost", we keep sending them "pings" and wait for the response. If we receive the response within a timeout we consider the peer as online again
- When the timeout expires and we have not received a response, we send the ping again and double the timeout
- We repeat this process a couple of times
- If we don't receive any response even after all the attempts are exhausted, we consider the peer as "gone" and vote it to be removed from the section.
- When that vote accumulates (that is, more than 2/3 of elders consider the peer "gone"), it is removed from the section

Closes #2037 

Draft, because some tests are failing.

### DIFF
--- a/src/messages/variant.rs
+++ b/src/messages/variant.rs
@@ -60,6 +60,11 @@ pub enum Variant {
     ParsecRequest(u64, parsec::Request),
     /// Parsec response message
     ParsecResponse(u64, parsec::Response),
+    /// Message sent to a lost peer to determine whether the disconnect is permanent or only
+    /// temporary.
+    PingRequest,
+    /// Response to `PingRequest` to notify the recipient that we are still online.
+    PingResponse,
 }
 
 impl Debug for Variant {
@@ -85,6 +90,8 @@ impl Debug for Variant {
             Self::MemberKnowledge(payload) => write!(f, "MemberKnowledge({:?})", payload),
             Self::ParsecRequest(version, _) => write!(f, "ParsecRequest({}, ..)", version),
             Self::ParsecResponse(version, _) => write!(f, "ParsecResponse({}, ..)", version),
+            Self::PingRequest => write!(f, "PingRequest"),
+            Self::PingResponse => write!(f, "PingResponse"),
         }
     }
 }
--- a/src/network_service/sending_targets_cache.rs
+++ b/src/network_service/sending_targets_cache.rs
@@ -8,7 +8,10 @@
 
 use crate::{quic_p2p::Token, ConnectionInfo};
 use log::LogLevel;
-use std::{collections::HashMap, net::SocketAddr};
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    net::SocketAddr,
+};
 
 const MAX_RESENDS: u8 = 3;
 
@@ -18,25 +21,6 @@ enum TargetState {
     Sending(u8),
     /// the last sending attempt (if any) failed; in total, the stored number of attempts failed
     Failed(u8),
-    /// sending to this target succeeded
-    Sent,
-}
-
-impl TargetState {
-    pub fn is_complete(&self) -> bool {
-        match *self {
-            Self::Failed(x) => x > MAX_RESENDS,
-            Self::Sent => true,
-            Self::Sending(_) => false,
-        }
-    }
-
-    pub fn is_sending(&self) -> bool {
-        match *self {
-            Self::Failed(_) | Self::Sent => false,
-            Self::Sending(_) => true,
-        }
-    }
 }
 
 #[derive(Default)]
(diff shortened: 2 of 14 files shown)


src/network_service/sending_targets_cache.rs
```diff
@@ -72,83 +56,78 @@ impl SendingTargetsCache {
...
+    pub fn target_succeeded(&mut self, token: Token, target: SocketAddr) {
+        let mut entry = if let Entry::Occupied(entry) = self.cache.entry(token) {
+            entry
+        } else {
+            return;
+        };
+
+        entry.get_mut().retain(|(info, _)| info.peer_addr != target);
+
+        if entry.get().is_empty() {
```
jeanphilippeD: Is that correct, how does it handle send to 3 out of 7 targets like the code did before?
Not 100% sure the refactoring makes thing clearer.

src/network_service/sending_targets_cache.rs
```diff
@@ -106,28 +100,58 @@ impl SendingTargetsCache {
     }
 }
 
-fn set_target_failed(targets: &mut Vec<(ConnectionInfo, TargetState)>, target: &SocketAddr) {
+/// Info about failed target.
+pub struct TargetFailed {
+    /// Next target to try to resend the message to, if any.
+    pub next: Option<ConnectionInfo>,
+    /// Whether the failed target should be considered as lost - that is - whether it failed more
+    /// than maximum allowed number of times.
```
jeanphilippeD: It seem that unless the message is for all targets, this will not return false unless multiple nodes do not respond.
Not clear what is the design here? do we get any information out of messages for 3 out of 7 targets?

src/network_service/sending_targets_cache.rs
madadam: We remove only the target that succeeded sending. The ones that are still in `Sending` state are kept until they too are done sending.

jeanphilippeD: For 3 out of 7 targets (i.e hop messages), not all nodes are in Sending at the start. i.e we are suppose to complete even if some node have not been sent too.

madadam: > Not 100% sure the refactoring makes thing clearer.

There is fewer unreachable states (so fewer log_and_panics), no dependency on implicit ordering of calls (`target_failed` and `take_next_target`) and it's slightly more efficient (fewer lookups). And the code size is slightly reduced too. Seems worthwhile to me, not sure.

madadam: > For 3 out of 7 targets (i.e hop messages), not all nodes are in Sending at the start. i.e we are suppose to complete even if some node have not been sent too.

Ah, I see what you mean now. Yeah, that's an oversight. We should remove when there are no more `Sending`.

jeanphilippeD: Not sure the functionality is the same. Additionally, we discard information as we progress which is not always ideal (i.e when the message is failed to be sent to all its required targets, it could be good to see what are original targets, and which succeeded). while one of the `log_or_panic` was removed, its condition is not removed, i.e it would still be a bug to receive an event regarding an already sent event, we just ignore it now instead of logging it out.

src/states/common/base.rs
```diff
@@ -311,16 +311,19 @@ pub trait Base: Display {
...
-        self.network_service_mut().send_message_to_next_target(
+        if self.network_service_mut().send_message_to_next_target(
             msg,
             token,
             conn_info.peer_addr,
             log_ident,
-        );
-        Transition::Stay
+        ) {
+            self.handle_peer_lost(conn_info.peer_addr, outbox)
```
jeanphilippeD: What if that message is to a client (from a vault)?

src/network_service/sending_targets_cache.rs
```diff
@@ -106,28 +100,58 @@ impl SendingTargetsCache {
...
             let _ = targets.remove(index);
+            true
         }
         TargetState::Sending(x) => {
             if x < MAX_RESENDS {
                 *state = TargetState::Failed(x + 1);
+                false
             } else {
                 let _ = targets.remove(index);
+                true
```
jeanphilippeD: I'm wondering if it would make more sense to return the final result on success or fail i.e if we remove the collection we send it back or the item that failed max number of times. (for both the success part and fail part)? not sure.

src/network_service/sending_targets_cache.rs
madadam: > when the message is failed to be sent to all its required targets, it could be good to see what are original targets, and which succeeded

We weren't doing that before so I didn't know that we want that. Btw, I actually added a log now which prints when a message completely fails to be sent. I doesn't print which targets failed and which succeeded, but it's better than nothing.

> one of the log_or_panic was removed, its condition is not removed

Fair enough. I'll add it back.

src/states/bootstrapping_peer.rs
```diff
@@ -465,11 +465,11 @@ mod tests {
         let mut sel = mpmc::Select::new();
         machine.register(&mut sel);
 
-        // Blocking step for the first one.
-        let op_index = sel.ready();
+        // Step for the first one.
+        let op_index = unwrap!(sel.try_ready());
```
jeanphilippeD: Why do we change that? does it make the name of the function wrong?

madadam: `ready` is blocking and I had this test hang due to error before. I think it's better to panic than to block.

jeanphilippeD: Make sense.

src/states/common/base.rs
madadam: `handle_unsent_message` is called only for nodes.

src/messages/variant.rs
```diff
@@ -60,6 +60,11 @@ pub enum Variant {
     ParsecRequest(u64, parsec::Request),
     /// Parsec response message
     ParsecResponse(u64, parsec::Response),
+    /// Message sent to a lost peer to determine whether the disconnect is permanent or only
+    /// temporary.
+    PingRequest,
```
jeanphilippeD: So I'm not clear why we need the PingRequest/Response?
I mean with previous change we get it lost if we cannot send a message to it trying multiple times (which i though was already counted as the acceptable margin for a node to be unresponsive). Why do we have this additional one?

src/states/common/base.rs
jeanphilippeD: Yes, indeed. wonder if that is an issue for vault messages (i.e does the message still end up in that cache).

src/messages/variant.rs
madadam: To give the node chance to come back. The unsent message thing might be exhausted too quickly.

jeanphilippeD: Why should it be exhausted too quickly, is it not what we should ensure is not exhausted too quickly instead of adding a new mechanism. (i.e ensure messages are passed through to all needed targets).

tests/mock_network/utils.rs
```diff
@@ -617,7 +617,7 @@ pub fn verify_section_invariants_for_node(node: &TestNode, elder_size: usize) {
 
     if !our_prefix.is_empty() {
         assert!(
-            our_section_elders.len() >= elder_size,
+            check_min_elder_size(our_section_elders.len(), elder_size),
```
jeanphilippeD: Ok, I think it would be good to explain in commit comment why we change that, why would we loose soo many nodes we have smaller elders. All tests are suppose to make this not happen, and in real life, with many adults this is really not suppose to be possible.

src/messages/variant.rs
madadam: The ping mechanism allow us to control the time interval for which we allow a node to be non-responsive. With the `UnsentUserMessage` mechanism we don't have that control. In some cases it can cycle through all the attempts (currently 3) too quickly and we might accuse the node as being lost even though it might not really be.

What I have some doubts about however is whether we actually need the `UnsentUserMessage` mechanism for this at all. Depending on how quic-p2p works, we might get `ConnectionFailure` in all the cases we need. In particular:
1. If we are connected to a peer and that peer goes offline, we should always get `ConnectionFailure` 
2. If we are not connected and we sent them a message, quic-p2p should establish the connection automatically, but if it fails to do so because the peer went offline it **should** emit `ConnectionFailure` too, but I'm not sure this is the case. I think it wasn't the case in the past, but maybe that changed.

src/states/common/base.rs
madadam: You mean message sent from a vault to another vault(s)? Such messages go through routing normally which I think is what we want.

src/network_service/sending_targets_cache.rs
madadam: > do we get any information out of messages for 3 out of 7 targets?

Probably not. Should work for direct messages though.

tests/mock_network/utils.rs
madadam: I'll add an explanation to the commit message.

src/messages/variant.rs
jeanphilippeD: - Well if we want a minimum time for retry we should add that i.e the retry is not just a number but a time elapsed.

- Regarding ConnectionFailure, yes, I think we should get this, but we will get this on the first fail attempt, I think, I think we fixed at least the mock to do the right thing and give an error. I think the real code was fixed too.

src/states/common/base.rs
jeanphilippeD: Well I mean both, but i think in this case I was more thinking vault -> client. Routing still have the quic_p2p instance so I think sending message to client also go through it?

src/messages/variant.rs
madadam: OK. I'm going to verify the quic-p2p behaviour regarding this and if it's what we want then I'll remove the unsent mechanism.

src/states/common/base.rs
madadam: Yes, but you need to use `Peer::Client` when sending to clients and that's what you'll get back in `UnsentUserMessage` and `SentUserMessage`, so we always know whether the event is for client or node and only handle the ones for node.

src/network_service/sending_targets_cache.rs
jeanphilippeD: Yes, it will work for direct message, seem a bit strange.

madadam: Yes. This might not be needed after all. See this: https://github.com/maidsafe/routing/pull/2038#discussion_r377130248

src/messages/variant.rs
jeanphilippeD: Sorry I do not understand how you came to that course of action? why would we remove the unsent mechanism.

madadam: By remove the unsent mechanism I meant don't use it for detecting lost peers. Not removing it completely.

src/states/common/base.rs
jeanphilippeD: But do we do the same check when sending a message and not adding it to the cache?

madadam: There is a separate function `send_message_to_client` for that and we don't insert it to the cache there.

src/messages/variant.rs
jeanphilippeD: Ok. Still not clear how we would do that (i.e even if we get ConnectionFailure can we use it because it does not say whether it is the first one etc...)

src/states/common/base.rs
jeanphilippeD: ok make sense.

src/messages/variant.rs
madadam: So there are two phases in this:
1. Detect that a node is lost, which means at least temporarily offline
2. Detect that it is gone, which means it is permanently offline and should be voted out of the section.

This discussion is mainly about point 1. This PR currently implements two mechanisms for that:
1. When we receive the `ConnectionFailure` event
2. Or when we repeatedly  fail to send a message to that peer 

What I'm trying to argue here is that we might not need the second mechanism because the first one might be sufficient. So we don't care whether it's the first time and so on because we only use it to trigger the second phase which is where we use the ping requests/responses to give the node chance to come back.

Does that make sense?

jeanphilippeD: It does to an extent, but it seem to provide something that essentially re-implement in a different way the resend mechanism. i.e the resend is design to send to all its expected target unless a node is gone, then we may only be able to send to part of the targets. (if it is broken we should fix it). It seem the reason to stop trying to re-send to a failing node should be the same as the reason we decide a node should be voted offline.

madadam: OK, I see what you mean. So maybe we should just modify the resend mechanism and not resend to a previously failed target immediately, but only after some delay (which we increase after every attempt)?

jeanphilippeD: That could be, though:
 - I think there is a timeout in quic-p2p when trying to connect to send a message, so there already is a delay (to confirm? / maybe not in mock?).
 - Whatever delay we introduce should not delay sending to other targets if available.
 - Not sure if we want an additional delay or whether we want to keep on retrying until enough time has passed and max try reached.

madadam: - There is no timeout in mock quic-p2p, but we probably don't need that.
- Retrying until enough time passed sounds like a decent idea and should be simpler to implement. I'll give it a try.

madadam:
Superseded by https://github.com/maidsafe/routing/pull/2050

PR #2050, opened 2020-02-11
Lost detection

### LINKED ISSUE #2037: Improve offline detection

madadam:
This PR changes the way lost peers are detected which allows nodes to go offline for a short time and come back without being kicked out of the section. This is needed to support node restarts, short-term network failures, and so on.

Depends on https://github.com/maidsafe/quic-p2p/pull/93
Supersedes https://github.com/maidsafe/routing/pull/2038
Closes #2037

### DIFF
--- a/src/network_service/sending_targets_cache.rs
+++ b/src/network_service/sending_targets_cache.rs
@@ -6,9 +6,14 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::quic_p2p::Token;
+use crate::{quic_p2p::Token, time::Duration};
 use log::LogLevel;
 use std::{collections::HashMap, net::SocketAddr};
+
+/// Maximal number of resend attempts to the same target.
+pub const RESEND_MAX_ATTEMPTS: u8 = 3;
+/// Delay before attempting to resend a previously failed message.
+pub const RESEND_DELAY: Duration = Duration::from_secs(10);
 
 enum TargetState {
     /// we don't know whether the last send attempt succeeded or failed
@@ -69,10 +74,11 @@ impl SendingTargetsCache {
     }
 
     fn fail_target(&mut self, token: Token, target: SocketAddr) {
-        let _ = self
+        if let Some((_addr, state)) = self
             .target_states_mut(token)
             .find(|(addr, _state)| *addr == target)
-            .map(|(_addr, state)| match *state {
+        {
+            match *state {
                 TargetState::Failed(_) => {
                     log_or_panic!(LogLevel::Error, "Got a failure from a failed target!");
                 }
@@ -85,27 +91,32 @@ impl SendingTargetsCache {
                         "A target that should no longer fail - failed!"
                     );
                 }
-            });
+            }
+        }
     }
 
     /// Finds a Failed target with the lowest number of failed attempts so far. If there are
     /// multiple possibilities, the one with the highest priority (earliest in the list) is taken.
-    /// Returns None if no such targets exist.
-    fn take_next_target(&mut self, token: Token) -> Option<NextTarget> {
-        self.target_states_mut(token)
+    /// Returns `Never` if no such targets exist.
+    fn take_next_target(&mut self, token: Token) -> Resend {
+        if let Some((addr, failed_attempts, state)) = self
+            .target_states_mut(token)
             .filter_map(|(addr, state)| match state {
-                TargetState::Failed(x) => Some((*addr, *x, state)),
-                TargetState::Sending(_) | TargetState::Sent => None,
+                TargetState::Failed(x) if *x < RESEND_MAX_ATTEMPTS => Some((*addr, *x, state)),
+                TargetState::Failed(_) | TargetState::Sending(_) | TargetState::Sent => None,
             })
             .min_by_key(|(_addr, failed_attempts, _state)| *failed_attempts)
-            .map(|(addr, failed_attempts, state)| {
-                *state = TargetState::Sending(failed_attempts);
+        {
+            *state = TargetState::Sending(failed_attempts);
 
-                NextTarget {
-                    addr,
-                    failed_attempts,
-                }
-            })
+            if failed_attempts == 0 {
+                Resend::Now(addr)
+            } else {
+                Resend::Later(addr, RESEND_DELAY)
+            }
+        } else {
+            Resend::Never
+        }
     }
 
     fn should_drop(&self, token: Token) -> bool {
@@ -118,18 +129,10 @@ impl SendingTargetsCache {
             .all(|(_info, state)| !state.is_sending())
     }
 
-    pub fn target_failed(&mut self, token: Token, target: SocketAddr) -> NextTarget {
+    pub fn target_failed(&mut self, token: Token, target: SocketAddr) -> Resend {
         self.fail_target(token, target);
-        let next_target = match self.take_next_target(token) {
-            Some(next_target) => next_target,
-            None => {
-                log_or_panic!(LogLevel::Error, "Next target not found");
-                NextTarget {
-                    addr: target,
-                    failed_attempts: 1,
-                }
-            }
-        };
+
+        let next_target = self.take_next_target(token);
 
         if self.should_drop(token) {
             let _ = self.cache.remove(&token);
@@ -151,7 +154,12 @@ impl SendingTargetsCache {
     }
 }
 
-pub struct NextTarget {
-    pub addr: SocketAddr,
-    pub failed_attempts: u8,
+// How to resend a previously failed message.
+pub enum Resend {
+    // Resend it now to the given target.
+    Now(SocketAddr),
+    // Resend it to the given target after the given delay.
+    Later(SocketAddr, Duration),
+    // Don't resend - all attempts failed.
+    Never,
 }
--- a/src/network_service/sending_targets_cache.rs
+++ b/src/network_service/sending_targets_cache.rs
@@ -10,8 +10,6 @@ use crate::quic_p2p::Token;
 use log::LogLevel;
 use std::{collections::HashMap, net::SocketAddr};
 
-const MAX_RESENDS: u8 = 3;
-
 enum TargetState {
     /// we don't know whether the last send attempt succeeded or failed
     /// the stored number of attempts already failed before
@@ -23,14 +21,6 @@ enum TargetState {
 }
 
 impl TargetState {
-    pub fn is_complete(&self) -> bool {
-        match *self {
-            Self::Failed(x) => x > MAX_RESENDS,
-            Self::Sent => true,
-            Self::Sending(_) => false,
-        }
-    }
-
     pub fn is_sending(&self) -> bool {
         match *self {
             Self::Failed(_) | Self::Sent => false,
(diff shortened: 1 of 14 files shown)


src/network_service/sending_targets_cache.rs
```diff
@@ -72,83 +65,91 @@ impl SendingTargetsCache {
...
+    };
+
+    match *state {
+        TargetState::Failed(_) => {
+            log_or_panic!(
+                LogLevel::Error,
+                "Got a failure from failed target: {}",
+                target
+            );
+            let _ = targets.remove(index);
```
jeanphilippeD: Why remove here?

src/network_service/sending_targets_cache.rs
```diff
@@ -6,28 +6,39 @@
...
+};
 use log::LogLevel;
 use std::{
     collections::{hash_map::Entry, HashMap},
     net::SocketAddr,
 };
 
-const MAX_RESENDS: u8 = 3;
+/// Time interval during which we attempt to resend a message to a given target before giving up.
+pub const RESEND_TIMEOUT: Duration = Duration::from_secs(30);
```
jeanphilippeD: In my mind I'd think I'd probably have both. In particular, lets say you start with node A1. then you'll try node A2, A3, A4, A5, then your back to A1, then time out is gone.

I guess I had more something like MAX_TRY, but if it reach it too fast, continue until timeout pass at least.

tests/mock_network/utils.rs
```diff
@@ -603,7 +603,7 @@ pub fn verify_section_invariants_for_node(node: &TestNode, elder_size: usize) {
 
     if !our_prefix.is_empty() {
```
jeanphilippeD: Relax the min elder size requirement in tests 

Looking at this, it seem that the problem is the test is broken. We should not get in situation where we would get an elder_size that is smaller. In this case, maybe the safe_section_size needs to increase then?

madadam: I think the test failure was false positive. There was nothing wrong with the network, yet the test still failed. So I think relaxing the test requirements is the right thing to do, to make them closer to reality.

src/network_service/sending_targets_cache.rs
madadam: Or maybe we should somehow pause the timeout while a target is in the failed state. I'll see which approach is better.

tests/mock_network/utils.rs
jeanphilippeD: The problem is the way we relax the test requirements, we are moving away from reality where we would never have a case of not having enough nodes to populate the elders (i.e if we lost all adults already we are already gone.

madadam: I guess there are two ways to look at it then. My view is that the section is still functional so there is no reason for the test to fail. You are right that we don't expect a section to fall this low in the real world, but then that isn't something caused by the routing logic itself, rather by external circumstances. So maybe the actual right thing to do would be to modify the test to never cause this situation in the first place. Although that could be tricky.

That said, increasing `safe_section_size` would probably work too and would bring the test a bit closer to reality too. So I'm OK with doing that if you want me to.

jeanphilippeD: > That said, increasing safe_section_size would probably work too and would bring the test a bit closer to reality too. So I'm OK with doing that if you want me to.

Sounds good.

src/network_service/sending_targets_cache.rs
dirvine: I think if we are trying a node, we should try a few times with a pause between each try. However, rather than wait on that node, we should just move on to the next one. So we try them all quickly then a short pause and go around again, we can do this twice (so 3 attempts per node in total)  and if it fails then we need to error.

src/network_service/sending_targets_cache.rs
```diff
@@ -20,18 +20,21 @@ use std::{
 /// Time interval during which we attempt to resend a message to a given target before giving up.
 pub const RESEND_TIMEOUT: Duration = Duration::from_secs(30);
 
+struct Target {
+    // What state the target is at - sending or failed.
+    state: TargetState,
+    // Timestamp of the first send attempt to this target.
+    timestamp: Instant,
```
dirvine: If we try in a round robbin fashion, say 3 times then perhaps the time between each attempt should be at least 10 secs? So we keep cycling round until we have exhausted each attempt. I am wondering if there is a clever way we can arrange to code for a typical `send` method that can do these waits, but when we have multiple targets it wll not wat on each target ut cycle them all, but till not try to send until the min time between tries exhausts? 
However, if we try a 100 times in that period then it's also OK just as long as we have a min of 3 10 sec (say) timeouts between what we consider then `Failed attempt`. So each attempt may fail, but we have a timeout that says we could not send in TIMEOUT duration (and in that duration we have tried at least once)

src/network_service/sending_targets_cache.rs
```diff
@@ -53,102 +69,129 @@ impl SendingTargetsCache {
...
+        TargetState::Failed { .. } => {
+            log_or_panic!(
+                LogLevel::Error,
+                "Got a failure from failed target: {}",
+                failed_target_addr
+            );
+            true
+        }
+        TargetState::Sending(start) => {
+            target.total += start.elapsed();
```
maqi: So this `start` is the time when this target being seclected for a sending attempt.
which means the `total` actually is the : accumulated sum of the failure detection period, i.e.  `start.elapsed()` means the time that elapsed since we send message to the target to the time we received the failure from it. 

This makes `RESEND_TIMEOUT` doesn't mean a fixed interval (say a fixed 30s since first sending attempt), but a network performance dependent factor? i.e. in a fast network there could be thousands of attempts been made?

Or did I misunderstand anything?

src/states/bootstrapping_peer.rs
```diff
@@ -465,11 +469,11 @@ mod tests {
         let mut sel = mpmc::Select::new();
         machine.register(&mut sel);
 
-        // Blocking step for the first one.
-        let op_index = sel.ready();
+        // Step for the first one.
+        let op_index = unwrap!(sel.try_ready());
```
maqi: so now switched from a blocking step to a non-blocking step?

madadam: Yes. The reason is that I had this test fail in the past and it resulted in hang. This would instead panic which is preferable.

src/network_service/sending_targets_cache.rs
madadam: No, you are right. Good point. Wonder how much of an issue this is...

madadam:
Fixed a bug in the unsent message handling where the decision whether to resend the message now or delayed wasn't correctly computed.

src/states/common/base.rs
```diff
@@ -304,13 +318,52 @@ pub trait Base: Display {
...
+                trace!(
+                    "{} - Sending message ID {} to {} failed - resending to {} in {:?}",
+                    self,
+                    msg_token,
+                    addr,
+                    next_target.addr,
+                    RESEND_DELAY
+                );
+                let timer_token = self.timer().schedule(RESEND_DELAY);
+                self.network_service_mut().send_later(
```
maqi: would like to clarify that the total detection time of a failing peer via a particular message will then be: 
    RESEND_MAX_ATTEMPTS * (RESEND_DELAY + quic-p2p_detect_sending_failure_time) * number_of_failed_peers_at_the_same_time

src/states/common/base.rs
```diff
@@ -304,13 +318,52 @@ pub trait Base: Display {
...
-        token: Token,
-        _outbox: &mut dyn EventBox,
+        msg_token: Token,
+        outbox: &mut dyn EventBox,
     ) -> Transition {
-        let log_ident = LogIdent::new(self);
-        self.network_service_mut()
-            .send_message_to_next_target(msg, token, addr, log_ident);
-        Transition::Stay
+        let next_target = self.network_service_mut().target_failed(msg_token, addr);
```
maqi: just to confirm that the `failing addr` could be picked as `next_target` as well.
I don't think it's a problem as the `send` shall just be scheduled later.

madadam: That's right. Also If the number of previously failed attempts to that target is zero, the message will be sent to them immediately, otherwise with a delay.

src/states/common/base.rs
madadam: In case there is only one target, or if the number of targets is the same as the delivery group, then the total time to detect them offline is `RESEND_MAX_ATTEMPTS * (RESEND_DELAY + quic-p2p send failure time)`. If there are multiple targets and the delivery group is not all of them, the it gets more complicated. For example, let's say we send to 3 of 7. Two of them succeed and one fails. Then we pick next target to resend to, one which we haven't sent to before. If this succeeds then the message is successfully sent and we don't need to retry the previously failed one. So we would not detect them being offline at all. But that's OK, because there are always some other direct messages / swarm messages on which we would detect them.

maqi: inside the function of `take_next_target` at https://github.com/maidsafe/routing/pull/2050/files#diff-79a9599b4b67afb5d69b448a0c38fedbR97 , 
the filter seems only pick the `failed` target as candidate for the `next target`.
that's why it is `number_of_failed_peers_at_the_same_time` mentioned in my formula.

src/states/common/base.rs
```diff
@@ -304,12 +316,50 @@ pub trait Base: Display {
...
+                    "{} - Sending message ID {} to {} failed - resending to {} now",
+                    self,
+                    msg_token,
+                    addr,
+                    next_target.addr
+                );
+                self.network_service_mut()
+                    .send_now(next_target.addr, msg, msg_token);
+            }
+            n if n < RESEND_MAX_ATTEMPTS => {
```
jeanphilippeD: Commit: Delay resending to previously failed target 

Wonder if it would make sense to keep most of it where it was, i.e out of base. and maybe use the result to trigger a scheduling for example something like:

```rust
if let Some(shedule) = self.network_service_mut().send_message_to_next_target(msg, token, addr, log_ident) {
    let timer_token = self.timer().schedule(RESEND_DELAY);
    self.network_service_mut().send_later(timer_token, shedule);
}
```

i.e the RESEND_MAX_ATTEMPTS is still where it was as well as most of the code?

madadam: Sure. Do you think that would be clearly better than the current way? If so, why?

src/states/elder/mod.rs
```diff
@@ -1321,6 +1321,29 @@ impl Base for Elder {
         Transition::Stay
     }
 
+    fn handle_connection_failure(
+        &mut self,
+        addr: SocketAddr,
+        _outbox: &mut dyn EventBox,
+    ) -> Transition {
```
jeanphilippeD: Commit: Send Ping message to trigger lost detection on connection failure 

Should that happen for Adults as well? i.e to detect their Elders are gone? Or maybe we don't really care in this case I guess.

src/states/common/base.rs
jeanphilippeD: The initial design was trying to keep the resend logic completely separated from the state.
Now it is no longer completely possible because of the timer, but the reason I think it would be better is that States already do a lot, and it seem we should strive to avoid adding code to it when not necessary (i.e push it to the helper code).

src/states/elder/mod.rs
madadam: No because adults currently don't do anything when a peer gets lost.

maqi: Being informed by other Elders shall be enough for Adults?

src/states/common/base.rs
madadam: OK, that makes sense. I'll see what I can do.

tests/mock_network/churn.rs
```diff
@@ -314,36 +324,22 @@ fn drop_random_nodes<R: Rng>(
...
-        // If there already are other drops scheduled, make sure we remain with at least one
-        // more node above elder_size. This is because one of those other drops might trigger
-        // relocation of one of the existing elders and we wouldn't have anyone to replace it with
-        // otherwise.
-        if section.all_dropped() > 0 && section.all_remaining() <= elder_size + 1 {
-            continue;
-        }
-
-        if node.inner.is_elder() {
-            // Don't drop elder if a non-elder is already scheduled for drop.
```
maqi: so this situation is no longer an issue now?

madadam: I'd say so. I think we should be able to drop any number of elders as long as we have enough non-elders to replace them with.

PR #2051, opened 2020-02-11
Remove Gossip timer

### LINKED ISSUE #2049: Remove gossip timeout, send gossip as soon as something changes
Send `parsec::Request` to a random node on:
    - receiving parsec request (in addition to sending the response)
    - receiving parsec response
    - creating a vote

Also, consider modifying parsec to create gossip only if we have something the recipient doesn't know yet.

Note: we might also consider some kind of throttling in case this causes network flooding. For example limiting the number of in-flight `parsec::Request`s.

bochaco:
Resolves #2049


bochaco:
Closing this as this change is being covered by #2054

PR #2052, opened 2020-02-12
pull quic-p2p with separate channel for client and node

jeanphilippeD:
Depends on https://github.com/maidsafe/quic-p2p/pull/95

Closes https://github.com/maidsafe/routing/issues/2039

This allow to provide user, safe_vault with a receiver for quic_p2p
client events directly.
This will avoid having routing mediating between quic-p2p and vaults for
client events.

### DIFF
--- a/src/states/common/base.rs
+++ b/src/states/common/base.rs
@@ -10,7 +10,6 @@ use crate::{
     action::Action,
     chain::SectionKeyInfo,
     error::{Result, RoutingError},
-    event::Client,
     id::{FullId, PublicId},
     location::{DstLocation, SrcLocation},
     messages::{Message, MessageWithBytes, Variant},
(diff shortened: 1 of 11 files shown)


tests/mock_network/utils.rs
```diff
@@ -191,7 +191,7 @@ impl<'a> TestNodeBuilder<'a> {
     }
 
     pub fn create(self) -> TestNode {
-        let (inner, user_event_rx) = self
+        let (inner, user_event_rx, _client_rx) = self
```
dirvine: Is this correct, perhaps only naming. should they be client_rx and node_rx or similar? (just user and client seems quite similar in name)

jeanphilippeD: They are different channel, user_event is the channel where routing send event (before it also contained the client message). client_rx is the new channel that contains message for client directly from quic-p2p (which we don't care in our tests since we don't do client stuff in routing.

jeanphilippeD:
Remove now unnecessary code and expose needed type that is now part of the API.

src/states/common/base.rs
```diff
@@ -139,54 +138,24 @@ pub trait Base: Display {
...
-            ConnectedTo {
-                peer: Peer::Client(peer_addr),
-            } => {
-                let client_event = Client::Connected { peer_addr };
-                outbox.send_event(From::from(client_event));
-                Transition::Stay
-            }
+            ConnectedTo { peer } => match peer {
+                Peer::Client(_) => Transition::Stay,
+                Peer::Node(peer_addr) => self.handle_connected_to(peer_addr, outbox),
```
madadam: This is what I was talking about here https://github.com/maidsafe/quic-p2p/pull/95#discussion_r378197863

src/states/elder/tests.rs
```diff
@@ -544,7 +544,11 @@ struct JoiningPeer {
 
 impl JoiningPeer {
     fn new(rng: &mut MainRng) -> Self {
-        let (network_event_tx, network_event_rx) = crossbeam_channel::unbounded();
+        let (network_event_tx, network_event_rx) = {
+            let (node_tx, node_rx) = crossbeam_channel::unbounded();
+            let (client_tx, _) = crossbeam_channel::unbounded();
+            (quic_p2p::EventSenders { node_tx, client_tx }, node_rx)
+        };
```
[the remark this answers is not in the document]
jeanphilippeD: I explicitly did not make EventReceivers part of the API.
We have to do it because for tests we create it all at once, but in production we do not.
I considered created a helper function in tests, but it did not seem to bring much here.

src/states/common/base.rs
jeanphilippeD: Yes, indeed, I think it make more sense as a follow up PR. I don't think I'll have time to do it if we hope to merge it before end of thursday. I'll raise an issue for it.

jeanphilippeD: Raised as https://github.com/maidsafe/routing/issues/2053

src/states/elder/tests.rs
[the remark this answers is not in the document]
madadam: > but in production we do not.

Not sure what you mean by that, because in the above code you seem to be creating both channels at once anyway.

jeanphilippeD:
Updated to use the merged PR in quic-p2p

PR #2054, opened 2020-02-13
Remove gossip timer (take 2)

### LINKED ISSUE #2049: Remove gossip timeout, send gossip as soon as something changes

madadam:
~~Tests are currently failing. The test framework needs to be adapted.~~

Supersedes #2051 
Closes #2049

### DIFF
--- a/src/chain/elders_info.rs
+++ b/src/chain/elders_info.rs
@@ -81,8 +81,8 @@ impl EldersInfo {
         &self.members
     }
 
-    pub fn is_member(&self, pub_id: &PublicId) -> bool {
-        self.members.contains_key(pub_id.name())
+    pub fn is_member(&self, name: &XorName) -> bool {
+        self.members.contains_key(name)
     }
 
     pub fn member_nodes(&self) -> impl Iterator<Item = &P2pNode> + ExactSizeIterator {
@@ -125,12 +125,12 @@ impl EldersInfo {
 
     /// Returns `true` if the proofs are from a quorum of this section.
     pub fn is_quorum(&self, proofs: &ProofSet) -> bool {
-        proofs.ids().filter(|id| self.is_member(id)).count() >= quorum_count(self.len())
+        proofs.ids().filter(|id| self.is_member(id.name())).count() >= quorum_count(self.len())
     }
 
     /// Returns `true` if the proofs are from all members of this section.
     pub fn is_total_consensus(&self, proofs: &ProofSet) -> bool {
-        proofs.ids().filter(|id| self.is_member(id)).count() == self.len()
+        proofs.ids().filter(|id| self.is_member(id.name())).count() == self.len()
     }
 
     /// Returns `true` if `self` is a successor of `other_info`, according to its hash.
--- a/src/mock/parsec/mod.rs
+++ b/src/mock/parsec/mod.rs
@@ -188,6 +188,11 @@ where
     }
 
     pub fn poll(&mut self) -> Option<Block<T, S::PublicId>> {
+        // Make sure we only return blocks for which we reached consensus.
+        if self.first_unpolled >= self.first_unconsensused {
+            return None;
+        }
+
         state::with(self.section_hash, |state| {
             if let Some((block, holder)) = state
                 .get_block(self.first_unpolled)
--- a/src/mock/parsec/mod.rs
+++ b/src/mock/parsec/mod.rs
@@ -122,6 +122,12 @@ where
         Ok(())
     }
 
+    pub fn see_as(&self, peer_id: &S::PublicId) {
+        state::with(self.section_hash, |state: &mut SectionState<T, _>| {
+            state.see(peer_id)
+        })
+    }
+
     pub fn vote_for_as(&mut self, observation: Observation<T, S::PublicId>, vote_id: &S) {
         state::with(self.section_hash, |state| {
             let holder =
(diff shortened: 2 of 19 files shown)


src/states/common/approved.rs
```diff
@@ -154,36 +145,19 @@ pub trait Approved: Base {
...
-            .parsec_map_mut()
-            .handle_response(msg_version, par_response, pub_id, &log_ident)
-        {
-            self.parsec_poll(outbox)
-        } else {
-            Ok(Transition::Stay)
-        }
+        self.parsec_map_mut()
+            .handle_response(msg_version, par_response, pub_id, &log_ident);
+        self.schedule_parsec_gossip();
```
jeanphilippeD: So we are doing it both on response & request (on top of vote). So Start with 1 request will trigger 1 gossip from destination + 1 new gossip from source. i.e 1 -> 2 -> 4 -> 8 etc. So it seem very quickly everyone will gossip to everyone all the time. I don't see how the current schedule_parsec_gossip/send_scheduled_parsec_gossip would avoid this. Am I missing something?

madadam: > Am I missing something?

Not necessarily. This is PR is very much work in progress and there are most likely issues. I just wanted to put it out so it can be tested already.

That said, perhaps we should not send that additional request on receiving request, only the response?

src/parsec.rs
```diff
@@ -117,32 +120,38 @@ impl ParsecMap {
         request: Request,
         pub_id: id::PublicId,
         log_ident: &LogIdent,
-    ) -> (Option<Variant>, bool) {
+    ) -> Option<Variant> {
```
dirvine: Much better

src/states/adult/mod.rs
```diff
@@ -387,6 +387,16 @@ impl Base for Adult {
         &mut self.rng
     }
 
+    fn finish_handle_input(&mut self, outbox: &mut dyn EventBox) -> Transition {
```
dirvine: Yes, I reckon these states will go eventually as here perhaps we should return an error. Happy for that later though.

src/states/elder/mod.rs
```diff
@@ -1339,24 +1332,27 @@ impl Base for Elder {
         };
 
         if self.chain.is_peer_our_member(&pub_id) {
+            self.parsec_map.block_gossip_recipient(pub_id.name());
```
maqi: This shall be fine for the fleming branch.
However, once your PR 2050 (lost detection) got merged, may be the `block_gossip_recipient` here is not enough anymore? and shall be at the place of detecting the first failure?

madadam: Yes, it will be moved to `handle_connection_failure` then.

madadam: Maybe I can put it there already...

maqi: yeah, that might be better

src/parsec.rs
```diff
@@ -151,38 +163,51 @@ impl ParsecMap {
...
-
-        self.last_version() == msg_version
     }
 
     pub fn create_gossip(
         &mut self,
         version: u64,
         target: &id::PublicId,
     ) -> Result<Variant, CreateGossipError> {
+        if !self.pending_requests.insert(*target.name()) {
```
maqi: This actually means we cannot have more than one pending request to the same target. right?
In case of that, shall the returned error be handled to pick another candidate?

And, at https://github.com/maidsafe/routing/pull/2054/files#diff-98b4a6ae0c53dc7ababf2d21eb815263R264
seems when choosing the candidate, we already carried out filtering, so this case shall not be happen?

src/mock/parsec/mod.rs
```diff
@@ -188,6 +188,10 @@ where
     }
 
     pub fn poll(&mut self) -> Option<Block<T, S::PublicId>> {
+        if self.first_unpolled >= self.first_unconsensused {
```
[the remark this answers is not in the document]
madadam: This was actually a bug which was never exposed, because we never called `poll` more than once in a row without handling gossip in between. But now we sometimes do that (because `parsec_poll` is now called after every network_event or action). This is a simple fix that makes sure things are sane. That is, we only return a block from `poll` if we previously reached consensus on that block.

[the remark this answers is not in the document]
maqi: Ah, right.
maybe better put your explaination as a comment?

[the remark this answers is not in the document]
madadam: Will do.

src/chain/elders_info.rs
```diff
@@ -81,8 +81,8 @@ impl EldersInfo {
         &self.members
     }
 
-    pub fn is_member(&self, pub_id: &PublicId) -> bool {
-        self.members.contains_key(pub_id.name())
+    pub fn is_member(&self, name: &XorName) -> bool {
```
jeanphilippeD: Commit: Limit the number of in-flight parsec gossip requests 

Considering how often it is used, it is probably a nice quality of life to still have the one taking a PublicId.
Maybe consider renaming `is_member` to `is_member_id`, and adding a new `is_member_name` function (that you can use when you already have the name and from within `is_member_id`.

madadam:
Tests with `mock` passing now, but taking prohibitively long with `mock_base`.

madadam:
Superseded by #2056. Closing.

PR #2061, opened 2020-02-27
Reject duplicate votes in mock parsec

madadam:


src/mock/parsec/mod.rs
```diff
@@ -111,22 +111,24 @@ where
...
+            }
+        })
     }
 
     pub fn vote_for_as(&mut self, observation: Observation<T, S::PublicId>, vote_id: &S) {
         state::with(self.section_hash, |state| {
             let holder =
                 ObservationHolder::new(observation, vote_id.public_id(), self.consensus_mode);
-            state.vote(vote_id, holder);
+            let _ = state.vote(vote_id, holder);
```
dirvine: can`state.vote` actually fail?

src/mock/parsec/mod.rs
```diff
@@ -360,6 +362,7 @@ impl<T: NetworkEvent, P: PublicId> Response<T, P> {
 #[derive(Debug)]
 pub enum Error {
     InvalidPeerState,
+    DuplicateVote,
```
dirvine: Ah ignore last comment

PR #2071, opened 2020-03-17
Remove state machine - part 1

madadam:
This PR is the first part of the effort to remove the state machine pattern in order to simplify the crate. In this PR the state machine isn't removed yet. It only merges the `Adult` and `Elder` states into one state called `ApprovedPeer`. The actual removal will happen in a followup PR.

This PR also includes some minor refactoring and code cleanup.


src/chain/shared_state.rs
```diff
@@ -104,116 +96,30 @@ impl SharedState {
...
-            &self.handled_genesis_event,
-            &false,
-        );
-        self.handled_genesis_event = true;
-
-        if related_info.is_empty() {
-            return Ok(());
+    pub fn update(&mut self, new: Option<Self>, log_ident: &LogIdent) {
+        if self.handled_genesis_event {
+            log_or_panic!(
```
dirvine: Is this an example of a straigh log message. I am not sure we would want to panic in a test here?

src/chain/shared_state.rs
```diff
@@ -104,116 +96,30 @@ impl SharedState {
...
-            );
-            update_with_genesis_related_info_check_same(
-                log_ident,
-                "relocate_queue",
-                &self.relocate_queue,
-                &relocate_queue,
-            );
+        if let Some(new) = new {
+            if self.our_infos.len() > 1 && *self != new {
+                log_or_panic!(
```
dirvine: Perhaps same??

src/chain/shared_state.rs
```diff
@@ -276,14 +182,13 @@ impl SharedState {
         }
     }
 
-    /// Remove all entries from `out_members` whose name does not match our prefix.
-    pub fn remove_our_members_not_matching_our_prefix(&mut self) {
-        let (our_members, post_split_sibling_members) =
-            mem::replace(&mut self.our_members, BTreeMap::new())
-                .into_iter()
-                .partition(|(name, _)| self.our_prefix().matches(name));
+    /// Remove all entries from `out_members` whose name does not match our prefix and returns them.
```
dirvine: out -> our

src/node.rs
```diff
@@ -330,30 +336,33 @@ impl Node {
...
+    /// Returns whether the node is Elder.
     pub fn is_elder(&self) -> bool {
-        self.elder_state().is_some()
+        self.chain()
+            .map(|chain| chain.is_self_elder())
+            .unwrap_or(false)
     }
 
-    /// Returns whether the current state is `Elder` or `Adult`.
+    /// Returns whether the node is Elder or Adult.
```
dirvine: We should consider Infant as well I reckon, although infants should not cause us to work (very much), so no parsec and not in our members.

src/states/approved_peer/mod.rs
```diff
@@ -271,52 +266,62 @@ impl Elder {
...
-                "{} - Routing Table size: {:3}",
+    ////////////////////////////////////////////////////////////////////////////
+    // Public API
+    ////////////////////////////////////////////////////////////////////////////
+
+    pub fn our_prefix(&self) -> &Prefix<XorName> {
+        self.chain.our_prefix()
+    }
+
+    pub fn our_elders(&self) -> impl Iterator<Item = &P2pNode> {
```
dirvine: Nice :+1:

src/chain/shared_state.rs
madadam: This exists in fleming too. We can change it (as per the ongoing discussion about error handling), but best to do it in a separate PR focused on just that.

src/chain/shared_state.rs
madadam: Yes. Same response too: https://github.com/maidsafe/routing/pull/2071#discussion_r393615197

src/states/elder/mod.rs
```diff
@@ -1089,30 +1089,16 @@ impl Elder {
...
         trace!(
             "{}: Sending message {:?} via targets {:?}",
             self,
             msg,
-            target_p2p_nodes
+            targets
         );
 
-        let targets: Vec<_> = target_p2p_nodes
+        let targets: Vec<_> = targets
```
dirvine: Is this type not inferred by the compiler?

src/chain/chain.rs
```diff
@@ -1051,7 +1059,7 @@ impl Chain {
     /// Returns `true` if we are not in the process of waiting for a pfx change
     /// or if incoming event is a vote for the ongoing pfx change.
     fn can_handle_vote(&self, event: &NetworkEvent) -> bool {
-        if !self.state.split_in_progress {
+        if !self.split_in_progress {
```
dirvine: Not sure we need the if statement here, only the predicate

src/chain/shared_state.rs
```diff
@@ -104,116 +96,30 @@ impl SharedState {
...
-                &churn_event_backlog,
-            );
-            update_with_genesis_related_info_check_same(
-                log_ident,
-                "relocate_queue",
-                &self.relocate_queue,
-                &relocate_queue,
-            );
+        if let Some(new) = new {
+            if self.our_infos.len() > 1 && *self != new {
```
maqi: hmm...   so when `self == new`, the later on `self = new` will still be executed ?

src/node.rs
```diff
@@ -125,9 +124,16 @@ impl Builder {
                 if first {
                     debug!("Creating a first node in the Elder state");
```
maqi: maybe this log msg also need to be udpated?

src/states/elder/mod.rs
```diff
@@ -125,15 +125,14 @@ impl Elder {
         let public_id = *full_id.public_id();
         let connection_info = network_service.our_connection_info()?;
         let p2p_node = P2pNode::new(public_id, connection_info);
-        let mut first_ages = BTreeMap::new();
-        let _ = first_ages.insert(public_id, MIN_AGE_COUNTER);
+        let mut ages = BTreeMap::new();
+        let _ = ages.insert(public_id, MIN_AGE_COUNTER);
```
dirvine: What would be the consequences of this failing

src/state_machine.rs
```diff
@@ -277,15 +267,7 @@ pub enum Transition {
...
-    IntoAdult {
-        gen_pfx_info: GenesisPfxInfo,
-    },
-    // `Adult` state transition to `Elder`.
-    IntoElder {
-        old_pfx: Prefix<XorName>,
-    },
-    // `Elder` state transition to `Adult` as a result of demotion
-    Demote {
+    Approve {
```
maqi: the comment says `JoiningPeer state transitioning to Adult`, maybe shall be `to Approved` ?
also, `Approve` to `IntoApproved` ?

src/states/approved_peer/tests/mod.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright 2020 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+// Copyright 2019 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
```
maqi: duplicated licensing paragraph?

src/states/tests/mod.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright 2020 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+// Copyright 2019 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
```
maqi: same as above, duplicated licensing paragraph?

src/states/tests/mod.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright 2020 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+// Copyright 2019 MaidSafe.net limited.
```
maqi: shall removed?

src/states/tests/utils.rs
```diff
@@ -0,0 +1,99 @@
+// Copyright 2020 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+// Copyright 2019 MaidSafe.net limited.
```
maqi: same as above, shall removed? and the following is a duplicated licensing paragraph?

src/states/elder/mod.rs
madadam: Apparently not. Probably because we pass it by reference to `send_message_to_targets` and it takes a slice so rust has no way of knowing whether the thing we pass is a `&Vec<...>` or `&[...]`. Or something that just `Deref`s to a slice.

src/states/tests/utils.rs
```diff
@@ -0,0 +1,99 @@
...
+    GenesisPfxInfo {
+        elders_info,
+        public_keys,
+        state_serialized: Vec::new(),
+        ages,
+        parsec_version,
+    }
+}
+
+pub fn elder_age_counters<'a, I>(elders: I) -> BTreeMap<PublicId, AgeCounter>
```
maqi: wondering shall this be `approved_age_counter` ?

src/states/elder/mod.rs
madadam: Not sure what you mean by failing, but we can safely ignore the return value of `insert` because we just created the collection one line above so it's always going to return `None`.

src/state_machine.rs
madadam: :+1 for the comment. Regarding `Approve` vs `IntoApproved` - I kind of like the transition variants to be verbs describing what is happening rather than what are we transitioning into. But this whole enum is going away in the next PR (or the one after) anyway, so I wouldn't worry about it too much.

src/states/tests/utils.rs
madadam: No, it's still the age counters of elders, so the name is appropriate.

src/states/approved_peer/mod.rs
```diff
@@ -1087,55 +1810,29 @@ impl Elder {
...
-        self.vote_for_event(AccumulatingEvent::User(event));
+        AccumulatingMessage::new(content, secret_key, pk_set, proof)
     }
 
-    /// Returns the set of peers that are responsible for collecting signatures to verify a message;
-    /// this may contain us or only other nodes. If our signature is not required, this returns
-    /// `None`.
+    // Returns the set of peers that are responsible for collecting signatures to verify a message;
+    // this may contain us or only other nodes. If our signature is not required, this returns
+    // `None`.
```
maqi: The function actually returns an empty vector ?

src/states/approved_peer/mod.rs
```diff
@@ -1161,34 +1858,110 @@ impl Elder {
...
+    ////////////////////////////////////////////////////////////////////////////
+    // Miscellaneous
+    ////////////////////////////////////////////////////////////////////////////
 
-        let content = PlainMessage {
-            src: *self.our_prefix(),
-            dst,
-            variant,
-        };
+    fn print_rt_size(&self) {
```
maqi: actually prints more?

src/states/approved_peer/mod.rs
```diff
@@ -1161,34 +1858,110 @@ impl Elder {
...
 
-    pub fn in_src_location(&self, src: &SrcLocation) -> bool {
-        self.chain.in_src_location(src)
+    // Connect to all elders from our section or neighbour sections that we are not yet connected
+    // to and disconnect from peers that are no longer elders of neighbour sections.
+    fn update_peer_connections(&mut self, change: &EldersChange) {
+        if !self.chain.split_in_progress() {
+            let our_needed_connections: HashSet<_> = self
+                .chain
+                .known_nodes()
```
maqi: just to confirm this `known_nodes` return only elders ?
as the comment says `no longer elders of neighbour sections`

src/states/approved_peer/tests/elder.rs
```diff
@@ -0,0 +1,598 @@
+// Copyright 2018 MaidSafe.net limited.
```
maqi: 2020
and duplicated paragraph

src/states/approved_peer/tests/elder.rs
```diff
@@ -0,0 +1,598 @@
...
+    let new_info = env.new_elders_info_with_candidate();
+    env.accumulate_online(env.candidate.clone());
+    env.accumulate_start_dkg(&new_info);
+    env.accumulate_section_info_if_vote(&new_info);
+    env.accumulate_voted_unconsensused_events();
+
+    env.other_ids = new_info.new_other_ids;
+    let new_info = env.new_elders_info_without_candidate();
+    env.accumulate_offline(*env.candidate.public_id());
+    env.accumulate_start_dkg(&new_info);
```
maqi: what if not accumulate start_dkg, just accumulate offline ?

src/chain/shared_state.rs
madadam: Yes, but this same logic which exists in fleming too. I think the idea is that if we are already elder, we don't expect the shared state that comes from the genesis event to be different from the shared state we already have. We panic in tests when this happens because it indicates a bug in the code, but in production we just apply the new shared state as we should always trust whatever comes out of parsec.

src/chain/chain.rs
```diff
@@ -71,6 +70,10 @@ pub struct Chain {
     /// then members are disjoint. We are working around not having access to the prefix for the
     /// DkgResult but only the list of participants.
     new_section_bls_keys: BTreeMap<XorName, DkgResult>,
+    // The accumulated info during a split pfx change.
+    split_cache: Option<SplitCache>,
+    // members that we had before last split.
```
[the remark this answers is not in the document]
madadam: The comment is wrong, good catch.

src/chain/chain.rs
```diff
@@ -1431,18 +1432,26 @@ impl Chain {
...
     ///     - if the destination name is an entry in the routing table, returns it; otherwise
     ///     - returns the `N/3` closest members of the RT to the target
-    pub fn targets(&self, dst: &DstLocation) -> Result<(Vec<&P2pNode>, usize), RoutingError> {
-        let (best_section, dg_size) = match *dst {
-            DstLocation::Node(ref target_name) => {
+    pub fn targets(&self, dst: &DstLocation) -> Result<(Vec<P2pNode>, usize), RoutingError> {
+        if !self.is_self_elder() {
+            // We are not Elder - return all the elders of our section, so the message can be properly
+            // relayed through them.
+            let targets: Vec<_> = self.our_info().member_nodes().cloned().collect();
```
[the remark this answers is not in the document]
madadam: `our_info` returns `EldersInfo` which contains only elders, so yes.

src/states/approved_peer/mod.rs
```diff
@@ -51,6 +52,9 @@ use std::{
     net::SocketAddr,
 };
 
+// Send our knowledge in a similar speed as GOSSIP_TIMEOUT
+const KNOWLEDGE_TIMEOUT: Duration = Duration::from_secs(2);
```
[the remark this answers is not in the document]
madadam: Possibly, but this is how it is in fleming and I'd prefer not to change it in this PR. Also the plan is to eventually not use any timeout and make the `MemberKnowledge` event-based instead of time-based. But there are some corner cases that neat to be handled first. In any case, it's not in the scope of this PR anyway.

src/states/approved_peer/mod.rs
madadam: Actually it never returns empty vec, even if we are not among the signature targets. Will fix the comment.

src/states/approved_peer/mod.rs
madadam: Yeah, there is more that could be improved in this function, but it's probably out of scope for this PR.

src/states/approved_peer/mod.rs
madadam: No, `known_nodes` returns all nodes (elders or not) that we know of. So the comment is wrong. Will fix it.

madadam: Ah, but we only know about *elders* of the neighbour sections, so it isn't wrong.

src/states/approved_peer/tests/elder.rs
madadam: Not sure what you are asking. That we should have a test for that? If so, then yeah, maybe. But again, this would be out of scope of this PR.

src/chain/chain.rs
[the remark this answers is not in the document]
maqi: then seems the comment is bit incorrect? `return all the elders of our section`

src/states/approved_peer/mod.rs
maqi: I am just suggesting to change the function name  :)

src/states/approved_peer/mod.rs
maqi: hmm...  still worth to at least put a comment to clarify?

src/states/approved_peer/tests/elder.rs
maqi: it's just not clear to me that whether it is `accumulate_offline` or `accumulate_start_dkg` cause the node to be removed from members but still retain as an elder.
especially the next test shows that to be removed fron elders, there are extra steps after `accumulate_start_dkg`. (I thought)

madadam: No, I think it's accumulate Offline first, then accumulate DKG and then accumulate SectionInfo. Only then is the node no longer considered elder. But in any case, the logic of these tests is still the same as in fleming (only some type/variable names changed), so I don't think it's appropriate to change it here.

src/chain/chain.rs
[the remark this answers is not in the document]
madadam: I'm not sure I understand. What is incorrect about it? What would you suggest?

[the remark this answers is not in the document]
maqi: Ah, I mis-understood your previous comment. It's all right then.

PR #2072, opened 2020-03-17
Remove state machine - part 2

madadam:
This PR is a followup to #2071 . It merges the `BootstrappingPeer` and `JoiningPeer` states into one, leaving just two states now. The next PR will merge those two states and actually remove the state machine.


src/states/joining_peer.rs
```diff
@@ -254,51 +463,38 @@ impl Base for JoiningPeer {
...
+        // Ignore messages from peers we didn't send `BootstrapRequest` to.
+        if let Stage::Bootstrapping(stage) = &self.stage {
+            let sender = msg.src.to_sender_node(sender)?;
+
+            if !stage.pending_requests.contains(sender.peer_addr()) {
+                debug!(
+                    "{} - Ignoring message from unexpected peer: {}: {:?}",
+                    self, sender, msg,
+                );
+                self.network_service.disconnect(*sender.peer_addr());
```
maqi: it's not only `ignore messages` but also disconnect from it?

src/states/joining_peer.rs
```diff
@@ -373,30 +567,78 @@ impl Base for JoiningPeer {
...
+    }
+
+    fn is_joining(&self) -> bool {
+        matches!(self, Self::Joining(_))
+    }
+}
+
+struct BootstrappingStage {
+    // Using `FxHashSet` for deterministic iteration order.
+    pending_requests: FxHashSet<SocketAddr>,
```
maqi: this `pending_requests` is more like: peers that we have pending bootstrap requests to 
right?

madadam:
> Wondering why those tests for elder got removed. Thought we still need them?

They didn't. They just moved to `src/states/approved_peer/tests/elder.rs`

src/states/mod.rs
```diff
@@ -1,52 +1,44 @@
 // Copyright 2018 MaidSafe.net limited.
-//
+//─
```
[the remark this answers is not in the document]
madadam: Must got there by accident. Will remove.

src/states/joining_peer.rs
```diff
@@ -25,19 +24,24 @@ use crate::{
...
 use bytes::Bytes;
+use fxhash::FxHashSet;
 use std::{
+    collections::HashMap,
     fmt::{self, Display, Formatter},
     net::SocketAddr,
     time::Duration,
 };
 
+/// Time after which bootstrap is cancelled (and possibly retried).
```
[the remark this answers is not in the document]
madadam: BOOTSTRAP_TIMEOUT applies to the bootstrapping phase, that is when the node is trying to find a section to join. There is a separate timeout per each bootstrap request sent. JOIN_TIMEOUT applies to the join phase which is when it sent the JoinRequests and is waiting for being approved. This one is global.

maqi:
> > Wondering why those tests for elder got removed. Thought we still need them?
> 
> They didn't. They just moved to `src/states/approved_peer/tests/elder.rs`

Ah, I put the comment in wrong PR. It's in PR 2073 that seems the tests/elder.rs got removed totally.

src/states/joining_peer.rs
```diff
@@ -47,52 +51,67 @@ pub struct JoiningPeerDetails {
     pub network_cfg: NetworkParams,
     pub timer: Timer,
     pub rng: MainRng,
-    pub elders_info: EldersInfo,
-    pub relocate_payload: Option<RelocatePayload>,
 }
 
 // State of a node after bootstrapping, while joining a section
 pub struct JoiningPeer {
+    stage: Stage,
```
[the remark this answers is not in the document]
madadam: Yes, but that's the point. We are inverting the pattern we use - instead of having the top level `StateMachine` type which is just responsible for switching the states and the actual logic is implemented in the states, we will have one top level type where all the logic will be and it will contain this `Stage` enum which just contains the stuff that's different between the stages (I call them stages now instead of states which seems more appropriate).

src/states/joining_peer.rs
```diff
@@ -168,6 +206,141 @@ impl JoiningPeer {
...
+
+        let relocate_details = stage.relocate_details.take();
+        let destination = match &relocate_details {
+            Some(details) => *details.destination(),
+            None => *self.name(),
+        };
+        let old_full_id = self.full_id.clone();
+
+        // Use a name that will match the destination even after multiple splits
+        let extra_split_count = 3;
```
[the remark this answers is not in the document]
madadam: Well it's sort of already is, it just uses `let` instead of `const`. But again, this is one of those things that I just moved over from the previous code and did not change. For the sake of keeping the amount of change in the PR minimal, I'd prefer to not change stuff that's not necessary for the purpose of this PR, no matter how trivial that change might be.

src/states/joining_peer.rs
madadam: Yes, this is because of the implicit connection that quic-p2p does, to avoid keeping unneeded connections.

src/states/joining_peer.rs
madadam: Yes, but I find the name descriptive enough, given it scope. Calling it `peers_with_pending_requests` or something like that would be unnecessarily verbose.

madadam:
> Ah, I put the comment in wrong PR. It's in PR 2073 that seems the tests/elder.rs got removed totally.

Hm, OK. Will check it, because that should not have happened.

src/states/joining_peer.rs
maqi: Then, I mean shall the log mention the disconnection as well?

PR #2073, opened 2020-03-18
Improve log identifier

madadam:
Followup to #2072. Changes start at the"Implement thread-local log identifier" commit.

When running the integration tests, it's very useful to be able to tell from the log output which node each log line belongs to. To achieve this, we prefix each log line with a short identifier of the node. It looks like this:
```
Elder(2d390f..(001))
```
The way this was implemented before this PR was to have a `Display` impl in each node state which produces the identifier and then manually output `self` in every log statement:
```rust
debug!("{} - handle SectionInfo: {:?}", self, info)
``` 
This works, but has several issues:
- works only if `self` is a node state. If we need to log inside one of the other structs, we need to capture the ident before and pass it explicitly:
    ```rust
    let log_ident = format!("{}", self);
    self.parsec_map.handle_request(..., log_ident)
    ```
- interferes with the borrow checker - can only be used if `self` is not already mutably borrowed and requires ugly workarounds otherwise.
- is verbose and unergonomic

So this PR improves the situation by making the log ident implicit. It is set only once before handling any input (using `log_utils::set_ident`) and stored in a thread-local variable. Then it is automatically fetched from this variable in each log statement (using provided custom logging macros: `error!`, `info!`, `debug!`, ...). This solves all the above issues.

### DIFF
--- a/src/log_utils.rs
+++ b/src/log_utils.rs
@@ -1,0 +1,169 @@
+// Copyright 2020 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+//! Logging with log identifier
+//!
+//! This module implement logging macros that automatically prefix each output with a custom log
+//! identifier. To log identifier is scoped to the current thread.
+//!
+//! # Example
+//!
+//! ```ignore
+//! // No ident set so this outputs just "foo"
+//! trace!("foo");
+//!
+//! // Set the ident
+//! let guard = log_utils::set_ident(|buffer| {
+//!     use std::fmt::Write;
+//!     write!(buffer, "MY IDENT: ")
+//! });
+//!
+//! // This now outputs "MY IDENT: bar"
+//! trace!("bar");
+//!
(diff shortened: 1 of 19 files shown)


src/log_utils.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+// Internal function but it needs to be `pub` so it can be used in the macros.
+#[doc(hidden)]
+pub fn with_ident<F: FnOnce(&str)>(f: F) {
+    LOG_IDENT.with(|ident| f(&*ident.borrow()))
+}
+
+// NOTE: an alternative to having custom logging macros would to to override the logger format and
```
maqi: `would to to` -> `would be to` ?

src/states/common/base.rs
```diff
@@ -70,6 +64,8 @@ pub trait Base: Display {
     fn unhandled_message(&mut self, sender: Option<SocketAddr>, msg: Message, msg_bytes: Bytes);
 
     fn handle_action(&mut self, action: Action, outbox: &mut dyn EventBox) -> Transition {
+        let _log_ident = self.set_log_ident();
```
maqi: is this ident only works within the scope of `_log_ident` ?
which actually means for each incoming action/network_event, this `_log_ident` is a new instance?
and, what for `handle_message` ?

madadam: Yes, it's a new instance every time, but it's a zero-sized type so there is no overhead. The actual buffer which stores the log ident (which is kept in a thread-local variable) is reused, so there is no allocation after the first couple of times.

> and, what for handle_message ?

`handle_message` is called in `handle_network_event` which also sets the log ident.

madadam:
> The PR shows that src/states/tests/elder.rs got removed totally?

No, it's just the accidentally duplicated files that were removed. The tests are in `src/states/approved_peer/tests` now.

PR #2077, opened 2020-04-02
Treat infants differently from adults

### LINKED ISSUE #1953: Treat infant differently for split etc
With #1929 we did not implement treating infant differently.
Decide what is the right thing and do the right thing with infant.

madadam:
This PR makes infants different from adults in these two ways:

1. Infant joining or leaving a section no longer causes ageing of the other section nodes
2. Infants are not taken into account when deciding splits

This change makes the network more resilient against [sybil attacks](https://en.wikipedia.org/wiki/Sybil_attack) because it makes it much more difficult for an attacker to attempt to take over a section by spawning large number of nodes under their control. 

The bulk of this PR however is in the test code, namely in the utilities to trigger splits. This now requires more work as it's no longer sufficient to just add new nodes until the section splits.

This PR also contains two seemingly unrelated changes:

1. It fixes a bug that was probably introduced in one of the recent PRs where adults would not handle `MemberKnowledge` messages. This caused issues when a section split and none of the new elders were elder before the split. Then the new to-be-elders would try to send `MemberKnowledge` to the old elders in order to get parsec gossip back (to make them see the event which promoted them), but the old elders would ignore the message thus stalling the section.

2. It introduces a new shared value called `section_version` which is a number that gets incremented every time a node joins or leaves the section. This number is then assigned to every node in the section when it joins and serves as additional differentiation when deciding which node to promote in case there are multiple candidates with the same age. Previously we would accidentally prefer nodes with smaller names (according to the xor metric) because the member infos are internally stored in a container sorted by the name. So with this change the process is more "just" as it prefers nodes that joined earlier. The main reason for this change though is to make the split triggering test utility work. 

Finally this PR contains some minor refactoring of the test code.

Closes #1953

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -524,6 +524,7 @@ impl Chain {
             }
             Entry::Vacant(entry) => {
                 // Node joining for the first time.
+
                 let _ = entry.insert(MemberInfo::new(
                     age,
                     p2p_node.clone(),
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -421,6 +421,7 @@ impl Chain {
             details_to_add.push(PartialRelocateDetails {
                 pub_id: *member_info.p2p_node.public_id(),
                 destination,
+                // TODO: why the +1 ?
                 age: member_info.age() + 1,
             })
         }
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -371,6 +371,8 @@ impl Chain {
         let our_section_size = self.state.our_joined_members().count();
         let safe_section_size = self.safe_section_size();
 
+        // As a measure against sybil attacks, we don't increment the age counters on infant churn
+        // once we reach safe_section_size.
         if our_section_size >= safe_section_size
             && self
                 .state
@@ -378,15 +380,11 @@ impl Chain {
                 .map(|persona| persona == MemberPersona::Infant)
                 .unwrap_or(true)
         {
-            // FIXME: skipping infants churn for ageing breaks tests for node ageing, as once a
-            // section reaches a safe size, nodes stop ageing at all, because all churn in tests
-            // is Infant churn.
-            // Temporarily ignore until we either find a better way of preventing churn spam,
-            // or we change the tests to provide some Adult churn at all times.
             trace!(
-                "FIXME: should do nothing for infants and unknown nodes {:?}",
-                trigger_node
+                "Not incrementing age counters on infant churn (section size: {})",
+                our_section_size,
             );
+            return;
         }
 
         let our_prefix = *self.state.our_prefix();
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -809,12 +809,11 @@ impl Chain {
         self.state.our_info().member_nodes()
     }
 
-    fn elders_and_adults(&self) -> impl Iterator<Item = &PublicId> {
+    /// Returns joined adults and elders from our section.
+    fn our_mature_members(&self) -> impl Iterator<Item = &PublicId> {
         self.state
             .our_joined_members()
-            // FIXME: we temporarily treat all section
-            // members as Adults
-            //.filter(|(_, info)| info.is_mature())
+            .filter(|(_, info)| info.is_mature())
             .map(|(_, info)| info.p2p_node.public_id())
     }
 
@@ -1294,7 +1293,7 @@ impl Chain {
         let our_name = self.our_id.name();
         let our_prefix_bit_count = self.our_prefix().bit_count();
         let (our_new_size, sibling_new_size) = self
-            .elders_and_adults()
+            .our_mature_members()
             .map(|id| our_name.common_prefix(id.name()) > our_prefix_bit_count)
             .fold((0, 0), |(ours, siblings), is_our_prefix| {
                 if is_our_prefix {
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -510,6 +510,9 @@ impl Chain {
                     // new age to max(old_age, new_age)
                     entry.get_mut().state = MemberState::Joined;
                     entry.get_mut().set_age(age);
+                    entry.get_mut().section_version = self.state.section_version;
+
+                    self.state.increment_section_version();
                 } else {
                     // Node already joined - this should not happen.
                     log_or_panic!(
@@ -521,7 +524,12 @@ impl Chain {
             }
             Entry::Vacant(entry) => {
                 // Node joining for the first time.
-                let _ = entry.insert(MemberInfo::new(age, p2p_node.clone()));
+                let _ = entry.insert(MemberInfo::new(
+                    age,
+                    p2p_node.clone(),
+                    self.state.section_version,
+                ));
+                self.state.increment_section_version();
             }
         }
     }
@@ -540,11 +548,16 @@ impl Chain {
             .filter(|info| info.state != MemberState::Left)
         {
             let member_state = info.state;
+            let member_addr = *info.p2p_node.peer_addr();
+
             info.state = MemberState::Left;
+
             self.state
                 .relocate_queue
                 .retain(|details| &details.pub_id != pub_id);
-            (Some(*info.p2p_node.peer_addr()), member_state)
+            self.state.increment_section_version();
+
+            (Some(member_addr), member_state)
         } else {
             log_or_panic!(
                 log::Level::Error,
@@ -809,7 +822,7 @@ impl Chain {
         let mut elders: BTreeMap<_, _> = self
             .state
             .our_joined_members()
-            .sorted_by(|&(_, info1), &(_, info2)| Ord::cmp(&info2.age_counter, &info1.age_counter))
+            .sorted_by(|(_, info1), (_, info2)| self.cmp_elder_candidates(info1, info2))
             .into_iter()
             .map(|(name, info)| (*name, info.p2p_node.clone()))
             .take(self.elder_size())
@@ -831,6 +844,15 @@ impl Chain {
         );
 
         elders
+    }
+
+    // Compare candidates for the next elders. The one comparing `Less` is more likely to become
+    // elder.
+    fn cmp_elder_candidates(&self, lhs: &MemberInfo, rhs: &MemberInfo) -> Ordering {
+        // Older nodes are preferred. In case of a tie, nodes joining earlier are preferred.
+        rhs.age_counter
+            .cmp(&lhs.age_counter)
+            .then(lhs.section_version.cmp(&rhs.section_version))
     }
 
     fn eldest_members_matching_prefix(
--- a/src/chain/member_info.rs
+++ b/src/chain/member_info.rs
@@ -54,15 +54,18 @@ pub struct MemberInfo {
     pub age_counter: AgeCounter,
     pub state: MemberState,
     pub p2p_node: P2pNode,
+    // Section version at the time this node joined.
+    pub section_version: u64,
 }
 
 impl MemberInfo {
     /// Create new `MemberInfo` in the `Joined` state.
-    pub fn new(age: u8, p2p_node: P2pNode) -> Self {
+    pub fn new(age: u8, p2p_node: P2pNode, section_version: u64) -> Self {
         Self {
             age_counter: AgeCounter::from_age(age),
             state: MemberState::Joined,
             p2p_node,
+            section_version,
         }
     }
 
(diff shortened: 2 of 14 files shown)


tests/mock_network/mod.rs
```diff
@@ -39,57 +39,12 @@ fn test_nodes(percentage_size: usize) {
     let size = LOWERED_ELDER_SIZE * percentage_size / 100;
     let env = Environment::new(NetworkParams {
         elder_size: LOWERED_ELDER_SIZE,
-        safe_section_size: LOWERED_ELDER_SIZE,
+        safe_section_size: LOWERED_ELDER_SIZE + 1,
```
oetyng, line 42: Why is it 5 now instead of 3?

tests/mock_network/mod.rs
```diff
@@ -23,7 +23,7 @@ use routing::{
 };
 use std::collections::BTreeMap;
 
-pub const LOWERED_ELDER_SIZE: usize = 3;
+pub const LOWERED_ELDER_SIZE: usize = 4;
```
oetyng, line 26: Why is it 4 now instead of 3?

tests/mock_network/mod.rs
```diff
@@ -219,64 +174,52 @@ fn multiple_joining_nodes() {
...
+    for node in nodes.iter() {
+        assert!(node.inner.is_approved());
     }
 }
 
+#[test]
+fn single_split() {
+    let env = Environment::new(NetworkParams {
+        elder_size: LOWERED_ELDER_SIZE,
+        safe_section_size: LOWERED_ELDER_SIZE + 3,
```
oetyng, line 210: Why is it 7? (What is the context here in the test)

tests/mock_network/mod.rs
```diff
@@ -219,64 +174,52 @@ fn multiple_joining_nodes() {
...
+    let mut nodes = vec![];
+    trigger_split(&env, &mut nodes, &Prefix::default());
+}
+
 #[test]
 fn multi_split() {
     let env = Environment::new(NetworkParams {
         elder_size: LOWERED_ELDER_SIZE,
-        safe_section_size: LOWERED_ELDER_SIZE,
+        safe_section_size: LOWERED_ELDER_SIZE + 1,
```
oetyng, line 220: Why is it 5? (What is the context here in the test)

src/chain/chain.rs
```diff
@@ -371,22 +371,20 @@ impl Chain {
         let our_section_size = self.state.our_joined_members().count();
         let safe_section_size = self.safe_section_size();
 
+        // As a measure against sybil attacks, we don't increment the age counters on infant churn
+        // once we reach safe_section_size.
         if our_section_size >= safe_section_size
```
maqi, line 376: just to confirm that:
1, when `section_size < safe_section_size`, infant_churn causes age increment
2, non-infant churn always causes aging
right ?

src/chain/member_info.rs
```diff
@@ -54,15 +54,18 @@ pub struct MemberInfo {
...
     pub state: MemberState,
     pub p2p_node: P2pNode,
+    // Section version at the time this node joined.
+    pub section_version: u64,
 }
 
 impl MemberInfo {
     /// Create new `MemberInfo` in the `Joined` state.
-    pub fn new(age: u8, p2p_node: P2pNode) -> Self {
+    pub fn new(age: u8, p2p_node: P2pNode, section_version: u64) -> Self {
```
maqi, line 63: this gives me a slight confusion that this `section_version` is our knowledge of that member's current section version.
However, I don't get a better name so far.  :(

src/chain/shared_state.rs
```diff
@@ -310,6 +315,10 @@ impl SharedState {
             index
         }
     }
+
+    pub fn increment_section_version(&mut self) {
+        self.section_version = self.section_version.wrapping_add(1);
```
maqi, line 320: will `saturating_add` be more suitable ?

src/node/mod.rs
```diff
@@ -1037,9 +1037,10 @@ impl Node {
     }
 
     /// Returns the age counter of the given node if it is member of the same section as this node,
-    /// `None` otherwise.
+    /// and this node is elder. Returns `None` otherwise.
```
maqi, line 1040: `if it is member of the same section as this node, and this node is elder.`  => `if it is an elder of the same section as own.`
btw. I think then better to update the function name to reflect this scope change?

src/xor_space/prefix.rs
```diff
@@ -196,6 +205,24 @@ impl<T: Clone + Copy + Default + Binary + Xorable> Prefix<T> {
             *self
         }
     }
+
+    /// Returns the ancestors of this prefix that has the given bit count.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `bit_count` is not less than the bit count if this prefix.
```
maqi, line 213: if => of

src/xor_space/prefix.rs
```diff
@@ -120,6 +120,15 @@ impl<T: Clone + Copy + Default + Binary + Xorable> Prefix<T> {
         }
     }
 
+    /// Compares the prefixes using breadth-first order. That is, shorter prefixes are ordered
+    /// before longer. This is in contrast with the default `Ord` impl of `Prefix` uses depth-first
+    /// order.
+    pub fn cmp_breadth_first(&self, other: &Self) -> Ordering {
```
maqi, line 126: just curious, why this different ordering is needed?

src/chain/chain.rs
```diff
@@ -798,20 +811,19 @@ impl Chain {
         self.state.our_info().member_nodes()
     }
 
-    fn elders_and_adults(&self) -> impl Iterator<Item = &PublicId> {
+    /// Returns joined adults and elders from our section.
```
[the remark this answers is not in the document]
madadam: A node can leave the section and then rejoin which would maintain its original age (although it's not fully implemented yet). To support this, we keep the info about the node, just mark it as `Left`. So yes, there can be nodes that are "not joined".

src/chain/chain.rs
```diff
@@ -798,20 +811,19 @@ impl Chain {
...
+    fn our_mature_members(&self) -> impl Iterator<Item = &PublicId> {
         self.state
             .our_joined_members()
```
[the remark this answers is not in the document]
madadam: Same explanation.

src/chain/chain.rs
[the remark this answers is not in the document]
madadam: Also, there is a third state - `Relocating`. That is for nodes that's already been decided to be relocated, but have not relocate yet for various reasons. A `Relocating` node is not considered as "joined".

src/chain/chain.rs
```diff
@@ -835,6 +847,15 @@ impl Chain {
         elders
     }
 
+    // Compare candidates for the next elders. The one comparing `Less` is more likely to become
+    // elder.
+    fn cmp_elder_candidates(&self, lhs: &MemberInfo, rhs: &MemberInfo) -> Ordering {
+        // Older nodes are preferred. In case of a tie, nodes joining earlier are preferred.
+        rhs.age_counter
+            .cmp(&lhs.age_counter)
+            .then(lhs.section_version.cmp(&rhs.section_version))
```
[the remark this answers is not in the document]
madadam: Yeah, I remember us discussing this at some point, just not sure what it was exactly anymore. So chance is this will become useful for other stuff too.

tests/mock_network/mod.rs
madadam: So `elder_size` changed from 3 to 4 because 4 is the smallest number where we can still lose 1 elder and still be able to reach consensus on it (because 3 is more than 2/3 of 4) which is what we need to do now (to trigger splits). I changed `safe_section_size` to be one more than `elder_size` just so it's not the same value. It would probably still work, but would be less realistic.

tests/mock_network/mod.rs
madadam: https://github.com/maidsafe/routing/pull/2077#discussion_r402318285

tests/mock_network/node_ageing.rs
```diff
@@ -334,36 +291,77 @@ fn node_relocated(
...
+        source_prefix,
+        target_prefix
+    );
+}
+
+// Returns the age counter of the node with the given name.
+fn node_age_counter(nodes: &[TestNode], name: &XorName) -> u32 {
+    if let Some(counter) = nodes
+        .iter()
+        .filter_map(|node| node.inner.member_age_counter(name))
```
maqi, line 356: I thought a node's age counter shall be from its own perspective.
However this code block seems scanning through all members' perspective?

tests/mock_network/mod.rs
madadam: So 7 is the smallest number where a certain bug was exposed  - the one caused by adults not handling the `MemberKnowledge` message. I think it's because it's the smallest number where when a split happens, the elders post-split might be completely different nodes than the elders pre-split which is exactly the situation which triggered the bug.

tests/mock_network/mod.rs
madadam: https://github.com/maidsafe/routing/pull/2077#discussion_r402318285

src/chain/chain.rs
madadam: That's right.

src/chain/member_info.rs
madadam: It's not the section_version we think the node has, rather it's the section_version at which the node joined which is something the whole section agrees on.

src/chain/shared_state.rs
madadam: I wanted this number to keep changing even when we reach the max value. That would for a brief time break the logic of preferring nodes that joined earlier to those that joined later when selecting who to promote, but that should not be a big problem. That said, this number wraps around after ~1 quintillion churn events, so we won't hit it any time soon :)

src/node/mod.rs
madadam: OK, that might be awkwardly worded, but what it tried to say is that if **we** are elder, this returns the age counter of that node, regardless of whether they are elder not not, otherwise it returns `None`. I'll try to reword it.

src/xor_space/prefix.rs
madadam: This is because the function `create_connected_nodes_until_split` takes as the input a list of prefix lengths. This is then turned into actual prefixes which we want the network to eventually consist of and that is turned into a sequence of prefixes to split in order to get there. And it's this final sequence that needs to be ordered using this breadth-first order. I was considering changing the `Ord` impl instead, but eventually decided against it as I wasn't sure it would not break something.

tests/mock_network/node_ageing.rs
madadam: No, it's from the section perspective, because only elders store the age counters.

src/chain/member_info.rs
maqi: Yeah, I know what it `want to do`.
it's just, within the chain::shared_state, there is a member field called the same name `section_version`. This causes some bit confustion to me at the beginning.

madadam: OK. It should be something like `section_version_at_the_time_of_join`, but that seems too verbose :)

maqi: yeah, I agree. I don't get a better name either. :(

tests/mock_network/mod.rs
oetyng: OK. I think these things are good to have as comments in there when some magic number is used/updated. It becomes very obscure for someone looking at it otherwise.

tests/mock_network/mod.rs
oetyng: OK. I think these things are good to have as comments in there when some magic number is used/updated. It becomes very obscure for someone looking at it otherwise.

src/chain/member_info.rs
oetyng: `joined_at_version: SectionVersion`, where `type SectionVersion = u64`

src/chain/chain.rs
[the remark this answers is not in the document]
oetyng: Hm. But `joined` means it _has_ at some point joined. It doesn't mean that it "is" joined (since there is no such thing as "being" joined), and that it leaves doesn't mean that it didn't join, so "not joined" is a grammatical hick-up. The wording is unfortunate.
We're describing a **current** state with a mix of _past tense_ and _present tense_.
`Joined` is past tense, `Left` is past tense. `Relocating` is present.
For describing current state (which this does), it would be better to stay consistent with that.

Suggestion for a more consistent and straight forward language (no past tense used):
`Member` replaces `Joined`.
`Away` replaces `Left`. `Away` describes a current state, and would indicate that we consider the possibility that it will come back - more so than `Left` I would say. It is not a `Member` while `Away`, per definition.
`Relocating` would stay as is. It also per definition means it is not a `Member`.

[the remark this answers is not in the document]
madadam: Those are good suggestions, but I'd probably make them in a different PR as this one has different focus.

PR #2079, opened 2020-04-14
Test improvements

madadam:
This PR changes the testing strategy. Instead of using `poll_and_resend` which kept polling the network until some mythical "idle" state, we now have `poll_until` where the termination condition is specified explicitly. This condition differs from test to test - for example: "poll until node joined", "poll until node becomes elder", "poll until section splits", "poll until a message gets delivered", and so on.

Also the `verify_section_invariants_between_nodes` function was removed because it was based on unrealistic assumptions that all the nodes in the network can ever get to the same state at the same time. Instead of this, we now check only that the nodes got to a given state at some point, but never that they all are in it at the same time.

These changes uncovered a bug in the `JoinRequest` handling logic which this PR fixes.

This PR also removes some no longer needed hacks and workarounds resulting in slightly cleaner code.

Note this PR contains almost the minimal amount of change needed to achieve these goals. There is definitely room for further improvements, for example by making the tests more event-driven, etc...

### DIFF
--- a/src/node/stage/approved.rs
+++ b/src/node/stage/approved.rs
@@ -1582,9 +1582,6 @@ impl Approved {
     // Send message over the network.
     pub fn send_signed_message(&mut self, core: &mut Core, msg: &MessageWithBytes) -> Result<()> {
         let (targets, dg_size) = self.chain.targets(msg.message_dst())?;
-
-        trace!("Sending {:?} via targets {:?}", msg, targets);
-
         let targets: Vec<_> = targets
             .into_iter()
             .filter(|p2p_node| {
@@ -1592,9 +1589,15 @@ impl Approved {
                     .filter_outgoing(msg, p2p_node.public_id())
                     .is_new()
             })
-            .map(|node| *node.peer_addr())
             .collect();
 
+        if targets.is_empty() {
+            return Ok(());
+        }
+
+        trace!("Sending {:?} via targets {:?}", msg, targets);
+
+        let targets: Vec<_> = targets.into_iter().map(|node| *node.peer_addr()).collect();
         let cheap_bytes_clone = msg.full_bytes().clone();
         core.send_message_to_targets(&targets, dg_size, cheap_bytes_clone);
 
--- a/src/node/stage/approved.rs
+++ b/src/node/stage/approved.rs
@@ -349,8 +349,7 @@ impl Approved {
         }
 
         self.gen_pfx_info = gen_pfx_info.clone();
-        self.parsec_map
-            .init(&mut core.rng, core.full_id.clone(), &self.gen_pfx_info);
+        self.init_parsec(core);
         self.chain = Chain::new(self.chain.network_params(), *core.id(), gen_pfx_info, None);
 
         Ok(())
--- a/src/node/stage/approved.rs
+++ b/src/node/stage/approved.rs
@@ -248,14 +248,13 @@ impl Approved {
             | Variant::ParsecResponse(..)
             | Variant::Bounce { .. } => true,
 
-            Variant::NeighbourInfo(_)
-            | Variant::UserMessage(_)
-            | Variant::AckMessage { .. }
-            | Variant::JoinRequest(_) => self.chain.is_self_elder(),
+            Variant::JoinRequest(req) => self.should_handle_join_request(req),
 
-            Variant::GenesisUpdate(info) => {
-                !self.chain.is_self_elder() && self.is_genesis_update_new(info)
+            Variant::NeighbourInfo(_) | Variant::UserMessage(_) | Variant::AckMessage { .. } => {
+                self.chain.is_self_elder()
             }
+
+            Variant::GenesisUpdate(info) => self.should_handle_genesis_update(info),
             Variant::Relocate(_) => !self.chain.is_self_elder(),
 
             Variant::MessageSignature(accumulating_msg) => {
@@ -266,9 +265,7 @@ impl Approved {
                     | Variant::AckMessage { .. }
                     | Variant::Relocate(_) => true,
 
-                    Variant::GenesisUpdate(info) => {
-                        !self.chain.is_self_elder() && self.is_genesis_update_new(info)
-                    }
+                    Variant::GenesisUpdate(info) => self.should_handle_genesis_update(info),
 
                     // These variants are not be signature-accumulated
                     Variant::MessageSignature(_)
@@ -343,10 +340,6 @@ impl Approved {
         gen_pfx_info: GenesisPfxInfo,
     ) -> Result<()> {
         info!("Received GenesisUpdate: {:?}", gen_pfx_info);
-
-        if !self.is_genesis_update_new(&gen_pfx_info) {
-            return Ok(());
-        }
 
         self.gen_pfx_info = gen_pfx_info.clone();
         self.init_parsec(core);
@@ -465,6 +458,7 @@ impl Approved {
 
         if join_request.elders_version < self.chain.our_info().version() {
             self.resend_bootstrap_response_join(core, &p2p_node);
+            return;
         }
 
         let pub_id = *p2p_node.public_id();
@@ -603,13 +597,10 @@ impl Approved {
         msg: Message,
         msg_bytes: Bytes,
     ) {
-        let bounce = match msg.variant {
-            Variant::MessageSignature(_) => true,
+        let bounce = match &msg.variant {
+            Variant::MessageSignature(_) | Variant::JoinRequest(_) => true,
             Variant::Relocate(_) if self.chain.is_self_elder() => true,
-            Variant::JoinRequest(_)
-            | Variant::NeighbourInfo(_)
-            | Variant::UserMessage(_)
-            | Variant::AckMessage { .. }
+            Variant::NeighbourInfo(_) | Variant::UserMessage(_) | Variant::AckMessage { .. }
                 if !self.chain.is_self_elder() =>
             {
                 true
@@ -1673,8 +1664,15 @@ impl Approved {
     ////////////////////////////////////////////////////////////////////////////
 
     // Ignore stale GenesisUpdates
-    fn is_genesis_update_new(&self, gen_pfx_info: &GenesisPfxInfo) -> bool {
-        gen_pfx_info.parsec_version > self.gen_pfx_info.parsec_version
+    fn should_handle_genesis_update(&self, gen_pfx_info: &GenesisPfxInfo) -> bool {
+        !self.chain.is_self_elder()
+            && gen_pfx_info.parsec_version > self.gen_pfx_info.parsec_version
+    }
+
+    // Ignore `JoinRequest` if we are not elder unless the join request is outdated in which case we
+    // reply with `BootstrapResponse::Join` with the up-to-date info (see `handle_join_request`).
+    fn should_handle_join_request(&self, req: &JoinRequest) -> bool {
+        self.chain.is_self_elder() || req.elders_version < self.chain.our_info().version()
     }
 
     // Connect to all elders from our section or neighbour sections that we are not yet connected
--- a/src/node/stage/approved.rs
+++ b/src/node/stage/approved.rs
@@ -341,6 +341,8 @@ impl Approved {
     ) -> Result<()> {
         info!("Received GenesisUpdate: {:?}", gen_pfx_info);
 
+        core.msg_filter.reset();
+
         self.gen_pfx_info = gen_pfx_info.clone();
         self.init_parsec(core);
         self.chain = Chain::new(self.chain.network_params(), *core.id(), gen_pfx_info, None);
@@ -994,6 +996,8 @@ impl Approved {
         let info_version = elders_info.version();
         let is_elder = elders_info.is_member(core.id());
         let is_split = info_prefix.is_extension_of(&old_pfx);
+
+        core.msg_filter.reset();
 
         if was_elder || is_elder {
             info!("handle SectionInfo: {:?}", elders_info);
--- a/src/relocation.rs
+++ b/src/relocation.rs
@@ -194,25 +194,10 @@ mod overrides {
         /// relocated from that prefix will be relocated to the given destination.
         /// The override applies only to the exact prefix, not to its parents / children.
         pub fn set(&mut self, src_prefix: Prefix<XorName>, dst: XorName) {
-            let rc = if self.prefixes.insert(src_prefix) {
-                1
-            } else {
-                0
-            };
+            let _ = self.prefixes.insert(src_prefix);
 
             OVERRIDES.with(|map| {
-                let _ = map
-                    .borrow_mut()
-                    .entry(src_prefix)
-                    .and_modify(|info| {
-                        info.next = dst;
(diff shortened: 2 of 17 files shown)


tests/mock_network/messages.rs
```diff
@@ -73,69 +63,58 @@ fn send_and_receive() {
...
-
-    assert!(request_received_count >= quorum);
+    // For all expected recipients, poll until it receives the request message and send
+    // the response
+    let expected_req_recipients: Vec<_> = expected_recipients(&nodes, &req_dst).collect();
+    assert!(expected_req_recipients.len() >= quorum);
 
-    let _ = poll_all(&mut nodes);
+    for index in expected_req_recipients {
+        poll_until(&env, &mut nodes, |nodes| {
```
maqi, line 82: I think when execute the first iteration of poll_until `index = 0`, there will be high chance that every nodes got request received and responded.
Though it shall be no-harm for the later on iterations, it feels bit un-efficient?

tests/mock_network/messages.rs
```diff
@@ -73,69 +63,58 @@ fn send_and_receive() {
...
+        if let Err(err) = nodes[index]
+            .inner
+            .send_message(res_src, res_dst, res_content.clone())
+        {
+            trace!("Failed to send message: {:?}", err);
         }
     }
 
-    assert_eq!(response_received_count, 1);
+    // Poll until the response is delivered.
```
maqi, line 97: Poll until the sender received the response ?

tests/mock_network/messages.rs
```diff
@@ -73,69 +63,58 @@ fn send_and_receive() {
...
+}
+
+fn expected_recipients<'a>(
+    nodes: &'a [TestNode],
+    dst: &'a DstLocation,
+) -> impl Iterator<Item = usize> + 'a {
+    nodes
+        .iter()
+        .enumerate()
+        .filter(move |(_, node)| node.inner.is_elder() && node.inner.in_dst_location(dst))
```
maqi, line 110: just to ensure there is no test case that a message's expected recipient is a non-elder, i.e. all messages expected (or within the tests only care about) to be exchanged among elders?

tests/mock_network/messages.rs
```diff
@@ -73,69 +63,58 @@ fn send_and_receive() {
...
+        .map(|(index, _)| index)
+}
+
+fn message_received(node: &TestNode, expected_content: &[u8]) -> bool {
+    if let Some(Event::MessageReceived { content, .. }) = node.try_recv_event() {
```
maqi, line 115: just to ensure that membership change events (nodes joined or left) won't cause extra Event notification from this channel

tests/mock_network/mod.rs
```diff
@@ -433,96 +426,118 @@ fn carry_out_parsec_pruning() {
...
+    add_node_to_section(&env, &mut nodes, &Prefix::default());
+    poll_until(&env, &mut nodes, |nodes| {
+        node_joined(nodes, nodes.len() - 1)
+    });
+    let new_id = *nodes.last().unwrap().id();
+
+    nodes.push(TestNode::resume(paused_state));
+
+    // If the paused node is elder, verify it caught up to the new node joining.
+    if nodes.last().unwrap().inner.is_elder() {
```
maqi, line 490: is it guaranteed this situation will happen during the test?
I mean, maybe we shall ensure the paused node is the last to ensure the check of caught up to the new node will be properly always?

tests/mock_network/mod.rs
```diff
@@ -433,96 +426,118 @@ fn carry_out_parsec_pruning() {
...
-    let paused_id = *nodes[index].id();
+    let id = *nodes[index].id();
     let state = nodes.remove(index).inner.pause().unwrap();
 
-    // Verify the other nodes do not see the node as going offline.
-    poll_and_resend(&mut nodes);
+    // Poll the network for a while and verify the other nodes do not see the node as going offline.
+    let start_time = FakeClock::now();
+    poll_until(&env, nodes, |_| {
+        start_time.elapsed() >= Duration::from_secs(60)
```
maqi, line 534: where this magic 60s come from?

tests/mock_network/utils.rs
```diff
@@ -94,34 +90,15 @@ impl TestNode {
...
-    pub fn env(&self) -> &Environment {
-        &self.env
-    }
-
     pub fn poll(&mut self) -> bool {
         let mut result = false;
 
-        // Exhaust all the events/actions from the channels but return true only if at least one of
-        // those events/actions are considered as handled (that is there is at least one
-        // non-timeout).
```
maqi, line 114: why the comment removed?
any change of the situation?

tests/mock_network/utils.rs
```diff
@@ -174,127 +151,168 @@ impl<'a> TestNodeBuilder<'a> {
...
 
-        if let Some(continue_predicate) = options.continue_predicate.as_ref() {
-            if continue_predicate(nodes) {
-                // Advance time in case the predicate is timeout-triggered.
-                advance_time(time_step);
-                continue;
+    let mut pending = nodes
+        .iter()
+        .filter(|node| {
+            if sub_prefix0.matches(node.name()) && *node.our_prefix() != sub_prefix0 {
```
maqi, line 273: wondering why not just check:
```rust
prefix.matches(node.name()) && 
   !(node.our_prefix() == sub_prefix0 || node.our_prefix() == sub_prefix1)
```

tests/mock_network/utils.rs
```diff
@@ -460,18 +478,67 @@ pub fn add_mature_nodes(
...
-        add_node_to_section(env, nodes, &prefix);
+    add_nodes_to_section_and_poll(env, nodes, &prefix, temp_count);
+
+    // Of the remaining nodes, `count0` goes to the 0-ending sub-prefix and `count1` to the
+    // 1-ending.
+    info!("Adding {} final nodes", count0 + count1);
+    add_nodes_to_subsections_and_poll(env, nodes, &prefix, count0, count1);
+
+    // Remove 16 mature nodes to trigger 16 age increments.
+    info!("Removing {} mature nodes", remove_count);
```
maqi, line 489: assertion remove_count equals to 16 ?

tests/mock_network/utils.rs
```diff
@@ -491,60 +558,30 @@ pub fn add_mature_nodes(
             add_node_to_section(env, nodes, &sub_prefix0);
             remaining0 -= 1;
         }
-    }
 
-    poll_and_resend(nodes);
-
-    // Remove 16 mature nodes to trigger 16 age increments.
```
maqi, line 498: assert remove_count equals to 16 ?

tests/mock_network/utils.rs
```diff
@@ -491,60 +558,30 @@ pub fn add_mature_nodes(
...
 
 // -----  Small misc functions  -----
 
 /// Sorts the given nodes by their distance to `name`. Note that this will call the `name()`
 /// function on them which causes polling, so it calls `poll_all` to make sure that all other
 /// events have been processed before sorting.
-pub fn sort_nodes_by_distance_to(nodes: &mut [TestNode], name: &XorName) {
-    let _ = poll_all(nodes); // Poll
+pub fn sort_nodes_by_distance_to(env: &Environment, nodes: &mut [TestNode], name: &XorName) {
+    poll_all(env, nodes); // Poll
```
maqi, line 584: the comment tail could be removed

tests/mock_network/utils.rs
```diff
@@ -690,109 +715,20 @@ pub fn verify_section_invariants_for_node(node: &TestNode, elder_size: usize) {
...
     }
 }
 
-pub fn verify_section_invariants_for_nodes(nodes: &[TestNode], elder_size: usize) {
-    for node in nodes.iter() {
-        verify_section_invariants_for_node(node, elder_size);
-    }
-}
-
-pub fn verify_section_invariants_between_nodes(nodes: &[TestNode]) {
```
maqi, line 708: do we still need to verify the consistency among the nodes?

tests/mock_network/utils.rs
```diff
@@ -690,109 +715,20 @@ pub fn verify_section_invariants_for_node(node: &TestNode, elder_size: usize) {
...
-                continue;
-            }
-            let _ = sections.insert(*prefix, our_info);
-        }
+pub fn verify_invariants_for_nodes(env: &Environment, nodes: &[TestNode]) {
+    for node in nodes {
+        verify_invariants_for_node(env, node);
     }
-
-    // check that prefixes are disjoint
```
maqi, line 755: do we still need to check the prefixes are disjoint?

src/node/stage/approved.rs
```diff
@@ -344,13 +341,8 @@ impl Approved {
     ) -> Result<()> {
         info!("Received GenesisUpdate: {:?}", gen_pfx_info);
 
-        if !self.is_genesis_update_new(&gen_pfx_info) {
```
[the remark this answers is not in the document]
madadam: It's not needed because this function is not called if `should_handle_message` returns false.

src/relocation.rs
```diff
@@ -237,15 +222,8 @@ mod overrides {
...
-                        Entry::Occupied(mut entry) => {
-                            if entry.get().rc <= 1 {
-                                let _ = entry.remove();
-                            } else {
-                                entry.get_mut().rc -= 1;
-                            }
-                        }
-                        Entry::Vacant(_) => panic!("Unexpected missing overrides for {:?}", prefix),
+                    if let Some(info) = map.get_mut(&prefix) {
+                        info.next = None;
```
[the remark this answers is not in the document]
madadam: Yes, because we no longer clear the already assigned mappings of original to overridden names. We only clear the future overrides (that a node from some given prefix P gets relocated to a name from some other given prefix Q). Because of that we no longer need to track the reference counter.

tests/mock_network/churn.rs
```diff
@@ -311,10 +317,10 @@ fn drop_random_nodes<R: Rng>(
     rng: &mut R,
     nodes: &mut Vec<TestNode>,
     drop_probability: f64,
-) -> BTreeSet<XorName> {
+) -> HashSet<PublicId> {
```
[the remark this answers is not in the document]
madadam: I think `HashSet` is in general more efficient than `BTreeSet`, but that is likely irrelevant in this case. The actual reason for the change is that I wanted to use the `retain` method which is only defined in `HashSet`.

tests/mock_network/messages.rs
```diff
@@ -33,31 +34,20 @@ fn send() {
...
-                _ => panic!("{} - Event::MessageReceived not received", node.name()),
+    let mut expected_recipients: HashMap<_, _> = expected_recipients(&nodes, &dst)
+        .map(|index| (index, false))
+        .collect();
+    assert!(expected_recipients.len() >= quorum);
+
+    poll_until(&env, &mut nodes, |nodes| {
+        for (index, node) in nodes.iter().enumerate() {
+            if let Some(received) = expected_recipients.get_mut(&index) {
+                *received = *received || message_received(node, &content);
```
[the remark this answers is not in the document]
madadam: Yes, if `received` is true, keep it true, otherwise assign to it the result of `message_received` function.
Will add the comment though.

tests/mock_network/messages.rs
madadam: `poll_until` evaluates the termination predicate as the very first thing, so if it's already true, it won't run even single iteration. So efficiency-wise, this should be fine.

tests/mock_network/messages.rs
madadam: That's what the logic of the test were before this PR, but routing actually does support sending user messages to non-elders too. We should probably add tests for that too at some point. But not in this PR.

tests/mock_network/messages.rs
madadam: There might be extra events, but we don't care about them in this tests. I'll make it explicit.

tests/mock_network/mod.rs
madadam: Do you mean is it guaranteed that the paused node is elder? If so, then no, it is not. That's why there is this `if`. So this tests needs to be run multiple times to make sure this situation is hit.

On the other hand, maybe we can pick the node to pause only from elders? But then we probably do want to test pausing also non-elders. Maybe we should have two tests?

tests/mock_network/mod.rs
madadam: It's what the comment above says: "poll the network for a while". 60 seconds is "a while" :)

That said, this should probably be changed to something slightly less than 30 seconds because that's the interval after which the node would be detected as non-responsive (`RESEND_MAX_ATTEMPTS * RESEND_DELAY`). The reason why it still works is that the pause/resume is not fully implemented yet and we don't actually drop the quic-p2p instance when pausing. So the remaining nodes never receive the `UnsentUserMessage` events and thus never trigger the non-responsiveness check. This is something that will definitely have to be addressed when we get to finalizing the pause/resume implementation. 

That said, I'll change the interval already and add some explanation to the code.

tests/mock_network/utils.rs
madadam: Yes, I removed the hack where we ignored timeout events so the comment is no longer necessary.

tests/mock_network/utils.rs
madadam: I guess that works too.

tests/mock_network/utils.rs
madadam: `remove_count` is [defined as `16`](https://github.com/maidsafe/routing/pull/2079/files#diff-c62a654da29f73317ed68538f16ae97bR463), so what would be the point of that?

tests/mock_network/utils.rs
madadam: I'll remove the whole `poll_all` call as it shouldn't be needed.

tests/mock_network/utils.rs
madadam: Well that's one of the main points of this PR - this function was incorrect because we should never expect that all nodes have the same view of the network at the same time. That's not how the real network works and so we should not assume that in the tests either.

tests/mock_network/utils.rs
madadam: No because they might not necessarily be at any given time.

tests/mock_network/mod.rs
maqi: > Do you mean is it guaranteed that the paused node is elder? If so, then no, it is not.

Yes, I mean this.

> So this tests needs to be run multiple times to make sure this situation is hit.

Yes, this is where I concerned. For how many iterations will this situation be hit? I am concerned it may not be tested enough. Ideally, such situation shall be covered within every iteration.

madadam: OK, I agree. But given that the whole pause/resume functionality is still not finished, this issue is probably not that critical right now. What I'd do is to create an issue to address this later. How does that sound?

maqi: yeah, that will be ok.  :)

madadam: Created the issue: https://github.com/maidsafe/routing/issues/2081

PR #2080, opened 2020-04-16
expose method to retrieve current section adults

madadam:


src/node/mod.rs
```diff
@@ -299,6 +299,14 @@ impl Node {
             .flat_map(move |stage| stage.chain.closest_section_info(name).1.member_nodes())
     }
 
+    /// Returns the information about all the current section adults.
```
maqi, line 302: => Returns the information of out section's adults.

madadam: I just copied the format of the comment for `our_elders`.

Issue #2078, opened 2020-04-02
Changes to node ageing


madadam:
- Node age is a number between 4 and 255 (inclusive), so it's stored in `u8`.
- There is no `age_counter` anymore
- During network startup when there is only one section (the section with empty prefix) and the section size is less than `safe_section_size`, any churn event causes the ages of all the nodes in the section to increase by 1.
- After `safe_section_size`, the ageing logic changes as follow:
    - consider only churn events of mature nodes (adults or elders), ignore infants
    - calculate the BLS signature `sig` of the churn event
    - for each node, calculate `sig % 2^age`
    - if it is zero, increase age of that node by one and relocate it
    - otherwise do nothing

Notes:
- We will need to collect the BLS signatures of the churn events (`Online` and `Offline`) which we currently don't. The mechanism for that is already in place for `SectionInfo` events so this should not be hard
- We will need some way to perform 256-bit integer arithmetic. There is no built-in 256-bit integer type in rust, so we would need to find a crate or implement it ourselves (although the only operations we need is `2^u8` and `u256 % u256`, so it might not be too bad).

Edit: this task is actually a prerequisite for #2117 (parsec removal) because it removes the dependency of node-ageing/relocation on the order of churn events.

dirvine:
We can take the first 2 bytes of the signature here. So a u16 will work as the numerator and save us working with large integer maths.

madadam:
> We can take the first 2 bytes of the signature here. So a u16 will work as the numerator and save us working with large integer maths.

But `2^u8` is still potentially 256-bit number. Or do you want to use `wrapping_pow` or something like that?

madadam:
Closed via https://github.com/maidsafe/sn_routing/pull/2185

PR #2082, opened 2020-04-21
Node ageing during network startup

madadam:
This PR changes the node ageing logic so during the network startup, all churn events cause age increments. 

Partially addresses #2078

### DIFF
--- a/src/chain/chain.rs
+++ b/src/chain/chain.rs
@@ -369,11 +369,15 @@ impl Chain {
     // Increment the age counters of the members.
     pub fn increment_age_counters(&mut self, trigger_node: &PublicId) {
         let our_section_size = self.state.our_joined_members().count();
-        let safe_section_size = self.safe_section_size();
+        let our_prefix = *self.state.our_prefix();
+
+        // Is network startup in progress?
+        let startup =
+            our_prefix == Prefix::default() && our_section_size < self.safe_section_size();
 
         // As a measure against sybil attacks, we don't increment the age counters on infant churn
-        // once we reach safe_section_size.
-        if our_section_size >= safe_section_size
+        // once we are past the startup phase.
+        if !startup
             && self
                 .state
                 .get_persona(trigger_node)
@@ -387,7 +391,6 @@ impl Chain {
             return;
         }
 
-        let our_prefix = *self.state.our_prefix();
         let relocating_state = self.state.create_relocating_state();
         let mut details_to_add = Vec::new();
 
@@ -399,6 +402,12 @@ impl Chain {
 
         for (name, member_info) in self.state.our_joined_members_mut() {
             if member_info.p2p_node.public_id() == trigger_node {
+                continue;
+            }
+
+            // During network startup we go through accelerated ageing.
+            if startup {
+                member_info.increment_age();
                 continue;
             }
 
@@ -426,7 +435,19 @@ impl Chain {
             })
         }
 
-        trace!("increment_age_counters: {:?}", self.state.our_members);
+        trace!(
+            "increment_age_counters: {}",
+            self.state
+                .our_members
+                .values()
+                .map(|info| format!(
+                    "{}: ({:?}, {:?})",
+                    info.p2p_node.name(),
+                    info.state,
+                    info.age_counter
+                ))
+                .format(", ")
+        );
 
         for details in details_to_add {
             trace!("Change state to Relocating {}", details.pub_id);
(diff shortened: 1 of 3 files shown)


src/chain/member_info.rs
```diff
@@ -122,15 +128,18 @@ mod tests {
 
     #[test]
     fn age_counter_to_age() {
+        let max_age = 16;
         let mut age_counter = AgeCounter::default();
 
-        for age in MIN_AGE..16 {
+        for age in MIN_AGE..max_age {
```
dirvine, line 134: should max_age not be capitalised?

src/chain/chain.rs
```diff
@@ -369,11 +369,15 @@ impl Chain {
...
+        let our_prefix = *self.state.our_prefix();
+
+        // Is network startup in progress?
+        let startup =
+            our_prefix == Prefix::default() && our_section_size < self.safe_section_size();
 
         // As a measure against sybil attacks, we don't increment the age counters on infant churn
-        // once we reach safe_section_size.
-        if our_section_size >= safe_section_size
+        // once we are past the startup phase.
```
maqi, line 379: Once we completed the startup phase, and ... ?

src/chain/member_info.rs
```diff
@@ -82,6 +84,10 @@ impl MemberInfo {
         self.age_counter.increment()
     }
 
+    pub fn increment_age(&mut self) {
+        self.age_counter = AgeCounter::from_age(self.age_counter.age().saturating_add(1))
```
maqi, line 88: why not just call `self.increment_age_counter` or `self.age_counter.increment()` ?

tests/mock_network/node_ageing.rs
```diff
@@ -150,6 +151,68 @@ fn relocate_during_split() {
...
 }
 
+// Test that during startup phase all churn events cause age increments.
+#[test]
+fn startup_phase() {
+    let env = Environment::new(NETWORK_PARAMS);
+    let mut nodes = vec![];
+
+    // Only the first `safe_section_size - 1` adds cause age increments, the rest does not.
+    for i in 0..(env.safe_section_size() + 1) {
```
maqi, line 161: the comment says `safe_section_size - 1` but the code is `safe_section_size() + 1`
I thought the code shall be `0..safe_section_size()` ?

tests/mock_network/node_ageing.rs
```diff
@@ -150,6 +151,68 @@ fn relocate_during_split() {
...
+            node_joined(nodes, nodes.len() - 1)
+        });
+
+        poll_until(&env, &mut nodes, |nodes| {
+            check_root_section_age_counters_after_only_adds(&env, nodes)
+        })
+    }
+}
+
+// Verify that the age counters of all the nodes in the root section are as expected assuming we
```
maqi, line 175: ... are as expected, assuming we were ...
or 
have a bracket around `assuming we were ... not removing`

tests/mock_network/node_ageing.rs
```diff
@@ -150,6 +151,68 @@ fn relocate_during_split() {
...
+        assert!(
+            nodes[i]
+                .inner
+                .our_prefix()
+                .map(|prefix| *prefix == Prefix::default())
+                .unwrap_or(true),
+            "the root section has split"
+        );
+
+        // The number of churn events the i-th node experienced during the startup phase.
```
maqi, line 194: ... churn events that the i-th node experienced ...

src/chain/chain.rs
madadam: Sorry, not sure what are you trying to say. The complete sentence is 

> As a measure against sybil attacks, we don't increment the age counters on infant churn once we are past the startup phase.

is there something wrong with it?

src/chain/member_info.rs
madadam: because we need to increment age, not age counter.

tests/mock_network/node_ageing.rs
madadam: We iterate safe_section_size + 1 times to test both cases: while we are below safe_section_size, the ages are incremented, but at and past it they are not.

src/chain/chain.rs
maqi: Ah, missed the above sentence.

However, still think:
once we are past the startup phase => once we completed the startup phase
will be better?

maqi: Ah, missed the above sentence.

However, still think:
once we are past the startup phase => once we completed the startup phase
will be better?

src/chain/member_info.rs
maqi: ok. I got what you mean.
So basical we want the new age_counter to be reset with a new increased age, right?

madadam: Yes. So if the age counter is e.g. 16 (corresponding to age 4), we want to go to 32, not 17.

PR #2084, opened 2020-04-27
Routing unchained

madadam:
This PR removes the `chain` module and moves its functionality into several modules, each with more clearly defined responsibilities. The new modules are:

- `shared_state`: info about the section shared among all its elders. this existed before inside `Chain`, but is now more prominent and contained directly in the `Approved` stage.
- `consensus`: everything related to voting and reaching consensus. Contains mainly the `ConsensusEngine` struct which wraps `ParsecMap` and `EventAccumulator` (formerly called `ChainAccumulator`).
- `section`: everything related to sections and their members. Contains mainly `SectionMap` which is a container for all known sections in the network (including ours) and `SectionMembers` which is a container for the members of our section.
- `routing_table`: message routing logic


src/location.rs
```diff
@@ -35,6 +35,14 @@ impl SrcLocation {
     pub fn is_multiple(&self) -> bool {
         !self.is_single()
     }
+
+    /// Returns whether the given name is part of this location
+    pub fn contains(&self, name: &XorName) -> bool {
```
maqi, line 40: function name better to be `is_src`?

src/location.rs
```diff
@@ -96,4 +104,20 @@ impl DstLocation {
             Self::Node(_) | Self::Section(_) | Self::Direct => Err(RoutingError::BadLocation),
         }
     }
+
+    /// Returns whether the given name of the given prefix is part of this location.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `prefix` does not match `name`.
+    pub fn contains(&self, name: &XorName, prefix: &Prefix<XorName>) -> bool {
```
maqi, line 113: function name better to be is_dst ?

src/routing_table.rs
```diff
@@ -0,0 +1,205 @@
...
+///     - if our name *is* the destination, returns an empty set; otherwise
+///     - if the destination name is an entry in the routing table, returns it; otherwise
+///     - returns the `N/3` closest members of the RT to the target
+pub fn delivery_targets(
+    dst: &DstLocation,
+    our_id: &PublicId,
+    our_members: &SectionMembers,
+    sections: &SectionMap,
+) -> Result<(Vec<P2pNode>, usize)> {
+    if !sections.is_elder(our_id) {
```
maqi, line 54: I thought being a non-elder, we shall not relay message?
i.e. message flow from elder to adult/infant is allowed, but the backwards is not?

src/routing_table.rs
```diff
@@ -0,0 +1,205 @@
...
+        dg_size = delivery_group_size(len);
+
+        if prefix == sections.our().prefix() {
+            // Send to all connected targets so they can forward the message
+            nodes_to_send.retain(|node| node.name() != our_id.name());
+            dg_size = nodes_to_send.len();
+            break;
+        }
+        if idx == 0 && nodes_to_send.len() >= dg_size {
+            // can deliver to enough of the closest section
```
maqi, line 156: not quite understand this comment.
does it want to say:
Closest section already having enough candidates to deliver ?

src/section/section_members.rs
```diff
@@ -0,0 +1,253 @@
...
+    collections::{btree_map::Entry, BTreeMap},
+    mem,
+    net::SocketAddr,
+};
+
+/// Container for storing information about members of our section.
+#[derive(Debug, Eq, Serialize, Deserialize)]
+pub struct SectionMembers {
+    members: BTreeMap<XorName, MemberInfo>,
+    // Number that gets incremented every time a node joins or leaves our section - that is, every
```
maqi, line 29: incremented  -> increased ?

src/section/section_members.rs
```diff
@@ -0,0 +1,253 @@
...
+            .collect();
+
+        Self {
+            members,
+            section_version: 0,
+            post_split_siblings: Default::default(),
+        }
+    }
+
+    /// Returns an iterator over the members that have not state == `Left`.
```
maqi, line 62: that is not in the `Left` state

src/section/section_members.rs
```diff
@@ -0,0 +1,253 @@
...
+        elder_size: usize,
+    ) -> BTreeMap<XorName, P2pNode> {
+        elder_candidates(
+            elder_size,
+            self.joined()
+                .filter(|info| prefix.matches(info.p2p_node.name())),
+        )
+    }
+
+    /// Check if the given `PublicId` is a member of our section.
```
maqi, line 121: an active member ?

src/section/section_proof_chain.rs
```diff
@@ -0,0 +1,317 @@
...
+    #[cfg(all(test, feature = "mock"))]
+    pub fn all_prefix_version(&self) -> impl DoubleEndedIterator<Item = (&Prefix<XorName>, u64)> {
+        iter::once((&self.prefix, self.version)).chain(
+            self.blocks
+                .iter()
+                .map(|block| (block.prefix(), block.version())),
+        )
+    }
+
+    pub fn last_new_key_info(&self) -> Option<&SectionKeyInfo> {
```
maqi, line 86: last_key_info shall be enough?

src/section/section_proof_chain.rs
```diff
@@ -0,0 +1,317 @@
...
+                block
+            );
+            return;
+        }
+
+        self.blocks.push(block)
+    }
+
+    #[cfg(test)]
+    pub fn validate(&self) -> bool {
```
maqi, line 188: self_validate ?

src/section/section_proof_chain.rs
```diff
@@ -0,0 +1,317 @@
...
+            if !validate_next_block(current, block) {
+                return false;
+            }
+
+            current = block.key_info();
+        }
+        true
+    }
+
+    pub fn first_key_info(&self) -> &SectionKeyInfo {
```
maqi, line 200: in contrast to the `last_key_info` where we fetch from the blocks, here the first one is always being the genesis ?
this is bit misleading. maybe better have the function name as `genesis_key_info` to be explicit?

src/section/section_proof_chain.rs
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    pub fn last_key_info(&self) -> &SectionKeyInfo {
+        self.blocks
+            .last()
+            .map(|block| block.key_info())
+            .unwrap_or(&self.genesis_key_info)
+    }
+
+    pub fn slice_from(&self, first_index: usize) -> SectionProofSlice {
```
maqi, line 211: better have a comment of this function

src/section/section_map.rs
```diff
@@ -0,0 +1,461 @@
...
+            }
+        }
+        let _ = self.knowledge.insert(prefix, version);
+    }
+
+    /// Compute an estimate of the size of the network from the size of our routing table.
+    ///
+    /// Return (estimate, exact), with exact = true iff we have the whole network in our
+    /// routing table.
+    pub fn network_size_estimate(&self) -> (u64, bool) {
```
oetyng, line 430: Ah this thing, I know this is not new code, but just pointing out that it is a little bit confusing to call this network size when it is just counting Elders.

src/section/section_members.rs
```diff
@@ -0,0 +1,253 @@
...
+    net::SocketAddr,
+};
+
+/// Container for storing information about members of our section.
+#[derive(Debug, Eq, Serialize, Deserialize)]
+pub struct SectionMembers {
+    members: BTreeMap<XorName, MemberInfo>,
+    // Number that gets incremented every time a node joins or leaves our section - that is, every
+    // time `members` changes.
+    section_version: u64,
```
oetyng, line 31: `section_version`
Thinking out loud here..
I think `section_` is redundant. I would maybe go for just `version`, it's the version of the `SectionMembers`, that is perfectly clear and evident.
Maybe we want to call it `member_version` globally to be absolutely clear though (sticking to one term).

src/section/section_map.rs
```diff
@@ -0,0 +1,461 @@
...
+
+        (our_prefix.is_neighbour(other_prefix) || other_prefix.is_extension_of(our_prefix))
+            && self.is_new(elders_info)
+    }
+
+    /// Returns all elders from all known sections.
+    pub fn elders(&self) -> impl Iterator<Item = &P2pNode> {
+        self.all()
+            .map(|(_, info)| info)
+            .flat_map(EldersInfo::member_nodes)
```
maqi, line 152: I always feel bit confused with this `elders` and `member` mismatch   :(
same comment to the others.

src/section/section_map.rs
```diff
@@ -0,0 +1,461 @@
...
+
+    // Remove outdated neighbour infos.
+    fn prune_neighbours(&mut self) {
+        // Remove invalid neighbour pfx, older version of compatible pfx.
+        let to_remove: Vec<_> = self
+            .other
+            .iter()
+            .filter_map(|(pfx, elders_info)| {
+                if !self.our().prefix().is_neighbour(pfx) {
+                    // we just split making old neighbour no longer needed
```
maqi, line 264: old neighbour is no longer needed, as we just split ?

src/section/section_map.rs
```diff
@@ -0,0 +1,461 @@
...
+        let _ = self.keys.insert(*key_info.prefix(), key_info.clone());
+    }
+
+    pub fn get_knowledge(&self, prefix: &Prefix<XorName>) -> Option<u64> {
+        self.knowledge.get(prefix).copied()
+    }
+
+    /// Returns the index of the public key in our_history that will be trusted by the target
+    /// location
+    pub fn proving_index(&self, target: &DstLocation) -> u64 {
```
maqi, line 348: provable_index ?

src/section/section_map.rs
```diff
@@ -0,0 +1,461 @@
...
+            _ => self.proving_index(target),
+        }
+    }
+
+    /// Updates the entry in `knowledge` for `prefix` to the `version`; if a split
+    /// occurred in the meantime, the versions for sections covering the rest of the address space
+    /// are initialised to the old version that was stored for their common ancestor
+    pub fn update_knowledge(&mut self, prefix: Prefix<XorName>, version: u64) {
+        trace!(
+            "attempts to update knowledge of our elders_info with version {:?} for \
```
maqi, line 390: I think this comment is bit misleading.
The `knowledge` is our perspective that other section think what is our version is.
So, I think the comment shall better be:
("Section (prefix {:?}) shall now know about our section version {:?}", prefix, version)

src/consensus/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+                        | parsec::Observation::DkgMessage(_) => None,
+                    })
+                    .cloned(),
+            )
+            .filter(|event| !completed_events.contains(&event.payload))
+            .collect()
+    }
+
+    // Completes the reset of the consensus engine.
+    pub fn complete_reset(
```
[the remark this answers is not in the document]
madadam: That would be ideal but it needs some thinking. Will try to come back to this after I address the more straightforward stuff.

src/consensus/network_event.rs
```diff
@@ -191,40 +217,41 @@ impl Debug for NetworkEvent {
...
-/// The outcome of polling the chain.
-#[derive(Eq, PartialEq, Serialize, Deserialize)]
-pub struct AccumulatedEvent {
-    pub content: AccumulatingEvent,
-    pub elders_change: EldersChange,
+// Change to section elders.
+#[derive(Default, Eq, PartialEq, Serialize, Deserialize)]
+pub struct EldersChange {
+    // Neighbour peers that became elders.
+    pub neighbour_added: BTreeSet<P2pNode>,
```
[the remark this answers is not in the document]
madadam: Not quite. An elder might be added without being promoted (if its relocated from other section) and removed without being demoted (if it goes online), so I'd keep the current terms.

src/location.rs
madadam: The type itself is already called **Src**Location so repeating `src` seems redundant to me. Also `contains` seems more appropriate than `is`, because it's about whether the location *contains* the name, rather than the location *is* the name.

src/location.rs
madadam: Same as https://github.com/maidsafe/routing/pull/2084#discussion_r417312930

src/section/shared_state.rs
```diff
@@ -0,0 +1,854 @@
...
+            return Ok(None);
+        }
+
+        let our_prefix = self.our_prefix().pushed(next_bit);
+        let other_prefix = self.our_prefix().pushed(!next_bit);
+
+        let our_elders = self
+            .our_members
+            .elder_candidates_matching_prefix(&our_prefix, network_params.elder_size);
+        let other_elders = self
```
maqi, line 336: just to confirm that we don't need to ensure the number of elders within the split out sections, right?

src/routing_table.rs
madadam: We do relay them, but only to our elders. This is because a node might receive a message addressed to elder but the node is not elder yet or not elder anymore. To prevent losing the message, we relay it to the elders to let them deal with it.

src/section/shared_state.rs
```diff
@@ -0,0 +1,854 @@
...
+            let pub_id = *FullId::within_range(rng, &pfx.range_inclusive()).public_id();
+            let _ = members.insert(
+                pub_id,
+                P2pNode::new(pub_id, ([127, 0, 0, 1], 9000 + index).into()),
+            );
+        });
+        EldersInfo::new_for_test(members, pfx, version).unwrap()
+    }
+
+    // start_pfx: the prefix of our section as string
```
maqi, line 490: `start_pfx` no longer present?

src/routing_table.rs
madadam: To be honest I'm not sure what it the comment trying to say myself. I only copied this code from elsewhere. My suspicion is that this whole function might not be needed because we don't have the concept of "being connected" anymore. I'd come back to this in some other PR.

src/section/section_members.rs
madadam: Same thing, no? But I'll change it if you prefer.

madadam: Actually, I think "increment" usually means "increase by one" which is what's going on here. So I'd keep it unless you strongly disagree.

src/node/stage/approved.rs
```diff
@@ -236,6 +261,11 @@ impl Approved {
         self.vote_for_event(AccumulatingEvent::User(event));
     }
 
+    /// Is the node with the given id an elder in our section?
+    pub fn is_our_elder(&self, id: &PublicId) -> bool {
+        self.shared_state.sections.our().is_member(id)
```
maqi, line 266: I think there is function called `is_self_elder` shall do the job?
`is_member` seems could be an adult or infant?

src/node/stage/approved.rs
```diff
@@ -370,18 +396,22 @@ impl Approved {
         }
 
         let conn_infos: Vec<_> = self
-            .chain
+            .shared_state
+            .sections
```
maqi, line 400: `our_elders` shall be a direct function of the shared_state ?

src/section/section_proof_chain.rs
madadam: But the first block IS the genesis block. `blocks.first()` would actually be the second block. The genesis block is stored separately because it doesn't have signature, but it's still part of the same chain.

src/node/stage/approved.rs
```diff
@@ -677,225 +735,268 @@ impl Approved {
...
+        self.shared_state.update(new_state);
 
         Ok(())
     }
 
     fn handle_online_event(&mut self, core: &mut Core, payload: OnlinePayload) -> Result<()> {
-        if !self.chain.can_add_member(payload.p2p_node.public_id()) {
-            info!("ignore Online: {:?}.", payload);
-        } else {
+        assert!(!self.churn_in_progress);
```
maqi, line 952: I think we shall not have an assertion within the production code?
same comment to other places

src/consensus/event_accumulator.rs
```diff
@@ -77,21 +80,21 @@ impl VoteStatuses {
...
 
 #[derive(Default)]
-pub(super) struct ChainAccumulator {
+pub struct EventAccumulator {
     /// A map containing network events that have not been handled yet, together with their proofs
     /// that have been collected so far. We are still waiting for more proofs, or to reach a state
     /// where we can handle the event.
     // FIXME: Purge votes that are older than a given period.
-    chain_accumulator: BTreeMap<AccumulatingEvent, AccumulatingProof>,
+    incomplete_events: BTreeMap<AccumulatingEvent, AccumulatingProof>,
```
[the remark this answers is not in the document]
madadam: :+1: Will go with `accumulated` / `unaccumulated` to avoid using different terms for the same thing.

src/section/section_proof_chain.rs
dirvine: We can treat the genesis block as the network name as that is the ultimate identifier.

src/section/section_map.rs
dirvine: To be network size perhaps we need elder count * 60 as min and elder count * 120 as max estimate? To compromise elder count * our_current_section_size  and assume relatively even distribution.

src/section/section_members.rs
dirvine: I think version will be dropped now in favour of just key as that is unique and all we need is uniqueness here.

src/consensus/genesis_pfx_info.rs
```diff
@@ -0,0 +1,49 @@
...
+    id::PublicId,
+    section::{AgeCounter, EldersInfo},
+};
+use std::{
+    collections::BTreeMap,
+    fmt::{self, Debug, Formatter},
+};
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
+pub struct GenesisPfxInfo {
```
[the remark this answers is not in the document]
madadam: Yeah, it's supposed to be `GenesisPrefixInfo` which is still a pretty terrible name. Maybe you can suggest better? It contains info about the section and is used in these cases:
1. When a new node join a section, it receives this in the `NodeApproval` message. 
2. When a section elder set changes, all the non-elder nodes in the section receive this in the `GenesisUpdate` message

src/consensus/mod.rs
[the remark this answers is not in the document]
madadam: Agree. Will go with `finalize` which we already use elsewhere for similar things.

src/consensus/genesis_pfx_info.rs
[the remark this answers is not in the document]
dirvine: IMO This should just be the genesis identifier and the whole section chain for this section from genesis. No need to strip this down at all, especially not yet. So this this_prefix signed by the elders is an item, the genesis id is an item and the section_chain is the other item. 
That way we are crypto solid and as simple as we can make it. 
It does mean we need to (on split) create not only the agreed section_chain entry but at that time sign what the prefix is.

src/consensus/network_event.rs
```diff
@@ -79,9 +79,21 @@ pub struct OnlinePayload {
 #[allow(clippy::large_enum_variant)]
 #[derive(Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub enum AccumulatingEvent {
-    /// Vote to start a DKG instance
+    /// Genesis event. This is output-only event.
+    Genesis {
```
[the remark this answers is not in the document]
madadam: Agree almost wholly, but this thing is tricky. This is not really an event, neither is it a command, but at the same time it is kind of both. We can try thinking about how to express this better, but as we are trying to move routing away from parsec anyway, so at some point this whole thing will go away. So not sure it's worth it? 

Or maybe we can just do something simple for now: rename the event-like variants to past-tense verbs and the command-like variants to imperative-tense verbs?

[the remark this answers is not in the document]
dirvine: Does that then lean towards 2 enums? one for events and one for commands?

src/consensus/network_event.rs
```diff
@@ -191,40 +217,41 @@ impl Debug for NetworkEvent {
...
-    }
-
-    pub fn with_elders_change(self, elders_change: EldersChange) -> Self {
-        Self {
-            elders_change,
-            ..self
+impl EldersChange {
+    pub fn builder(sections: &SectionMap) -> EldersChangeBuilder {
+        EldersChangeBuilder {
+            old_neighbour: sections.other_elders().cloned().collect(),
```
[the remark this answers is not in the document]
madadam: It's about peers in this case. Maybe that should be reflected in the name, i.e.: `old_neighbour_elders`? I'll do that.

src/consensus/network_event.rs
```diff
@@ -191,40 +217,41 @@ impl Debug for NetworkEvent {
...
 }
 
-/// The outcome of polling the chain.
-#[derive(Eq, PartialEq, Serialize, Deserialize)]
-pub struct AccumulatedEvent {
-    pub content: AccumulatingEvent,
-    pub elders_change: EldersChange,
+// Change to section elders.
+#[derive(Default, Eq, PartialEq, Serialize, Deserialize)]
+pub struct EldersChange {
```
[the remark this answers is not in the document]
madadam: The way I look at it is an struct describing the change itself, that is, which nodes were added and which removed (a diff). So calling it `Change` seems OK? 

But apart from that, I like your suggestion. In the past, this actually contained separate fields for our-section elders and for neighbour elders so the naming was more appropriate then.

Or we can go the other way and make it generic: `SetChange` / `SetDiff` ?

src/node/mod.rs
```diff
@@ -944,23 +943,25 @@ impl Node {
...
     /// Returns `None` otherwise.
     pub fn section_elder_info_version(&self, prefix: &Prefix<XorName>) -> Option<u64> {
-        self.chain()
-            .and_then(|chain| chain.get_section(prefix))
+        self.shared_state()
+            .and_then(|state| state.sections.get(prefix))
             .map(|info| info.version())
     }
 
     /// Returns the elder of a section with the given prefix.
```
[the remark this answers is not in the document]
madadam: Yes, it should be "elders". Will fix.

src/node/mod.rs
```diff
@@ -972,78 +973,92 @@ impl Node {
...
+        self.core.network_params.elder_size
     }
 
     /// Size at which our section splits. Since this is configurable, this method is used to
     /// obtain it.
-    ///
-    /// Only if we have a chain (meaning we are elders) we will process this API
-    pub fn safe_section_size(&self) -> Option<usize> {
-        self.chain().map(|chain| chain.safe_section_size())
+    pub fn safe_section_size(&self) -> usize {
```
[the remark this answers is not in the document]
madadam: The "safe" here is not "SAFE" though. It's sort of like the minimal size each section should have, even though a section can still function with lot less members. So maybe it should be called something like 'desired_section_size` or `nominal_section_size` or `comfortable_section_size`, etc... ?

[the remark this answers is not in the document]
dirvine: minimum_section_size ?

src/node/stage/approved.rs
```diff
@@ -51,14 +54,20 @@ const INITIAL_RELOCATE_COOL_DOWN_COUNT_DOWN: i32 = 10;
...
     gen_pfx_info: GenesisPfxInfo,
     timer_token: u64,
     // DKG cache
     dkg_cache: BTreeMap<BTreeSet<PublicId>, EldersInfo>,
+    // The accumulated info during a split pfx change.
+    split_cache: Option<SplitCache>,
+    // Marker indicating we are processing churn event
+    churn_in_progress: bool,
+    // Marker indicating that elders may need to change,
+    members_changed: bool,
```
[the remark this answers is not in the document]
madadam: :+1 Will try to change the comment.

src/routing_table.rs
maqi: Ok. just to mark this

src/section/section_members.rs
maqi: It's just `increment` is a noun, not a verb. :(
I won't insist on it, so up to you then.  :)

src/section/section_proof_chain.rs
maqi: > The genesis block is stored separately because it doesn't have signature

Ah, ok then.

src/node/mod.rs
```diff
@@ -917,25 +918,23 @@ impl Node {
...
-        self.chain()
-            .map(|chain| {
-                chain
-                    .neighbour_infos()
-                    .map(|info| info.prefix())
+        self.shared_state()
+            .map(|state| {
+                state
+                    .sections
+                    .other()
```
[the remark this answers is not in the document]
madadam: I also try to avoid using different terms to refer to the same thing but that is not the case here. "other" means a section that is not our section and "neighbour" means a section whose prefix differs from our prefix by at most one bit. So all neighbours are also others, but not the other way around. 

That said, currently other actually contains only neighbours so the two terms seemingly mean the same thing. But that's going to change soon because we are going to keep all sections, not just neighbours and we might stop using the term "neighbour" at all.

src/node/stage/approved.rs
```diff
@@ -599,17 +636,24 @@ impl Approved {
         msg: Message,
         msg_bytes: Bytes,
     ) {
+        let is_self_elder = self.is_our_elder(core.id());
```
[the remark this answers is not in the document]
madadam: They are two different things: `self` means "this node" and `our` means "of our section". Maybe we should make it more explicit and call the method `is_our_section_elder` or `is_elder_of_our_section` or something like that?

src/node/stage/approved.rs
```diff
@@ -678,225 +729,268 @@ impl Approved {
...
+            return Ok(true);
+        }
+
+        Ok(false)
     }
 
-    fn chain_poll(&mut self, core: &mut Core) -> Result<()> {
-        let mut old_pfx = *self.chain.our_prefix();
-        let mut was_elder = self.chain.is_self_elder();
+    pub fn can_poll_churn(&self) -> bool {
```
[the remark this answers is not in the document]
madadam: Yeah :) It's supposed to mean something like: can we perform an action which can potentially result in churn (change of our elder membership) right now? Not sure how to express that succinctly.  Maybe `can_process_churn`? What do you think?

src/node/stage/approved.rs
```diff
@@ -678,225 +729,268 @@ impl Approved {
...
+            );
+
+            self.handle_accumulated_event(core, event, AccumulatingProof::default())?;
+            Ok(true)
+        } else {
+            Ok(false)
+        }
+    }
+
+    fn check_ready_or_backlog_churn_event(
```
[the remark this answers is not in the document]
madadam: And this is supposed to mean: can we process this event right now or do we need to put it into the backlog because we are in the middle of churn? Better naming suggestions welcome, but note that this function doesn't just return `true`/`false`. It either returns the event which means we can process it, or it backlogs it and returns `None`.

src/node/stage/approved.rs
```diff
@@ -678,225 +729,268 @@ impl Approved {
...
+        &mut self,
+        event: AccumulatingEvent,
+    ) -> Option<AccumulatingEvent> {
+        let start_churn_event = match &event {
```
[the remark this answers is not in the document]
madadam: Yeah, it means an event that can trigger churn. Will rename it according to your suggestion.

src/node/stage/approved.rs
```diff
@@ -678,225 +729,268 @@ impl Approved {
...
+    /// Polls and handles the next scheduled relocation, if any.
+    fn poll_relocation(&mut self, our_id: &PublicId) -> bool {
+        // Delay relocation until no additional churn is in progress.
+        if !self.can_poll_churn() {
+            return false;
+        }
+
+        if let Some(details) = self.shared_state.poll_relocation() {
+            if self.is_our_elder(our_id) {
+                self.vote_for_relocate_prepare(details, INITIAL_RELOCATE_COOL_DOWN_COUNT_DOWN);
```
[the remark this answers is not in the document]
madadam: I'd keep the phrase "cool down" because we've been using it for this specific thing. It's supposed to mean the number of iterations (rounds/steps/...) of the relocation cool-down period. So maybe just `RELOCATE_COOL_DOWN_STEPS`  /`RELOCATE_COOL_DOWN_ITERATIONS` / `RELOCATE_COOL_DOWN_ROUNDS`?

src/section/section_map.rs
dirvine: Perhaps 
members == all 
mature == all except infants 
Then Infant, Adults, Elders

maqi: Yeah, that will be more explicit I think.

src/consensus/genesis_pfx_info.rs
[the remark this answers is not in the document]
oetyng: > IMO This should just be the genesis identifier and the whole section chain for this section from genesis.

Aha, trying to understand.. So for every time this happens:
> When a section elder set changes, all the non-elder nodes in the section receive this in the `GenesisUpdate` message

.. would we then send the whole section chain to non-elders?

src/consensus/network_event.rs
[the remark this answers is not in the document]
oetyng: Yeah, if you allow me some time I'm going to look at the names and the actual contexts of these and try come up with a good enough solution for now.

From what I have seen before, I think there is a possible cmd/event separation even here, if only some things are straightened out (we'll see if that holds true, and if there's not too much straightening to do).

So, basically, I think this could be the case:
> Does that then lean towards 2 enums? one for events and one for commands?

But will say more when I have fully seen what is meant by this:
> This is not really an event, neither is it a command, but at the same time it is kind of both.

src/consensus/network_event.rs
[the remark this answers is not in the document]
oetyng: We're getting into some interesting details of distinction now :)

> So calling it `Change` seems OK?

This is actually the definition of an event. It is a change (see https://en.wikipedia.org/wiki/Event-driven_architecture). So, I would say we are kind of duplicating the language if we come up with an item called `Change`, that is not an event. We could come up with another one that is `Progress`, yet another one that is `Alteration` and so on.. Or, we identify all these things as events, fully embrace the event driven architecture that is half-way adopted already (and certainly is a very good fit for a decentralized network), and as a result we get a simpler structure/taxonomy.

Now, that said, maybe this `EldersChange` struct is part of something else that today is called an event? But if that event is not only about the elder change, then it seems to me we have bundled up several changes into one change.

*looking into the code*

Okay, so I attribute this impedance mismatch to the half-way adoption of the event driven architecture. So, what I see in the code is that developers have correctly used and identified event concepts, but it can be seen that all over, in details as well as overall, that there hasn't been much or any experience with event driven architectures. So there's a lot of "non-event driven" things mixed in there.

This for example:

```
// Connect to all elders from our section or neighbour sections that we are not yet connected
    // to and disconnect from peers that are no longer elders of neighbour sections.
    fn update_peer_connections(&mut self, core: &mut Core, change: &EldersChange) {
        let our_needed_connections: HashSet<_> = self
            .shared_state
            .known_nodes()
            .map(|node| *node.peer_addr())
            .collect();

        for p2p_node in &change.neighbours_removed {
            // The peer might have been relocated from a neighbour to us - in that case do not
            // disconnect from them.
            if our_needed_connections.contains(p2p_node.peer_addr()) {
                continue;
            }

            core.transport.disconnect(*p2p_node.peer_addr());
        }
    }
```

.. is what a Saga does. It listens for certain events, and sends cmds as a result of it.

So, being explicit with such systematization, actually makes the code simpler. There's not just a loose bunch of different objects leading to different things. There's a clearly defined pattern of flow: cmds become events, events are picked up and lead to new cmds. When we create a structure for that, actual types that do things in a specific way, it is much easier for developers to see the overall structure, as well as where they are in a specific flow at any given point.

> Or we can go the other way and make it generic: `SetChange` / `SetDiff` ?

I would not recommend it. Changes are about system behaviour, making a system event driven, and making events first class citizens, is about putting the changes - the behaviour - at the fore front. That means the events are granular and the event names are specific.

The problem here is deep, so I don't think we can fix it at this specific line of code :)

We have things like `handle_neighbour_info_event` and `handle_section_info_event`, and the "misinterpretation of events", the deviation from the principles I describe, that I consider be pervasive throughout the code, continues all the way. (Clarifying: `neighbour_info_event` is not considered a good naming of an event, it is not describing a change, it is not event centric, it is data centric).

src/consensus/genesis_pfx_info.rs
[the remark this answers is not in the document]
dirvine: In our own section we only need to send the update to our Adults etc. No need for the whole chain as they should get that on joining our sectoin. So we just send  a diff.

src/node/stage/approved.rs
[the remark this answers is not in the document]
oetyng: Aha :)
mm.. seems the most common reason why we would not be able to do this is if there is currently a churn in progress..
Having handled the genesis event seems like the base/edge case, that will most likely be true.

I propose: `is_ready_to_churn`

I think that is a straight forward language, that succinctly communicates that we need to be ready to churn before proceeding with what ever it is we need to do. And there is no unnecessary leaking of what ever details is going on under there.

src/node/stage/approved.rs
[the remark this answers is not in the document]
oetyng: aha ok, that whole thing seems a bit convoluted..

I would consider it better to keep the query part separated, so we get this:

in `fn poll_one()`:

```
let poll_one = if self.shall_backlog_event(event) {
            self.shared_state.churn_event_backlog.push_front(event);
            false
        } else {
            self.handle_accumulated_event(core, event, proof)?;
            true
        };

        Ok(poll_one)
```

and `check_ready_or_backlog_churn_event` is replaced with

```
fn shall_backlog_event(&mut self, event: AccumulatingEvent) -> bool {
        let is_churn_trigger = match &event {
            AccumulatingEvent::Online(_)
            | AccumulatingEvent::Offline(_)
            | AccumulatingEvent::Relocate(_) => true,
            _ => false,
        };
        is_churn_trigger && !self.can_poll_churn()
    }
```

More straight forward IMO, clearer.

src/node/stage/approved.rs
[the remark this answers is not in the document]
dirvine: I hope when we get to the `pull` method of seciton updates this problem actually goes away. We should not need to care about serialisation of events liek this and should be able to handle churn concurrently with any other action.

src/node/stage/approved.rs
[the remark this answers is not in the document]
oetyng: Ah ok, all those are better. So is any of those (rounds/steps/iterations) used more often to refer to this? I would go with that one then, or else any of the shorter ones.

src/routing_table.rs
madadam: I think what we should do is to write some (at least basic) unit tests for this module and then refactor and simplify it. Something I'd do in a different PR though.

src/section/section_members.rs
madadam: It is verb as well: https://en.wiktionary.org/wiki/increment#Verb

src/consensus/network_event.rs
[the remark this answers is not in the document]
madadam: OK. As I said, I'm sure it's worth spending time on this as it'll all go away eventually. On the other hand, in the post-parsec world some of these will become normal messages and so we would still need to name them. So maybe worth to spend a little bit of time on this, just not too much?

src/consensus/network_event.rs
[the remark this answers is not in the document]
madadam: OK, will make the changes as you are suggesting. 

Regarding things like `handle_section_info_event` - here "event" just means the thing that comes out of parsec (so `AccumulationEvent`) which we are already discussing elsewhere. So the convention we are currently using is `handle_foo` means handle a received message `Foo`, whereas `handle_foo_event` means handle the consensus/accumulation of an "event" 'Foo`. I admit it's not a particularly great convention. Maybe we should be more explicit:
`handle_foo_message` and `handle_foo_consensus` (or `handle_foo_accumulation`) ?

src/node/mod.rs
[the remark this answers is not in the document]
madadam: > minimum_section_size

That would imply that the section can't work correctly below that size which isn't the case. It's more like `recommended_section_size`.

src/node/stage/approved.rs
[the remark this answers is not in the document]
madadam: Yeah, I was thinking about this and I think we might be able to get rid of this churn serialization already. But I'd prefer to not do it in this PR which is already pretty huge. Will just apply @oetyng 's suggestion for now.

src/node/stage/approved.rs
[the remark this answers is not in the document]
madadam: Sounds good. Will do.

src/node/mod.rs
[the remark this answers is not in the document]
dirvine: It' a weird one to name. It is the minimum size that routing will keep adding nodes to get to and less than that we have all the age++ stuff happening. 
So this size is the min routing will push for and is therefore recommended by routing. After this point routing leaves it up to the consumer to ask for new members or get's relocates to here and so on. 
I am not sure what name makes sense, but think we need to go with something, so I would vote for `recommended_section_size` for now with a comment similar to above when we initialise that number.

src/consensus/network_event.rs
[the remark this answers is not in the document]
oetyng: Ok I see.
"handle" is a good description sometimes of what a method does. I use it at very high levels, when there is not much behaviour associated with the handling. Otherwise I try to instead name the method according to _what kind of handling_ is going on. Otherwise we run big risk of having the code littered with "handle_" this and that.

It is kind of implicit (often) that a method taking am event as arg handles an event so the information we're passing along at that point (prefixing with "handle") is not very useful and tend to become noise.

In safe_vaults this is something I have identified and I have started naming methods just straight up and down what the intended behaviour is, instead of a technical description of where something came from, what it is and how it will be handled.
As a very recent example from CH refactoring merged just yesterday:

`handle_[do_something_cmd]_client_req`
became
`initiate_[something to be done]`

and
`handle_[do_something_cmd]_vault_req`
became
`finalize_[something to be done]`

We that way went into describing the behaviour and flow (first cmd initiates, the other is a finalization of such a cmd), instead of the technical details around it.

So, to connect back to this. Could we simply call it `accumulate_foo`, or if it was already accumulated, describe what is actually done as a next step with this accumulated foo?

src/section/section_map.rs
madadam: I'd just rename the method to something like `network_elder_count_estimate` and let the upper layers do the total node count estimation if they want to.

src/section/section_members.rs
madadam: This is a different version though. This one tracks any change to the section, not just elders change. It's purpose is to break ties when selecting which node to promote in case there are multiple ones with the same age - the one with the lower version (which means the one that joined the section earlier) wins.

src/node/mod.rs
```diff
@@ -271,12 +274,20 @@ impl Node {
         }
     }
 
+    /// Returns whether the node is Elder.
+    pub fn is_elder(&self) -> bool {
+        self.stage
+            .approved()
+            .map(|stage| stage.shared_state.sections.our().is_member(self.core.id()))
```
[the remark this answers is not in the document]
madadam: `our_elders` returns just Iterator and so we can't call `is_member` on it.

src/section/section_map.rs
```diff
@@ -0,0 +1,461 @@
...
+    iter,
+};
+
+// Number of recent keys we keep: i.e how many other section churns we can handle before a
+// message send with a previous version of a section is no longer trusted.
+// With low churn rate, a ad hoc 20 should be big enough to avoid losing messages.
+const MAX_RECENT_KEYS: usize = 20;
+
+/// Container for storing information about sections in the network.
+/// Note: currently does not store our section, but that may change.
```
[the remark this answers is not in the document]
madadam: Good catch. It used to store only other sections but that changed and I forgot to change the comment.

src/node/stage/approved.rs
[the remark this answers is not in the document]
oetyng: Ah I see. ok so, the call to `self.is_our_elder(core.id());` answers if we are an elder. :+1:

> `is_our_section_elder`

This is clearer yep. 

So, not asking for big changes here, just continuing some thoughts:

I think "our" may have been used a bit inconsistently. Consider this:

> They are two different things: `self` means "this node" and `our` means "of our section".

..but in another part (also in node context) we have

> let our_id = self.core.id();

Here, `our` means "this node".


Anyway, it's a little bit tricky with nodes code in that way, when the code is describing  what happens on "this node" vs on "many nodes".
It's not a problem to solve here, but I got reminded that I want to find a clearer convention for this.

src/section/section_map.rs
```diff
@@ -0,0 +1,461 @@
...
+    }
+
+    /// Returns a known section whose prefix is compatible with the given prefix, if any.
+    pub fn compatible(&self, prefix: &Prefix<XorName>) -> Option<&EldersInfo> {
+        self.all()
+            .find(move |(pfx, _)| pfx.is_compatible(prefix))
+            .map(|(_, info)| info)
+    }
+
+    /// Find other section containing the given elder.
```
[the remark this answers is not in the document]
madadam: We only keep elders. I'm going to rename the method instead, to make it more explicit.

src/section/section_map.rs
madadam: Yeah, it's because the `EldersInfo` struct contains only elders, so `member` means actually elder. But I can see the confusion and will change it.

src/node/mod.rs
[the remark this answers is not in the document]
oetyng: > It' a weird one to name.

Yep, lot's of information in that little piece.

Writing out a previous comment that I hadn't sent:

> If it is the minimal size it should have, it seems natural to call it `minimal` size.
> If it is a target size (we strive towards it both from higher and lower numbers) then `target` size would be good. `Desired` I interpret as `target`.
> `Comfortable` has a bit other nuance than `minimal`, so if it's not actually the _minimal_ size, then. I can see what was intended with `safe`.
> 

`recommended_section_size` is quite informative, and with a comment as described, I think I'd vote for that as well now.

src/section/section_members.rs
oetyng: Ahh, several sorts of versions. Mm, a good distinction between the versions would be good then (if both are to be kept anyway).

If it tracks any change, then I go back on the previous suggestion, then it seems ok with just `section_version`. The other type of version, which is for elders change, would instead be distinguished somehow (if it is to be kept?),

src/section/section_map.rs
madadam: No, that wouldn't be accurate. But I'll rename it to something like `trusted_key_version` which is more explicit.

src/section/shared_state.rs
```diff
@@ -0,0 +1,854 @@
...
+                    "shared state update - mismatch: old: {:?} --- new: {:?}",
+                    self,
+                    new
+                );
+            }
+
+            *self = new;
+        }
+
+        self.handled_genesis_event = true;
```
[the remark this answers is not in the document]
madadam: Yes. A `None` input is sent to non-elders only because they don't need the full thing, but they still need the `handled_genesis_event` in case they are about to be promoted.

src/section/shared_state.rs
```diff
@@ -0,0 +1,854 @@
...
+    }
+
+    /// Checks if given `PublicId` is an elder in our section or one of our neighbour sections.
+    pub fn is_peer_elder(&self, pub_id: &PublicId) -> bool {
+        self.sections.is_elder(pub_id)
+    }
+
+    /// Returns whether the given peer is elder in our section.
+    pub fn is_peer_our_elder(&self, pub_id: &PublicId) -> bool {
+        self.our_info().is_member(pub_id)
```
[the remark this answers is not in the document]
madadam: Already addressed as per this discussion: https://github.com/maidsafe/routing/pull/2084#discussion_r417202840

src/section/shared_state.rs
madadam: Could of lines above is the check that both sub-sections have enough members which should be enough I think.

src/node/stage/approved.rs
madadam: `is_self_elder` is removed. `EldersInfo::is_member` is now renamed to `contains_elder` to avoid confusion.

src/node/stage/approved.rs
madadam: We used to have small delegating functions like that, but I removed them because I saw no point in having them when we can access the underlying container (`sections`) directly.

src/node/stage/approved.rs
madadam: The assertion existed before, but yeah, it's controversial. I'll remove it.

madadam: OK. Will replace it with log and early exit.

src/section/section_members.rs
dirvine: For ties, I would do this. 
1. Use age_counter
2. If still tied then select xor closest to the previous node that left to create this gap (or similar, even just the lex smallest one)

dirvine: Later I think it will make sense for any promote to elder that we force a name change and give the node a range to get an address for. That way on a split the elders should split circa 50/50. Needs some discussion though

madadam: > Use age_counter

1. I'm already using it, but there can still be ties.
2. I though we were going to remove age_counter and use only age (https://github.com/maidsafe/routing/issues/2078)

dirvine: Good point, in that case then selecting on some criteria we can compute and is not gamable will be the best. 
How about select the Adult to promote that is oldest Adult and on tie then the Adult from the tie that is xor closest to the current section key?

src/consensus/network_event.rs
```diff
@@ -217,41 +217,24 @@ impl Debug for NetworkEvent {
...
-// Change to section elders.
+// Neighbour section elders that got removed/demoted.
 #[derive(Default, Eq, PartialEq, Serialize, Deserialize)]
-pub struct EldersChange {
-    // Neighbour peers that became elders.
-    pub neighbours_added: BTreeSet<P2pNode>,
-    // Neighbour peers that ceased to be elders.
-    pub neighbours_removed: BTreeSet<P2pNode>,
-}
+pub struct NeighbourEldersRemoved(pub BTreeSet<P2pNode>);
```
oetyng, line 222: OK, cool. I like this one. And now we are getting into the really interesting discussions as we shift towards event-centric code; behaviour.

So previously we had `EldersChange` with added and removed Elders.
Generally I would not consider it a good event, because it is so broad, events should be as granular as possible, but no more than useful. It all depends on the behaviour.
If an Elder is always replaced when removed, then they do come in pairs (removed + added). 
In our case I think it's not very common that there are many of them at a time though? So a whole set of added and removed? Seems there's some serious turmoil for that to be ever used. Normally the Elder changes are so far between, that it would be a very delayed event if we were to wait for subsequent removals.
On the other hand, when they do come in bunches, there might be stress on the network, so I guess the thinking is that at that time it is better to lower stress by not sending more events than necessary.
And as a final thought on all that, there's not really much lost in having a set in the event. I've often designed events like that as well, even when it is most often just ever going to be one single item in it.

Moving onwards: But, Elders aren't always replaced when removed, are they? And even if they are, is it so immediate that we should hold off on the removed event just to await the added one, and bundle them in one event?
That's a good reason to split `NeighbourEldersChanged` to `NeighbourEldersRemoved` and `NeighbourEldersAdded`.
Performance considerations come to mind. Also: who needs this information and how important is it to get it sooner rather than later? I have used events like the former, more broad one, as well at times.

So, when deciding granularity, there's always the "it depends" playing in.

Having said all that, I'm now just curious to how you were reasoning behind this specific event, what motivates this level of granularity, and what happened to `NeighbourEldersAdded`?

madadam: Let me clarify what the purpose of this struct really is, which is very narrow: we want to identify peers that are no longer elders because we want to disconnect from them. This is because network connections are limited resource and we don't want to waste them. There is really not much more to it than this. Which is why I'm not entirely convinced that this struct is an event. I think it's more like utility / auxilliiary thing.

Couple of points:
1. There is a kind of inconsistency at the network transport layer in that connection are implicit, but disconnection are explicit. If we somehow made even disconnection implicit, then we wouldn't need this struct at all. 
2. Or if we had a way to list all currently active connections, then we would be able to compare this list with the list of known elders and disconnect the ones we no longer need.
2. Or maybe this whole logic could at least be moved somewhere else? For example, maybe the function that performs the update of the neighbour containers could return the set of no-longer-elders itself?

src/section/section_members.rs
madadam: OK, added an issue for this to be addressed in a different PR: https://github.com/maidsafe/routing/issues/2088

src/section/shared_state.rs
maqi: hmm...  what I mean is that the above check only ensure both sections got enough `members`.
However it doesn't check whether the new sections can all get enough qualified elders.

I am not sure it is required to carry out this check (or, if as you mentioned, `members` refer to `edlers` actually).

madadam: Got you. What matters for the split to happen is only that both sub-sections get enough *members*. The elders are selected only afterwards. There is no such thing as "qualified elder" - any node can become elder if it is among the 7 oldest ones of the section.

maqi: ok. then.

src/consensus/network_event.rs
oetyng: It seems my questions got lost.
So, apparently, you did the change from EldersChange to NeighbourEldersRemoved, my question is what happened to NeighbourEldersAdded?
The previous struct had info on both, now there's only one. 

So, what was the reasoning there? Don't we need to know about the added ones?

About being an event or not, if it described something that happened, then it essentially is an event. If that "something" was already described in an event, then there are numerous considerations: 
- Was this info part of that event? 
- Is this info passed between some subscriber and someone else?
- Why is "someone else" not also a subscriber to the event? (That's the normal way, one event often has multiple subscribers, with various responsibilities.)
If the event is not passed to the subscriber because it has too much info, the event might be having too much responsibility, or there is some key part missing in the call flow between these parts.


There are many unclarities with regards to what is considered with the design and the rationale.
Most often I would say that when we can't clearly describe a pattern and its rationale, it's because there is none.

madadam: Ah, sorry. There is no `NeighbourEldersAdded` because it's not needed. As said, we only need to track the removed elders so we can disconnect from them. We don't need to do anything with the added ones, because the connections are implicit (will be established on first message sent to them).

Regarding event/not event - there is no subscription going on with this thing. It's produced on one line and then consumed couple of lines later within the same function. But perhaps it's still justified to call it an event? Just one with very narrow scope?

oetyng: I don't think this part is very important, but I will have a closer look at the code when I'm at the machine to make some further comments and/or conclusions.

But just quick question, was the added fields in previous struct not used, or did the use of the struct change?

madadam: It wasn't used.

PR #2085, opened 2020-04-27
Handle user message as non elder

madadam:


src/xor_space/xor_name.rs
```diff
@@ -271,7 +271,6 @@ mod tests {
     use super::*;
     use crate::unwrap;
     use bincode::{deserialize, serialize};
-    use rand;
```
maqi, line 274: suprised it was not soptted by the clippy previously.

madadam: maybe the particular lint just wasn't enabled before v1.43?

maqi: yeah, maybe.

PR #2086, opened 2020-04-28
only handle UserMessage as non-elder if addressed directly to us

madadam:
Messages for `DstLocation::Section` or `DstLocation::Prefix` are only meant to be handled by elders and could fail validation if attempted to be handled by non-elder.

Also use the log ident in the panic branch of the `log_or_panic` macro.

### DIFF
--- a/src/node/stage/approved.rs
+++ b/src/node/stage/approved.rs
@@ -240,16 +240,16 @@ impl Approved {
     // Message handling
     ////////////////////////////////////////////////////////////////////////////
 
-    pub fn should_handle_message(&self, msg: &Message) -> bool {
+    pub fn should_handle_message(&self, our_id: &PublicId, msg: &Message) -> bool {
         match &msg.variant {
-            Variant::UserMessage(_)
-            | Variant::Relocate(_)
+            Variant::Relocate(_)
             | Variant::BootstrapRequest(_)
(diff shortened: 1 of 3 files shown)


src/node/stage/approved.rs
```diff
@@ -1681,6 +1681,12 @@ impl Approved {
         self.chain.is_self_elder() || req.elders_version < self.chain.our_info().version()
     }
 
+    // If elder, always handle UserMessage, otherwise handle it only if addressed directly to us
+    // as a node.
+    fn should_handle_user_message(&self, our_id: &PublicId, dst: &DstLocation) -> bool {
+        self.chain.is_self_elder() || dst.as_node().ok() == Some(our_id.name())
```
maqi, line 1687: why not sue `Ok(our_id.name())` so that the convertion of `ok()` (Result -> Option) can be skipped?

madadam: Good point, will change.

madadam: Actually no, it doesn't work because `Result<_, RoutingError>` doesn't implement `Eq` (probably because `RoutingError` doesn't implement it.

maqi: Ah, Ok then.

PR #2089, opened 2020-05-04
Miscellaneous minor refactoring

madadam:
- remove elders history of our section
- simplify `EldersInfo`
- simplify `GenesisPrefixInfo`
- simplify `SectionProofChain`
- use `XorName`, not `PublicId` to lookup peers

### DIFF
--- a/src/consensus/event_accumulator.rs
+++ b/src/consensus/event_accumulator.rs
@@ -354,7 +354,8 @@ mod test {
             EventType::WithSignature => {
                 let elders_info = empty_elders_info();
                 let (sig_payload, keys) = random_section_info_sig_payload(rng);
-                let key_info = SectionKeyInfo::from_elders_info(&elders_info, keys.public_key());
+                let key_info =
+                    SectionKeyInfo::new(elders_info.prefix, elders_info.version, keys.public_key());
                 let event = AccumulatingEvent::SectionInfo(elders_info, key_info);
 
                 TestData {
--- a/src/consensus/event_accumulator.rs
+++ b/src/consensus/event_accumulator.rs
@@ -303,7 +303,7 @@ mod test {
     }
 
     fn empty_elders_info() -> EldersInfo {
-        EldersInfo::new(Default::default(), Default::default(), Default::default()).unwrap()
+        EldersInfo::new(Default::default(), Default::default(), Default::default())
     }
 
     fn random_section_info_sig_payload(rng: &mut MainRng) -> (EventSigPayload, bls::PublicKeySet) {
--- a/src/consensus/event_accumulator.rs
+++ b/src/consensus/event_accumulator.rs
@@ -303,11 +303,7 @@ mod test {
     }
 
     fn empty_elders_info() -> EldersInfo {
-        unwrap!(EldersInfo::new_for_test(
-            Default::default(),
-            Default::default(),
-            Default::default(),
-        ))
+        EldersInfo::new(Default::default(), Default::default(), Default::default()).unwrap()
     }
 
     fn random_section_info_sig_payload(rng: &mut MainRng) -> (EventSigPayload, bls::PublicKeySet) {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -111,9 +111,7 @@ pub use self::{
     network_params::NetworkParams,
     relocation::Overrides as RelocationOverrides,
     routing_table::delivery_group_size,
-    section::{
-        elders_info_for_test, quorum_count, section_proof_slice_for_test, SectionKeyShare, MIN_AGE,
-    },
+    section::{quorum_count, section_proof_slice_for_test, EldersInfo, SectionKeyShare, MIN_AGE},
     xor_space::Xorable,
 };
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -169,11 +169,11 @@ const QUORUM_NUMERATOR: usize = 2;
 /// See `QUORUM_NUMERATOR`.
 const QUORUM_DENOMINATOR: usize = 3;
 
-/// Minimal safe section size. Routing will keep adding nodes until the section reaches this size.
+/// Recommended section size. Routing will keep adding nodes until the section reaches this size.
 /// More nodes might be added if requested by the upper layers.
 /// This number also detemines when split happens - if both post-split sections would have at least
 /// this number of nodes.
-const SAFE_SECTION_SIZE: usize = 100;
+const RECOMMENDED_SECTION_SIZE: usize = 100;
 
 /// Number of elders per section.
 const ELDER_SIZE: usize = 7;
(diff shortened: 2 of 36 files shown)


src/lib.rs
```diff
@@ -169,11 +169,11 @@ const QUORUM_NUMERATOR: usize = 2;
 /// See `QUORUM_NUMERATOR`.
 const QUORUM_DENOMINATOR: usize = 3;
 
-/// Minimal safe section size. Routing will keep adding nodes until the section reaches this size.
+/// Recommented section size. Routing will keep adding nodes until the section reaches this size.
```
maqi, line 172: recommended

src/section/section_map.rs
```diff
@@ -30,7 +29,7 @@ const MAX_RECENT_KEYS: usize = 20;
 #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct SectionMap {
     // Our section including its whole history.
-    our: NonEmptyList<EldersInfo>,
+    our: EldersInfo,
```
maqi, line 32: The comment says there is a history included.
shall the comment be updated?

src/section/section_proof_chain.rs
```diff
@@ -39,258 +38,209 @@ impl SectionProofBlock {
...
+    pub(crate) fn first_key_info(&self) -> &SectionKeyInfo {
+        &self.head
     }
 
-    fn last_trusted_key_info<'a>(
-        &'a self,
-        last_trusted: &'a SectionKeyInfo,
-    ) -> Option<&'a SectionKeyInfo> {
-        let block_offset = last_trusted.version().saturating_sub(self.version) as usize;
+    pub(crate) fn last_key_info(&self) -> &SectionKeyInfo {
```
maqi, line 97: when tail is empty, this function shall return the head ?
as the comment of `new`  function says `Creates new chain consisting of only one block`, where the first is set to head.

src/section/section_proof_chain.rs
```diff
@@ -39,258 +38,209 @@ impl SectionProofBlock {
...
-        } else {
-            // Root of trust not found
-            return None;
+    #[cfg(all(test, feature = "mock"))]
+    pub fn key_infos(&self) -> impl DoubleEndedIterator<Item = &SectionKeyInfo> {
+        iter::once(&self.head).chain(self.tail.iter().map(|block| block.key_info()))
+    }
+
+    /// Returns a slice of this chain starting at the given index.
+    pub(crate) fn slice_from(&self, first_index: usize) -> Self {
```
maqi, line 110: why not construct a vector from the head & tail, then utilize the standard `split_off` function?
so the belowing un-explicit code block can then be avoided?

src/section/shared_state.rs
```diff
@@ -46,10 +46,10 @@ pub struct SharedState {
 impl SharedState {
     pub fn new(
         elders_info: EldersInfo,
-        bls_keys: bls::PublicKeySet,
+        section_pk: bls::PublicKey,
         ages: BTreeMap<PublicId, AgeCounter>,
```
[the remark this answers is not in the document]
madadam: It is so far. The work to change that hasn't been started yet: https://github.com/maidsafe/routing/issues/2078

src/lib.rs
```diff
@@ -169,11 +169,11 @@ const QUORUM_NUMERATOR: usize = 2;
...
 /// More nodes might be added if requested by the upper layers.
 /// This number also detemines when split happens - if both post-split sections would have at least
 /// this number of nodes.
-const SAFE_SECTION_SIZE: usize = 100;
+const RECOMMENDED_SECTION_SIZE: usize = 100;
```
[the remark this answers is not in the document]
madadam: OK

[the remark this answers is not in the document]
madadam: Wonder if we should maybe at some point try to run some high-level network simulations to determine optimal value for this.

src/consensus/event_accumulator.rs
```diff
@@ -303,11 +303,7 @@ mod test {
     }
 
     fn empty_elders_info() -> EldersInfo {
-        unwrap!(EldersInfo::new_for_test(
-            Default::default(),
-            Default::default(),
-            Default::default(),
-        ))
+        EldersInfo::new(Default::default(), Default::default(), Default::default())
```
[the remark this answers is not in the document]
madadam: Yeah, maybe. On the other hand, `EldersInfo` with all default values isn't useful and is only used here because for the purpose of these tests it's sufficient. So I'd probably keep it as it, to at least highlight the fact that something non-standard is going on here.

src/section/section_proof_chain.rs
madadam: Yes, the `head`/`tail` thing is an implementation detail to make it statically checked that the chain always contains at least one block. So if the chain contains only one block, it's both the first and the last block.

src/section/section_proof_chain.rs
madadam: Sounds like a good idea. Will give it a try.

src/lib.rs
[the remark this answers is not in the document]
dirvine: It's a tough one as we can simulate the network storage side, but we would have to guess at mutations and number of clients and so on. The number of clients is the one I am wary of if the section is large. I would prefer lots of small sections, just to allow Elders to cope with mutations and clients. 
We need to work out something to check/test this though and with sections only adding nodes when vaults tell routing then it should mean the first section is close in work to the nth section, at least I imagine it will be.

src/section/section_proof_chain.rs
madadam: Ok, done. Didn't use `split_off`, but still simplified the impl.

src/lib.rs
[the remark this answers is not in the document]
madadam: Yeah, I see. Maybe we can experiment with it on alpha.

src/consensus/event_accumulator.rs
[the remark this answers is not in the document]
maqi: ok, got it

src/section/section_proof_chain.rs
```diff
@@ -39,258 +38,203 @@ impl SectionProofBlock {
...
         }
 
-        let mut current = last_trusted;
-        for block in &self.blocks[block_offset..] {
-            if !validate_next_block(current, block) {
-                return None;
-            }
+        let head_index = std::cmp::min(first_index, self.tail.len()) - 1;
+        let head = self.tail[head_index].key_info().clone();
+        let tail = self.tail[head_index + 1..].iter().cloned().collect();
```
maqi, line 117: say there are only two blocks (one in head, one in tail) currently, and the first_index to slice_from is 1.
Then, `head_index` is 0. so wondering `self.tail[head_index + 1..].iter()` will cause a panic or just an empty vec ?

madadam: Yeah, I also needed to check this, but turns our it returns empty vec, allowing us to keep the code simple: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b5c886758865c4be0b1eff24c90980aa

maqi: good to know that :)

PR #2091, opened 2020-05-06
Remove version and prefix from SectionProofChain

madadam:
This PR simplifies `SectionProofChain` by storing only the keys in it, not versions or prefixes.

Note: the concept of elders version isn't completely gone yet. It's still used in `EldersInfo` to be able to  determine causal relationship between them. We need that to prevent overwriting newer info with older info. After switching to the pull model (in a different PR), this will likely not be necessary anymore.


src/consensus/network_event.rs
```diff
@@ -29,16 +29,16 @@ pub struct AckMessagePayload {
...
     /// The prefix of our section when we acknowledge their SectionInfo of version ack_version.
     pub src_prefix: Prefix<XorName>,
-    /// The version acknowledged.
-    pub ack_version: u64,
+    /// The key acknowledged.
+    pub ack_key: bls::PublicKey,
 }
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub struct SendAckMessagePayload {
```
dirvine, line 36: I think this whole Ack approach can be deleted here actually. Perhaps when the `pull` mechanism in place would be more correct. Happy to waiton that PR

madadam: Yes, that will be part of the pull model.

src/section/section_map.rs
```diff
@@ -281,71 +281,80 @@ impl SectionMap {
...
     }
 
-    pub fn latest_compatible_key(&self, name: &XorName) -> Option<&SectionKeyInfo> {
+    #[cfg_attr(feature = "mock_base", allow(clippy::trivially_copy_pass_by_ref))]
+    pub fn has_key(&self, key: &bls::PublicKey) -> bool {
+        self.keys.values().any(|known_key| known_key == key)
+    }
+
+    pub fn latest_compatible_key(&self, name: &XorName) -> Option<&bls::PublicKey> {
+        // `keys` is already ordered newest to oldest.
```
maqi, line 296: `keys` is a `BTreeMap` keyed by prefix, why the comment says `is already ordered newest to oldest` ?
or you mean the `keys` function? 
however, the function seems chain self.keys first then the self.recent_keys ?

src/section/section_map.rs
```diff
@@ -362,46 +371,46 @@ impl SectionMap {
...
-        if let Some((&old_prefix, &old_version)) = self
+        if let Some((&old_prefix, &old_index)) = self
             .knowledge
             .iter()
             .find(|(other_prefix, _)| other_prefix.is_compatible(&prefix))
         {
-            if old_version >= version || old_prefix.is_extension_of(&prefix) {
-                // Do not overwrite newer version or prefix extensions
+            if old_prefix.is_extension_of(&prefix) || old_index > new_index {
+                // Do not overwrite newer inded or prefix extensions
```
maqi, line 390: index ?

src/section/section_map.rs
```diff
@@ -441,3 +450,235 @@ impl SectionMap {
...
+    fn update_their_knowledge_after_split_from_both_siblings() {
+        let mut rng = rng::new();
+        update_their_knowledge_and_check_proving_index(
+            &mut rng,
+            vec![("1", 1), ("10", 2), ("11", 2)],
+            vec![("10", 2), ("11", 2)],
+        )
+    }
+
+    // Create a `SectionMap` an apply a series of `update_keys` calls to it, then verify the stored
```
maqi, line 608: and apply?

src/section/section_proof_chain.rs
```diff
@@ -6,114 +6,85 @@
...
-use crate::{
-    error::Result,
-    xor_space::{Prefix, XorName},
-};
-
-#[cfg(all(test, feature = "mock"))]
-use std::iter;
+use std::{collections::HashSet, iter};
 
+/// Block of the section proof chain. Contains the section BLS public key and is signed by the
```
maqi, line 11: who will sign the first block?
or this block will only be in `tail`, not in `head`?
in that case, maybe the comment better to mention this explicitly?

src/section/section_proof_chain.rs
```diff
@@ -6,114 +6,85 @@
...
-            log_or_panic!(
-                log::Level::Error,
-                "Invalid next block: {:?} -> {:?}",
-                self.last_key_info(),
-                block
-            );
-            return;
-        }
-
+    /// Pushes a new block into the chain. No validation is performed.
```
maqi, line 45: wondering why the validation is removed?

src/section/section_map.rs
madadam: So `keys` contains only the latest (newest) keys for every section and it is then chained with `recent_keys` which has the previous keys in order newest to oldest. So the resulting iterator also yields the keys newest to oldest. The `find` function then returns the first matching entry which also happens to be the newest entry because of that. Maybe the comment should be clearer?

src/section/section_proof_chain.rs
madadam: The first block is not signed. If this chain is the full section chain, then the first block is the genesis block created by the very first node in the network. If this is just a slice of a chain, the presumably the first block's key is already trusted (by being verified previously) so it doesn't need to be signed.

> or this block will only be in tail, not in head?

That's right. The first block doesn't have signature so it's not stored as `SectionProofBlock` (which is `key` + `signature`) but only as the key alone.

src/section/section_proof_chain.rs
madadam: Mostly because I needed to be able to push invalid blocks for testing purposes. I could have instead created a separate `#[feature = "mock"]` function `push_without_validation` or so, but I decided to keep things simple instead. The validation itself weren't that useful anyway because even when you create an invalid chain, it's going to be rejected by the recipients anyway.

src/messages/accumulating_message.rs
```diff
@@ -261,21 +260,14 @@ mod tests {
...
+        SectionProofChain::new(pk_set.public_key())
     }
 
     fn make_their_key_infos(
         pk_set: &bls::PublicKeySet,
-    ) -> BTreeMap<Prefix<XorName>, SectionKeyInfo> {
-        let key_info = make_section_key_info(pk_set);
-        iter::once((key_info.prefix, key_info)).collect()
+    ) -> BTreeMap<Prefix<XorName>, bls::PublicKey> {
+        iter::once((Prefix::default(), pk_set.public_key())).collect()
```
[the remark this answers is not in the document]
madadam: Will remove this function as it's not really that useful.

src/section/section_map.rs
maqi: Yes, I think the comment shall be clearer.

acutally, according what you explained, there are actually two set of keys chained within the result from `fn keys()`: one is keys of other section, and the other is previous keys (of ours ?).
which, to me, is bit weird/confusing to chain these two together. :(

src/section/section_proof_chain.rs
maqi: In this case, I'd prefer to make the comment to be more explicit, say:
Block of the section proof chain (tail part only)

src/section/section_proof_chain.rs
maqi: hmm...  I thought the `validate` function will be used (though maybe not that useful as you say) for the production code. So, if the removal is due to the testing purpose, I'd prefer make a separate function under the `mock` feature guard.
It's not a strong opinion here, so leave it to you to decide.

src/section/section_map.rs
madadam: No, it contains only their keys. `keys` are the latest ones we know of and `recent_keys` is a few of previous ones, but still theirs. I'll try to make the comment clearer.

src/section/section_proof_chain.rs
madadam: What I meant is that we don't need to validate the block at the time we are pushing it, because we will validate the whole chain as part of message verification. But I'll add a separate function for it anyway.

PR #2095, opened 2020-05-12
API enhancements

### LINKED ISSUE #2092: Add API to retrieve adults from our section that are closest to a given name

### LINKED ISSUE #2093: Add events for node joining and leaving sections
Include the node name and whether it is elder, adult or infant.

### LINKED ISSUE #2094: Expose secret key share
Vaults need it to sign messages meant to be accumulated by clients.

madadam:
- Add method that returns the current section adults sorted by distance to a given name (Closes #2092)
- Add events raised when a node joins and leaves the section (Closes #2093)
- Add method that returns the current BLS secret key share (Closes #2094)

### DIFF
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -30,7 +30,7 @@ use crate::{
     rng::{self, MainRng},
     time::Duration,
     transport::PeerStatus,
-    xor_space::{Prefix, XorName},
+    xor_space::{Prefix, XorName, Xorable},
     TransportConfig, TransportEvent,
 };
 use bytes::Bytes;
@@ -304,7 +304,7 @@ impl Node {
     pub fn closest_known_elders_to<'a>(
         &'a self,
         name: &'a XorName,
-    ) -> impl Iterator<Item = &'a P2pNode> + 'a {
+    ) -> impl Iterator<Item = &'a P2pNode> {
         self.stage
             .approved()
             .into_iter()
@@ -317,6 +317,14 @@ impl Node {
             .approved()
             .into_iter()
             .flat_map(|stage| stage.shared_state.our_adults())
+    }
+
+    /// Returns the adults of our section sorted by their distance to `name` (closest first).
+    /// If we are not elder or if there are no adults in the section, returns empty vec.
+    pub fn our_adults_sorted_by_distance_to(&self, name: &XorName) -> Vec<&P2pNode> {
+        let mut output: Vec<_> = self.our_adults().collect();
+        output.sort_by(|lhs, rhs| name.cmp_distance(lhs.name(), rhs.name()));
+        output
     }
 
     /// Checks whether the given location represents self.
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -402,6 +402,15 @@ impl Node {
         Ok(())
     }
 
+    /// Returns the current BLS secret key share or `RoutingError::InvalidState` if we are not
+    /// elder.
+    pub fn secret_key_share(&self) -> Result<&bls::SecretKeyShare> {
+        self.stage
+            .approved()
+            .and_then(|stage| stage.secret_key_share())
+            .ok_or(RoutingError::InvalidState)
+    }
+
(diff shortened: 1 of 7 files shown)


src/node/mod.rs
```diff
@@ -319,6 +319,14 @@ impl Node {
             .flat_map(|stage| stage.shared_state.our_adults())
     }
 
+    /// Returns the adults of our section sorted by their distance to `name` (closest first).
+    /// If we are not elder or if there are no adults in the section, returns empty vec.
```
maqi, line 323: the comment says `If we are not elder ... returns empty vec `
so just want to confirm that adult won't have the info of other adults within the same section?

src/node/mod.rs
```diff
@@ -394,6 +402,15 @@ impl Node {
         Ok(())
     }
 
+    /// Returns the current BLS secret key share or `RoutingError::InvalidState` if we are not
+    /// elder.
+    pub fn secret_key_share(&self) -> Result<&bls::SecretKeyShare> {
+        self.stage
+            .approved()
+            .and_then(|stage| stage.secret_key_share())
```
maqi, line 410: is that only an elder will have a valid secet_key_share ?
if so, maybe make it more explicit in the comment?

src/node/stage/approved.rs
```diff
@@ -1004,15 +1016,21 @@ impl Approved {
...
         {
             info!("handle Offline: {}", pub_id);
 
             self.members_changed = true;
-            core.transport.disconnect(addr);
+
+            core.transport.disconnect(*info.p2p_node.peer_addr());
             let _ = self.members_knowledge.remove(pub_id.name());
+
+            core.send_event(Event::MemberLeft {
```
maqi, line 1030: do we need to check whether core is elder, before broadcasting this left event out?

src/section/section_members.rs
```diff
@@ -182,30 +181,26 @@ impl SectionMembers {
         }
     }
 
-    /// Remove a member from our section. Returns the SocketAddr and the state of the member before
-    /// the removal.
-    pub fn remove(&mut self, name: &XorName) -> (Option<SocketAddr>, MemberState) {
+    /// Remove a member from our section. Returns the `MemberInfo` from before the removal.
```
maqi, line 184: what is `from before the removal` ?

src/section/shared_state.rs
```diff
@@ -153,17 +155,16 @@ impl SharedState {
     }
 
     /// Removes a member with the given pub_id.
-    /// Returns the member's socket address (or `None` if if wasn't a member) and the member state
-    /// before the removal.
+    /// Returns its `MemberInfo` from before the removal.
```
maqi, line 158: `from before the removal` ?

src/node/mod.rs
madadam: Yes, only elders know about non-elders.

src/node/mod.rs
madadam: Either elders or nodes that are in the process of being promoted to elders and have already seen the DKG result but have not yet seen the `SectionInfo` event. WIll add some comment there.

src/node/stage/approved.rs
madadam: Good catch :+1:

src/section/section_members.rs
madadam: Means it returns the state as it were before the member was removed. But that's probably expected anyway and this commented is just redundant and confusing. I'll remove it.

src/node/mod.rs
maqi: it's just the function name of `approved` misleading me to think that an `adult` may have a `secret_key_share` as well.
yeah, give it a second thought and I don't get a better suggestion on comment here.
so, just leave it then.

madadam: Yeah, it's been brought up before that `approved` might not be the best name, because there is no real approval going on. Should probably be renamed to `joined` or something like that.

PR #2096, opened 2020-05-14
Update transport module and use latest quic-p2p

lionel1704:
This PR also updates the versions of bytes, crossbeam-channel and
parsec

### DIFF
--- a/src/messages/with_bytes.rs
+++ b/src/messages/with_bytes.rs
@@ -114,7 +114,7 @@ mod tests {
 
         let full_msg = unwrap!(Message::from_bytes(bytes));
         let partial_msg = unwrap!(PartialMessage::from_bytes(bytes));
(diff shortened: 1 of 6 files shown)


src/messages/with_bytes.rs
```diff
@@ -114,7 +115,9 @@ mod tests {
 
         let full_msg = unwrap!(Message::from_bytes(bytes));
         let partial_msg = unwrap!(PartialMessage::from_bytes(bytes));
-        let partial_msg_head = unwrap!(PartialMessage::from_bytes(&bytes.slice(0, 40)));
+        let partial_msg_head = unwrap!(PartialMessage::from_bytes(
+            &bytes.slice(Range { start: 0, end: 40 })
```
madadam, line 119: I think you can just do `&bytes.slice(0..40)`

PR #2099, opened 2020-05-19
Pull model

madadam:
This PR changes the mechanism by which sections get updated about the elders of other sections. Previously a section would send ("push") a `NeighbourInfo` message to all its neighbours every time its elder set changed. This PR changes that to a "pull" model:

- Every message now contains a `dst_key` field which contains the latest section key of the destination that the source knows about.
- On receiving a message, we compare the destination section key we know with the latest key of the message proof and if they differ, it means our knowledge of the destination section is out of date.
- We also compare our latest section key with the `dst_key` field of the message and if they differ it means their knowledge of us is out of date
- If either knowledge is out of date, we send them a `NeighbourInfo` message containing the current elder set of our section
- On receiving the message, the destination section updates its knowledge of us (if they were out of date) and/or sends a `NeighbourInfo` of their own to us (if we were out of date).

Currently sending a section message needs to be synchronized via parsec which means to send a `NeighbourInfo` message we must first accumulate a `SendNeighbourInfo` vote. This synchronization requirement will be relaxed in the future.

Additional changes:

- For neighbour sections, we now always update their keys and their elder sets together - that is - we never store their key unless we also have their elders set corresponding to that key. This simplifies some flows.
-  It's no longer necessary for the `NeighbourInfo` message to be successfully delivered on the first try (because it will be sent again as per the flow described above) which allows us to remove some complexity of resending the message to the sibling section in case of split.
- The `DstLocation::Prefix` variant is removed. A message is now always sent to a `XorName` which simplifies the routing logic.

Changes to tests:

- We now need to explicitly send some messages every now and then to trigger the `NeighbourInfo` exchanges as it doesn't happen implicitly. Not doing that would result in some sections knowledge to become so out of date that they would not be able to communicate with each other. This reflects  how the real network will behave which will have constant message traffic driven by the clients.
- We also relax some checks reflecting the fact that the knowledge of other sections might nor necessarily be up-to-date all the time.

### DIFF
--- a/src/consensus/network_event.rs
+++ b/src/consensus/network_event.rs
@@ -150,7 +150,9 @@ impl Debug for AccumulatingEvent {
             Self::Online(payload) => write!(formatter, "Online({:?})", payload),
             Self::Offline(id) => write!(formatter, "Offline({})", id),
             Self::SectionInfo(info, _) => write!(formatter, "SectionInfo({:?})", info),
-            Self::NeighbourInfo(info) => write!(formatter, "NeighbourInfo({:?})", info),
+            Self::NeighbourInfo(elders_info) => {
+                write!(formatter, "NeighbourInfo({:?})", elders_info)
+            }
             Self::SendNeighbourInfo(prefix) => write!(formatter, "SendNeighbourInfo({:?})", prefix),
             Self::TheirKeyInfo { prefix, key } => write!(
                 formatter,
--- a/src/consensus/network_event.rs
+++ b/src/consensus/network_event.rs
@@ -107,6 +107,12 @@ pub enum AccumulatingEvent {
         key: bls::PublicKey,
     },
 
+    // Voted to update their knowledge of our section.
+    TheirKnowledge {
+        prefix: Prefix<XorName>,
+        knowledge: u64,
+    },
+
     // Voted for received AckMessage to update their_knowledge
     AckMessage(AckMessagePayload),
 
@@ -172,6 +178,11 @@ impl Debug for AccumulatingEvent {
                 formatter,
                 "TheirKeyInfo {{ prefix: {:?}, key: {:?} }}",
                 prefix, key
+            ),
+            Self::TheirKnowledge { prefix, knowledge } => write!(
+                formatter,
+                "TheirKnowledge {{ prefix: {:?}, knowledge: {} }}",
+                prefix, knowledge
             ),
             Self::AckMessage(payload) => write!(formatter, "AckMessage({:?})", payload),
             Self::SendAckMessage(payload) => write!(formatter, "SendAckMessage({:?})", payload),
--- a/src/consensus/network_event.rs
+++ b/src/consensus/network_event.rs
@@ -21,26 +21,6 @@ use std::{
 };
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
-pub struct AckMessagePayload {
-    /// The name of the section that message was for. This is important as we may get a message
-    /// when we are still pre-split, think it is for us, but it was not.
-    /// (i.e sent to 00, and we are 01, but lagging at 0 we are valid destination).
-    pub dst_name: XorName,
-    /// The prefix of our section when we acknowledge their SectionInfo of version ack_version.
-    pub src_prefix: Prefix<XorName>,
-    /// The key acknowledged.
-    pub ack_key: bls::PublicKey,
-}
-
-#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
-pub struct SendAckMessagePayload {
-    /// The prefix acknowledged.
-    pub ack_prefix: Prefix<XorName>,
-    /// The key acknowledged.
-    pub ack_key: bls::PublicKey,
-}
-
-#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub struct EventSigPayload {
     /// The public key share for that signature share
     pub pub_key_share: bls::PublicKeyShare,
@@ -113,12 +93,6 @@ pub enum AccumulatingEvent {
         knowledge: u64,
     },
 
-    // Voted for received AckMessage to update their_knowledge
-    AckMessage(AckMessagePayload),
-
-    // Voted for sending AckMessage (Require 100% consensus)
-    SendAckMessage(SendAckMessagePayload),
-
     // Prune the gossip graph.
     ParsecPrune,
 
@@ -184,8 +158,6 @@ impl Debug for AccumulatingEvent {
                 "TheirKnowledge {{ prefix: {:?}, knowledge: {} }}",
                 prefix, knowledge
             ),
-            Self::AckMessage(payload) => write!(formatter, "AckMessage({:?})", payload),
-            Self::SendAckMessage(payload) => write!(formatter, "SendAckMessage({:?})", payload),
             Self::ParsecPrune => write!(formatter, "ParsecPrune"),
             Self::Relocate(payload) => write!(formatter, "Relocate({:?})", payload),
             Self::RelocatePrepare(payload, count_down) => {
--- a/src/consensus/network_event.rs
+++ b/src/consensus/network_event.rs
@@ -78,8 +78,11 @@ pub enum AccumulatingEvent {
 
     SectionInfo(EldersInfo, bls::PublicKey),
 
-    // Voted for received message with info to update neighbour_info.
+    // Voted for received message with info about a neighbour section.
     NeighbourInfo(EldersInfo),
+
+    // Voted to send info about our section to a neighbour section.
+    SendNeighbourInfo(Prefix<XorName>),
 
     // Voted for received message with keys to update their_keys
     TheirKeyInfo {
@@ -148,6 +151,7 @@ impl Debug for AccumulatingEvent {
             Self::Offline(id) => write!(formatter, "Offline({})", id),
             Self::SectionInfo(info, _) => write!(formatter, "SectionInfo({:?})", info),
             Self::NeighbourInfo(info) => write!(formatter, "NeighbourInfo({:?})", info),
+            Self::SendNeighbourInfo(prefix) => write!(formatter, "SendNeighbourInfo({:?})", prefix),
             Self::TheirKeyInfo { prefix, key } => write!(
                 formatter,
                 "TheirKeyInfo {{ prefix: {:?}, key: {:?} }}",
--- a/src/consensus/network_event.rs
+++ b/src/consensus/network_event.rs
@@ -9,6 +9,7 @@
 use crate::{
     consensus::{DkgResultWrapper, Observation, ParsecNetworkEvent},
     id::{P2pNode, PublicId},
+    messages::MessageHash,
     relocation::RelocateDetails,
     section::{EldersInfo, SectionMap},
     Prefix, XorName,
@@ -84,8 +85,9 @@ pub enum AccumulatingEvent {
     // Voted to send info about our section to a neighbour section.
     SendNeighbourInfo {
         dst: Prefix<XorName>,
-        dst_key: bls::PublicKey,
-        our_key_index: u64,
+        // Hash of the incoming message that triggered this vote. It's purpose is to make the votes
+        // triggered by different message unique.
+        nonce: MessageHash,
     },
 
     // Voted for received message with keys to update their_keys
@@ -157,14 +159,10 @@ impl Debug for AccumulatingEvent {
             Self::NeighbourInfo(elders_info, _) => {
                 write!(formatter, "NeighbourInfo({:?}, ..)", elders_info)
             }
-            Self::SendNeighbourInfo {
-                dst,
-                dst_key,
-                our_key_index,
-            } => write!(
+            Self::SendNeighbourInfo { dst, nonce } => write!(
                 formatter,
-                "SendNeighbourInfo {{ dst: {:?}, dst_key: {:?}, our_key_index: {:?} }}",
-                dst, dst_key, our_key_index
+                "SendNeighbourInfo {{ dst: {:?}, nonce: {:?} }}",
+                dst, nonce
             ),
             Self::TheirKeyInfo { prefix, key } => write!(
                 formatter,
--- a/src/consensus/network_event.rs
+++ b/src/consensus/network_event.rs
@@ -84,7 +84,7 @@ pub enum AccumulatingEvent {
 
     // Voted to send info about our section to a neighbour section.
     SendNeighbourInfo {
-        dst: Prefix<XorName>,
+        dst: XorName,
         // Hash of the incoming message that triggered this vote. It's purpose is to make the votes
         // triggered by different message unique.
         nonce: MessageHash,
--- a/src/consensus/network_event.rs
+++ b/src/consensus/network_event.rs
@@ -226,7 +226,7 @@ pub struct NeighbourEldersRemoved(pub BTreeSet<P2pNode>);
 
 impl NeighbourEldersRemoved {
     pub fn builder(sections: &SectionMap) -> NeighbourEldersRemovedBuilder {
-        NeighbourEldersRemovedBuilder(sections.other_elders().cloned().collect())
+        NeighbourEldersRemovedBuilder(sections.neighbour_elders().cloned().collect())
     }
 }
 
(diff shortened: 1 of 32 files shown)


src/section/section_map.rs
```diff
@@ -346,13 +308,15 @@ impl SectionMap {
         let _ = self.keys.insert(prefix, new_key);
     }
 
-    pub fn get_knowledge(&self, prefix: &Prefix<XorName>) -> Option<u64> {
-        self.knowledge.get(prefix).copied()
+    /// Returns the the index of the public key in our_history that will be trusted by the given
+    /// section.
+    pub fn knowledge_by_section(&self, prefix: &Prefix<XorName>) -> u64 {
+        self.knowledge.get(prefix).copied().unwrap_or(0)
```
maqi, line 314: just to ensure that `index 0` won't point to a valid entry, right?

src/section/section_proof_chain.rs
```diff
@@ -92,6 +92,11 @@ impl SectionProofChain {
         1 + self.tail.len()
     }
 
+    /// Index of the last key in the chain.
+    pub(crate) fn last_key_index(&self) -> u64 {
```
maqi, line 96: empty tail (return with 0) will pointing to `head`, right?

src/section/shared_state.rs
```diff
@@ -314,6 +310,67 @@ impl SharedState {
...
+    ) -> Vec<AccumulatingEvent> {
+        let (&prefix, new_key) = if let Ok(pair) = src.as_section_prefix_and_key() {
+            pair
+        } else {
+            return vec![];
+        };
+
+        let is_neighbour = self.our_prefix().is_neighbour(&prefix);
+
+        let mut events = Vec::with_capacity(2);
```
maqi, line 329: so, does it mean `AccumulatingEvent::TheirKeyInfo` and `AccumulatingEvent::TheirKnowledge` won't to be voted for at the same time?

tests/mock_network/mod.rs
```diff
@@ -547,3 +550,57 @@ fn pause_node_and_poll(env: &Environment, nodes: &mut Vec<TestNode>) -> PausedSt
...
+        all_nodes_joined(nodes, 0..nodes.len())
+    });
+
+    // Remove some elders from B.
+    for _ in 0..num_elders_to_remove {
+        let name = remove_elder_from_section(&mut nodes, &prefix_b);
+        poll_until(&env, &mut nodes, |nodes| node_left(nodes, &name));
+    }
+
+    // Send a message from B to A to trigger the update request.
```
maqi, line 601: shall we have a check before sending the message that the knowledge is now out of date ?

tests/mock_network/node_ageing.rs
```diff
@@ -296,18 +296,32 @@ fn churn_until_age_counter(
 
         match churn {
             Churn::Add => {
-                add_node_to_section(env, nodes, prefix);
+                // We are making lot of churn here and so it might happen that other section's view
+                // of this section could become so out-of-date that none of the nodes they know is
+                // still online. If we picked such node for the bootstrap node, the bootstrapping
+                // would fail because the node would keep redirecting the joining node to
+                // non-existing peers. To avoid this and to keep things simple, we make sure we
+                // bootstrap off a node from the same section.
```
maqi, line 304: `bootstrap off a node from the same section` , don't get it, `same section` as what?

src/consensus/network_event.rs
```diff
@@ -98,20 +79,28 @@ pub enum AccumulatingEvent {
 
     SectionInfo(EldersInfo, bls::PublicKey),
 
-    // Voted for received message with info to update neighbour_info.
-    NeighbourInfo(EldersInfo),
+    // Voted for received message with info about a neighbour section.
+    NeighbourInfo(EldersInfo, bls::PublicKey),
```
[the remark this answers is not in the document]
madadam: No, both the `EldersInfo` and the `PublicKey` are of their section.

src/messages/variant.rs
```diff
@@ -25,21 +25,19 @@ use std::{
 /// Message variant
 pub enum Variant {
     /// Inform neighbours about our new section.
-    NeighbourInfo(EldersInfo),
+    NeighbourInfo {
+        /// `EldersInfo` of the neighbour section.
+        elders_info: EldersInfo,
+        /// Nonce that is derived from the incoming message that triggered sending this
+        /// `NeighbourInfo`. It's purpose is to make sure that `NeighbourInfo`s that are otherwise
```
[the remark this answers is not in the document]
madadam: The idea was to make the typename highlighted (using backticks) in the rendered docs but also put it into plural which works better grammatically. I didn't want to backtick the "s" too as it's not part of the type name.

src/node/mod.rs
```diff
@@ -671,22 +669,11 @@ impl Node {
...
-                    ack_key,
-                } => {
-                    stage.handle_ack_message(
-                        src_prefix,
-                        ack_key,
-                        *msg.src.as_section()?,
-                        *msg.dst.as_section()?,
-                    )?;
+                    let src_key = *msg.src.as_section_key()?;
+                    let _: &XorName = msg.dst.as_section()?;
```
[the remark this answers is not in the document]
madadam: Exactly.

src/node/mod.rs
```diff
@@ -671,22 +669,11 @@ impl Node {
                 _ => unreachable!(),
             },
             Stage::Approved(stage) => match msg.variant {
-                Variant::NeighbourInfo(elders_info) => {
+                Variant::NeighbourInfo { elders_info, .. } => {
```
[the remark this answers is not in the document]
madadam: No, it's only needed to make the message unique, but not needed for processing it.

src/section/section_map.rs
madadam: Yes, index `0` points to the genesis key which every section knows.

src/section/section_proof_chain.rs
madadam: Yes.

src/section/shared_state.rs
madadam: So the only possible combination of events that can be voted are these:
- `TheirKeyInfo` only
- `TheirKeyInfo` and `TheirKnowledge`
- `SendNeighbourInfo` only
- `SendNeighbourInfo` and `TheirKnowledge`
which is why it's enough to pre-allocate the vec to 2 elements.

tests/mock_network/mod.rs
madadam: We could, but I don't think we should. Although the knowledge is out of date at this point, we don't require it to be. What we require it only that the knowledge is eventually updated after the message is sent.

tests/mock_network/node_ageing.rs
madadam: from the same section that we are joining. I'll update the comment.

src/consensus/network_event.rs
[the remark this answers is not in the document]
maqi: ah, ok.
then, just want to ensure that here we only need `bls::PublicKey`, not `bls::PublicKeySet`, right?
I think bls::PublicKeySet is the info of the neighbour section, meanwhile bls::PublicKey is only about one node of it?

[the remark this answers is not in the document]
madadam: No, both `PublicKey` and `PublicKeySet` are about the whole section. `PublicKey` is all you need to verify a message against a given `Signature`. You only need `PublicKeySet` when you also need to combine `SignatureShare`s.

src/messages/variant.rs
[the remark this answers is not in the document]
maqi: Ah, I am ok with the backtick. I am just wondering about the `that are otherwise ...` part  :)

src/node/mod.rs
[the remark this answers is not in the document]
maqi: in that case, I'd prefer a more explicit check & error return, instead of this unused unwrap.

src/section/shared_state.rs
maqi: maybe better put the above as a comment in the code ?
it's really bit hard to figure this out by only reading the judgements  :)

tests/mock_network/mod.rs
maqi: > Although the knowledge is out of date at this point, we don't require it to be

So you mean the knowledge could already be updated (due to some other messaging flow). And the extra message is just a `safe-net` to guarantee that update happens, right?

src/messages/variant.rs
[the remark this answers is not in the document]
madadam: I see. The full sentence is  "make sure that NeighbourInfo`s that are otherwise identical but triggered by different messages are not filtered out". Not sure what's wrong with it?

src/node/mod.rs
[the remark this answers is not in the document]
madadam: Something like this:
```rust
if !msg.dst.is_section() {
    return Err(BadLocation);
}
```
?

src/messages/variant.rs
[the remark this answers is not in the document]
maqi: you mean `ensure that identical NeighbourInfos but triggered by different messages are not filtered out` ?
the word of `otherwise` made me bit confused.  :(

src/node/mod.rs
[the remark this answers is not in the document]
maqi: yeah, better  :)

tests/mock_network/mod.rs
madadam: It could in theory, but there is no other message flow in this case. My point was that we don't really care about the knowledge being out of date before the message. We only care about it being updated afterwards. And I think we should assert only things we care about. 

That said, adding the assertion would slightly change the meaning of the test as it would be about explicitly requiring that it's the message that causes the knowledge to get updated and not something else. There can be some value in that, I'm not sure. If you think it's worthwhile, I'll add it there.

src/messages/variant.rs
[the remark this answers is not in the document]
madadam: Yeah. Will change the wording.

src/node/stage/approved.rs
```diff
@@ -331,15 +356,37 @@ impl Approved {
...
+            | Variant::ParsecRequest(..)
+            | Variant::ParsecResponse(..)
+            | Variant::Bounce { .. } => {
+                if self.verify_message(msg)? {
+                    Ok(MessageAction::Handle)
+                } else {
+                    Ok(MessageAction::Discard)
+                }
+            }
+            Variant::Ping => Ok(MessageAction::Discard),
```
maqi, line 385: seems this entry can now be merged with BootstrapResponse (both just for discarding) ?

tests/mock_network/utils.rs
```diff
@@ -322,6 +325,41 @@ pub fn consensus_reached(
...
 
+// Returns whether all elders of section `a` know at least `threshold` online nodes from
+// section `b`.
+pub fn section_knowledge_is_up_to_date(
+    nodes: &[TestNode],
+    a: &Prefix<XorName>,
+    b: &Prefix<XorName>,
+    threshold: usize,
+) -> bool {
+    let names_b: Vec<_> = nodes_with_prefix(nodes, b)
```
maqi, line 336: just wondering that for `a`, it is the function of `elders_with_prefix` being used, but here is `nodes_with_prefix`, is it ok?

tests/mock_network/utils.rs
```diff
@@ -322,6 +325,41 @@ pub fn consensus_reached(
...
+    for node_a in elders_with_prefix(nodes, a) {
+        let count = names_b
+            .iter()
+            .filter(|name_b| node_a.inner.is_peer_elder(name_b))
+            .take(threshold)
+            .count();
+
+        if count < threshold {
+            trace!(
+                "Node {}({:b}) knows only {}/{} online nodes from {:?}",
```
maqi, line 349: online elders ?

tests/mock_network/utils.rs
```diff
@@ -743,38 +802,137 @@ pub fn verify_invariants_for_nodes(env: &Environment, nodes: &[TestNode]) {
...
+    for node in elders_with_prefix_mut(nodes, &src) {
+        node.inner
+            .send_message(src_location, dst_location, content.clone())
+            .unwrap()
+    }
+}
+
+// Poll until all sections have up-to-date knowledge of their neighbour sections.
+// We consider section A's knowledge of its neighbour section B as up-to-date if every elder from A
+// knows at least `threshold` online nodes from B.
```
maqi, line 831: online elders from B ?

tests/mock_network/utils.rs
```diff
@@ -743,38 +802,137 @@ pub fn verify_invariants_for_nodes(env: &Environment, nodes: &[TestNode]) {
...
+pub fn update_neighbours_and_poll(env: &Environment, nodes: &mut [TestNode], threshold: usize) {
+    info!("update_neighbours_and_poll start");
+
+    let mut rng = env.new_rng();
+
+    // Max number of times we will try to send messages and wait until the knowledge gets updated.
+    let max_sends = 3;
```
maqi, line 838: Is it not guaranteed that one message will be enough to trigger the update?

tests/mock_network/utils.rs
```diff
@@ -743,38 +802,137 @@ pub fn verify_invariants_for_nodes(env: &Environment, nodes: &[TestNode]) {
...
+    let prefixes: Vec<_> = current_sections(nodes).collect();
+    prefixes
+        .into_iter()
+        .tuple_combinations()
+        .filter(|(a, b)| a.is_neighbour(b))
+        .filter(move |(a, b)| {
+            !section_knowledge_is_up_to_date(nodes, a, b, threshold)
+                || !section_knowledge_is_up_to_date(nodes, b, a, threshold)
+        })
+        .map(|(a, b)| if a < b { (a, b) } else { (b, a) })
```
maqi, line 885: I think when there is an out-of-date happens, there will be `(a, b)` and `(b, a)` both be reported. 
However, with this ordering, it means within the iterator, there will be two same entries ?

tests/mock_network/mod.rs
maqi: > That said, adding the assertion would slightly change the meaning of the test as it would be about explicitly requiring that it's the message that causes the knowledge to get updated and not something else. There can be some value in that, I'm not sure. If you think it's worthwhile, I'll add it there.

Yes, I agree adding an assertion changes the meaning slightly as you say. However, if so, we may need to know what that extra cause the update? So far, as I understand, it shall only be the message flow causing the update?

tests/mock_network/utils.rs
madadam: Yes, that's on purpose. It's enough to know nodes of any type, not necessarily elders. If we only know non-elders, we can still send them a message and the non-elders will relay it to their elders so everything still works.

tests/mock_network/utils.rs
madadam: See https://github.com/maidsafe/routing/pull/2099#discussion_r427918835

tests/mock_network/utils.rs
madadam: Same.

tests/mock_network/utils.rs
madadam: No, because `tuple_combinations` ignores order: https://docs.rs/itertools/0.9.0/itertools/trait.Itertools.html#method.tuple_combinations

The ordering is actually unnecessary now. I put it there to make sure the function doesn't return `(a, b)` once and `(b, a)` the second time, but then I changed the code that required it. Will remove it to simplify the code and avoid confusion.

tests/mock_network/mod.rs
madadam: OK, will add the assert.

tests/mock_network/utils.rs
maqi: ah, good to know that :)

tests/mock_network/utils.rs
maqi: ah, ok then

tests/mock_network/utils.rs
madadam: Yes, it can fail for at least these reasons:
1) the `UserMessage` might fail to accumulate at the source if the section is undergoing churn
2) the `SendNeighbourInfo` and/or `NeighbourInfo` votes might fail to accumulate if the section doing the voting is undergoing split

We could solve 1 by synchronizing the send via parsec and 2 by resending the message to the sibling and re-voting, which is roughly what we were doing previously. We needed to do that because the `NeighbourInfo` message was sent only once and so we needed to make sure it's delivered. That was a flawed approach, because the delivery can never be 100% guaranteed. Now the message is sent every time we detect the knowledge is out of date and so it isn't that critical that its delivery succeeds. This allows as to remove some complexity.

maqi: Ah, ok. Thanks for the clarification.

PR #2111, opened 2020-06-08
feat/routing: using BLS-DKG instead of parsec::DKG

maqi:


src/consensus/dkg.rs
```diff
@@ -0,0 +1,133 @@
...
+            self.secret_key_share.is_some()
+        )
+    }
+}
+
+#[derive(Clone)]
+/// Wrapper providing necessary functionality to be stored in an Observation.
+/// Do not add these directly to DkgResult as they are not semantically correct for it.
+/// Ignore secret key for all of these: blocks are expected to be the same between peers.
+pub struct DkgResultWrapper(pub DkgResult);
```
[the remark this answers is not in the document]
maqi: ok, will check if can get this removed.

src/consensus/parsec.rs
```diff
@@ -232,7 +231,7 @@ impl ParsecMap {
         &mut self,
         participants: BTreeSet<PublicId>,
         vote_id: &FullId,
-    ) -> Option<DkgResult> {
+    ) -> Option<parsec::DkgResult> {
```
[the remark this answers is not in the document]
maqi: will check if can get this removed.

src/messages/variant.rs
```diff
@@ -75,6 +77,15 @@ pub enum Variant {
         /// The latest parsec version of the recipient of `message`.
         parsec_version: u64,
     },
+    /// Message exchanged for DKG process.
+    DKGMessage {
+        /// The identifier of the key_gen instance this message is about.
+        /// Currently just using the participants.
+        /// TODO: may need to consider using other unique identifying approach.
+        participants: BTreeSet<PublicId>,
```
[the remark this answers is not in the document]
maqi: it turned out this field format is useful for having newly promoted node to handle DKGMessage.
so maybe will leave this as it is.

src/messages/variant.rs
```diff
@@ -75,6 +77,15 @@ pub enum Variant {
...
+        /// The serialized DKG message.
+        message: Bytes,
```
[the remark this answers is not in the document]
maqi: avoiding a hashing requirement.
try best avoiding updating DKG crate.

src/mock/parsec/tests.rs
```diff
@@ -221,85 +219,85 @@ fn reevaluate_previously_insufficient_votes() {
...
-            (index, sig_share)
-        })
-        .tuple_combinations()
-    {
-        let sig = public_key_set
-            .combine_signatures(vec![(a_index, &a_sig_share), (b_index, &b_sig_share)]);
-        assert!(public_key_set.public_key().verify(&sig.unwrap(), message));
-    }
-}
+// // Distributed key generation
```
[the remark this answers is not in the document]
maqi: yeah, will do.

src/node/stage/approved.rs
```diff
@@ -68,6 +69,8 @@ pub struct Approved {
     // Flag indicating that our section members changed (a node joined or left) and we might need
     // to change our elders.
     members_changed: bool,
+    // Map for DKG generators
+    key_gen_map: BTreeMap<BTreeSet<PublicId>, KeyGen<FullId>>,
```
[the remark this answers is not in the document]
maqi: will have a try with merge these two into one.
regarding the DkgId, as mentioned above, probably will remain as using BTreeSet<PublicId>

src/messages/variant.rs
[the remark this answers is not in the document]
madadam: I see, makes sense.

src/node/stage/approved.rs
```diff
@@ -979,12 +1045,42 @@ impl Approved {
         for info in new_infos {
             let participants: BTreeSet<_> = info.elder_ids().copied().collect();
             let _ = self.dkg_cache.insert(participants.clone(), info);
-            self.vote_for_event(AccumulatingEvent::StartDkg(participants));
+
+            // TODO: allow threshold to be configured.
+            if let Ok((key_gen, message)) = KeyGen::initialize(
+                &core.full_id,
+                participants.len() * 2 / 3,
```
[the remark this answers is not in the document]
dirvine: Also we should test this at >50% as I cannot see any more why we need 2/3 which is a graph/gossip related constraint.

src/node/stage/approved.rs
```diff
@@ -245,6 +272,85 @@ impl Approved {
...
             }
+            let ori_ticks = self.dkg_ticks;
+            if let Some(ticks) = ori_ticks {
+                if ticks > 5 {
+                    self.dkg_ticks = Some(0);
+                    self.progress_dkg(core);
+                } else {
+                    self.dkg_ticks = Some(ticks + 1);
+                }
+            }
```
madadam, line 283: It would be cleaner to have a separate timer token for this stuff. Something like:
```rust
if token == self.gossip_timer_token { // rename `timer_token` to `gossip_timer_token` for clarity
    // ...
} else if token == self.dkg_timer_token {
    self.dkg_timer_token = core.timer.schedule(DKG_PROGRESS_INTERVAL);
    self.progress_dkg(core);
}
```

src/node/stage/approved.rs
```diff
@@ -971,18 +1222,118 @@ impl Approved {
...
+            if participants
+                .iter()
+                .all(|participant| !elder_ids.contains(participant))
+            {
+                trace!(
+                    "Triggering DKG process for sibling section during splitting {:?}",
+                    participants
+                );
+                // TODO: allow threshold to be configured.
+                let threshold = participants.len() * 2 / 3;
```
madadam, line 1253: Use the `QUORUM_NUMERATOR` and `QUORUM_DENOMINATOR` constants here.

src/node/stage/approved.rs
```diff
@@ -971,18 +1222,118 @@ impl Approved {
...
+                "already have key_gen of {:?}-{:?}",
+                parsec_version,
+                participants
+            );
+            self.check_dkg(core);
+            return;
+        }
+
+        // TODO: allow threshold to be configured.
+        let threshold = participants.len() * 2 / 3;
```
madadam, line 1298: ...same here. Or maybe define a function for it: `fn dkg_threshold(num_participants: usize) -> usize`

src/node/stage/approved.rs
```diff
@@ -1423,14 +1842,33 @@ impl Approved {
...
+    fn add_force_gossip_peer(
+        &mut self,
+        elders_info: &EldersInfo,
+        old_prefix: Prefix<XorName>,
+        was_elder: bool,
+    ) {
+        if was_elder
+            && (elders_info.prefix == old_prefix || elders_info.prefix.is_extension_of(&old_prefix))
+        {
+            for (_, id) in elders_info.elders.iter() {
```
madadam, line 1854:
```rust
for id in elders_info.elders.values()
```
(I think clippy should be complaining about this)

src/node/stage/approved.rs
```diff
@@ -1935,6 +2377,21 @@ impl Approved {
...
+            .shared_state
+            .our_members
+            .elder_candidates(core.network_params.elder_size)
+            .contains_key(core.id().name())
+        {
+            self.is_our_elder(core.id())
+        } else {
+            true
+        }
+    }
```
madadam, line 2393: Why do we need this?

src/node/tests/elder.rs
```diff
@@ -482,6 +482,7 @@ fn when_accumulate_online_then_node_is_added_to_our_members() {
 }
 
 #[test]
+#[ignore]
```
madadam, line 485: Put `FIXME` comments to all these `ignore`s, to remind us to fix them later.

src/section/section_update_barrier.rs
```diff
@@ -55,12 +55,12 @@ impl SectionUpdateBarrier {
         self.try_get_details(our_prefix)
     }
 
-    fn try_get_details(&mut self, our_prefix: &Prefix<XorName>) -> Option<SectionUpdateDetails> {
+    fn try_get_details(&self, our_prefix: &Prefix<XorName>) -> Option<SectionUpdateDetails> {
```
madadam, line 58: What's the reason for this change? The idea was that once we collect all the events, we return them and reset the barrier into its default state. Now we are keeping the events in forever? Seems like it can cause issues when old events are mixed with new ones, no?

src/node/stage/approved.rs
madadam: Or use the `quorum_count` (in `src/section/elders_info.rs`)  function we already have, just subtract one from the result?

tests/mock_network/accumulate.rs
```diff
@@ -43,10 +43,11 @@ fn messages_accumulate_with_quorum() {
     let dst = DstLocation::Node(*nodes[closest_elder_index].name()); // The closest node.
     let content = gen_bytes(&mut rng, 8);
 
-    // The BLS scheme will require more than `participants / 3`
+    // The BLS scheme will require more than `participants * 2 / 3`
     // shares in order to construct a full key or signature.
     // The smallest number such that `quorum > threshold`:
-    let threshold = elder_size.saturating_sub(1) / 3;
+    // TODO: make this configurable.
+    let threshold = elder_size * 2 / 3;
```
madadam, line 50: Same here: https://github.com/maidsafe/routing/pull/2111#discussion_r447573283

src/node/stage/approved.rs
```diff
@@ -68,6 +69,13 @@ pub struct Approved {
     // Flag indicating that our section members changed (a node joined or left) and we might need
     // to change our elders.
     members_changed: bool,
+    // Map for DKG generators
+    key_gen_map: BTreeMap<(BTreeSet<PublicId>, u64), KeyGen<FullId>>,
+    dkg_ticks: Option<usize>,
+    pending_dkg_msg: BTreeMap<(BTreeSet<PublicId>, u64), VecDeque<Bytes>>,
+    pending_accumulated_events: VecDeque<(AccumulatingEvent, Proof)>,
+    completed_dkg: BTreeSet<(BTreeSet<PublicId>, u64)>,
+    sibling_dkg_result_cache: BTreeMap<BTreeSet<PublicId>, DkgResultWrapper>,
```
madadam, line 78: Quite a lot of stuff added here. To maintain better separation of concerns, I'd consider moving this (and also the new dkg-related methods) over to `SectionKeysProvider` (or a separate `Dkg` module perhaps if that makes more sense).

src/consensus/parsec.rs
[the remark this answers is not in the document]
dirvine: Yip let's not mock here at all if possible. Critical to not use parsec methods as well. We will remove parsec soon.

src/node/stage/approved.rs
maqi: will do. however strangely clippy doesn't complain about this.

src/node/stage/approved.rs
maqi: the exsiting testing code only checks `joined` for the new node and will bootstrap next immediately.
doing so, some new node could be bootstrapped too early which messed up with its own DKG voting round.
This is to ensure that every supposed to have DKG node does completed, then to bootstrap next.

Non elder node shall not be affected.

src/section/section_update_barrier.rs
maqi: Barrier will still be reset after updating key got completed.
The reason of this change is that:

1, in case we get the SectionInfo first, then complete the DKG, the taken events from barrier will be `lost` even the later on update of section key got failed. Which means the node lost the chance to recover.

src/node/stage/approved.rs
maqi: I have updated BLS_DKG to carry out caching internally, so some shall be removed.
Also, I will squash some into one struct once all tests settled.

src/node/stage/approved.rs
```diff
@@ -191,14 +211,21 @@ impl Approved {
...
             Ok(event) => self.consensus_engine.vote_for(event),
-            Err(error) => log_or_panic!(
-                log::Level::Error,
-                "Failed to create NetworkEvent: {}",
-                error
-            ),
+            Err(error) => {
+                if AccumulatingEvent::ParsecPrune == event {
+                    debug!("ParsecPrune doesn't cause panic");
+                } else {
```
madadam, line 220: Why is this special handling needed? Why would the `ParsecPrune` event fail? Is this because you might not have any key share at this point yet? If so, then I think it would be better (cleaner) to do the check before calling `vote_for_event(ParsecPrune)`.

src/node/stage/approved.rs
madadam: OK, I see. So given the current ongoing discussion about reducing the use of mocks, I think it would be much better to not do this and instead modify the tests themselves. For example, I recently added `poll_until_minimal_elder_count` which waits until the section reaches `MIN_ELDER_SIZE` elders before proceeding. Perhaps that, or something similar to that could be used in this case too?

maqi: ok, I will have a check if can use `poll_until_minimal_elder_count ` instead.

src/node/stage/approved.rs
maqi: > Is this because you might not have any key share at this point yet?

Yes. will move the check to the calling place.

src/node/stage/approved.rs
```diff
@@ -323,6 +364,11 @@ impl Approved {
                     return Ok(MessageStatus::Useless);
                 }
             }
+            Variant::DKGMessage { .. } => {
+                if !self.is_our_elder(our_id) {
+                    return Ok(MessageStatus::Unknown);
+                }
```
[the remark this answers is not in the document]
maqi: Ah, missed to answer this comment.
yes, they shall join the DKG process. This has been fixed in the later on tweaks.

src/node/stage/approved.rs
maqi: this is actually being used by the `node_joined` testing function call, which inside doesn't know whether the targeted node will be an elder or not.
refactoring this will mean modify many calling places within the testing code. 
I'd prefer to do it in a separate PR.

src/consensus/parsec.rs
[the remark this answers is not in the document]
maqi: Had a check with attempt to remove this.
However, it turned out to affects many mock_parsec functions, and testing logic as well.
I'd prefer get this done in a separate PR, which will be much cleaner by itself.
Or even just wait for the removal of the parsec, so all the mock_parsec stuff can be gone at once.

src/consensus/dkg.rs
```diff
@@ -0,0 +1,223 @@
...
+    fmt::{self, Debug, Formatter},
+    mem,
+};
+
+/// Generate a BLS SecretKeySet for the given number of participants.
+/// Used for generating first node, or for test.
+pub fn generate_secret_key_set(rng: &mut MainRng, participants: usize) -> bls::SecretKeySet {
+    // The BLS scheme will require more than `participants / 3`
+    // shares in order to construct a full key or signature.
+    let threshold = participants.saturating_sub(1) / 3;
```
madadam, line 29: Use the `threshold_count` function here too.

src/node/stage/approved.rs
madadam: OK, I see. I'm just worried this might potentially hide some issues. For example, there is a test `simultaneous_joining_nodes` which tries to test that the network handles many nodes joining at the same time. With this change, we are basically implicitly limiting the number of simultaneous nodes which makes the test less aggressive and thus less likely to catch bugs. There might be other tests similarly affected by this.

That said, I'm OK with postponing this to a separate PR. Just maybe add a `TODO` comment there to remind us about this.

src/consensus/parsec.rs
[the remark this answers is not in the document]
madadam: I'm OK with dealing with this in a separate PR. Maybe just add a `TODO` with something like "remove this function because we no longer use parsec for DKG".

src/node/stage/approved.rs
maqi: Adding multiple adult nodes is not affected. However, I agree with the intial phase that all new nodes are elder, this part could be affected.
I will add a TODO comment in.

src/mock/parsec/tests.rs
```diff
@@ -710,15 +628,15 @@ where
...
-}
+// fn extract_dkg_result(node: &mut Parsec<Payload, PeerId>) -> (BTreeSet<PeerId>, DkgResult) {
+//     match node.poll().unwrap().payload() {
+//         Observation::DkgResult {
+//             participants,
+//             dkg_result,
+//         } => (participants.clone(), dkg_result.0.clone()),
+//         x => panic!("Unexpected block {:?}", x),
+//     }
+// }
```
madadam, line 639: Better delete this if it's not needed.

tests/mock_network/accumulate.rs
```diff
@@ -43,10 +44,11 @@ fn messages_accumulate_with_quorum() {
     let dst = DstLocation::Node(*nodes[closest_elder_index].name()); // The closest node.
     let content = gen_bytes(&mut rng, 8);
 
-    // The BLS scheme will require more than `participants / 3`
+    // The BLS scheme will require more than `participants * 2 / 3`
     // shares in order to construct a full key or signature.
     // The smallest number such that `quorum > threshold`:
-    let threshold = elder_size.saturating_sub(1) / 3;
+    // TODO: make this configurable.
```
madadam, line 50: Is this `TODO` still needed?

maqi: think so, as currently just caculated from const?

madadam: Maybe that's enough? We can always change the consts if we want to. Or do you think we might need to change the threshold at runtime? Or have different threshold in different part of the code?

maqi: I thought we may have different threshold in different part of code.
But I agree this can be adjusted via const easily.
I will remove the TODO

src/node/stage/approved.rs
```diff
@@ -902,18 +1016,95 @@ impl Approved {
...
+
         for info in new_infos {
             let participants: BTreeSet<_> = info.elder_ids().copied().collect();
-            let _ = self.dkg_cache.insert(participants.clone(), info);
-            self.consensus_engine.vote_for(NetworkEvent {
-                payload: AccumulatingEvent::StartDkg(participants),
-                proof_share: None,
-            });
+            let parsec_version = self.consensus_engine.parsec_version();
+            let dkg_key = (participants.clone(), parsec_version);
```
madadam, line 1026: Why does the parsec version need to be a part of the dkg key?

src/node/stage/approved.rs
```diff
@@ -902,18 +1016,95 @@ impl Approved {
...
+                        "Failed handle sibling notified dkg_result {:?} - {:?}",
+                        dkg_key, err
+                    );
+                    self.dkg_voter.insert_dkg_result(dkg_key, dkg_result);
+                }
+                continue;
+            }
+
+            // In case all the current elders split into one side of the sub-section, triggering the
+            // sibling section's DKG process by send them an Initial DKG message.
```
madadam, line 1041: I wonder whether we should do this always, not just when all elders went only to one subsection. For example, say 6 elders went to one subsection and 1 to the other, but that 1 is faulty/malicious and never broadcasts the necessary dkg messages to the rest. Wouldn't that be a problem?

tests/mock_network/accumulate.rs
madadam: Yeah, currently we have 2/3 for parsec votes and 1/3 for message signatures. But I think we eventually want to have 1/2 everywhere. For now however, we will only have 2/3 everywhere.

src/node/stage/approved.rs
dirvine: Yes I think we should always do this and then we are consistent.

tests/mock_network/accumulate.rs
dirvine: Yes, until parsec is gone then we can get to >50% :+1:

src/node/stage/approved.rs
maqi: just to avoid the case that voting of {A,B,C} then {A,B,C,D} then {A,B,C} got messed up. 
quite rare, but put just as a safty.

src/node/stage/approved.rs
maqi: yes, it could be a problem.
Was just want to reduce some messaging.
will make the notification happens always.

dirvine: Later I would like to force new node names when nodes become elders. So we try and spread them evenly in the address space. IT will help but not guarantee this never happens.

src/node/stage/approved.rs
dirvine: Hmm, if we can only vote a node joined or left (for simplicity) then this is OK. If D is voted joined then he is until he is voted to remove. This is the crdt part here really and we should use that. So if we had a section with ABC and were adding D then we can easily etc. 
In any case as the section chain is the agreed order of groups (each section key == a specific and unique group) then it's also cemented in place there.

madadam: OK. Maybe better to use the current section key index instead of parsec version, to avoid making the impression that the dkg is somehow still tried to parsec. But that's a trivial thing which can be addressed anytime, doesn't have to be in this PR.

Also, this will likely get sorted when we implement the thing where we require nodes to change their name before promotion.

dirvine: Either in this PR or a FIXME in this PR? I am happy either way.

src/node/stage/approved.rs
```diff
@@ -1136,23 +1354,52 @@ impl Approved {
         self.send_routing_message(core, src, dst, content, Some(knowledge_index))
     }
 
+    fn notify_sibling(
```
madadam, line 1357: It seems this method is not only for siblings, but also for any nodes that are currently elders but are about to get demoted, is it? If so, then maybe it should be renamed to reflect it. That probably applied to the `DKGSibling` variant. Not sure what a better name should be though...

src/node/stage/approved.rs
maqi: I will put a FIXME comment in for using section_key index

> this will likely get sorted when we implement the thing where we require nodes to change their name before promotion.

will still be issue if section doesn't have enough nodes. i.e. elder_size =5, then the 5th newly added elder dropped immediately. The other four never change, hence won't have name changed, leaving `praticipants` remain the same.

dirvine: Think that's Ok @maqi  the remaining 4 will do a new dkg and therefor a new unique section key.

src/node/stage/approved.rs
maqi: Yes you are right.
However I think the current name reflects the `purpose` and the case you described is a un-necessary messaging which will be ignored.
I'd prefer remain with this naming.

src/node/stage/approved.rs
dirvine: That's a nice feature of a group key, It means the original A B C are not the same as A B C if D joined and left quick. So first A B C are all using different secretShares than the last A B C 
Where as if we just used ED25519 keys and counted them to get quorum then we can have confusion as both A B C seem the same.

maqi: Yes. but this round of vote could be messed up with the previous voting round, if parsec_version/section_key not being part of the DkgKey, as the participants are the same.

src/node/stage/approved.rs
dirvine: Yes but sibling is like children of the Elders really and they are not. They are perhaps `NonVoting` nodes?

src/node/stage/approved.rs
dirvine: Not if the group are recognised by the section key as that's unique.  So the voting round should include the section key as the group identifier, just like the parsec version would.

src/node/stage/approved.rs
madadam: Right. If the purpose of this is to notify siblings, then it's fine. But I wonder, isn't this used also in the non-split case to notify nodes that are currently elders, but won't be in the next one?

maqi: They are just `old elders` that not vote for this because got split into the other side during split.

How about change it to NotifyOldElders ? and same to the variant?

maqi: > isn't this used also in the non-split case to notify nodes that are currently elders, but won't be in the next one?

yes. the message will be sent. However it shall be ignored on the receiver side.

dirvine: I feel we are playing with this a bit. A node either has voting rights in a section or not, so old_elders would be just non voting nodes (no longer elders) in this section. I think it would be best to stick with Elder as meaning voting node. So notifying new section elders perhaps should be OK as long as we identify the correct voting group and that is implicit really if the section key is used as an index and more in a split as the prefix has got longer.

src/node/stage/approved.rs
```diff
@@ -231,16 +231,55 @@ impl Approved {
...
+            } else {
+                self.dkg_voter.remove_voter(&dkg_key);
             }
         }
+
+        // To avoid the case that DKG was completed after received certain accumulated events.
+        while let Some((event, proof)) = backlog_events.pop_back() {
+            trace!("handle cached accumulated event {:?}", event);
+            // In case of error, event got cached inside `handle_accumulated_event`.
+            let _ = self.handle_accumulated_event(core, event.clone(), Some(proof.clone()));
```
madadam, line 273: Maybe safer would to only ignore the `InvalidDkgResult` error and still propagate the rest?

src/node/stage/approved.rs
```diff
@@ -782,6 +836,66 @@ impl Approved {
         }
     }
 
+    // TODO: carry out accumulation
```
madadam, line 839: What does this mean?

src/node/stage/approved.rs
madadam: > yes. the message will be sent. However it shall be ignored on the receiver side.

Ah OK. This is what I was unsure about. So this is only meant for the split case then, right?

maqi: > So this is only meant for the split case then, right?

yes

src/consensus/dkg.rs
```diff
@@ -0,0 +1,223 @@
...
+    ) -> Vec<DkgMessage<PublicId>> {
+        let mut messages = Vec::new();
+
+        if let Some(mut key_gen) = self.key_gen_map.remove(dkg_key) {
+            if let Ok(responses) = key_gen.handle_message(rng, message) {
+                messages = responses;
+            }
+
+            let _ = self.key_gen_map.insert(dkg_key.clone(), key_gen);
+        }
```
madadam, line 148: Why are you removing and then reinserting it here? Can't you do just:
```rust
        if let Some(key_gen) = self.key_gen_map.get_mut(dkg_key) {
            if let Ok(responses) = key_gen.handle_message(rng, message) {
                messages = responses;
            }
        }
```

src/consensus/dkg.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    pub fn take_info(&mut self, dkg_key: &DkgKey) -> Option<EldersInfo> {
+        self.dkg_cache.remove(dkg_key)
+    }
+
+    pub fn info_keys(&self) -> impl Iterator<Item = &DkgKey> {
+        self.dkg_cache.keys()
+    }
+
+    pub fn insert_dkg_result(&mut self, dkg_key: DkgKey, dkg_result: DkgResult) {
```
madadam, line 208: Maybe better to rename this to `insert_sibling_dkg_result` or something like that?

src/node/stage/approved.rs
maqi: it's actually handling a cached event. propagate the error will be tricky (handle_time_out function will be affected).
Actually, this cached event shall only cause `InvalidDkgResult` error, as this is why the event got cached.

src/consensus/dkg.rs
```diff
@@ -0,0 +1,223 @@
...
+/// Returns the number of minumn responsive participants expected for the DKG process
+#[inline]
+pub const fn threshold_count(elder_size: usize) -> usize {
+    // TODO: allow threshold to be configurable.
+    elder_size * QUORUM_NUMERATOR / QUORUM_DENOMINATOR
+}
+
+pub type DkgKey = (BTreeSet<PublicId>, u64);
+
+#[derive(Default)]
```
madadam, line 74: I think this whole module could use more documentation comments. It's not always clear from the method names alone what their purpose is.

src/node/stage/approved.rs
maqi: here we trust the first guy give us this info.
It actually shall accumulate the incoming notification and trust by group.

src/consensus/dkg.rs
```diff
@@ -0,0 +1,223 @@
...
+pub struct DkgVoter {
+    dkg_cache: BTreeMap<DkgKey, EldersInfo>,
+    key_gen_map: BTreeMap<DkgKey, KeyGen<FullId>>,
+    pending_accumulated_events: VecDeque<(AccumulatingEvent, Proof)>,
+    completed_dkg: BTreeSet<DkgKey>,
```
madadam, line 79: Are we ever removing entries from this set?

src/consensus/dkg.rs
maqi: Ah, previously there is extra work be done by `approved`, which causing borrow check error.
with the squashing, it shall now be ok.
I will change it.

src/consensus/dkg.rs
maqi: ok, I will change to `insert_sibling_dkg_result`

src/consensus/dkg.rs
maqi: Ah, right. I will put more comments

src/consensus/dkg.rs
maqi: how about use `LruCache` instead?

src/node/stage/approved.rs
madadam: OK, then maybe at least log it here?

src/node/stage/approved.rs
madadam: Hm, so is this the thing we discussed, where the DKG is performed by the new elders, but then has to be sent to the old ones to sign?

src/node/stage/approved.rs
maqi: Ok, will have it logged

src/node/stage/approved.rs
maqi: yes

src/section/section_update_barrier.rs
madadam: 1. Ok, I see that you are resetting it in `update_our_section`, so that's fine. Maybe in the future we can try to refactor it so that the barrier also waits for the DKG result. With that, we might be able to also remove the pending events cache. But I'm OK with this for now.
2. Yeah, good point. It should have been called `try_take_details`.

src/consensus/dkg.rs
madadam: Maybe. But maybe we can remove it completely? If I understand it correctly, it's to prevent us from processing DKG messages for an already completed DKG session so that it doesn't accidentally start it over? If so, then maybe if we used the current section key (or its index) instead of parsec version in the DKG key, the we would be able to just check whether that key / index exists anywhere in the section chain apart from the last block and if it does, reject it?

maqi: could be.
How about this: I change it to use LruCache, then put a TODO comment says: if using section_key index as part of the DkgKey, and check agaisnt the section chain, then this can be removed.

madadam: Yeah, works for me.

src/node/stage/approved.rs
madadam: Hm, but that's that mean that `notify_sibling` then isn't only for siblings after all? Say the current elders are `A`, `B`, `C` and `D` and the new ones are `A`, `B`, `E` and `F`. Then first `A`, `B`, `E`, `F` will complete the DKG, then send the result to `C` and `D` who (together with `A` and `B`) will then vote to sign it? If so, then there is no split involved and the term sibling is misleading.

madadam: Also maybe expand the TODO a bit. Something like: "accumulate at least quorum of these messages and only then proceed to handle the DKG result."

maqi: How about change `sibling` to `old_elders` ?

madadam: Yeah, sounds good.

src/node/stage/approved.rs
```diff
@@ -781,6 +840,67 @@ impl Approved {
...
+        pub_id: PublicId,
+    ) -> Result<()> {
+        trace!(
+            "handle dkg message of p{:?}-{:?} from {}",
+            participants,
+            parsec_version,
+            pub_id
+        );
+        self.dkg_voter
+            .set_timer_token(core.timer.schedule(DKG_PROGRESS_INTERVAL));
```
madadam, line 884: I wonder, could a malicious node keep sending invalid DKG messages to us to prevent the DKG from ever expiring? Shouldn't we somehow verify the message is valid before extending the timer? Not sure if that makes sense. Or alternatively, don't extend the timer here, only in `handle_timeout`, so `progress_dkg` is always called?

maqi: an incoming DkgMessage could case key_generator to transit into next phase as well, which means the ticker shall be set to a new token.
but your concern is valid.
I will move it into the place that only a valid DkgMessage will reset the ticker.

PR #2113, opened 2020-06-11
feat/node: add API to fetch our index in current BLS group

Yoga07:


src/node/mod.rs
```diff
@@ -418,6 +418,15 @@ impl Node {
             .ok_or(RoutingError::InvalidState)
     }
 
+    /// Returns our index in the current BLS group or `RoutingError::InvalidState` if we are not
+    /// elder.
```
maqi, line 422: better change the comment to `if section key was not generated yet.`
the `approved` function also includes the stage of `adult`

Issue #2125, opened 2020-06-17
Remove relocate_queue / overhaul relocation


madadam:
`relocate_queue` cannot be part of `SharedState` because its entries are not section-signed and their order is not verifiable. We should ideally find a way to remove it completely or at least move it out of `SharedState`.

madadam:
Closed via https://github.com/maidsafe/sn_routing/pull/2185

Issue #2088, opened 2020-05-04
Improve the tie-breaking rule when deciding which node to promote


madadam:
More info in this discussion: https://github.com/maidsafe/routing/pull/2084#discussion_r417179603

Issue #2126, opened 2020-06-17
Overhaul node promotion / remove `Genesis` event


madadam:
Currently the node promotion works like this:

- The section votes via parsec on `SectionInfo` which contains the new elders
- When that accumulates, the existing elders start sending parsec gossip messages to the new elders
- The new elders receive such messages and first see the consensus on the `Genesis` event which contains the whole `SharedState` inside it
- They modify their `SharedState` instance according to it
- Then they eventually see the consensus on the `SectionInfo` itself at which point they transition themselves to elders

This has several issues:
- It requires non elders to still maintain the parsec machinery
- It requires that the quorum of the section elders insert exactly identical copies into parsec in the `Genesis` event which means they need to be exactly synchronized at that point

Instead we should do this:

- On `SectionInfo` accumulation, the existing elders send their `SharedState` in a `Promote` message to the new elders. This message doesn't need to go through accumulation because `SharedState` is now self-verifiable (requires #2110). The `SharedState`s sent by different nodes don't even need to be identical, because `SharedState` is CRDT-compliant now and different instances can be simply merged into one (requires #2120)

PR #2128, opened 2020-06-17
Shared state signatures

### LINKED ISSUE #2047: Confirm when stuff has PARSEC agreed it's also section signed

### LINKED ISSUE #2110: Make every piece of data in SharedState individually signed
The idea is that would would be able to just send the data as is to anyone (say a newly joined node) and they would be able to instantly verify it. Right now the data needs to be fed to parsec (in the `Genesis` event) and re-accumulated for anyone to be able to trust it. This is a step towards removing Parsec.

madadam:
All entries in all the container in `SharedState` are now individually signed with BLS signatures. The only exceptions are these:

- The `relocate_queue`. The entries in this queue are not created via parsec consensus, so there is currently no way to accumulate the signatures for them. Even if we did, we would still have no way to verify the order the entries are in. All of this would be possible to change, but there are some changes to the relocation logic in the pipeline which would be best to tackle first, to avoid wasted work. See also #2125.
- The `section_version` in `SectionMembers` and `MemberInfo`. It will be removed in #2088 so not worth working on it.
- Node age. Currently very tricky to do as the age counter of every member changes on every churn event. There are changes to node ageing in the pipeline which will affect this, so best to handle them first
- Node socket address. Need to decide whether we want this. Not hard to do if we do.

Closes #2047
Closes #2110


src/node/mod.rs
```diff
@@ -313,6 +314,7 @@ impl Node {
     pub fn our_adults_sorted_by_distance_to(&self, name: &XorName) -> Vec<&P2pNode> {
         self.our_adults()
             .sorted_by(|lhs, rhs| name.cmp_distance(lhs.name(), rhs.name()))
+            .collect()
```
dirvine, line 318: Did this used to work without .collect() ?

src/section/prefix_map.rs
```diff
@@ -0,0 +1,239 @@
...
+#[derive(Default, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub(crate) struct PrefixMap<T>(BTreeMap<Prefix<XorName>, T>);
+
+impl<T: Clone> PrefixMap<T> {
+    /// Create empty `PrefixMap`.
+    pub fn new() -> Self {
+        Self(Default::default())
+    }
+
+    /// Inserts new entry into the map. Replaces previous entry at `prefix` or any of its ancestors
```
dirvine: I wonder if it makes sense to only remove Prefix when we have both descendants here?  (i.e. after a split)

src/node/stage/approved.rs
```diff
@@ -57,7 +57,6 @@ pub struct Approved {
     pub shared_state: SharedState,
     section_keys_provider: SectionKeysProvider,
     sig_accumulator: SignatureAccumulator,
-    genesis_prefix_info: GenesisPrefixInfo,
     timer_token: u64,
     // DKG cache
```
dirvine, line 62: We could remove this comment

src/section/prefix_map.rs
dirvine: Ah this is covered in a later commit :+1:

src/node/mod.rs
madadam: Yes, due to a change in the itertools crate. Previously `sorted_by` returned `Vec`, but now it return an iterator. But it's the `vec::IntoIter` iterator, so collecting it back into `Vec` is zero-cost.

src/xor_space/prefix.rs
```diff
@@ -94,6 +94,14 @@ impl<T: Clone + Copy + Default + Binary + Xorable> Prefix<T> {
         }
     }
 
+    /// Returns `true` if the `other` prefix differs only in the last bit from this one.
+    pub fn is_sibling(&self, other: &Self) -> bool {
+        let i = self.name.common_prefix(&other.name);
+        self.bit_count() == other.bit_count()
+            && self.bit_count() == i + 1
+            && self.name.bit(i) != other.name.bit(i)
```
maqi, line 102: why need this last check? 
does `i != self.bit_count()` already check it ?

tests/mock_network/mod.rs
```diff
@@ -396,11 +396,7 @@ fn sibling_knowledge_update_after_split() {
                 return false;
             }
 
-            if !node
-                .inner
-                .get_their_knowledge()
-                .contains_key(&node.our_prefix().sibling())
-            {
+            if node.inner.get_their_knowledge(&node.our_prefix().sibling()) == 0 {
```
maqi, line 399: == 0 ?

src/node/stage/approved.rs
```diff
@@ -1244,22 +1291,153 @@ impl Approved {
...
+            return Ok(());
+        }
+
+        let key = Proven::new(key, proof);
+
+        if let Some(details) = self
+            .section_update_barrier
+            .handle_our_key(self.shared_state.our_prefix(), key)
+        {
+            self.update_our_section(core, details)
```
maqi, line 1359: looks like this `self.update_our_section` is always inside the barrier check,
then why not have the barrier check inside the `update_our_section` function to reduce duplication?

src/node/stage/approved.rs
```diff
@@ -1322,305 +1499,125 @@ impl Approved {
...
-        )
-    }
+    fn reset_parsec(&mut self, core: &mut Core, new_parsec_version: u64) -> Result<()> {
+        let is_elder = self.is_our_elder(core.id());
 
-    fn handle_their_key_info_event(&mut self, prefix: Prefix<XorName>, key: bls::PublicKey) {
-        self.shared_state.sections.update_keys(prefix, key);
-    }
+        let events = self.consensus_engine.prepare_reset(core.name());
+        let events = self.filter_events_to_revote(events);
```
maqi, line 1537: better not to use same variable name
why not just do:
```
let events = self.filter_events_to_revote(self.consensus_engine.prepare_reset(core.name()));
```

src/section/section_map.rs
```diff
@@ -6,101 +6,88 @@
...
-    // pre-split parent section, so the value's prefix doesn't always match the key.
-    neighbours: BTreeMap<Prefix<XorName>, EldersInfo>,
-    // BLS public keys of known sections
-    keys: BTreeMap<Prefix<XorName>, bls::PublicKey>,
-    // Recent keys removed from `keys`. Contains at most `MAX_RECENT_KEYS` entries.
-    recent_keys: VecDeque<(Prefix<XorName>, bls::PublicKey)>,
-    // Indices of our section keys that are trusted by other sections.
-    knowledge: BTreeMap<Prefix<XorName>, u64>,
+    neighbours: PrefixMap<Proven<EldersInfo>>,
+    // BLS public keys of known sections excluding ours. The proof is for the whole `(prefix, key)`
```
maqi, line 28: `The proof` refers to `Proven<(Prefix<XorName>, bls::PublicKey)>` , right?

src/section/section_map.rs
```diff
@@ -6,101 +6,88 @@
...
+    // pair.
+    keys: PrefixMap<Proven<(Prefix<XorName>, bls::PublicKey)>>,
+    // Indices of our section keys that are trusted by other sections. The proof is for the
+    // `(prefix, knowledge)` pair.
```
maqi, line 32: `The proof is for the (prefix, knowledge) pair` , hmm...  is that `u64` the `knowledge` ?

src/section/section_map.rs
```diff
@@ -543,25 +502,80 @@ mod tests {
...
+    }
+
     // Create a `SectionMap` and apply a series of `update_keys` calls to it, then verify the stored
     // keys are as expected.
     //
-    // updates:  updates to `SectionMap::keys` as a sequence of (prefix, key) pairs. The first pair
-    //           is our section and its initial key. The following ones are then applied in sequence
-    //           by calling `update_keys`
+    // updates:  updates to `SectionMap::keys` as a sequence of (prefix, key) pairs that are
+    //           applied in sequenceby calling `update_keys`
```
maqi, line 560: in sequence by calling `update_keys`.

src/consensus/network_event.rs
```diff
@@ -95,21 +99,104 @@ pub enum AccumulatingEvent {
...
+                bincode::serialize(key)
+            }
+            Self::TheirKey { prefix, key } => bincode::serialize(&(prefix, key)),
+            Self::TheirKnowledge { prefix, knowledge } => bincode::serialize(&(prefix, knowledge)),
+            Self::SectionInfo(info) => bincode::serialize(info),
+            Self::Online {
+                p2p_node,
+                age: _,
+                their_knowledge: _,
+            } => bincode::serialize(&member_info::to_sign(p2p_node.name(), MemberState::Joined)),
```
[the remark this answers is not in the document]
madadam: knowledge doesn't because it is not stored. It is used only to create correct proof chain for the `NodeApproval` message. `age` should be signed eventually, but I decided to postpone doing that because there is a change to the node ageing and relocation logic planned and it would most likely affect the way we do this and I wanted to avoid spending time on something that's likely going to change soon anyway.

src/consensus/parsec.rs
```diff
@@ -281,11 +286,8 @@ impl ParsecMap {
         parsec.has_unpolled_observations()
     }
 
-    pub fn prune_if_needed(&mut self) {
-        if self.size_counter.needs_pruning() {
-            self.vote_for(AccumulatingEvent::ParsecPrune.into_network_event());
```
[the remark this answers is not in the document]
madadam: I moved the voting to `Approved::finish_handle_input`. The idea was that the `ParsecPrune` event would be signed too and there was no way to sign it here. The actual signing of this event is not implemented yet though. I plan to implement it probably as part of this task: https://github.com/maidsafe/routing/issues/2126.

src/node/tests/elder.rs
```diff
@@ -108,28 +128,29 @@ impl Env {
...
-                    );
-                });
+            for (full_id, secret_key_share) in self.other_ids.iter().take(count) {
+                let event = if event.needs_signature() {
+                    let index = self
+                        .elders_info
+                        .position(full_id.public_id().name())
+                        .unwrap();
+                    event
+                        .clone()
```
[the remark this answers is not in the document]
madadam: `into_signed_network_event` takes `self` by value, so the clone is necessary because we are voting for it by multiple nodes.

src/node/tests/elder.rs
```diff
@@ -227,14 +248,16 @@ impl Env {
...
-        let section_key = new_info.new_pk_set.public_key();
+    fn accumulate_our_key_and_section_info_if_vote(&mut self, new_info: &DkgToSectionInfo) {
         let _ = self.n_vote_for_gossipped(
             NOT_ACCUMULATE_ALONE_VOTE_COUNT,
-            iter::once(AccumulatingEvent::SectionInfo(
-                new_info.new_elders_info.clone(),
-                section_key,
-            )),
+            vec![
+                AccumulatingEvent::OurKey {
```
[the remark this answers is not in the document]
madadam: Yes, I had to split the `SectionInfo` into two votes, because there are two distinct pieces of info that are signed - the `EldersInfo` and the section key and they need separate signatures.

src/section/prefix_map.rs
```diff
@@ -0,0 +1,378 @@
...
+    pub fn get_matching(&self, name: &XorName) -> Option<&T> {
+        self.0
+            .iter()
+            .filter(|entry| entry.prefix().matches(name))
+            .max_by_key(|entry| entry.prefix().bit_count())
+            .map(|entry| &entry.0)
+    }
+
+    /// Returns whether the map contain at least one entry whose prefix matches `name`.
+    pub fn contains_matching(&self, name: &XorName) -> bool {
```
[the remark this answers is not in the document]
madadam: Yes, because `(0)` matches it.

src/signature_accumulator.rs
```diff
@@ -197,7 +198,7 @@ mod tests {
         let env = Env::new();
 
         // Add each message with the section list added - none should accumulate.
-        env.msgs_and_sigs.iter().foreach(|msg_and_sigs| {
```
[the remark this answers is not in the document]
madadam: I think `foreach` is from the `itertools` crate and `for_each` is `std`.

tests/mock_network/mod.rs
madadam: `get_their_knowledge` returns the index in **our** `SectionProofChain` of the key **they** know. If it returns 0, that means they know* only the root/genesis key which every section implicitly knows, so in other words, they don't know us yet.

*) more precisely: "we know that they know"

src/consensus/event_accumulator.rs
```diff
@@ -80,194 +90,168 @@ impl VoteStatuses {
...
-        &self.parsec_proofs
-    }
+    fn add(
+        &mut self,
+        voter_name: XorName,
+        voter_index: usize,
+        signature_share: bls::SignatureShare,
+    ) -> Result<(), AccumulatingError> {
+        if self.shares.insert(voter_index, signature_share).is_some() {
+            return Err(AccumulatingError::ReplacedAlreadyInserted);
```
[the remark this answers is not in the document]
madadam: Yeah, I'm not sure this error even makes sense. Because the only way a vote could be overwritten is with same vote. Otherwise the verification would fail. And if it's the same vote, there is no harm overwriting it. Will try to remove it.

src/consensus/event_accumulator.rs
```diff
@@ -80,194 +90,168 @@ impl VoteStatuses {
...
+        }
 
-    pub fn into_sig_shares(self) -> BTreeMap<PublicId, bls::SignatureShare> {
-        self.sig_shares
+        if !self.voters.insert(voter_name) {
+            return Err(AccumulatingError::ReplacedAlreadyInserted);
```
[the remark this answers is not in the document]
madadam: Yeah, which I don't think can really happen, because if we had two votes with the same index but different signature share, then at most one of them could be valid. Will remove this error as mentioned here: https://github.com/maidsafe/routing/pull/2128#discussion_r443420474

src/consensus/event_accumulator.rs
```diff
@@ -80,194 +90,168 @@ impl VoteStatuses {
...
-                    .map(|sig_share| (index, sig_share))
-            })
-            .filter(|&(index, sig_share)| {
-                pk_set
-                    .public_key_share(index)
-                    .verify(sig_share, signed_bytes)
-            });
-        pk_set.combine_signatures(fr_and_shares).ok()
+    fn has_enough_votes(&self, threshold: usize, elders_info: &EldersInfo) -> bool {
+        self.shares.len() > threshold && elders_info.is_quorum(&self.voters)
```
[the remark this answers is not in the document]
madadam: Yeah. I think eventually we want to check only `shares.len() > threshold`. But checking the voters too for now because the threshold and the quorum are currently different (1/3 vs 2/3). But that's going to change with the DKG replacement, right?

src/node/stage/approved.rs
```diff
@@ -987,7 +975,10 @@ impl Approved {
         for info in new_infos {
             let participants: BTreeSet<_> = info.elder_ids().copied().collect();
             let _ = self.dkg_cache.insert(participants.clone(), info);
-            self.vote_for_event(AccumulatingEvent::StartDkg(participants));
+            self.consensus_engine.vote_for(NetworkEvent {
+                payload: AccumulatingEvent::StartDkg(participants),
+                proof_share: None,
```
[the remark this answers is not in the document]
madadam: It's just not used for the `StartDkg` variant, but it's used for the other ones.

src/node/stage/approved.rs
madadam: Because there are actually three different barrier checks: `handle_our_key`, `handle_their_key` and `handle_info`. Each is for different event.

src/node/stage/approved.rs
madadam: Your suggestion fails the borrow checker

src/section/section_map.rs
madadam: Yes. The comment was written before the introduction of the `Proven` type and it's a bit redundant now. I'll change it.

src/section/section_map.rs
madadam: Yes, but again, the comment is redundant now.

PR #2134, opened 2020-06-23
Update to use xor_name crate and remove xor_space in this crate

dirvine:


Cargo.toml
```diff
@@ -57,7 +58,10 @@ mock_base = [
     "env_logger",
 ]
 mock = ["mock_base"]
+test = ["mock"]
```
maqi, line 61: does this mean use `mock` feature for test by default?
if so, maybe the CI script shall be updated as well?

dirvine: Yes this was a local change that I mistakenly put in, I will remove it though,

PR #2135, opened 2020-06-25
Signature accumulator

### LINKED ISSUE #2131: Create generic SignatureAccumulator and expose it for the consumers of routing
To be used by routing itself and also by the upper layers. Should have API similar to this:

```rust
struct SignatureAccumulator<T> { ... }

impl<T> SignatureAccumulator<T> {
    pub fn accumulate(
        &mut self, 
        payload: T, 
        signature_share: bls::SignatureShare, 
        share_index: usize
    ) -> Option<(T, bls::PublicKey, bls::Signature)>;

    pub fn remove_expired(&mut self) 
        -> Vec<(T, bls::PublicKeySet, HashMap<usize, bls::SignatureShare)>;
}
```

madadam:
This PR implement the `SignatureAccumulator` abstraction and exposes it to the upper layers.

The `SignatureAccumulator` is also used internally to implement `MessageAccumulator`. It's not yet used for `EventAccumulator` though, because we currently need a 2/3 quorum for events, but the BLS threshold currently is only 1/3 which is insufficient for voting. It requires the DKG replacement to land first.

Closes #2131


src/consensus/signature_accumulator.rs
```diff
@@ -0,0 +1,343 @@
...
+        match self {
+            Self::Accumulating {
+                shares, modified, ..
+            } => {
+                if shares
+                    .insert(proof_share.index, proof_share.signature_share)
+                    .is_none()
+                {
+                    *modified = Instant::now();
+                }
```
maqi, line 197: could return early for the else branch, with NotEnough error or a new DuplicatedShare error ?

src/messages/message_accumulator.rs
```diff
@@ -0,0 +1,76 @@
+// Copyright 2018 MaidSafe.net limited.
```
maqi, line 1: 2020

src/messages/message_accumulator.rs
```diff
@@ -0,0 +1,76 @@
...
+use serde::{Serialize, Serializer};
+
+#[derive(Default)]
+pub struct MessageAccumulator(SignatureAccumulator<Payload>);
+
+impl MessageAccumulator {
+    pub fn add(&mut self, accumulating_msg: AccumulatingMessage) -> Option<MessageWithBytes> {
+        let payload = Payload {
+            content: accumulating_msg.content,
+            proof_chain: accumulating_msg.proof_chain,
```
maqi, line 23: I am not quite understand why this whole proof_chain need to be part of the accumulating scheme?

src/messages/message_accumulator.rs
```diff
@@ -0,0 +1,76 @@
...
+        match self.0.add(payload, accumulating_msg.proof_share) {
+            Ok((payload, proof)) => {
+                // TODO: should we verify that `proof.public_key` is the same as
+                // `proof_chain.last_key()`?
+
+                let msg = Message {
+                    src: SrcAuthority::Section {
+                        prefix: payload.content.src,
+                        signature: proof.signature,
+                        proof: payload.proof_chain,
```
maqi, line 35: so the `proof` needs to be `proof_chain` actually? not something like: the last proof within the proof_chain?
better make the naming consistent to avoid confusion?

src/consensus/signature_accumulator.rs
```diff
@@ -0,0 +1,343 @@
...
+    #[error(display = "signature already accumulated")]
+    AlreadyAccumulated,
+    /// The signature share being added is invalid. Such share is rejected but the already collected
+    /// shares are kept intact. If enough new valid shares are collected afterwards, the
+    /// accumulation might still succeed.
+    #[error(display = "signature share is invalid")]
+    InvalidShare,
+    /// The payload failed to be serialised and can't be inserted into the accumulator. This doesn't
+    /// affect other entries already present and the accumulator can still be used normally
+    /// afterwards.
```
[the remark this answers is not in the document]
madadam: In practice I don't think a serialization can ever fail. But we still need to handle the error, to make the compiler happy. Unless we want to panic instead.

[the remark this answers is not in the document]
maqi: I just mean as you put the same word (shall never happen) for the below error, maybe shall put it here as well?

src/messages/message_accumulator.rs
madadam: because It's part of the resulting `Message`.

src/lib.rs
```diff
@@ -79,16 +79,18 @@ mod macros;
 // Public API
 // ############################################################################
 pub use self::{
+    consensus::{Proof, ProofShare, SignatureAccumulator},
```
lionel1704, line 82: `AccumulationError` needs to be exposed  in the public API as well.

PR #2137, opened 2020-06-26
Remove unneeded parts

### LINKED ISSUE #2088: Improve the tie-breaking rule when deciding which node to promote

### LINKED ISSUE #2123: Remove churn_event_backlog from SharedState
It comes from the times when we didn't yet have the distinction between elders and non-elders and was needed to make sure the churn events are serialized. It should not be needed now because a joined node doesn't automatically becomes elder.

### LINKED ISSUE #2124: Remove "relocation cool-down period" / remove RelocatePrepare event
The purpose of this period was to make sure that a node that is being relocated processes all the messages that it needs to before actually leaving the section. This should not be needed now because such messages should now be handled by the message bounce mechanism.

madadam:
Closes #2088 
Closes #2123 
Closes #2124

### DIFF
--- a/src/section/section_members.rs
+++ b/src/section/section_members.rs
@@ -6,7 +6,10 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use super::member_info::{MemberInfo, MemberState};
+use super::{
+    member_info::{MemberInfo, MemberState},
+    EldersInfo,
+};
 use crate::{consensus::Proof, id::P2pNode};
 
 use itertools::Itertools;
@@ -70,8 +73,12 @@ impl SectionMembers {
     }
 
     /// Returns the candidates for elders out of all the nodes in this section.
-    pub fn elder_candidates(&self, elder_size: usize) -> BTreeMap<XorName, P2pNode> {
-        elder_candidates(elder_size, self.joined())
+    pub fn elder_candidates(
+        &self,
+        elder_size: usize,
+        current_elders: &EldersInfo,
+    ) -> BTreeMap<XorName, P2pNode> {
+        elder_candidates(elder_size, current_elders, self.joined())
     }
 
     /// Returns the candidates for elders out of all nodes matching the prefix.
@@ -79,9 +86,11 @@ impl SectionMembers {
         &self,
         prefix: &Prefix<XorName>,
         elder_size: usize,
+        current_elders: &EldersInfo,
     ) -> BTreeMap<XorName, P2pNode> {
         elder_candidates(
             elder_size,
+            current_elders,
             self.joined()
                 .filter(|info| prefix.matches(info.p2p_node.name())),
         )
@@ -182,13 +191,17 @@ impl PartialEq for SectionMembers {
 }
 
 // Returns the nodes that should become the next elders out of the given members.
-fn elder_candidates<'a, I>(elder_size: usize, members: I) -> BTreeMap<XorName, P2pNode>
+fn elder_candidates<'a, I>(
+    elder_size: usize,
+    current_elders: &EldersInfo,
+    members: I,
+) -> BTreeMap<XorName, P2pNode>
 where
     I: IntoIterator<Item = &'a MemberInfo>,
 {
     members
         .into_iter()
-        .sorted_by(|lhs, rhs| cmp_elder_candidates(lhs, rhs))
+        .sorted_by(|lhs, rhs| cmp_elder_candidates(lhs, rhs, current_elders))
         .map(|info| (*info.p2p_node.name(), info.p2p_node.clone()))
         .take(elder_size)
         .collect()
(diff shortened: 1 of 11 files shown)


src/section/shared_state.rs
```diff
@@ -437,12 +417,16 @@ impl SharedState {
...
-        let our_elders = self
-            .our_members
-            .elder_candidates_matching_prefix(&our_prefix, network_params.elder_size);
-        let other_elders = self
-            .our_members
-            .elder_candidates_matching_prefix(&other_prefix, network_params.elder_size);
+        let our_elders = self.our_members.elder_candidates_matching_prefix(
+            &our_prefix,
+            network_params.elder_size,
+            self.sections.our(),
```
maqi, line 423: this `our` function returns with `our_elders`, right?

tests/mock_network/messages.rs
```diff
@@ -77,7 +82,7 @@ fn send_and_receive() {
     // For all expected recipients, poll until it receives the request message and sends
     // the response
     let expected_req_recipients: Vec<_> = expected_recipients(&nodes, &req_dst).collect();
-    assert!(expected_req_recipients.len() >= quorum);
+    assert!(expected_req_recipients.len() >= quorum,);
```
maqi, line 85: this comma is added by autofmt?

tests/mock_network/utils.rs
```diff
@@ -374,11 +374,16 @@ pub fn create_connected_nodes(env: &Environment, size: usize) -> Vec<TestNode> {
...
-
-        poll_until(env, &mut nodes, |nodes| node_joined(nodes, nodes.len() - 1));
-        verify_invariants_for_nodes(&env, &nodes);
     }
 
+    poll_until(env, &mut nodes, |nodes| {
+        all_nodes_joined(nodes, 0..nodes.len())
+    });
+    poll_until(env, &mut nodes, |nodes| {
+        elder_count_reached(nodes, &Prefix::default(), env.elder_size().min(size))
```
maqi, line 383: why not just poll this second? seems the above poll is duplicated?

tests/mock_network/utils.rs
```diff
@@ -629,12 +636,66 @@ fn add_nodes_to_subsections_and_poll(
...
 }
 
+// Poll until the section with `prefix` has at least 4 elders which is the miminum so that if one
+// elder is removed, the consensus on it can still be reached.
+fn poll_until_minimal_elder_count(
+    env: &Environment,
+    nodes: &mut [TestNode],
+    prefix: &Prefix<XorName>,
+) {
+    let expected = 4;
```
maqi, line 665: why 4?
better not to use a magic number here

src/section/section_members.rs
```diff
@@ -182,36 +182,52 @@ impl SectionMembers {
...
+    // signature and therefore game its chances of promotion.
     rhs.age_counter
         .cmp(&lhs.age_counter)
-        .then(lhs.section_version.cmp(&rhs.section_version))
+        .then_with(|| {
+            let lhs_is_elder = current_elders.elders.contains_key(lhs.p2p_node.name());
+            let rhs_is_elder = current_elders.elders.contains_key(rhs.p2p_node.name());
+
+            match (lhs_is_elder, rhs_is_elder) {
+                (true, false) => Ordering::Less,
```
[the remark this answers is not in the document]
madadam: Yes, `Less` is better. The idea is if you sort a collection by this, then the best candidates will be at the beginning.

src/section/shared_state.rs
madadam: It returns the current `EldersInfo` of our section.

tests/mock_network/messages.rs
madadam: Yeah, will remove it.

tests/mock_network/utils.rs
madadam: Because we need both: all the new nodes joining and the section reaching the minimal size of elders. Neither one implies the other.

tests/mock_network/utils.rs
madadam: It's explained in the comment of this function. But maybe better to use a constant?

src/section/shared_state.rs
maqi: I'd prefer this function name to be more explicit reflecting the stuff returned, so maybe make it as `our_elders`.
Especially there are now adults/infancts within the section. (same idea to the function `our_member`, which actually also returns elders only)

However, I think this better to be done in a separate PR (if decide to do so).
So won't make this as a blocker to this PR.

tests/mock_network/messages.rs
maqi: just wondering where it comes from, and how to disable it.
I also see the similar stuff couple of times some where else.

tests/mock_network/utils.rs
maqi: ah, I see what they means.

tests/mock_network/utils.rs
maqi: ah, sry, I missed the comment.
And, yes, will prefer it to be a constant

tests/mock_network/messages.rs
madadam: No idea. Probably rustfmt doesn't handle macros very well.

PR #2139, opened 2020-06-28
Make Message immutable and complete on construction

dirvine:
Remove MessageWithBytes and PartialMessage. Now message serialialises at most once and calculates
signature once as well as calculate hash once. These were memoized in previous iterations but deemed to be always useful as we must sign message, we must confirm sig on receipt. We deserialise from the wire as bytes and keep that in case it's required (to forward or bounce the message). We filter via hash too (except Direct messages) therefor it made snse to not complexify via memoization but instead keep the serialised and hash fields correct.

The message type now has unfailable to_bytes() and hash() methods which does help with cleaner code.


src/node/tests/elder.rs
```diff
@@ -706,7 +707,7 @@ impl OtherNode {
 
     fn expect_bootstrap_response(&self) -> BootstrapResponse {
         self.received_messages()
-            .find_map(|msg| match msg.variant {
+            .find_map(|msg| match msg.variant().clone() {
                 Variant::BootstrapResponse(response) => Some(response),
```
madadam, line 711: I'd move the `clone` here, to avoid cloning the variants we don't care about:
```rust
match msg.variant() {
    Variant::BootstrapResponse(response) => Some(response.clone()),
    _ => None,
}
```

src/node/tests/elder.rs
```diff
@@ -633,6 +633,7 @@ fn handle_bounced_unknown_message() {
 }
 
 #[test]
+#[ignore] //FIXME Any message invalidly signed will not deserialise / be created
```
madadam, line 636: This test isn't about invalidly signed messages, but about messages with untrusted proof which we should handle via the bounce thing.

src/relocation.rs
```diff
@@ -46,15 +46,15 @@ pub struct SignedRelocateDetails {
 
 impl SignedRelocateDetails {
     pub fn new(signed_msg: Message) -> Result<Self, RoutingError> {
-        if let Variant::Relocate(_) = &signed_msg.variant {
+        if let Variant::Relocate(_) = &signed_msg.variant() {
```
madadam, line 49: The `&` is unnecessary now (clippy should have complained about it I think).

tests/mock_network/secure_message_delivery.rs
```diff
@@ -109,6 +109,7 @@ fn message_with_invalid_security(fail_type: FailType) {
 }
 
 #[test]
+#[ignore] // We cannot create a message with invalid sig FIXME (possibly we want to do so via mock)
```
madadam, line 112: Note that when https://github.com/maidsafe/routing/pull/2135 is merged, this whole module is gone and is instead replaced with unit tests, so faking an invalid message would be easier there.

src/node/mod.rs
```diff
@@ -535,51 +532,32 @@ impl Node {
...
-        &mut self,
-        sender: SocketAddr,
-        mut msg_with_bytes: MessageWithBytes,
-    ) -> Result<()> {
-        trace!("try handle message {:?}", msg_with_bytes);
+    fn try_handle_message(&mut self, sender: SocketAddr, msg: Message) -> Result<()> {
+        trace!("try handle message {:?}", msg);
 
-        self.try_relay_message(&msg_with_bytes)?;
+        self.try_relay_message(&msg)?;
```
madadam, line 538: So we now lost the optimization we had where we only deserialize the `dst` first and only if the message is for us, we deserialize the rest of the message. Probably OK for now but we might want to add it back at some point, in some form.

src/messages/mod.rs
```diff
@@ -37,62 +35,114 @@ use xor_name::{Prefix, XorName};
...
-        Ok(bincode::deserialize(&bytes[..])?)
-    }
+    /// Serialised message, this is a signed and fully serialised message ready to send.
+    #[serde(skip)]
+    serialized: Bytes,
+    #[serde(skip)]
+    hash: MessageHash,
 }
 
 impl Message {
```
madadam, line 58: Maybe we should add a method like this:
```rust
fn into_parts(self) -> (SrcAuthority, DstLocation, Option<bls::PublicKey>, Variant)
```
to avoid the `clone`s?

src/messages/mod.rs
```diff
@@ -37,62 +35,114 @@ use xor_name::{Prefix, XorName};
...
-    /// Serialize the message.
-    pub(crate) fn to_bytes(&self) -> Result<Bytes> {
-        Ok(bincode::serialize(self)?.into())
+    /// send across wire
+    pub(crate) fn to_bytes(&self) -> Bytes {
+        self.serialized.clone()
+    }
+
+    /// Creates a signed message where signature is assumed valid.
+    pub fn new_signed(
```
madadam, line 100: This probably should't be `pub`, because it can be used to violate the invariant (that the signature is always valid).

src/messages/mod.rs
```diff
@@ -37,62 +35,114 @@ use xor_name::{Prefix, XorName};
...
+                } else {
+                    Err(RoutingError::FailedSignature)
+                }
+            }
+            SrcAuthority::Section {
+                prefix,
+                signature,
+                proof,
+            } => {
+                // FIXME Assumes the nodes proof last key is the one signing this message
```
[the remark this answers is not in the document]
dirvine: Yes, I mostly agree but was not 100%. Are we sure there's no situation where a proof would go beyond the current? I mean the sender signed with a section key as churn was happening and sends us more than we need?

src/node/mod.rs
dirvine: Yes, however now we don't pass on any unsigned message either and this is the trade-off. As we amplify messages per-hop then it may be best to stop invalid signed messages as close to the source as possible.  I think this also allows the sections to keep catching up with each other as we progress this area of the code.

src/messages/mod.rs
dirvine: I am not so sure, perhaps we can discuss this? I mean I am not seeing how that would work in practice and allowing the same functionality. Unless that method cloned each element itself. I may be missing the obvious though.

src/node/tests/elder.rs
dirvine: Good point

src/node/tests/elder.rs
dirvine: Yes IIRC it was the untrusted proof signature failure here. Not the src beign recodnised, just the sig failed.

src/messages/mod.rs
```diff
@@ -37,62 +35,114 @@ use xor_name::{Prefix, XorName};
 #[derive(Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
 pub struct Message {
     /// Destination location.
-    pub dst: DstLocation,
+    dst: DstLocation,
     /// Source authority.
-    pub src: SrcAuthority,
+    /// Messages do not need to sign this field as it is all verifiable (i.e. if the sig validates
+    /// agains the pub key and we know the pub key then we are good. If the proof is not recodnised we
```
maqi, line 41: against , public key

recognised ?
same to the line below

src/messages/mod.rs
```diff
@@ -37,62 +35,114 @@ use xor_name::{Prefix, XorName};
...
+    }
+
+    /// Creates a signed message where signature is assumed valid.
+    pub fn new_signed(
+        dst: DstLocation,
+        src: SrcAuthority,
+        variant: Variant,
+        dst_key: Option<bls::PublicKey>,
+    ) -> Result<Message> {
+        let mut msg = Message {
```
maqi, line 106: better use a construct to populate serialized & hash field internally? which can avoid using default as well?

src/messages/src_authority.rs
```diff
@@ -19,6 +19,10 @@ use std::net::SocketAddr;
 use xor_name::{Prefix, XorName};
 
 #[derive(Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
+/// Src of message and authority to send it. Authority is validate by the signature.
```
maqi, line 22: is validated by

src/signature_accumulator.rs
```diff
@@ -212,25 +205,23 @@ mod tests {
             for signature_msg in msg_and_sigs.signature_msgs {
                 let old_num_msgs = sig_accumulator.msgs.len();
 
-                let result = match signature_msg.variant {
-                    Variant::MessageSignature(msg) => sig_accumulator.add_proof(*msg),
+                let result = match signature_msg.variant() {
+                    Variant::MessageSignature(msg) => sig_accumulator.add_proof(*msg.clone()),
```
maqi, line 209: `*` shall not be required anymore?

src/messages/mod.rs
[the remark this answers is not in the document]
madadam: Well either we enforce that or we need to include the index of the key that was actually used for the signature. Probably easier to go with the first option. Basically when selecting the part of the proof chain to attach to the message, just make sure it ends at the same key as the one used to sign the message. 

So maybe keep the FIXME there for now, so we don't forger to address it at some point.

src/node/tests/elder.rs
madadam: Ah, this is the mysterious failure you wanted me to investigate? I'll look into it.

src/node/mod.rs
madadam: Ok, makes sense. Let's make things correct first and worry about optimizations later.

src/messages/mod.rs
madadam: Note it takes `self` by value so it can move from it:
```rust
fn into_parts(self) -> (SrcAuthority, DstLocation, Option<bls::PublicKey>, Variant) {
    (self.src, self.dst, self.dst_key, self.variant)
}
```

src/signature_accumulator.rs
dirvine: Ah it was because this field is a box, so a deref call to the msg then clone(). I tried without.

maqi: ah, right.

src/messages/mod.rs
dirvine: Yes, I played a lot with this, sub classes, options, and more. The default() trick is a serde thing. So it deserialised to the type by populating the ignored fields with Default::default() 

I am not sure this is the best route yet, but it's solid and allows us to serialise ignoring these fields. It started as a mechanism to memoize them via options, but that messed up a lot of code that read the values a lot. So many match statements of map deconstructs. So I ended up with this. It may be we end up with a custom serialise/deserialize but it feels a bit messier doing that. 
I am sure this will improve over time but for now I want to create a barrier that if a message sxists then it's signed correctly with the key supplied in src. Still have to consider the section message being sperate and see if we can bring it an and then consider validation again to be sig + recodnised.

src/messages/mod.rs
dirvine: Does that invalidate the message though when we move data from it?

madadam: No, it consumes it, so it can't be used afterwards. It's just how moving works in rust (unlike C++ where you can still used a value after it has been moved from).

src/messages/mod.rs
maqi: Ah, ok, I get it. Thx for the clarification.

src/messages/mod.rs
dirvine: Yes that's what I mean, So if you read msg.src() then msg.dst() the latter would fail unless all your data was in that parts tuple. But then you want to get the hash or serliazed msg are you not in trouble? I mean is it not more complex to take bits out like this? Even if we read msg.src more than once then we are back to clone() one of the calls. 
I am perhaps not understanding the value. Maybe it will take a bit of investigating where we clone() do we ever read it again? But then just making src dst return self.src etc. would achieve the same?

dirvine: My goal is the msg is immutable all the way through our code and we never have to call serialize or hash after construction/reception of the Message. I suppose if we did a `to_parts` we would need to be sure we don't any longer want to serialize or hash, maybe that's when the message is being handled? Although we would need to take care with bounced where we pass the received bytes back.

madadam: > So if you read msg.src() then msg.dst() the latter would fail

No, you can't read `msg` after you called `into_parts` (or any other function that takes it by value). Rust statically prevents that. 

>  But then you want to get the hash or serliazed msg are you not in trouble

We can return these in the tuple as well if we think we might need them. But I think once we are in `dispatch_message` we no longer do.

> I am perhaps not understanding the value

It's just to avoid the `clone`. But maybe we don't have to worry about that just yet?

dirvine: Ah yes, I get you now, thanks. If you feel it's OK I would prefer to delay that decision until we finalise the message module proper. So in a later PR if at all?

src/messages/mod.rs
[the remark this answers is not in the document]
dirvine: I agree we should probably enforce that. I will keep the FIXME as a reminder.

PR #2142, opened 2020-06-30
feat/api: add SectionProofChain getter api for Node

### LINKED ISSUE #2140: Expose SectionProofChain to the upper layers
- Add a `pub` function to return the current `SectionProofChain`
- Change the visibility of the `SectionProofChain` methods from `pub(crate)` to `pub`, but only those that are useful for the upper layers.

Yoga07:
Resolves #2140

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,6 +88,7 @@ pub use self::{
     pause::PausedState,
     quic_p2p::Config as TransportConfig,
     quic_p2p::Event as TransportEvent,
+    section::SectionProofChain,
 };
 
 pub use xor_name::{Prefix, XorName, Xorable, XOR_NAME_LEN}; // TODO remove pub on API update
--- a/src/section/section_proof_chain.rs
+++ b/src/section/section_proof_chain.rs
@@ -20,6 +20,7 @@ pub struct SectionProofChain {
     tail: Vec<Block>,
 }
 
+#[allow(clippy::len_without_is_empty)]
 impl SectionProofChain {
     /// Creates new chain consisting of only one block.
     pub fn new(first: bls::PublicKey) -> Self {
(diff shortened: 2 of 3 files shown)


src/section/section_proof_chain.rs
```diff
@@ -52,30 +53,33 @@ impl SectionProofChain {
...
     }
 
-    pub(crate) fn first_key(&self) -> &bls::PublicKey {
+    /// Returns the first key(head) from the chain
+    pub fn first_key(&self) -> &bls::PublicKey {
         &self.head
     }
 
-    pub(crate) fn last_key(&self) -> &bls::PublicKey {
+    /// Returns the last key from tail of the chain (returns head as last if there is no tail).
```
madadam, line 61: Same here. Just say "Returns the last key of the chain".

src/section/section_proof_chain.rs
```diff
@@ -52,30 +53,33 @@ impl SectionProofChain {
         self.tail.push(Block { key, signature })
     }
 
-    pub(crate) fn first_key(&self) -> &bls::PublicKey {
+    /// Returns the first key(head) from the chain
```
madadam, line 56: The head / tail thing is supposed to be an implementation detail. I'd just say "Returns the first key of the chain" here.

src/section/section_proof_chain.rs
```diff
@@ -85,7 +89,7 @@ impl SectionProofChain {
     ///
     /// Note: unlike `std::slice`, if the range is invalid or out of bounds, it is silently adjusted
     /// to the nearest valid range and so this function never panics.
-    pub(crate) fn slice<B: RangeBounds<u64>>(&self, range: B) -> Self {
+    pub fn slice<B: RangeBounds<u64>>(&self, range: B) -> Self {
```
madadam, line 92: Worth expanding this comment a bit, now the function is `pub`. Maybe
"Returns a subset of this chain specified by the given index range" or something like that.

src/lib.rs
```diff
@@ -86,6 +86,7 @@ pub use self::{
     pause::PausedState,
     quic_p2p::Config as TransportConfig,
     quic_p2p::Event as TransportEvent,
+    section::SectionProofChain,
```
madadam, line 89: Need to also remove `SectionProofChain` from the `[cfg(feature = "mock_base")]` block below (line 115)

PR #2143, opened 2020-07-01
Fix proof chain extension on received bounced untrusted message

madadam:
In one of the recent changes, we accidentally disabled the proof chain extension which needs to happen when receiving `BouncedUntrustedMessage`. This PR fixes that.


src/messages/mod.rs
```diff
@@ -170,6 +169,48 @@ impl Message {
...
+
+        let index_to = if let Some(index) = section_proof_chain.index_of(proof_chain.last_key()) {
+            index
+        } else {
+            return Err(ExtendProofChainError::InvalidLastKey);
+        };
+
+        *proof_chain = section_proof_chain.slice(index_from..=index_to);
+
+        self.reserialise()
```
dirvine, line 204: PErhaps rather than this method, we use the `new_signed` method as it will serialise and hash, etc. Then we are more in line with immutable messages. So we can never change interior code of a msg, but always create a new message?

src/messages/mod.rs
```diff
@@ -170,6 +169,48 @@ impl Message {
     pub fn hash(&self) -> &MessageHash {
         &self.hash
     }
+
+    // Extend the current message proof so it starts at `new_first_key` while keeping the last key
+    // (and therefore the signature) intact.
+    #[cfg_attr(feature = "mock_base", allow(clippy::trivially_copy_pass_by_ref))]
```
dirvine, line 175: Just a quick one, do we need this flag here?

madadam: Because `feature=mock_base` also enables the mock version of threshold-crypto and with it the `bls::PublicKey` type is only 4 bytes long. And clippy things that for types that small, it's more efficient to pass them by value. But with the normal production version of threshold-crypto, the type has ~40 bytes so it's definitely more efficient to pass it by reference.

src/messages/mod.rs
madadam: Yeah,I was considering that, but the issue is with error handling. `new_signed` returns `Result<Self, RoutingError>` but I wanted to return a more specific error here. There are ways to have both, but I'm not sure it was worth spending time on it.

dirvine: Can we then match or map this error to allow us to pass a new error? Mind you I suppose that kinda breaks our error handling / types. So the choice is to break error handling or break msg immutability? 
I would perhaps vote that if possible can we keep msg immutability and if required create a new error?

src/messages/mod.rs
dirvine: Be nice from yesterdays meet to see just how much of a hitr it is here to not use mock threshold crypto. Esepecially so as the dkg PR is en route.

madadam: As far as I remember, it was pretty significant. But we can try to measure it again.

dirvine: Yes thanks @madadam  For sure crypto is slow and bls is a factor slower than ed crypto. So likely they will be treacle in dev mode but should be acceptable in release mode.

src/messages/mod.rs
madadam: What I think I can do is to change `new_signed` to return a separate error, say `CreateError` and define `From` conversions of `CreateError` -> `RoutingError` and `CreateError` -> `ExtendProofChainError`. That should give us everything we need without breaking anything. How does that sound?

dirvine: I think we can actually use the From macro in the derive_more crate to allow that conversion, just to make it easier.

madadam: Btw, `err-derive` has it's own `From` thing now, so we don't even need `derive_more` anymore.

dirvine: Ah nice one.  I had an issue using it for errors form bls, but that is due to bls using Failure which is not a Error derived form std::Error, but shoudl be cool here.

madadam: So it seems that normally you can use the `#[error(source)]` attribute which gives you `From` + `std::Error`'s `source`, but if the underlying type doesn't implement `std::Error` you can use `#[error(from)]` which just gives you `From`.

dirvine: Ah, good to know, I wrestled with that for an hour last wekk. Cheers Adam :+1:

PR #2150, opened 2020-07-04
Add tests to crypto.rs

dirvine:
Test hash/sign/into PublicKey 

Much of this file can be removed when parsec is gone.


src/crypto.rs
```diff
@@ -119,3 +119,44 @@ pub mod signing {
 pub mod encryption {
     pub use bls::{Ciphertext, PublicKey, SecretKey};
 }
+
+#[cfg(test)]
+mod test {
```
madadam, line 124: I'm confused about the purpose of these tests. It seems we are basically testing other crate. Aren't the tests in that crate enough?

dirvine: As we wrap other crates these are testing we did that properly. My preference is this does not exist really, but the hash wrapper may be OK, it used to exist in the tiny keccak crate. I will check. When we are parsec free we can look at removing this file and tests I recon.

PR #2151, opened 2020-07-06
fix/dkg: use section_key_index as part of the DkgKey

maqi:


src/consensus/dkg.rs
```diff
@@ -178,7 +169,7 @@ impl DkgVoter {
         full_id: &FullId,
         dkg_key: &DkgKey,
     ) -> Vec<DkgMessage<PublicId>> {
-        if self.key_gen_map.contains_key(dkg_key) || self.completed_dkg.contains_key(dkg_key) {
+        if self.key_gen_map.contains_key(dkg_key) {
```
madadam, line 172: Let's say a DKG instance completes, but the new key is not yet appended to the section chain (because the signatures are still being accumulated). And then a dkg message with the dkg_key of the just completed instance is received (maybe because it was delayed, maybe due to malice, ...). Would that start the dkg instance again?

maqi: >but the new key is not yet appended to the section chain (because the signatures are still being accumulated

maqi: >but the new key is not yet appended to the section chain (because the signatures are still being accumulated

the completed DkgResult will be stored within section_key_provider.new , and be `taken` only after the vote of OurKey and SectionInfo got completed.

maqi: fixing commit pushed

Issue #2119, opened 2020-06-17
Ensure all mutations to SharedState are approved by the section


madadam:
Any mutation to `SharedState` must be accompanied by a valid BLS signature which must correspond to a known key. Failing that, the mutation must be rejected.

PR #2155, opened 2020-07-08
feat/shared_state: ensure mutations to SharedState are approved by the section

maqi:
Close https://github.com/maidsafe/routing/issues/2119

### DIFF
--- a/src/consensus/proof.rs
+++ b/src/consensus/proof.rs
@@ -6,6 +6,8 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use crate::section::SectionProofChain;
+use serde::Serialize;
 use std::{
     borrow::Borrow,
     fmt::{self, Debug, Formatter},
@@ -75,20 +77,28 @@ impl Debug for ProofShare {
 
 /// A value together with the proof that it was agreed on by the quorum of the section elders.
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Serialize, Deserialize)]
-pub struct Proven<T> {
+pub struct Proven<T: Serialize> {
     pub value: T,
     pub proof: Proof,
 }
 
-impl<T> Proven<T> {
+impl<T: Serialize> Proven<T> {
     pub fn new(value: T, proof: Proof) -> Self {
         Self { value, proof }
+    }
+
+    pub fn verify(&self, history: &SectionProofChain) -> bool {
+        if let Ok(bytes) = bincode::serialize(&self.value) {
+            history.has_key(&self.proof.public_key) && self.proof.verify(&bytes)
+        } else {
+            false
+        }
     }
 }
 
 impl<T> Borrow<Prefix> for Proven<T>
 where
-    T: Borrow<Prefix>,
+    T: Borrow<Prefix> + Serialize,
 {
     fn borrow(&self) -> &Prefix {
         self.value.borrow()
--- a/src/section/shared_state.rs
+++ b/src/section/shared_state.rs
@@ -6,9 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use super::{
-    EldersInfo, MemberInfo, MemberState, SectionMap, SectionMembers, SectionProofChain, TrustStatus,
-};
+use super::{EldersInfo, MemberInfo, MemberState, SectionMap, SectionMembers, SectionProofChain};
 use crate::{
     consensus::{AccumulatingEvent, Proof, Proven},
     error::RoutingError,
@@ -69,7 +67,12 @@ impl SharedState {
                 );
                 return Err(RoutingError::InvalidState);
             }
-        } else if !self.our_history.self_verify() {
+        } else if !new.self_verify()
+            && !self
+                .our_history
+                .keys()
+                .any(|key| new.our_history.has_key(key))
+        {
             error!("shared state update - invalid new history: {:?}", new);
             return Err(RoutingError::InvalidState);
         }
@@ -77,6 +80,12 @@ impl SharedState {
         *self = new;
         self.handled_genesis_event = true;
         Ok(())
+    }
+
+    fn self_verify(&self) -> bool {
+        self.our_history.self_verify()
+            && self.our_members.verify(&self.our_history)
+            && self.sections.verify(&self.our_history)
     }
 
     // Clear all data except that which is needed for non-elders.
@@ -191,18 +200,14 @@ impl SharedState {
             return false;
         }
 
-        // The section public key of the proof shall be trusted by us.
-        match self.our_history.check_trust(iter::once(&proof.public_key)) {
-            TrustStatus::Trusted => {}
-            result => {
-                trace!(
-                    "not adding node {} / {:?} - Proof {:?} is not trusted.",
-                    p2p_node.name(),
-                    result,
-                    proof
-                );
-                return false;
-            }
+        // The section public key of the proof shall be known to us.
+        if !self.our_history.has_key(&proof.public_key) {
+            trace!(
+                "not adding node {} - Proof {:?} section_key is not known to us.",
+                p2p_node.name(),
+                proof
+            );
+            return false;
         }
 
         let name = *p2p_node.name();
@@ -221,18 +226,14 @@ impl SharedState {
         proof: Proof,
         recommended_section_size: usize,
     ) -> Option<MemberInfo> {
-        // The section public key of the proof shall be trusted by us.
-        match self.our_history.check_trust(iter::once(&proof.public_key)) {
-            TrustStatus::Trusted => {}
-            result => {
-                trace!(
-                    "not removing member {} / {:?} - Proof {:?} is not trusted.",
-                    name,
-                    result,
-                    proof
-                );
-                return None;
-            }
+        // The section public key of the proof shall be known to us.
+        if !self.our_history.has_key(&proof.public_key) {
+            trace!(
+                "not removing member {} - Proof {:?} section_key is not known to us.",
+                name,
+                proof
+            );
+            return None;
         }
 
         match self.our_members.get(name).map(|info| &info.state) {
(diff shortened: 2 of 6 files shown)


src/section/shared_state.rs
```diff
@@ -60,11 +62,19 @@ impl SharedState {
             );
         }
 
-        if self.our_history.len() > 1 && *self != new {
+        if self.our_history.len() > 1 {
+            if *self != new {
```
dirvine, line 66: I really want to get this log_or_panic gone ASAP. In this case here will it matter? I mean, it's an Error and a valid Error, why log or panic, if we want to test this we can test for that error. 
If these are semilattice then perhaps we can be even more smart?

maqi: was just following the previous implementation.
I think the idea here is that shall not be happen, and if happens during test, we want it to be exposed immediately and explicitly.

However, I think there is missing `return` for these `error branches` ? @madadam

dirvine: Yes, I understand but let's get to launch now and no more log or panic. We don't panic unless we have an unrecoverable error. I am not sure we have that here so we should return the error or fix the error. If we return it we should log Error and if we fix it then perhaps log Debug? 
Panic in tests would mean we did neither and the production code has to do one of these things (return or fix). I say now is the time to stop using log_or_panic and getting rid of it when we are working on those bits of code. 

It's a better cleaner solution to not use this from this point forward.  :+1:

maqi: ok, I will change it to log and return with an error

dirvine: Or just `?` if it can be handled upstream.

src/section/shared_state.rs
```diff
@@ -52,25 +55,28 @@ impl SharedState {
...
-            );
+        if self.our_history.len() > 1 {
+            if *self != new {
+                error!(
+                    "shared state update - mismatch: old: {:?} --- new: {:?}",
+                    self, new
+                );
+                return Err(RoutingError::InvalidState);
+            }
+        } else if !self.our_history.self_verify() {
```
madadam, line 72: Two points:
1. You should do the check on `new`, not `self`. `self` is presumably valid, because we only allow approved mutations. But we don't know whether `new` is, because we just received it from someone.
2. You should verify everything, not just `our_history`.

src/section/shared_state.rs
```diff
@@ -173,15 +179,32 @@ impl SharedState {
         // FIXME: we should perform these checks before voting, but once the vote accumulates we
         // must obey it:
         if !self.our_prefix().matches(p2p_node.name()) {
-            trace!("not adding node {} - not matching our prefix", p2p_node);
+            trace!(
+                "not adding node {} - not matching our prefix",
+                p2p_node.name()
```
madadam, line 184: Why this change?

src/section/shared_state.rs
```diff
@@ -173,15 +179,32 @@ impl SharedState {
...
         }
 
         if self.our_members.contains(p2p_node.name()) {
-            trace!("not adding node {} - already a member", p2p_node);
+            trace!("not adding node {} - already a member", p2p_node.name());
             return false;
         }
 
+        // The section public key of the proof shall be trusted by us.
+        match self.our_history.check_trust(iter::once(&proof.public_key)) {
```
madadam, line 195: So this checks that the proof key is known to us, but we should also check the proof itself is valid by verifying the signature. This applies to other places too.

madadam: Also, `check_trust` isn't supposed to be used like this, although it might still happen to work. What `check_trust` does, is: given a collection of keys that we already trust, return whether the proof chain itself is trusted. So you would use it to verify the proof chain itself, not the keys. But we don't need to verify the proof chain here, we can assume it is valid already. What you should use is just `self.our_history.has_key(&proof.public_key)`.

src/section/shared_state.rs
```diff
@@ -264,6 +301,22 @@ impl SharedState {
...
+            result => {
+                trace!(
+                    "not updating our section {:?} - Old section key {:?} is not trusted.",
+                    result,
+                    section_key.proof.public_key
+                );
+                return;
+            }
+        }
+
```
madadam, line 319: Again, just use `self.our_history.has_key(&section_key.proof.public_key)`. But also verify the signatures and also verify `elders_info`, not just `section_key`. It might be useful to add a `verify` method to `Proven`, to avoid duplication.

src/section/shared_state.rs
maqi: > You should do the check on new

Ah, right. 

> You should verify everything, not just our_history

That means a check against the serialised info itself? I thought such check shall be carried out inside the message accumulation already?
Regarding checking each field individually, I am not sure for each field, which value is `valid`.
For example, `handled_genesis_event` shall be `false` or `true` for new? and `our_members` or `relocate_queue` shall be what? etc.

src/section/shared_state.rs
maqi: just to make the log message cleaner and consistent with outputs in other places

src/section/shared_state.rs
maqi: > but we should also check the proof itself is valid by verifying the signature. This applies to other places too

I thought this has been carried out during message accumulation already?

> What you should use is just self.our_history.has_key(&proof.public_key)

Ok, will change to this

src/section/shared_state.rs
maqi: > But also verify the signatures and also verify elders_info, not just section_key. It might be useful to add a verify method to Proven, to avoid duplication.

I thought such signature check has been carried out already when accumulation. Will have a check through.

src/section/shared_state.rs
madadam: > I thought such check shall be carried out inside the message accumulation already

No, the idea is that any node can send us a message containing a `SharedState` and we should be able to merge it with outs if it is valid. There will be no accumulation for this, because all the info contained in `SharedState` is already signed by the section quorum. Currently this isn't necessary, because we only call `update` in response to accumulated `Genesis` event, but that will eventually change.

madadam: > For example, handled_genesis_event shall be false or true for new? and our_members or relocate_queue shall be what? etc.

Right, so not everything is ready for parsec removal yet, but we can do at least those parts that are. So:
- `handled_genesis_event` can be ignored (it will always be `false` for incoming shared state)
- `relocate_queue` can be ignored too, for now, at least until we figure out what to do with it (we have a separate issue for it and I think you started looking into it?)
- `our_members` can be verified normally, because the entries in it (`MemberInfo`) all have proofs attached.
- Same with `SectionMap`

src/section/shared_state.rs
madadam: OK. I think it might sometimes be useful to see the full `P2pNode` in the log, not just the name, but it's probably fine here.

src/section/shared_state.rs
maqi: Ok, will do the checks.

>it will always be false for incoming shared state

Then maybe just check the incoming `new`'s `handled_genesis_event ` shall be `false` ?

src/section/shared_state.rs
madadam: > I thought this has been carried out during message accumulation already?

Yeah that's true. Hm. Then maybe it's not necessary to do those checks here. Only where we know the input might come from untrusted sources (so for example directly from a received message, not accumulated).

maqi: > so for example directly from a received message, not accumulated

yes.
So, basically, if the current mutation function has a `proof` passed it, it indicates it's `accumulated`, wo we don't need to check the signature validity anymore.
If no `proof` provided, then it maybe a direct message, then shall be checked. (however, maybe due to the serialisation/derialization, the checking place shall be in the calling place, not here.)

src/consensus/proof.rs
```diff
@@ -6,6 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use serde::Serialize as SerdeSerialize;
```
madadam, line 9: Why not just `use serde::Serialize` ?

maqi: avoid confliction with std::Serialize

src/section/shared_state.rs
```diff
@@ -52,25 +53,34 @@ impl SharedState {
...
-            );
+        if self.our_history.len() > 1 {
+            if *self != new {
+                error!(
+                    "shared state update - mismatch: old: {:?} --- new: {:?}",
+                    self, new
+                );
+                return Err(RoutingError::InvalidState);
+            }
+        } else if !new.self_verify() {
```
madadam, line 70: One more thing: verify that `self.our_history` and `new.our_history` have at least one key in common. Without it someone could in theory send us a completely fabricated `SharedState` which would still pass `self_verify`.

src/section/shared_state.rs
```diff
@@ -52,25 +53,34 @@ impl SharedState {
...
 
         *self = new;
         self.handled_genesis_event = true;
+        Ok(())
+    }
+
+    fn self_verify(&self) -> bool {
+        self.our_history.self_verify()
+            && self.our_members.self_verify()
+            && self.sections.self_verify()
```
madadam, line 83: We are getting there :) But we should also check that all the proofs in all the containers are signed with keys that exists in `self.our_history`. WIth that and https://github.com/maidsafe/routing/pull/2155/files#r452245185, we should be good.

src/consensus/proof.rs
madadam: What is `std::Serialize` ?

madadam: Do you mean `derive(Serialize)`? If yes, then that is not `std` but also `serde` and its a different namespace (trait vs proc-macro) so it doesn't conflict.

maqi: ah, my mistake, will change it

src/section/shared_state.rs
maqi: ok. will do

src/section/shared_state.rs
maqi: that will requires the key history to be pass down across couple of structs.  :(
will have a try

src/consensus/proof.rs
```diff
@@ -75,20 +77,28 @@ impl Debug for ProofShare {
...
     pub proof: Proof,
 }
 
-impl<T> Proven<T> {
+impl<T: Serialize> Proven<T> {
     pub fn new(value: T, proof: Proof) -> Self {
         Self { value, proof }
     }
+
+    pub fn self_verify(&self, history: &SectionProofChain) -> bool {
```
madadam, line 90: Very minor point: maybe this should no longer be called `self_verify` as it does the trusted keys check too? Maybe call it just `verify`?.

madadam:
Also maybe add some unit tests for the new logic? Could be done in a separate PR too.

PR #2158, opened 2020-07-13
feat/event: include previous name if relocated

maqi:


src/consensus/network_event.rs
```diff
@@ -37,6 +37,8 @@ pub enum AccumulatingEvent {
     Online {
         /// Identifier of the joining node.
         p2p_node: P2pNode,
+        /// Previous name if relocated.
+        previous_name: Option<XorName>,
```
dirvine, line 41: Only Infants will have None here.

maqi: mention this within the comment?

PR #2159, opened 2020-07-15
refactor/accumulator: move SignatureAccumulator logic to safe-network-signature-accumulator

Yoga07:


src/consensus/proof.rs
```diff
@@ -6,75 +6,11 @@
...
-            "ProofShare {{ public_key: {:?}, index: {}, .. }}",
-            self.public_key_set.public_key(),
-            self.index
-        )
-    }
-}
-
 /// A value together with the proof that it was agreed on by the quorum of the section elders.
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Serialize, Deserialize)]
 pub struct Proven<T: Serialize> {
```
maqi, line 16: better change the file name to Proven

src/messages/message_accumulator.rs
```diff
@@ -7,15 +7,15 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::{
-    consensus::{AccumulationError, SignatureAccumulator},
     messages::{AccumulatingMessage, Message, PlainMessage, SrcAuthority},
     section::SectionProofChain,
 };
+use safe_network_signature_aggregator::{AccumulationError, SignatureAggregator};
```
maqi, line 13: these have been re-exposed from the consensus mod.
no need to change here, or no need to re-expose inside consensus

Cargo.toml
```diff
@@ -36,6 +36,7 @@ rand_crypto = { package = "rand", version = "~0.6.5" }
 rand_core = "~0.5.1"
 rand_os = "~0.2.2"
 rand_xorshift = "~0.2.0"
+safe-network-signature-aggregator = { git = "https://github.com/Yoga07/safe-network-signature-aggregator" }
```
maqi, line 39: need to point to maidsafe repository before merge.

src/lib.rs
```diff
@@ -79,7 +79,6 @@ mod macros;
 // Public API
 // ############################################################################
 pub use self::{
-    consensus::{AccumulationError, Proof, ProofShare, SignatureAccumulator},
     error::RoutingError,
```
dirvine, line 82: I suspect this crate should have it's own error type and not RotuingError. If we use err_derive then it will autmatically impl From for us.

PR #2160, opened 2020-07-15
Overhaul promotion

### LINKED ISSUE #2126: Overhaul node promotion / remove `Genesis` event

maqi:
Closes #2126

Closes https://github.com/maidsafe/routing/pull/2157


src/node/stage/approved.rs
```diff
@@ -1611,45 +1642,69 @@ impl Approved {
...
         if new_prefix.is_extension_of(&old_prefix) {
             info!("Split");
-        } else if old_prefix.is_extension_of(&new_prefix) {
+
+            self.send_promote(
+                core,
+                old_shared_state,
+                self.consensus_engine.parsec_version() + 1,
+                promoted_nodes.clone(),
+            )?;
```
madadam, line 1690: Shouldn't we send this only to the nodes from the sibling section?

src/node/stage/approved.rs
```diff
@@ -1611,45 +1642,69 @@ impl Approved {
...
+        self.section_update_barrier = Default::default();
         self.reset_parsec(core, self.consensus_engine.parsec_version() + 1)?;
 
-        if !is_elder {
+        self.send_promote(
+            core,
+            self.shared_state.clone(),
+            self.consensus_engine.parsec_version(),
+            promoted_nodes,
+        )?;
```
madadam, line 1704: Similarly, in case of a split, shouldn't we send this only to the nodes from our section?

src/node/stage/approved.rs
```diff
@@ -1660,22 +1715,21 @@ impl Approved {
         // on our `OurKey` so they know our latest key. Need to vote for it first though, to
         // accumulate the signatures.
         if let Some(prefix) = sibling_prefix {
+            self.send_send_neighbour_info()?;
```
madadam, line 1718: Why are we doing this? We are not proactively updating neighbour sections about our section in any other case, so why here?

src/node/stage/approved.rs
```diff
@@ -568,19 +570,61 @@ impl Approved {
...
+                // Ignore `InvalidElderDkgResult` because it just means the DKG hasn't completed
+                // for us yet.
+                // TODO: check that we have an ongoing DKG instance corresponding to the latest
+                // EldersInfo.
+            }
+            Err(error) => return Err(error),
+        }
+
+        self.send_elders_update(core)?;
+        self.send_send_neighbour_info()?;
```
madadam, line 619: I think the idea was that we would only send `NeighbourInfo` to other section if we detect their knowledge of us is outdated which happens when we receive a message from them. Why are we doing it proactively here now?

src/node/stage/approved.rs
```diff
@@ -568,19 +570,61 @@ impl Approved {
...
+        self.send_their_knowledge();
```
madadam, line 620: I'm confused by this. This basically means we are voting to update the knowledge of our section by other sections to be up to date, but how do we know that to be the case?

src/node/stage/approved.rs
maqi: This is mainly for the split case. During split, especially for the testing setup, there will be no message flowing from one new-sub-section to the other. Hence we will have to send NeighbourInfo to other section actively.

src/node/stage/approved.rs
maqi: Same as above, this is mainly for the split case. Without this, there will be not-enough votes to consensus the knowledge update.

src/node/stage/approved.rs
maqi: the receiver need to carry out a check to `ignore` the incoming message if it's not in the sibling sectionm anyway.
Didn't carry out filtering here, on sender's side, is mainly to simplify the coding, as there is another `send_promote` call immediately below.

I will consider adding a check inside the send_promote function to see if works.

src/node/stage/approved.rs
maqi: same comment as above

src/node/stage/approved.rs
maqi: same reason as mentioned above. mainly for the split case

src/node/stage/approved.rs
dirvine: Do you mean SendSibling as opposed to SendNeighbor in this case?

maqi: Yes. However, didn't want to introduce extra SendSibling message/variant, so used the `neighbour` stuff at that moment.

dirvine: Yes I see, but I mean should that not be renamed as currently it is specific only to Siblings and we don't have a neighbor as a concept unless/until we have to again have neighbors when the network is several millinois of ndoes or whatever.

src/node/stage/approved.rs
madadam: OK, but how do we know that the knowledge is really up-to-date? Because what we are doing here is voting that every other section has up-to-date knowledge of out section, but I don't think that's necessarily the case. We can only know that if we receive a message from them and the message's `dst_key` field points to our latest key.

src/node/stage/approved.rs
madadam: > During split, especially for the testing setup, there will be no message flowing from one new-sub-section to the other.

This sounds wrong. So you are basically adjusting the production code to accommodate the tests, instead of the other way around. In the real network there will always be some messages flowing all the time, so we should not need this. In the tests, we should simulate those message instead. And we actually already do, there is a `update_neighbours_and_poll` function in the test utils for this very purpose.

madadam: > as currently it is specific only to Siblings 

It's not specific to siblings, as far as I can see. Siblings get updated about each other immediately, because they process the `SectionInfo` vote of both subsections. It's the non-siblings that we need to update. Although my claim is that we actually don't as it happens automatically as messages are being sent over the network.

> and we don't have a neighbor as a concept

We still do. That change hasn't been implemented yet.

dirvine: > Although my claim is that we actually don't as it happens automatically as messages are being sent over the network.
I agree completely with this. 
>We still do. That change hasn't been implemented yet.
Yes cheers I was meaning as a design concept, but yea we need to remove that code.

maqi: >there is a update_neighbours_and_poll function in the test utils for this very purpose

src/node/stage/approved.rs
maqi: addressed by PR https://github.com/maidsafe/routing/pull/2166

src/node/stage/approved.rs
maqi: addressed by PR https://github.com/maidsafe/routing/pull/2166

src/node/stage/approved.rs
maqi: The newly promoted nodes won't vote, so sometimes during split there will be not enough votes.
So here will have to vote during handle_promote (the updated shared_state shall contains the latest info)

src/node/stage/approved.rs
maqi: The newly promoted nodes won't vote, so sometimes during split there will be not enough votes.

src/node/stage/approved.rs
madadam: So the `send_their_knowledge` call should be replaced with a vote for `TheirKnowledeg` but only for the sibling section and only if there is a split.

src/node/stage/approved.rs
madadam: Yes, but that's a different issue - it's about the knowledge update. This issue here is about us sending the `NeighbourInfo` proactively which we shouldn't do.

madadam: So the resolution of this is that the `send_send_neighbour_info` call should be removed.

src/node/stage/approved.rs
maqi: addressed by PR https://github.com/maidsafe/routing/pull/2170

src/node/stage/approved.rs
maqi: addressed by PR https://github.com/maidsafe/routing/pull/2170

src/node/stage/approved.rs
maqi: addressed by PR https://github.com/maidsafe/routing/pull/2170

PR #2161, opened 2020-07-20
Message voting

### LINKED ISSUE #2121: Implement message-based voting (parsec-less voting)
Implement voting that doesn't require parsec. It should work something like this:

- A wants to vote on some payload `X`
- It creates a message which contains:
    - `X`
    - the signature share of `X` created with the node's secret key share
    - the corresponding public key set
- The node then broadcasts this message to the rest of the elders
- On receiving this message, the node verifies the signature share is valid and the public key set is known
- Then it inserts it into some kind of accumulator
- Once enough shares are received, it combines them into full signature
- At this point the voting is over and `X` can be actioned 

This gives us consensus without strong order which should be enough for everything routing needs.

maqi:
Implements the framework for message based voting.

Closes #2121
Closes https://github.com/maidsafe/routing/pull/2153

### DIFF
--- a/src/section/section_keys.rs
+++ b/src/section/section_keys.rs
@@ -16,7 +16,7 @@ use std::collections::{BTreeMap, BTreeSet};
 use xor_name::XorName;
 
 /// All the key material needed to sign or combine signature for our section key.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct SectionKeyShare {
     /// Public key set to verify threshold signatures and combine shares.
     pub public_key_set: bls::PublicKeySet,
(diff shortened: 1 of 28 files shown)


src/section/section_keys.rs
```diff
@@ -38,13 +39,25 @@ pub struct SectionKeysProvider {
 }
 
 impl SectionKeysProvider {
+    pub fn has_key_or_candidate(&self) -> bool {
+        self.current.is_some() || !self.new.is_empty()
```
dirvine, line 43: Is this `|| !self.new.is_empty()` valid in this sense, could it have content but not the content we wish?

maqi: > could it have content but not the content we wish

There is check when populating this field, and this field will be clear once got updated properly.
So, I think we don't need to carry a further check here.

dirvine: If the populated field is immutable then I think this is fine, but if it can be mutated, perhaps we can look a bit further? 
i.e. I think correct and immutable on construction is perfectly valid and then this all is valid as well if that makes sense?

maqi: Ok, will have a check.

PR #2162, opened 2020-07-22
feat/event: send event when promoted to Adult

### LINKED ISSUE #2156: Fire event from routing when a Vault is promoted to an Adult
There is no way for a Vault to know when it has been promoted to Adult status. In the community testnet, it was observed that the lack of this distinction gave room for doubt since chunks were not being stored yet.

Yoga07:
Resolves #2156

### DIFF
--- a/src/section/member_info.rs
+++ b/src/section/member_info.rs
@@ -94,7 +94,10 @@ impl MemberInfo {
         self.age_counter >= AgeCounter(2_u32.pow(MAX_INFANT_AGE + 1))
     }
 
-    #[cfg(feature = "mock_base")]
+    pub fn is_new_adult(&self) -> bool {
+        self.age_counter_value() == 2_u32.saturating_pow(MAX_INFANT_AGE + 1)
+    }
+
     pub fn age_counter_value(&self) -> u32 {
         self.age_counter.0
     }
--- a/src/section/shared_state.rs
+++ b/src/section/shared_state.rs
@@ -194,14 +194,16 @@ impl SharedState {
     }
 
     /// Adds new member if its name matches our prefix and it's not already joined.
-    /// Returns whether the member was actually added.
+    /// Returns whether the member was actually added and whether our node gets promoted to Adult
+    /// as a part of this churn.
     pub fn add_member(
         &mut self,
         p2p_node: P2pNode,
         age: u8,
         proof: Proof,
         recommended_section_size: usize,
-    ) -> bool {
+        our_name: &XorName,
+    ) -> (bool, bool) {
         // FIXME: we should perform these checks before voting, but once the vote accumulates we
         // must obey it:
         if !self.our_prefix().matches(p2p_node.name()) {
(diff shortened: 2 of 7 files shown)


src/section/member_info.rs
```diff
@@ -94,6 +94,10 @@ impl MemberInfo {
         self.age_counter >= AgeCounter(2_u32.pow(MAX_INFANT_AGE + 1))
     }
 
+    pub fn is_new_adult(&self) -> bool {
+        self.age_counter == AgeCounter(2_u32.pow(MAX_INFANT_AGE + 1))
```
maqi, line 98: why not using the following `age_counter_value` method?
so that no need to construct a AgeCounter

src/section/shared_state.rs
```diff
@@ -568,6 +578,10 @@ impl SharedState {
                 continue;
             }
 
+            if member_info.is_new_adult() {
+                core.send_event(Event::PromotedToAdult);
```
maqi, line 582: I'd prefer make core sending this notification within the approved.rs, instead of having core passed down.
you could make `increment_age_counters` return a bool indicating whether it `becomes` a new adult

src/section/shared_state.rs
```diff
@@ -201,20 +202,21 @@ impl SharedState {
         age: u8,
         proof: Proof,
         recommended_section_size: usize,
-    ) -> bool {
+        our_name: &XorName,
+    ) -> (bool, bool) {
```
maqi, line 206: better expand the function comment to explain the meaning of the retuned value.

src/section/shared_state.rs
```diff
@@ -241,34 +243,39 @@ impl SharedState {
         &mut self,
         name: &XorName,
         proof: Proof,
-        recommended_section_size: usize,
```
maqi, line 244: better still use this recommended_section_size as input.
avoid importing `Core` for this file.

src/section/shared_state.rs
```diff
@@ -241,34 +243,39 @@ impl SharedState {
...
-                return None;
+                return (None, new_adult);
             }
             Some(MemberState::Relocating { .. }) => (),
             Some(MemberState::Joined) => {
-                self.increment_age_counters(name, recommended_section_size)
+                new_adult = self.increment_age_counters(
+                    name,
+                    core.network_params.recommended_section_size,
+                    core.name(),
```
maqi, line 272: better pass by name directly

src/section/member_info.rs
```diff
@@ -94,7 +94,10 @@ impl MemberInfo {
         self.age_counter >= AgeCounter(2_u32.pow(MAX_INFANT_AGE + 1))
     }
 
-    #[cfg(feature = "mock_base")]
+    pub fn is_new_adult(&self) -> bool {
+        self.age_counter_value() == MAX_INFANT_AGE + 1
```
maqi, line 98: self.age_counter_value() is actually the real number of churned times, which is `2's power of age`.
I think the function here needs to be :
`self.age_counter_value() == 2_u32.saturating_pow(u32::from(MAX_INFANT_AGE + 1))`

madadam:
This doesn't seem to be correct. It seems the `PromotedToAdult` is only fired when the current node is simultaneously elders and infant, which can happen only when the section has less than elder_size nodes. In the normal case of going from infant to adult, it won't be fired. This is because the logic to fire it is inside the handling of the `Online` / `Offline` events which only the elders do.

Also, there are no tests for the new functionality. I think we agreed that any new feature must come with tests?

Anyway, I think the correct way to implement this is that the elders would send some kind of message to the new adults and the new adults would then fire the event. This is because normally nodes don't know their own age unless they are elders.

Issue #2122, opened 2020-06-17
Improve message signature accumulation to not implicitly require parsec


madadam:
Currently message signature accumulation implicitly requires parsec because it assumes that all the nodes that create the signatures do so with the same key and they choose the same nodes to accumulate them. If the signature sending is synchronized with parsec (either explicitly like `SendNeighbourInfo` or implicitly via some other event) then these assumptions are guaranteed. When we remove parsec we lose this guarantee and we must implement some alternative. 

Some ideas:
- Use all the elders as the accumulation targets. That way the probability of successful accumulation should be high even in the case of ongoing churn
- Make sure to not mix message signatures that use different public key
- When receiving message signature with outdated yet, bounce it back together with all the info necessary for the sender to catch up re-sign it with the up-to-date key (reuse as much as possible from the current message bounce mechanism)

maqi:
> Use all the elders as the accumulation targets. That way the probability of successful accumulation should be high even in the case of ongoing churn

The DKG voting process enforce all participants to contribute. And the threshold could be updated according to network status.

>Make sure to not mix message signatures that use different public key

This has been covered by https://github.com/maidsafe/safe-network-signature-aggregator/pull/4

> When receiving message signature with outdated yet, bounce it back together with all the info necessary for the sender to catch up re-sign it with the up-to-date key (reuse as much as possible from the current message bounce mechanism)

We further define the situations to be: `the receiver is lagging` and `the sender is lagging` . And had https://github.com/maidsafe/routing/pull/2163 to cover these two cases.

PR #2163, opened 2020-07-27
feat/approved: notify lagging elders

maqi:
Closes https://github.com/maidsafe/routing/issues/2122

### DIFF
--- a/src/messages/variant.rs
+++ b/src/messages/variant.rs
@@ -49,6 +49,11 @@ pub(crate) enum Variant {
         shared_state: SharedState,
         parsec_version: u64,
     },
+    /// Message sent to a lagging peer.
+    NotifyLagging {
+        shared_state: SharedState,
+        parsec_version: u64,
+    },
     /// Send from a section to the node being relocated.
     Relocate(RelocateDetails),
     /// Sent from members of a section message's source location to the first hop. The
@@ -142,6 +147,15 @@ impl Debug for Variant {
                 parsec_version,
             } => f
                 .debug_struct("Promote")
+                .field("elders_info", shared_state.sections.our())
+                .field("section_key", shared_state.our_history.last_key())
+                .field("parsec_version", parsec_version)
+                .finish(),
+            Self::NotifyLagging {
+                shared_state,
+                parsec_version,
+            } => f
+                .debug_struct("NotifyLagging")
                 .field("elders_info", shared_state.sections.our())
                 .field("section_key", shared_state.our_history.last_key())
                 .field("parsec_version", parsec_version)
--- a/src/node/stage/approved.rs
+++ b/src/node/stage/approved.rs
@@ -254,6 +254,23 @@ impl Approved {
         }
     }
 
+    pub fn check_lagging(&mut self, core: &mut Core, peer: &SocketAddr, proof_share: &ProofShare) {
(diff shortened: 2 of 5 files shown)


src/messages/variant.rs
```diff
@@ -49,6 +49,11 @@ pub(crate) enum Variant {
         shared_state: SharedState,
         parsec_version: u64,
     },
+    /// Message sent to a lagging peer.
+    NotifyLagging {
+        shared_state: SharedState,
+        parsec_version: u64,
```
dirvine, line 55: Why parsec_version if that's what we are removing?  I think this field can be removed completely. All we need send the lagging peer is the `ProofChain` he needs plus current elders AFIAK.

src/node/stage/approved.rs
```diff
@@ -624,7 +653,9 @@ impl Approved {
         // With the removal of parsec, sending parsec_gossip won't update the peer anymore.
         // We will have to send the message later on to allow the peer be updated by other delayed
         // messages.
-        // TODO: use other duration schedule.
+        // TODO: When the testing framework is updated to use real messaging underlayer, i.e. not
+        // using fake clock, the bounced unknown message could be resent immediately. As the network
+        // already took time to exchange messages, which allows the lagging node to be updated.
         self.gossip_timer_token = Some(core.timer.schedule(self.consensus_engine.gossip_period()));
```
dirvine, line 659: I would also remove this gossip line and we will test this with real network if necessary. We cannot have test code in our production code. So I would go for the real deal here as well.

src/messages/variant.rs
maqi: > Why parsec_version if that's what we are removing?

This just to be consistent with other variant (and the handling logic as well). I think it will be easy for the `parsec removing` PR to remove all this stuff at once.

> All we need send the lagging peer is the ProofChain he needs plus current elders

SharedState is almost about this, especially there is further plan to remove its `relocate_queue` further.

src/node/stage/approved.rs
maqi: I agree with not having test code in the production code. 
However I think that will be out of the scope of this PR.

Another option is put this under a feature flag, but I think it shall be avoided as well?

Maybe the PR of using real deal for test shall be priorized ?

dirvine: Yes I agree, we must just get well away from mock as it's cluttering up logic now. It's actually worse here as it's forcing us to adopt logic we defo do not want. Fake time is a real problem for logic in decentralised network so we must get away from it very fast. 
I think we need to get parsec gone, go async and then do real tests as we cannot rely on the test suite for real world validation. At least we are all on the same boat here and if mock is used at all it will be restricted to unit tests.

src/messages/variant.rs
dirvine: I feel we should make the consistency removing this from the other variant. It would make more sense I feel.

maqi: Well, this is bit like a chicken-egg issue: this PR is supposed to be a pre-condition of removal parsec, and consistency of removing parsec_version will mean remove of parsec.

I do feel it will be better have this PR contain this, and remove all in a later on PR.

dirvine: Ok then I suppose we can get the remove parsec done very soon and this can stop hurting our eyes :-)

PR #2165, opened 2020-07-30
fix/routing: remove SendNeighbourInfo votes

maqi:


src/node/stage/approved.rs
```diff
@@ -2431,9 +2414,16 @@ impl Approved {
...
+                SendNeighbourInfo { dst, nonce } => self.send_routing_message(
+                    core,
+                    SrcLocation::Section(*self.shared_state.our_prefix()),
+                    DstLocation::Section(dst),
+                    Variant::NeighbourInfo {
+                        elders_info: self.shared_state.our_info().clone(),
+                        nonce,
+                    },
+                    None,
+                )?,
```
madadam, line 2426: I think we can do better than this. The elders info of our section is already signed, so we don't need to re-accumulate the messages. We can just send it as `Node` -> `Section` message. Just need to change the `NeighbourInfo` variant to contain `Proven<EldersInfo>` instead of just `EldersInfo`. Also we need to attach a sufficient proof to the message and implement the proof verification for this variant (see `Variant::verify`).

src/section/shared_state.rs
```diff
@@ -403,8 +403,6 @@ impl SharedState {
         // - `[]`
         // - `[VoteTheirKey]`
         // - `[VoteTheirKey, VoteTheirKnowledge]`
-        // - `[VoteSendNeighbourInfo]`
-        // - `[VoteSendNeighbourInfo, VoteTheirKnowledge]`
```
madadam, line 407: Better edit these comments to replace `VoteSendNeighbourInfo` -> `SendNeighbourInfo` instead of deleting them.

src/node/stage/approved.rs
```diff
@@ -2105,7 +2079,7 @@ impl Approved {
         Ok(())
     }
 
-    fn send_send_neighbour_info(&mut self) -> Result<()> {
+    fn send_neighbour_info(&mut self, core: &mut Core) -> Result<()> {
```
madadam, line 2082: As discussed elsewhere, I don't think we need this at all.

src/node/stage/approved.rs
maqi: Ok, will have a try with the change

src/node/stage/approved.rs
maqi: will have a double check once we settled with the other discussion.

maqi: covered by routing PR https://github.com/maidsafe/routing/pull/2170

src/messages/variant.rs
```diff
@@ -126,6 +128,21 @@ impl Variant {
...
+                let proof_chain = proof_chain.ok_or(RoutingError::InvalidMessage)?;
+                if !proof_chain.has_key(&elders_info.proof.public_key) {
+                    return Err(RoutingError::UntrustedMessage);
+                }
+
+                match proof_chain.check_trust(trusted_keys) {
+                    TrustStatus::Trusted => Ok(VerifyStatus::Full),
+                    TrustStatus::Unknown => Ok(VerifyStatus::Unknown),
+                    TrustStatus::Invalid => Err(RoutingError::UntrustedMessage),
+                }
```
madadam, line 144: Can we replace all this with just `elders_info.verify(proof_chain, trusted_keys)` ?

src/messages/variant.rs
```diff
@@ -30,8 +30,10 @@ use xor_name::XorName;
 pub(crate) enum Variant {
     /// Inform neighbours about our new section.
     NeighbourInfo {
-        /// `EldersInfo` of the neighbour section.
-        elders_info: EldersInfo,
+        /// `EldersInfo` of the sender's section, with the proof chain.
+        elders_info: Proven<EldersInfo>,
+        /// Latest section key of the sender's section
+        section_key: bls::PublicKey,
```
madadam, line 36: This might not be needed because we already have the key - it's the latest key of the proof chain of this message.

src/node/mod.rs
```diff
@@ -653,10 +653,19 @@ impl Node {
...
+                    section_key,
+                    nonce,
+                } => {
                     msg.dst().check_is_section()?;
-                    let src_key = *msg.proof_chain_last_key()?;
-                    stage.handle_neighbour_info(&mut self.core, elders_info.clone(), src_key)?;
+                    stage.handle_neighbour_info(
+                        &mut self.core,
+                        elders_info.value.clone(),
+                        *section_key,
```
madadam, line 665: So I think this can be replaced with something like `msg.proof_chain_last_key()?` and then we wouldn't need to include the `section_key` in the `NeighbourInfo` variant.

src/node/stage/approved.rs
```diff
@@ -676,10 +678,19 @@ impl Approved {
...
             .is_neighbour(self.shared_state.our_prefix())
         {
-            self.cast_unordered_vote(core, Vote::SectionInfo(elders_info))?;
+            self.cast_unordered_vote(core, Vote::SectionInfo(elders_info.clone()))?;
         }
 
-        Ok(())
+        if let Some(dst_key) = dst_key {
+            if self.shared_state.our_history.has_key(&dst_key)
+                && dst_key == *self.shared_state.our_history.last_key()
```
madadam, line 686: `dst_key == our_history.last_key()` implies `our_history.has_key(dst_key)`, so the first check is redundant.

src/node/mod.rs
```diff
@@ -653,10 +653,19 @@ impl Node {
...
+                        *msg.dst_key(),
+                        *nonce,
```
madadam, line 667: The `nonce` should be derived from the message that triggered sending the `NeighbourInfo`. in this case that's the incoming `NeighbourInfo` message itself, not the message that triggered it. So this should be `msg.hash()`.

src/node/stage/approved.rs
```diff
@@ -2430,15 +2415,40 @@ impl Approved {
                 VoteTheirKnowledge { prefix, key_index } => {
                     self.cast_unordered_vote(core, Vote::TheirKnowledge { prefix, key_index })?
                 }
-                VoteSendNeighbourInfo { dst, nonce } => {
-                    self.cast_ordered_vote(AccumulatingEvent::SendNeighbourInfo { dst, nonce })
+                SendNeighbourInfo { dst, nonce } => {
+                    self.send_neighbour_info(core, dst, nonce, None)?
```
madadam, line 2419: The `dst_key` should be obtained from `shared_state.sections.key_by_name(dst)`.

src/node/stage/approved.rs
```diff
@@ -2430,15 +2415,40 @@ impl Approved {
...
     }
 
+    fn send_neighbour_info(
+        &mut self,
+        core: &mut Core,
+        dst: XorName,
+        nonce: MessageHash,
+        dst_key: Option<bls::PublicKey>,
+    ) -> Result<()> {
+        let proof_chain = self.shared_state.create_proof_chain_for_our_info(None);
```
madadam, line 2434: I think we should pass `shared_state.sections.knowledge_by_location(DstLocation::Section(dst))` to this function to make sure the attached proof chain would be trusted by the destination.

tests/mock_network/utils.rs
```diff
@@ -896,6 +901,11 @@ pub fn update_neighbours_and_poll(env: &Environment, nodes: &mut [TestNode], thr
             for (a, b) in outdated {
                 let content = gen_vec(&mut rng, 32);
                 send_user_message(nodes, a, b, content);
+                // For the case that sectoin a got churned, but section b not.
```
madadam, line 904: Typo: "sectoin" -> "section"

src/node/mod.rs
maqi: I thought as this is a B->A NeighbourInfo triggered by a A-B NeighbourInfo, and the elders_info is different, so using the ame nonce can `safely` link them.
I am ok with your suggestion, will change it.

src/node/stage/approved.rs
maqi: I remember it was previous doing so however having issue, will roll back to check what was the issue (or maybe no longer have)

src/node/stage/approved.rs
maqi: You mean the proof_chain that passed to `Message::single_src` shall be `proof_chain = self.shared_state.sections.knowledge_by_location(DstLocation::Section(dst))`

madadam: Like this:
```rust
let proof_chain = self.shared_state.create_proof_chain_for_our_info(Some(DstLocation::Section(dst));
```

src/messages/variant.rs
maqi: the current `EldersInfo::verify` function only check against the proof_chain.
the part of `proof_chain.check_trust(trusted_keys)` is actually nothing related to `EldersInfo`

Also `EldersInfo::self_verify` is currently a separate function

I can make all this as a combined verify function within EldersInfo, any name suggestion for it?
how about `self_and_proof_chain_check` ?

madadam: Sorry, that's won't work. We should do something like this I think:
```rust
elders_info.verify(proof_chain)...
proof_chain.check_trust(trusted_keys)...
```

maqi: Then that will be almost the same as the current ?

maqi: maybe just change
```rust
 if !proof_chain.has_key(&elders_info.proof.public_key) {
                    return Err(RoutingError::UntrustedMessage);
                }
```
to
```rust
 if !elders_info.verify(proof_chain) {
      return Err(RoutingError::UntrustedMessage);
 }
```

madadam: Yeah, I mean the two calls: `elders_info.self_verify` and `proof_chain.has_key(elders_info.proof.public_key)` can be replaced by just `elders_info.verify(proof_chain)`. The rest should stay.

src/node/stage/approved.rs
```diff
@@ -676,10 +678,17 @@ impl Approved {
...
 
-        Ok(())
+        if let Some(dst_key) = dst_key {
+            if dst_key == *self.shared_state.our_history.last_key() {
+                // The sender's knowledge of us is latest.
+                return Ok(());
+            }
+        }
+
+        self.send_neighbour_info(core, elders_info.prefix, nonce, Some(src_key))
```
madadam, line 691: Wonder if this whole part is necessary. Isn't this going to be handled by `update_section_knowledge` triggered by the reception of this `NeighbourInfo` message?

maqi: Ah, it was required maybe because I improperly populated the dst_key field previously.

Got this block of code removed.

PR #2167, opened 2020-08-03
fix/approved: sending DKG message to the participants directly

maqi:
The previous broadcasting DKG messages by send_routing_message
will send DKG messages to elders multiple times un-necessarily.
Also could send to non-participants elder during split.
This commit reduce the duplication of messages by sending to
participants directly. Also fixed a bug that DKG messages between
two non-elder participants not get transmitted previously.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,6 @@ mock-quic-p2p = { git = "https://github.com/maidsafe/quic-p2p", optional = true 
 parsec = { git = "https://github.com/maidsafe/parsec" }
 quic-p2p = { version = "~0.7.0", features = ["upnp"] }
 rand = "~0.7.3"
-# rand in the versions used by the threshold-crypto crate
 rand_core = "~0.5.1"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -31,7 +31,6 @@ mock-quic-p2p = { git = "https://github.com/maidsafe/quic-p2p", optional = true
 parsec = { git = "https://github.com/maidsafe/parsec" }
 quic-p2p = { version = "~0.7.0", features = ["upnp"] }
 rand = "~0.7.3"
-# rand in the versions used by the threshold-crypto crate
 rand_core = "~0.5.1"
```
dirvine, line 34: Are these differnt rand versions still required? (cargo-udeps)

maqi: The comment was left over from last PR https://github.com/maidsafe/routing/pull/2164, when removing rand_ctypto, by mistake
rand_core is really in different version.

PR #2168, opened 2020-08-04
chore(deps): depend upon random xor branch

joshuef:


Cargo.toml
```diff
@@ -38,7 +38,7 @@ rand_xorshift = "~0.2.0"
 safe-network-signature-aggregator = { git = "https://github.com/maidsafe/safe-network-signature-aggregator" }
 serde = { version = "1.0.111", features = ["derive" ,"rc"] }
 tiny-keccak = { version = "2.0.2", features = ["sha3"] }
-xor_name = "1"
+xor_name = { git="https://github.com/joshuef/xor-name.git", branch="AddRandom"}
```
madadam, line 41: I think we agreed we would be using the patch section for this. So:

```
[dependencies]
...
xor_name = "1"

[patch.crates-io]
xor_name = { git = "https://...", branch = "AddRandom" }
```

madadam:
Also, do we actually need this? Routing is happily using the `rng.gen()` way of generating random `XorName`s and I don't see a reason to change that. So maybe this can wait until the change is merged into xor_name master and new version is released?

joshuef:
Yeh that makes sense. We can just shim it in other libs for now

madadam:
Seems like this change is already in master. Closing.

PR #2169, opened 2020-08-04
Remove Genesis event handling

madadam:
This PR removes the logic around handling of the `Genesis` parsec event which isn't needed now we have the `Promote` message. This should have been part of the PR to address https://github.com/maidsafe/routing/issues/2126, but it must have been missed. 

Also remove the explicit gossip peer adding which isn't needed either, because now only elders participate in parsec gossip.

### DIFF
--- a/src/node/stage/approved.rs
+++ b/src/node/stage/approved.rs
@@ -193,7 +193,7 @@ impl Approved {
         match self
             .section_keys_provider
             .key_share()
-            .and_then(|share| event.clone().into_signed_network_event(share))
+            .and_then(|share| event.clone().into_network_event(share))
         {
             Ok(event) => self.consensus_engine.vote_for(event),
             Err(error) => log_or_panic!(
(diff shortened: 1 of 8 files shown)


src/node/stage/approved.rs
```diff
@@ -2025,6 +1975,10 @@ impl Approved {
             }
         }
 
+        // FIXME: this should not be necessary because elder membership should always be up to date
+        // after parsec reset. But without this almost all the tests fail. Figure out why.
+        self.members_changed = true;
```
maqi, line 1980: reset_parsec will be called for pruning as well. 
setting members_change to true for such case sounds bit weird?

madadam: Yes, that's why I put in the FIXME. It's not that critical though, because the flag is there only to prevent calling `promote_and_demote_elders` too often, as it's quite expensive. So worst case is we might call it unnecessarily a few times, wasting some CPU cycles.

PR #2170, opened 2020-08-04
fix/approved: remove unnessary send_neighbour

maqi:
also only vote their_knowledge for sibling section when being a
newly promoted elder during split

### DIFF
--- a/src/node/stage/approved.rs
+++ b/src/node/stage/approved.rs
@@ -726,6 +726,7 @@ impl Approved {
         }
 
         info!("Promoted To Elder");
+        let old_prefix = self.shared_state.our_info().prefix;
 
         core.msg_filter.reset();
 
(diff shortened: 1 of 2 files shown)


src/node/stage/approved.rs
```diff
@@ -750,8 +752,13 @@ impl Approved {
...
         self.send_elders_update(core)?;
-        self.send_send_neighbour_info()?;
-        self.send_their_knowledge(core)?;
+
+        // Only need to vote during split and for sibling section, to accumulate enough votes.
+        if old_prefix != self.shared_state.our_info().prefix {
+            let prefix = self.shared_state.our_info().prefix.sibling();
+            let key_index = self.shared_state.our_history.last_key_index();
+            self.cast_unordered_vote(core, Vote::TheirKnowledge { prefix, key_index })?;
+        }
```
madadam, line 761: Nice, but need to do the same in `handle_lagging` too I think. Also at some point we should refactor these two methods, because there is some repetition between them. Not necessary in this PR though.

PR #2171, opened 2020-08-05
fix/approved: remove bounced_unknown_messages cache

maqi:


src/node/stage/approved.rs
```diff
@@ -642,23 +635,14 @@ impl Approved {
...
-        // With the removal of parsec, sending parsec_gossip won't update the peer anymore.
-        // We will have to send the message later on to allow the peer be updated by other delayed
-        // messages.
-        // TODO: When the testing framework is updated to use real messaging underlayer, i.e. not
-        // using fake clock, the bounced unknown message could be resent immediately. As the network
-        // already took time to exchange messages, which allows the lagging node to be updated.
-        self.gossip_timer_token = Some(core.timer.schedule(self.consensus_engine.gossip_period()));
-        self.bounced_unknown_messages
-            .push((*sender.peer_addr(), bounced_msg_bytes));
+        core.send_direct_message(
```
madadam, line 638: Would be nice to preserve the original comment that was here, just replace "parsec gossip" with "NotifyLagging":

>        // First send parsec gossip to update the peer, then resend the message itself. If the
>        // messages arrive in the same order they were sent, the gossip should update the peer so
>        // they will then be able to handle the resent message. If not, the peer will bounce the
>        // message again.

src/node/mod.rs
```diff
@@ -727,15 +727,18 @@ impl Node {
...
+                Variant::BouncedUnknownMessage(message) => {
+                    if let Some(sender_last_key) = msg.dst_key() {
+                        stage.handle_bounced_unknown_message(
+                            &mut self.core,
+                            msg.src().to_sender_node(sender)?,
+                            message.clone(),
+                            sender_last_key,
+                        )
+                    } else {
+                        unreachable!()
```
madadam, line 739: This is not really unreachable. A malicious/faulty peer can send us a message with missing `dst_key` and we would hit this branch and crash. Just log something and return `Err(InvalidMessage)` or something.

src/node/stage/approved.rs
```diff
@@ -558,12 +551,9 @@ impl Approved {
...
             &core.full_id,
             DstLocation::Direct,
-            Variant::BouncedUnknownMessage {
-                message: msg_bytes,
-                parsec_version: self.consensus_engine.parsec_version(),
-            },
-            None,
+            Variant::BouncedUnknownMessage(msg_bytes),
             None,
+            Some(*self.shared_state.our_history.last_key()),
```
madadam, line 556: Hm, I just realized that I might have misled you :flushed:  . We should not be using the `dst_key` field for this. The `dst_key` field is to indicate the latest key of the destination that we know. So if we are handling an unknown message from our own section, it would be OK to use `dst_key`, but this function is also used for handling messages from other sections and then it would make no sense to pass out own key. So I'm afraid it must be added back to `Variant::BouncedUnknownMessage`. Might call is `src_key` or something like that.

Alternatively, we can sort of abuse the `proof_chain` field for that, but that might be confusing, so not sure. Your call.

maqi: I think a `src_key` field within the `Variant::BouncedUnknownMessage` will be more clear.

PR #2172, opened 2020-08-05
fix/adult: fix firing event when new Adults are formed

Yoga07:


dirvine:
Will be addressed soon as too many changes have taken place in the codebase.

PR #2174, opened 2020-08-17
feat(api): expose an async API

bochaco:


src/node/stage/approved.rs
```diff
@@ -180,24 +150,28 @@ impl Approved {
...
+
+        // We need to relay it to ourself as well
+        // TODO: don't send it over the wire but just internally,
+        // making sure we don't create a recursive call.
+        self.comm
+            .send_message_to_target(&self.comm.our_connection_info()?, message.to_bytes())
+            .await?;
 
-        self.handle_unordered_vote(core, vote, proof_share)
+        Ok(())
```
madadam, line 168: Just call `handle_unordered_vote` here just like it was before. No need to worry about recursion, because accumulated vote might trigger another vote, but it's not going to be the same vote.

src/node/stage/approved.rs
```diff
@@ -924,68 +1079,72 @@ impl Approved {
...
+        if self.msg_filter.contains_incoming(&msg) {
             trace!("not handling message - already handled: {:?}", msg);
             return Ok(());
         }
 
-        match self.decide_message_status(core.id(), &msg)? {
+        match self.decide_message_status(&msg)? {
             MessageStatus::Useful => {
-                core.msg_filter.insert_incoming(&msg);
-                core.msg_queue.push_back(msg.into_queued(None));
```
madadam, line 979: Just call `handle_useful_message` here too.

src/node/stage/approved.rs
```diff
@@ -1628,34 +1811,48 @@ impl Approved {
...
+            .await?;
 
-        self.handle_dkg_message(core, dkg_key, dkg_message_bytes, *core.id())
+        // TODO: don't send it over the wire but just internally,
+        // making sure we don't create a recursive call.
+        self.comm
+            .send_message_to_target(&self.comm.our_connection_info()?, message.to_bytes())
+            .await?;
+
+        Ok(())
```
madadam, line 1840: Again, just call `handle_dkg_message` and don't worry about infinite recursion. It won't happen.

src/node/stage/bootstrapping.rs
```diff
@@ -6,73 +6,84 @@
...
     section::EldersInfo,
-    time::Duration,
 };
-
-use fxhash::FxHashSet;
-use std::{collections::HashMap, iter, mem, net::SocketAddr};
+use std::{iter, net::SocketAddr};
 use xor_name::Prefix;
 
+// TODO: review if we still need to set a timeout for joining
```
madadam, line 20: Would be good to keep it, but it can also be pushed up to vault.

src/node/stage/bootstrapping.rs
```diff
@@ -6,73 +6,84 @@
...
-pub const BOOTSTRAP_TIMEOUT: Duration = Duration::from_secs(20);
+// pub const BOOTSTRAP_TIMEOUT: Duration = Duration::from_secs(20);
 
 // The bootstrapping stage - node is trying to find the section to join.
 pub(crate) struct Bootstrapping {
-    // Using `FxHashSet` for deterministic iteration order.
-    pending_requests: FxHashSet<SocketAddr>,
-    timeout_tokens: HashMap<u64, SocketAddr>,
     relocate_details: Option<SignedRelocateDetails>,
-    msg_backlog: Vec<QueuedMessage>,
```
madadam, line 32: Why was this removed?

src/node/stage/bootstrapping.rs
```diff
@@ -187,12 +180,18 @@ impl Bootstrapping {
             *relocate_details.destination(),
         );
 
-        let new_full_id = FullId::within_range(&mut core.rng, &name_prefix.range_inclusive());
-        let relocate_payload =
-            RelocatePayload::new(relocate_details, new_full_id.public_id(), &core.full_id)?;
+        // FIXME: do we need to reuse MainRng everywhere really??
```
madadam, line 183: The rng reuse is there to maintain reproducible tests. But it only makes sense for tests that use the mock network. Because we seem to be moving away from it, we can maybe remove it too.

src/node/stage/comm.rs
```diff
@@ -0,0 +1,170 @@
...
+use std::{boxed::Box, net::SocketAddr, sync::Arc};
+
+// Number of Connections to maintain in the cache
+const CONNECTIONS_CACHE_SIZE: usize = 1024;
+
+// Communication component of the node to interact with other nodes.
+#[derive(Clone)]
+pub(crate) struct Comm {
+    quic_p2p: Arc<Box<QuicP2p>>,
+    endpoint: Arc<Box<Endpoint>>,
```
madadam, line 29: Again, why the `Box`? Simple `Arc<QuicP2p>` would be enough.

src/node/stage/comm.rs
```diff
@@ -0,0 +1,170 @@
...
+                HexFmt(&msg)
+            );
+        }
+
+        trace!(
+            "Sending message to {:?}",
+            &conn_infos[..delivery_group_size.min(conn_infos.len())]
+        );
+
+        // initially only send to delivery_group_size targets
```
madadam, line 111: We should also implement the resending to the other targets when the send fails. And report failure when the resend fails too many times. And then use that to detect nodes that went offline.

src/node/stage/joining.rs
```diff
@@ -6,24 +6,23 @@
...
     relocation::RelocatePayload,
-    section::EldersInfo,
+    section::{EldersInfo, SharedState},
 };
-
-use std::{mem, net::SocketAddr, time::Duration};
+use std::net::SocketAddr;
 use xor_name::Prefix;
 
+// TODO: review if we still need to set a timeout for joining
```
madadam, line 23: Same here. Would be good to keep this.

src/node/stage/joining.rs
```diff
@@ -33,47 +32,104 @@ pub(crate) struct Joining {
...
 
 impl Joining {
-    pub fn new(
-        core: &mut Core,
+    pub async fn new(
+        comm: Comm,
         elders_info: EldersInfo,
         section_key: bls::PublicKey,
         relocate_payload: Option<RelocatePayload>,
-        msg_backlog: Vec<QueuedMessage>,
```
madadam, line 48: Why was this removed?

src/node/stage/mod.rs
```diff
@@ -8,39 +8,362 @@
...
     Approved(Approved),
-    Terminated,
+}
+
+// Node's information.
+#[derive(Clone)]
+pub(crate) struct NodeInfo {
+    pub full_id: FullId,
+    pub network_params: Arc<Box<NetworkParams>>,
+    pub rng: Arc<Box<MainRng>>,
```
madadam, line 56: Again, not need to `Box` them.

src/section/member_info.rs
```diff
@@ -36,12 +36,15 @@ impl MemberInfo {
     }
 
     // Converts this info into one with the state changed to `Left`.
+    // TODO: review if we still need this function
```
madadam, line 39: We do. The reason why it seems like we don't is because you commented out all the processing of nodes going offline.

src/section/shared_state.rs
```diff
@@ -185,10 +184,13 @@ impl SharedState {
         self.our_members.is_adult(name) || self.is_peer_our_elder(name)
     }
 
+    // TODO: review if we still need this function
```
madadam, line 187: This was used to identify nodes that disconnected from us, so we can handle them properly (vote them Offline, etc...). So probably still needed.

src/node/stage/mod.rs
bochaco: agreed

src/section/member_info.rs
bochaco: yeah, i just put a generic comment in these places to remove clippy issues as suggested by Qi, so whenever we re-enable the logic for nodes going offline this can be re-enabled.

src/node/stage/joining.rs
bochaco: Basically to not cache/accumulate anything, if a message arrives and it was not for us we should ignore, if it arrived in a different order than expected we should be able to process it anyways, so there shouldn't be a need to keep messages in a backlog/cache for future processing. E.g. If we are bootstrapping still, and we receive a message which we'll need when we are in a diff stage, well it will need to be resent, too bad it was sent earlier which may be weird or a bug as well.

src/node/mod.rs
```diff
@@ -94,761 +84,223 @@ impl Node {
...
-        self.handle_messages();
-
-        Ok(())
-    }
-
-    /// Returns whether this node is running or has been terminated.
-    pub fn is_running(&self) -> bool {
-        !matches!(self.stage, Stage::Terminated)
+    /// Is this the genesis node or not
+    pub fn is_genesis(&self) -> bool {
```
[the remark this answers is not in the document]
bochaco: We don't, it was one way Josh found to fix this but we then fixed it later on by reporting the correct event, I'll remove it now. If eventually it's needed for informative purposes for the user we can add it back.

src/node/mod.rs
```diff
@@ -94,761 +84,223 @@ impl Node {
...
-        self.stage
-            .approved()
-            .map(|stage| stage.is_our_elder(self.core.id()))
-            .unwrap_or(false)
+    pub async fn is_elder(&self) -> bool {
+        let name = self.name().await;
+        match self.stage.lock().await.approved() {
+            None => false,
+            Some(stage) => stage.shared_state.sections.our().elders.contains_key(&name),
+        }
```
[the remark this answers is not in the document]
bochaco: nice! aside, those Ids/pub_ids are really confusing to me, i cannot still get my head around them.

src/node/stage/approved.rs
bochaco: This does create a recursion, you can try it and the compiler will complain, async doesn't like recursion unless you Box the Future apparently, so we either need to use the macro @Yoga07 used in other place or remove the recursion. I think we should remove the recursion with the proposal about not accumulating messages we discussed, and we'll be fine.

src/node/stage/approved.rs
bochaco: I put the `#[async_recursion]`, and kept a TODO to remove the recursion eventually as discussed before.

src/event.rs
```diff
@@ -95,8 +97,19 @@ pub enum Event {
...
+        content: Bytes,
+        /// The address of the client that sent the message.
+        src: SocketAddr,
+        /// The destination location that receives the message.
+        dst: DstLocation,
+        /// Stream to send messages back to the client that sent the message
+        send: SendStream,
+        /// Stream to receive more messages from the client on the same channel
+        recv: RecvStream,
+    },
```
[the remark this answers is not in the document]
madadam: I think it would be nicer from a consistency point of view, but it's not critical. There are more pressing issues to solve first. We can always come back to this later.

src/node/event_stream.rs
```diff
@@ -0,0 +1,132 @@
...
+                debug!("Failed to deserialize message: {:?}", error);
+            }
+            Ok(msg) => {
+                trace!("try handle message {:?}", msg);
+                // Process the message according to our stage
+                if let Err(err) = stage
+                    .lock()
+                    .await
+                    .process_message(sender, msg.clone())
+                    .await
```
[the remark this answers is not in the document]
madadam: Yeah, I was thinking about something similar, but not necessarily tied to CRDTs. Something like this:
- Keep the node state in a `RwLock` instead of `Mutex`, to allow multiple concurrent readers
- Each message handler is passed only immutable reference to the state (with `RwLock::read`)
- Instead of mutating the state, the handlers would push an `Op` into the op queue/channel
- There is going to be one dedicated task which will keep popping from the op queue and applying the op to the state. This is the only task that will have writable access to the state (using `RwLock::write`)
- We can also use this mechanism as sort of general purpose async notification system. So for example, when a message handler `tokio::spawn`s a task and that task needs to pass some info back to the system, it can also push it into the op queue. Some examples of where this would be useful is the DKG (which can run in its own task) and the message resending which needs some way to notify about failed peers.

How does that sound?

src/node/mod.rs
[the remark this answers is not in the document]
madadam: I hear you. We use `id` and `public_id` interchangeably, which is probably not ideal. This has been brought up before and I think it will be simplified soon.

src/node/mod.rs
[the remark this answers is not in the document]
madadam: Right. The reason I would prefer to not have this is that there shouldn't really be any difference between the first node and the other nodes apart from the bootstrapping process. So once they are online, they should be indistinguishable.

src/node/stage/approved.rs
madadam: I see, didn't know there are issues with async recusrion. Good to know.

src/node/stage/joining.rs
madadam: So on one hand, we should never have a situation where we absolutely need some message to be processed while also having the message being sent only once. So it shouldn't be necessary to backlog anything, because any missed message will be sent again. On the other hand, the backlog is an optimization to avoid the resends which can make things faster and involving less traffic. 

But OK, maybe we should first make sure things work even without the backlog, and possibly add it later purely as an optimization. Let's remove it then.

src/node/stage/mod.rs
```diff
@@ -8,39 +8,373 @@
...
 
 #[cfg(feature = "mock")]
 pub use self::{bootstrapping::BOOTSTRAP_TIMEOUT, joining::JOIN_TIMEOUT};
 
-// Type to represent the various stages a node goes through during its lifetime.
+// Maximum number of events to be buffered internally, when the buffer is full
+// no new events can be generated by this crate
+// TODO: if external connections or messages are arriving when
+// the buffer is full, they need to be rejected.
+const MAX_EVENTS_BUFFERED: usize = 1024;
```
madadam, line 42: We wouldn't need this if we used unbounded channels: https://docs.rs/tokio/0.2.22/tokio/sync/mpsc/fn.unbounded_channel.html

PR #2177, opened 2020-08-19
Fix incorrect section key assigned to non-elders

madadam:
When creating the new `SharedState` (on `EldersUpdate` or `NodeApproval`), we incorrectly used the previous section key instead of the latest one for the section chain. This made every `EldersUpdate` message bounce back as untrusted.

### DIFF
--- a/src/section/shared_state.rs
+++ b/src/section/shared_state.rs
@@ -40,9 +40,14 @@ pub(crate) struct SharedState {
 }
 
 impl SharedState {
-    pub fn new(elders_info: Proven<EldersInfo>) -> Self {
+    /// Creates a minimal `SharedState` initially contains only info about our section elders
+    /// (`elders_info`) and whose section chain contains only a single key (`section_key`).
+    /// Note: currently the `EldersInfo`s are not signed with the latest section key, but with the
+    /// one before that. Because of that, we need to pass the latest key explicitly via the
+    /// `section_key` argument. This might change in the future.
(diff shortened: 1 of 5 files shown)


src/section/shared_state.rs
```diff
@@ -40,9 +40,9 @@ pub(crate) struct SharedState {
 }
 
 impl SharedState {
-    pub fn new(elders_info: Proven<EldersInfo>) -> Self {
+    pub fn new(section_key: bls::PublicKey, elders_info: Proven<EldersInfo>) -> Self {
         Self {
-            our_history: SectionProofChain::new(elders_info.proof.public_key),
```
maqi, line 45: just want to confirm that `elders_info.proof.public_key` is the DKG key with participants of `elders_info.value (i.e. the elders)`? i.e. the section_key of the elders of the `elders_info`, right ?

madadam: No, that's the issue this PR is trying to fix. The key the `elders_info` is signed with is not the latest section key - it's the one before that. The reason for this is that we vote for the elders_info and the new section key at the same time, so we don't yet have the new key in the section chain. We could probably change it so that the elders_info is signed with the latest key (because we do have the new DKG result at that time), and maybe we should do it, but that's not how things currently work.

maqi: Ok, I got it.
Thx for the clarification

maqi: maybe worth to put your above wording as a comment in the code ?

madadam: :+1:  will do.

maqi (approved):

PR #2178, opened 2020-08-21
WIP - Vaults test debug

maqi:
The work mainly contains two fixes:
1, It is observed that some node got ordered Online vote consensused before a parsec version increase, meanwhile some other not and carried the vote into the new parsec instance. Which make those node never got consensused on the vote. So here casting it as unordered as well as a quick patch.

2, After a DKG completion, routing will carry out votes of OurKey and SectionInfo. Only after these votes got consensused, will then the section_key_index got updated. So, a routing level check of section_index will not be enough within the gap. i.e. any DKG message received during the gap will cause a new round of DKG for the same. Hence the info of completed highest section_key_index shall be recorded and checked against to cover the gap.


maqi:
shall be addressed by other work already.

PR #2179, opened 2020-08-24
fix/dkg_voter: record completed DKG section_key_index

maqi:
After a DKG completion, routing will carry out votes of OurKey and
SectionInfo. Only after these votes got consensused, will then the
section_key_index got updated. So, a routing level check of
section_index will not be enough within the gap. i.e. any DKG message
received during the gap will cause a new round of DKG for the same.
Hence the info of completed highest section_key_index shall be recorded
and checked against to cover the gap.

### DIFF
--- a/src/consensus/dkg.rs
+++ b/src/consensus/dkg.rs
@@ -16,7 +16,7 @@ use crate::{
 use bls::{PublicKeySet, SecretKeyShare};
 use bls_dkg::key_gen::{message::Message as DkgMessage, KeyGen};
 use std::{
-    collections::{BTreeMap, BTreeSet, VecDeque},
+    collections::{BTreeMap, BTreeSet, HashMap, VecDeque},
     fmt::{self, Debug, Formatter},
     mem,
 };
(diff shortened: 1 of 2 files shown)


src/consensus/dkg.rs
```diff
@@ -164,7 +176,10 @@ impl DkgVoter {
         full_id: &FullId,
         dkg_key: &DkgKey,
     ) -> Vec<DkgMessage<PublicId>> {
-        if self.key_gen_map.contains_key(dkg_key) {
+        if self.key_gen_map.contains_key(dkg_key)
+            || (self.current_section_key_index > 0 && self.current_section_key_index >= dkg_key.1)
+            || self.key_gen_map.iter().any(|(key, _)| key.1 == dkg_key.1)
```
madadam, line 181: Seems like the `self.key_gen_map.contains_key(dkg_key)` check is redundant now?

maqi: Ah, yes, you are right.
Will remove it.

madadam (approved):

Issue #2173, opened 2020-08-06
Handle DKGOldElders message only after accumulating a quorum of them


madadam:
Currently we handle it after only one message which is vulnerable to malice.

PR #2181, opened 2020-08-26
fix/dkg: Accumulate DKGOldElders messages

maqi:
Handle DKGOldElders message only after accumulating quorum of them.
To avoid vulnerable to malice.

Close https://github.com/maidsafe/routing/issues/2173

### DIFF
--- a/src/consensus/dkg.rs
+++ b/src/consensus/dkg.rs
@@ -15,11 +15,15 @@ use crate::{
 };
 use bls::{PublicKeySet, SecretKeyShare};
 use bls_dkg::key_gen::{message::Message as DkgMessage, KeyGen};
+use lru_time_cache::LruCache;
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap, VecDeque},
     fmt::{self, Debug, Formatter},
     mem,
+    time::Duration,
 };
+
+const OLD_ELDERS_EXPIRY_DURATION: Duration = Duration::from_secs(2 * 60);
 
 /// Returns the number of minumn responsive participants expected for the DKG process
 #[inline]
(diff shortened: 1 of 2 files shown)


src/consensus/dkg.rs
```diff
@@ -260,4 +267,24 @@ impl DkgVoter {
...
+            return false;
+        }
+        if let Some(senders) = self.old_elders_notifications.get_mut(participants) {
+            let _ = senders.insert(*src_id);
+            return senders.len() == threshold_count(participants.len());
+        }
+        let senders: BTreeSet<_> = iter::once(*src_id).collect();
+        let _ = self
+            .old_elders_notifications
+            .insert(participants.clone(), senders);
```
madadam, line 287: This can be made more succinct:

```rust
let senders = self.old_elders_notifications.entry(participants.clone()).or_default();
let _ = senders.insert(*src_id);
senders.len() == threshold_count(participants.len())
```

src/consensus/dkg.rs
```diff
@@ -15,12 +15,16 @@ use crate::{
...
+use lru_time_cache::LruCache;
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap, VecDeque},
     fmt::{self, Debug, Formatter},
-    mem,
+    iter, mem,
+    time::Duration,
 };
 
+const OLDELDERS_EXPIRY_DURATION: Duration = Duration::from_secs(2 * 60);
```
madadam, line 26: I would put underscore between `OLD` and `ELDERS`: `OLD_ELDERS_EXPIRY_DURATION`. It's more in line with rust style.

madadam (commented):
Looks good, just some style comments.

src/consensus/dkg.rs
maqi: still has to be something like:
```rust
        match self.old_elders_notifications.entry(participants.clone()) {
            Entry::Vacant(entry) => {
                let senders: BTreeSet<_> = iter::once(*src_id).collect();
                let _ = entry.insert(senders);
                false
            }
            Entry::Occupied(entry) => {
                let value = entry.into_mut();
                let _ = value.insert(*src_id);
                value.len() == threshold_count(participants.len())
            }
        }
```
as LruTimeCache has it's own definition of Entry

madadam (approved):

PR #2185, opened 2020-08-27
(WIP) Remove parsec

madadam:
This work-in-progress PR removes parsec from routing.

## Notes:

- There are some failing tests (more on this below)
- The relocation changes are based on this document: https://hackmd.io/7UecFpsiRmCg2Ck297GBRA but differ from it in some places. In particular, this PR currently doesn't limit the number of nodes to relocate at a time. This might be revisited in the future. This PR also additionally implements delayed relocation of elders to prevent relocating too many elders before the section has chance to elect the new ones. 
- The shared state is not yet full CRDT, although this is probably OK for now (doesn't need to addressed in this PR)
- The non-responsiveness detection is currently disabled because it was tied to parsec. It should be possible to adapt it to the message based voting too. It doesn't need to happen in this PR.

## Failing tests

Currently all integration tests that involve relocation fail most of the time (although sometimes they pass :)). This is probably because the relocation logic still has bugs or maybe there are issues with the tests helpers. The best way to debug it is to start with the `single_split` test. Pick a seed that fails and start debugging. When it seems like it's consistently passing, more on to other tests (for example `multi_split`).

The way relocation testing works is that there is now a new user `Event` which is raised when a node decides to relocate someone. In the tests, we then listen to these events and when we collect at least quorum of them for a single node, we assume that node will be relocated and we poll until that happens. It's possible this approach has flaws. 

With the new relocation logic, it's no longer possible to tell in advance how many churn events we need to trigger to cause relocation. So the test work by looping until the relocation happens. This means they can now take arbitrarily long to execute. This should be addressed eventually (I have some ideas), but more important is to make the tests pass first.

There are some tests that are currently commented out. Once the failing tests are resolved, we should uncomment and fix those as well.


maqi:
Merge as it is, as we need to carry out further issue resolving work together with the async API as well.

PR #2188, opened 2020-09-01
fixing parsec removal refactoring

maqi:
This PR contains some fixing commits trying to make the parsec removal work passing with the CI tests.

### DIFF
--- a/src/node/tests/adult.rs
+++ b/src/node/tests/adult.rs
@@ -55,6 +55,7 @@ impl Env {
         let (subject, ..) = Node::approved(
             NodeConfig {
                 network_params: NETWORK_PARAMS,
+                rng: rng::new_from(&mut rng),
                 ..Default::default()
             },
             shared_state,
--- a/src/node/tests/elder.rs
+++ b/src/node/tests/elder.rs
@@ -345,7 +345,7 @@ impl Env {
             .expect("subject is not approved");
 
         let proof_start_index = state.sections.knowledge_by_location(&dst);
-        let proof_chain = state.our_history.slice(..proof_start_index);
+        let proof_chain = state.our_history.slice(proof_start_index..);
 
         let dst_key = *state.section_key_by_location(&dst);
 


src/node/stage/approved.rs
```diff
@@ -1022,12 +1021,14 @@ impl Approved {
...
                     if index > dkg_key.1 {
                         trace!("DKG for {:?} already completed", dkg_key);
+                        return;
                     }
                 }
             }
         }
 
         for message in self.dkg_voter.init_dkg_gen(&core.full_id, dkg_key) {
+            self.churn_in_progress = true;
```
madadam, line 1044: Should we also set `self.chun_in_progress = false` here, instead of in `promote_and_demote_elders` if `init_dkg_gen` returns empty vec?

src/section/section_keys.rs
```diff
@@ -76,7 +76,7 @@ impl SectionKeysProvider {
         &mut self,
         section_key_index: u64,
         our_name: &XorName,
-        elders_info: &EldersInfo,
+        _elders_info: &EldersInfo,
```
madadam, line 79: Why not remove this parameter?

src/section/section_keys.rs
```diff
@@ -102,8 +102,10 @@ impl SectionKeysProvider {
         self.current = dkg_result
             .secret_key_share
             .and_then(|secret_key_share| {
-                elders_info
-                    .position(our_name)
+                dkg_key
+                    .0
+                    .iter()
+                    .position(|id| id.name() == our_name)
```
madadam, line 108: This won't always work, because the actual participants might be different from the expected participants (in case of some non-cooperating nodes) and this will then return incorrect index. I have a fix for this in the other PR: https://github.com/maidsafe/routing/pull/2185/commits/d5cf05db1ede665ef3134ecfb7675ab54f8bba5c

src/node/stage/approved.rs
```diff
@@ -1124,7 +1124,13 @@ impl Approved {
...
             .compute_relocations(churn_name, churn_signature);
 
-        for (info, action) in relocations {
+        // During startup, members are all at age 4. When a churn_signature has more than 4 trailing
+        // zeros (actually encountered 6 or more quite often), it means all members will be
+        // relocated at the same time, even the elders. So here changed to only relocate the first.
+        // TODO: ideally, it shall be : only one existing/potential elder and multiple adults (not
+        //       include the potential elder if already got one existing elder relocating) can be
+        //       relocated at same time.
+        if let Some((info, action)) = relocations.iter().next() {
```
madadam, line 1137: This is both insufficient and unnecessary and also wrong:

**Insufficient**:
Even if we limit the number of simultaneous relocations, what can happen is that some relocations are in progress while another churn event happens which triggers another relocation(s), effectively bypassing any limits.

**Unnecessary**:
- During startup, we don't do relocation, we increment the ages directly. Startup is until the section reaches `recommended_section_size` which is currently 60. So We would need a signature with at least 60 trailing zeroes in order to trigger relocation of all the nodes once the startup phase is over. This is extremely unlikely. 
- In tests the `recommended_section_size` is much lower, but it still shouldn't be a problem in practice. This is because of the "delayed relocation" feature. The way it works is that when we relocate an elder, that elder will keep participating in the section until it is demoted and only then will proceed with the relocation. This will happen even if the elder is actually voted offline. The reason this works is that the relocating elders have incentive to keep participating in the form of a `RelocatePromise` message which they can exchange for actual `Relocate` (which is what increases their age) once they are demoted. 
- Even if we end up relocating all the nodes, the elders will stay in the section until new elders are elected. If there are no nodes to pick the new elders from, the current elders will be blocked from leaving until new nodes join the section and can be promoted to elders

**Wrong**:
We can't limit the relocations to just one at a time, because then we would never be able to progress past the empty prefix. This is because relocations happen only on mature churn (non-infant) and while there is only one section, the only way to cause mature churn is if a mature node leaves the section. So with a limit of only one relocation per churn, that would mean every removed node would cause at most one relocation which means the number of mature nodes would at best stay constant but would actually decrease in practice, therefore never reaching the threshold necessary for split.

src/section/section_keys.rs
```diff
@@ -31,7 +35,7 @@ pub struct SectionKeysProvider {
     /// Our current section BLS keys.
     current: Option<SectionKeyShare>,
     /// The new keys to use when section update completes.
-    pending: BTreeMap<u64, DkgResult>,
+    pending: BTreeMap<DkgKey, DkgResult>,
```
madadam, line 38: Not quite sure why this change is necessary. In what case would we get multiple completed DKGs for the same key index but different participants?

src/node/stage/approved.rs
maqi: we cannot do so, as `init_dkg_gen` returns empty could mean there is on-going DKG process, which shall still have `churn_in_progress` as `true` ?

src/section/section_keys.rs
maqi: this is covered/overwriten by your new fixing commit.
it will be removed after rebase/cherry-pick

src/section/section_keys.rs
maqi: Yeah, I see your fix, will cherry-pick that into this PR to overwrite mine.

src/node/stage/approved.rs
maqi: > So We would need a signature with at least 60 trailing zeroes in order to trigger relocation of all the nodes once the startup phase is over. This is extremely unlikely.

I am not sure about this. As mentioned in the comment, I do observe the situation that a signature with just 6 trailing zeros cause ALL elders got relocated. maybe there is some calculation got mistaken? will double check this

> This is because of the "delayed relocation" feature.

Will the DKG process be affected? The situation here is that: DKG is looking for ALL `after-relocated` nodes as participants

> the elders will stay in the section until new elders are elected

You mean such elder routing node instance keeps two approved instance (i.e. two ids) ?

> if a mature node leaves the section

The current relocation within `prefix ()` happens as `relocate to self with different id`. so that mature node with old id does left the section ?

src/section/section_keys.rs
maqi: previously, we start DKG whenever promote_demote_elders returns with a new_elder_info , which means multiple DKG with different participants with same section_index.
However, this may no longer be the case as there is your more fixing commits.
I will double check this once cherry-picked your work into PR 2188.

src/node/stage/approved.rs
madadam: > I am not sure about this. As mentioned in the comment, I do observe the situation that a signature with just 6 trailing zeros cause ALL elders got relocated

I think this is because in tests, `recommended_section_size` is much lower. For example in the `single_split` test it's just 7.

madadam: > Will the DKG process be affected?

The existing elders, even when already voted offline, will still keep fulfilling their duties as elders. So they will try to select new nodes to promote and will trigger the DKG for them. Then the new nodes will run the DKG and send the results back to the existing elders who will then vote for the new `SectionInfo` and `OurKey` and only when that reaches consensus, will they actually leave, at which point the new nodes will become elders.

madadam: > You mean such elder routing node instance keeps two approved instance (i.e. two ids) ?

No, they still have their old name. Only after they are demoted will they proceed with the relocation and name change.

maqi: > I think this is because in tests, recommended_section_size is much lower. For example in the single_split test it's just 7

Then we shall resolve this by changing the testing setup or have some testing specific handling?

> So they will try to select new nodes to promote and will trigger the DKG for them. Then the new nodes will run the DKG. Only after they are demoted will they proceed with the relocation and name change.

But they are actuallly the same physical instance. so at the time that the DKG to be triggered, there is no new nodes to pick up such trigger and starts voting for it. :(

src/node/stage/approved.rs
```diff
@@ -1001,20 +1002,23 @@ impl Approved {
             // malicious elders within one side don't carry out vote, triggering the sibling
             // section's DKG process by send them an Initial DKG message.
             if !dkg_key.0.contains(core.id()) {
-                trace!(
-                    "Triggering DKG process for sibling section during splitting {:?}",
-                    dkg_key.0
-                );
+                // Avoiding the case that a staled elder trigger DKG votes by mistake.
+                if num_of_new_infos > 1 {
```
madadam, line 1011: Not sure about this. Sometimes it can happen that the new elders are completely different from the old ones even outside of split (for example, due to relocation) and we still need to notify them.

src/node/stage/approved.rs
```diff
@@ -292,11 +292,17 @@ impl Approved {
...
         for (dkg_key, message) in self.dkg_voter.progress_dkg(&mut core.rng) {
             let _ = self.broadcast_dkg_message(core, &dkg_key, message);
         }
 
         self.check_dkg(core);
+
+        if prev_has_dkg_voter && !self.dkg_voter.has_live_dkg_voter() {
+            self.churn_in_progress = false;
+            self.promote_and_demote_elders(core);
+        }
```
madadam: What if the above `check_dkg` results in a successful DKG completion? Should we still do this anyway?

src/node/stage/approved.rs
madadam: Also, could this be the reason for the failure that happens when we relocate all elders?

maqi: >Sometimes it can happen that the new elders are completely different from the old ones even outside of split (for example, due to relocation) and we still need to notify them.
>could this be the reason for the failure that happens when we relocate all elders?

Yes, I think so. and even worse, the NotifyOldElder message won't be picked up, as at that time, the node is still using the old-name, not the new-name. Meanwhile the `EldersInfo` is already looking for the new names. :(

src/node/stage/approved.rs
maqi: yes. but the `promote_and_demote_elders` only triggers a new DKG when there is new_elders_info (i.e. some node change happened).

src/section/section_keys.rs
dirvine: This is also fixed by forcing name change on promote/demote. We really should do this to become crdt complient in any case.

src/node/stage/approved.rs
```diff
@@ -1561,6 +1561,16 @@ impl Approved {
...
 
+            // For the case that adult got promoted into elder via update, EldersInfo needs to be
+            // inserted into dkg_voter manually. So that the voted dkg_result can be fetched and
+            // update section_key_provider properly.
+            let info = self.shared_state.our_info().clone();
+            let participants: BTreeSet<_> = info.elder_ids().copied().collect();
+            let section_key_index = self.shared_state.our_history.last_key_index() - 1;
+            let dkg_key = (participants, section_key_index);
+            trace!("inserting dkg info {:?} {:?}", dkg_key, info);
+            self.dkg_voter.insert_info(&dkg_key, info);
```
madadam, line 1578: Could you explain why is this necessary? If I understand it correctly, it's for the case where a node receives `Sync` with the new section info before its DKG completes. But isn't that already covered by this: https://github.com/maidsafe/routing/blob/master/src/node/stage/approved.rs#L1353-L1360 ? Or is it something else?

maqi: This is mainly for the case that ALL elders got relocated, i.e. all new elders are previous adults.
In which, those adults doesn't have the new_elders_info to populate the dkg_voter, hence no one will vote for the OurKey and SectionInfo.

The code block you pointing to doesn't work in this case, as the shared_state updated via sync doesn't contain this section_key of all new_elders.

maqi: Ok. It seems due to the issue as address by [only call promote_demote_elders when elders_info updated](https://github.com/maidsafe/routing/pull/2188/commits/d2aa569b7ef28d1a5f3e770812d164ffd9565aeb) , there is un-necessary additional increase of section_index, which cause the shared_state doesn't updated properly.

I tried with revert this change [insert EldersInfo when being updated](https://github.com/maidsafe/routing/pull/2188/commits/4f7428b832f3cf0e1d613b859bc018cdb28c375f), which still passing all unit tests and most of the integration tests.

So, as long as there is no un-necessary extra votes, this change is not required.

I will update the PR with this commit removed.

src/node/stage/approved.rs
```diff
@@ -1539,7 +1541,10 @@ impl Approved {
             self.churn_in_progress = false;
             core.msg_filter.reset();
 
-            if new_is_elder {
+            // During the handling of OurKey, there is only section_key_index got updated, meanwhile
+            // the edlers info doesn't got updated. Calling `promote_and_demote_elders` will make
+            // un-necessary DKG process of same participants but different section_key_index.
+            if new_is_elder && (old_elders_info != new_elders_info) {
```
madadam, line 1547: Normally this would be handled by `SectionUpdateBarrier` which makes sure we only update our shared state if the new elders info is in sync with the section chain. That means that the current elders info must be signed by the key previous to the last key. This is the case always except for the very first elders info, which is signed by the last key. This is because there is no previous key in that case. There is a special handling in `SectionUpdateBarrier` for this but it assumes that there is only one elders in that case. This assumption does not hold in some of the unit tests, because we create multiple elders but still have only one key in the chain.

I think a better fix for this would be one of these:
1. Either modify the unit tests so the initial section chain has more than one key
2. Or move this logic to `SectionUpdateBarrier::take`: https://github.com/maidsafe/routing/blob/master/src/section/section_update_barrier.rs#L100

The reason why I think it would be better is that the purpose of `SectionUpdateBarrier` is specifically to only allow complete updates to the shared state. Having additional check for that here can cause confusion for people reading the code.

maqi: > Normally this would be handled by SectionUpdateBarrier which makes sure we only update our shared state if the new elders info is in sync with the section chain

maqi: >Normally this would be handled by SectionUpdateBarrier which makes sure we only update our shared state if the new elders info is in sync with the section chain

You mean the check at https://github.com/maidsafe/routing/blob/master/src/node/stage/approved.rs#L1466-1468 ?

However, it seems during the handling of sync, there is no such check? https://github.com/maidsafe/routing/blob/master/src/node/stage/approved.rs#L593

madadam: That's right, good point. Maybe we should address that, but I don't think that's the issue here. If I'm understanding it correctly, the problem is we reach conssesus on `OurKey` while having only one key in the section chain yet. So the `SectionUpdateBarrier` checks and sees that the current elders info is signed with the previous key and considers the update as complete even though we are still expecting a consensus on `SectionInfo`.

maqi: > the problem is we reach conssesus on OurKey while having only one key in the section chain yet. So the SectionUpdateBarrier checks and sees that the current elders info is signed with the previous key and considers the update as complete even though we are still expecting a consensus on SectionInfo

Ok. I will check about this.

> Maybe we should address that

right, I will have a check to see whether addressing that will get something fixed

madadam: > right, I will have a check to see whether addressing that will get something fixed

I doubt it. Non-malicious nodes should never send a `Sync` with incomplete `SharedState`. So maybe all that's necessary is reject `Sync`s if the contained shared state is not complete (that is, it's elders info is not signed with the previous key).

src/node/tests/elder.rs
```diff
@@ -219,7 +219,7 @@ impl Env {
         new_info: &DkgToSectionInfo,
     ) -> Result<()> {
         self.cast_unordered_votes(
-            self.quorum_count() - 1,
+            quorum_count(new_info.new_elders_info.elders.len()),
```
madadam, line 222: Don't think this is correct. This function is supposed to aggregate enough votes that if **this node votes** too, the vote  accumulates. But with this change it's going to accumulate regardless of this node voting.

src/node/tests/adult.rs
```diff
@@ -55,6 +55,7 @@ impl Env {
         let (subject, ..) = Node::approved(
             NodeConfig {
                 network_params: NETWORK_PARAMS,
+                rng: rng::new_from(&mut rng),
```
madadam, line 58: Just curious: why was this needed?

maqi: previously the adult's rng was using default rng, which is generated from local thread, which means it has the same seed as to the main rng within the testing code generating elder ids. 
This results in the adult's id is duplicated to one of the generated elder id, which fails the `is_elder` check, i.e. the adult think itself as an elder.

src/node/tests/elder.rs
```diff
@@ -214,19 +214,56 @@ impl Env {
...
         &mut self,
         new_info: &DkgToSectionInfo,
     ) -> Result<()> {
+        let our_key_vote = Vote::OurKey {
+            prefix: new_info.new_elders_info.prefix,
+            key: new_info.new_pk_set.public_key(),
+        };
+        let section_info_vote = Vote::SectionInfo(new_info.new_elders_info.clone());
+
+        self.self_cast_unordered_votes(vec![our_key_vote.clone(), section_info_vote.clone()])?;
```
madadam, line 262: This is still wrong. This function is not supposed to vote for the `subject` node. The idea is that something else has already triggered that vote and here we are only voting by the other nodes so together those votes accumulate. So if the tests fail without this it means that `subject` did not cast the vote it was supposed to. My guess is that the `simulate_dkg` function is probably not working as expected. Probably because it does not actually simulate receiving enough `DKGResult` messages from the other nodes.

maqi: Hmm...  right. 
How about I just remove this fixing commit and merge PR 2188 only with the other two.
And trying to resolve this in a separate PR ?

PR #2192, opened 2020-09-17
Fix some bugs around DKG and relocation

madadam:
This PR refactors and modifies the DKG module with the aim of fixing bugs in it and making it simpler. The new flow goes somehow like this:

1. On every churn, the current elders propose the new elders as the `elder_size` oldest nodes of the section.
2. If the proposed elders are the same as current elders, we are done
3. Otherwise they send `DKGStart` message (signed with their BLS key share) to the proposed elders (also called DKG paricipants)
4.When the `DKGStart` message accumulates, the participant initiates a DKG session and starts exchanging DKG messages with the other participants
5. When the DKG session completes successfully or fails, the participants send `DKGResult` message carrying the result to the current elders
6. The current elders accumulate this message and when they get quorum of results, they action it:
7. If the result is a success, they cast `SectionInfo` and `OurKey` votes to update the section
8. If the result is a failure, they restart the DKG by repeating step 3. 

Unsolved issues:

The above mechanism can sometimes lead to more than one successful DKG outcome (usually during heavy churn). There is currently a simple measure in place which blocks voting for section update if another one is in progress. but it is not completely sufficient because the DKG results can arrive in any order and it can cause section to stall. This is quite rare but still happens and needs to be solved.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,3 +52,6 @@ mock = [
     "env_logger",
     "bls_signature_aggregator/mock_timer"
 ]
+
+[patch.crates-io]
+bls_signature_aggregator = { git = "https://github.com/madadam/bls_signature_aggregator.git", branch = "repetition" }
--- a/examples/minimal.rs
+++ b/examples/minimal.rs
@@ -353,12 +353,7 @@ fn handle_event(
             dst,
             HexFmt(content)
         ),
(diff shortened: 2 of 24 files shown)


dirvine (approved):
Merging this, not passing all tests but clears up a lot of code. We will investigate CRDT types in dkg to allow us to strongly order the dkg rounds in flight.

PR #2195, opened 2020-09-28
WIP fix/clippy: resolve clippy errors after using async

maqi:


src/lib.rs
```diff
@@ -152,9 +156,6 @@ const RECOMMENDED_SECTION_SIZE: usize = 60;
 const ELDER_SIZE: usize = 7;
 
 // Quic-p2p
-#[cfg(feature = "mock")]
-use mock_qp2p as qp2p;
-#[cfg(not(feature = "mock"))]
 use qp2p::{self};
```
bochaco, line 159: I think you don't need `{self}` anymore

PR #2196, opened 2020-09-29
Message resend and lost peer detection

madadam:


src/node/stage/comm.rs
```diff
@@ -6,136 +6,447 @@
...
+            .recipients
+            .iter()
+            .position(|recipient| recipient.addr == *addr)
+        {
+            let _ = self.recipients.swap_remove(index);
+            self.remaining -= 1;
+        }
+    }
+
+    // Consumes the state and returns the list of recipients that failed all attempts (if any).
```
maqi, line 289: state is not consumed?

src/node/stage/comm.rs
```diff
@@ -6,136 +6,447 @@
...
+    fn finish(self) -> Vec<SocketAddr> {
+        self.recipients
+            .into_iter()
+            .filter(|recipient| !recipient.sending && recipient.attempt >= RESEND_MAX_ATTEMPTS)
```
maqi, line 293: if a recipient is still sending, it will not be counted as `failed` ?
but after `finish` being called, the state is supposed to be consumed and that `recipient`'s result won't be handled?

src/node/stage/comm.rs
```diff
@@ -6,136 +6,447 @@
...
+
+    impl Peer {
+        fn new() -> Result<Self> {
+            let transport = QuicP2p::with_config(Some(transport_config()), &[], false)?;
+
+            let endpoint = transport.new_endpoint()?;
+            let addr = endpoint.local_addr()?;
+            let mut incoming_connections = endpoint.listen()?;
+
+            let (tx, rx) = mpsc::channel(1);
```
maqi, line 422: make this `1` as a const? 
maybe it shall be the same as the above idle_timeout_msec ?

tests/drop.rs
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }
+
+        Err(format_err!(
+            "Event::MemberLeft not received for {}",
+            dropped_name
+        ))
+    };
+
+    time::timeout(TIMEOUT, expect_event).await?
```
maqi, line 54: so this test will last TIMEOUT seconds (5) ?

src/node/executor.rs
```diff
@@ -0,0 +1,208 @@
...
+    });
+}
+
+// A single consumer, multiple producer one-shot channel that sends when the sender gets dropped.
+// Used to observe termination of some object from any number of tasks simultaneously.
+//
+// Note: it seems we could have used `tokio::sync::watch` for this exact purpose. The reason why we
+// didn't is that `watch` interacts poorly with the `select!` macro. It requires the future
+// returned from `recv` to be pinned which leads to convoluted code.
+mod terminate {
```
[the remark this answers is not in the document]
madadam: > why not have this as a separate file?

It's quite short and used only in this file, but I don't mind putting it in a separate file either.

> It seems this is the struct in responsible of detecting the termination of spawned tasks, right ?

It's the other way around. It notifies the spawned tasks about the termination of the `Executor` object.

> maybe termination_handler or just observer ?

Maybe something like `termination_notifier` or `drop_notifier`?

src/node/stage/bootstrapping.rs
```diff
@@ -153,30 +154,38 @@ impl Bootstrapping {
...
-        }
+        debug!("Sending BootstrapRequest to {}", dst);
+        self.comm
+            .send_message_to_target(&dst, message.to_bytes())
+            .await?;
 
         Ok(())
     }
 
+    async fn reconnect_to_new_section(&self, new_conn_infos: Vec<SocketAddr>) -> Result<()> {
```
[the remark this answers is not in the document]
madadam: It was called like this before. Don't mind renaming it though.

src/node/stage/approved.rs
```diff
@@ -586,7 +544,7 @@ impl Approved {
         let bounce_msg = bounce_msg.to_bytes();
 
         if let Some(sender) = sender {
-            self.comm.send_message_to_target(&sender, bounce_msg).await
```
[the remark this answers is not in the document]
madadam: > any special handling required?

Yes, it needs to vote `Offline` for lost peers, because they are now detected inside `Comm::send_message_to_targets`.

src/node/stage/comm.rs
madadam: It means it takes `self` by value.

src/node/stage/comm.rs
madadam: Yeah, you are not really supposed to call `finish` while there are still sends in progress, but the api doesn't prevent that either.

src/node/stage/comm.rs
madadam: It's the channel capacity. It means it can buffer at most one item. Anything more, or even using unbounded channel here would be overkill. It has nothing to do with the idle timeout. Might move it a const, but doesn't seem it would bring us that much.

tests/drop.rs
madadam: No, the test should take less. But if there is a bug and the event isn't raised, then it would run forever. So to prevent that, there is an arbitrary timeout now and it expiring most likely indicates a bug.

src/node/stage/comm.rs
madadam: Maybe we can count sending as failed too. Not sure. Doesn't really matter in practice, because we never call `finish` while there are still pending sends.

src/node/executor.rs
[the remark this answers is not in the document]
maqi: `termination_notifier` sounds more tech

src/node/stage/bootstrapping.rs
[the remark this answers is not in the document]
maqi: up to you then

src/node/stage/approved.rs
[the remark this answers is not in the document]
maqi: but seems the most functions are duplicated.

anyway, it can be addressed later on to see if can have a common `Comm` to be used by all.

src/node/stage/comm.rs
maqi: I thought `cosumes` means the entries of recipients will be cleaned-up. 
but seems the belowing code does do so?

src/node/stage/mod.rs
```diff
@@ -217,48 +227,57 @@ impl Stage {
         recipient: &SocketAddr,
         msg: Bytes,
     ) -> Result<()> {
-        self.comm.send_message_to_target(recipient, msg).await
+        self.comm.send_message_to_target(recipient, msg).await?;
+        Ok(())
```
bochaco, line 231: Is the above change for somehow more clarity and/or preference?

tests/utils/mod.rs
```diff
@@ -101,27 +102,84 @@ impl<'a> TestNodeBuilder {
...
-                stringify!($pattern),
-                other
-            )),
-            Err(_) => Err(anyhow::format_err!(
-                "Timeout when expecting {}",
-                stringify!($pattern)
-            )),
+        match tokio::time::timeout($crate::utils::TIMEOUT, $node.next()).await {
+            Ok(Some($pattern)) => {}
+            Ok(other) => panic!("Expecting {}, got {:?}", stringify!($pattern), other),
```
bochaco, line 114: Not sure if this was pre/post our chat about returning Result from tests, we should not panic but return an error, is there anything now not allowing you to return a Result?

src/node/stage/mod.rs
madadam: No, it would not compile without it. `Comm::send_message_to_target` returns different error type. The `?` operator automatically coverts it to `Error`.

tests/utils/mod.rs
madadam: This is maybe up to a philosophical debate, but the way I see it, this macro is essentially a form of assert and asserts do panic. Maybe we can rename it to `assert_next_event` to make it cleaner?

bochaco: ah right i see, I'd vote for renaming it if that's enough for the philosophical debate :smile: , I was seeing it simply as a helper to avoid code duplication

PR #2197, opened 2020-09-30
Remove message accumulator

madadam:


src/consensus/vote.rs
```diff
@@ -54,6 +55,12 @@ pub enum Vote {
 
     // Voted to change the age of the given node.
     ChangeAge(MemberInfo),
+
+    // Voted to send user message on behalf of the whole section.
```
maqi, line 59: bit comfused with `on behalf of the whole section`
thought the on concensus of the Vote, elders will send out the contained message, then that is already `whole` section doing so?

maybe just remove `on behalf of ...` to make the comment same pattern as others ?

src/node/stage/approved.rs
```diff
@@ -1316,6 +1294,24 @@ impl Approved {
...
+        message: PlainMessage,
+        proof_chain: SectionProofChain,
+        proof: Proof,
+    ) -> Result<()> {
+        let message = Message::section_src(
+            message.src,
+            proof.signature,
+            message.dst,
+            message.variant,
+            proof_chain,
```
maqi, line 1308: just to confirm, that when we carry out aggregator on this vote, is there any check to be carried out to verify the proofshare is from the proof_chain ?
it seems the `Message::section_src` doesn't carry out any check whether the proof is from the proof_chain ?

also, wondering shall we also carry out a check that the `proof_chain` is a valid chain? i.e. at least contains the knowledge that we have of the srouce section?

bochaco (approved):

src/consensus/vote.rs
madadam: Yeah, you are right. It's badly worded. Will change it.

PR #2199, opened 2020-10-01
Command paradigm

madadam:


src/node/executor.rs
```diff
@@ -100,53 +91,26 @@ fn spawn_messages_handler(
...
-                        msg,
-                        err
-                    );
-                }
+            Ok(message) => {
+                let command = Command::HandleMessage {
+                    message,
+                    sender: Some(sender),
+                };
+                let _ = stage.handle_commands(command).await;
```
maqi, line 109: it's actually `handle_command` ?

src/node/mod.rs
```diff
@@ -6,26 +6,40 @@
...
 pub use self::event_stream::EventStream;
-use self::{executor::Executor, stage::Stage};
+#[cfg(test)]
+use self::stage::State;
+use self::{
+    command::Command,
+    executor::Executor,
+    stage::{Approved, Bootstrapping, Comm, NodeInfo, Stage},
+};
+#[cfg(all(test, feature = "mock"))]
```
maqi, line 24: feature `mock` ?

src/node/mod.rs
```diff
@@ -6,26 +6,40 @@
...
     network_params::NetworkParams,
     peer::Peer,
-    rng::MainRng,
+    rng,
     section::{EldersInfo, SectionProofChain},
     TransportConfig,
 };
 use bytes::Bytes;
 use itertools::Itertools;
+#[cfg(feature = "mock")]
```
maqi, line 39: is the plan to bring back the `mock` feature?
However I don't see it be added in the Cargo.toml ?

src/node/tests/elder.rs
```diff
@@ -456,67 +447,6 @@ fn add_and_promote_member() {
     assert!(env.is_candidate_elder());
 }
 
-#[test]
-fn remove_member() {
```
maqi, line 460: why these tests are removed?

src/section/shared_state.rs
```diff
@@ -53,6 +58,39 @@ impl SharedState {
...
+        // Note: `ElderInfo` is normally signed with the previous key, but as we are the first node
+        // of the network there is no previous key. Sign with the current key instead.
+        let elders_info = create_first_elders_info(&public_key_set, &secret_key_share, peer)?;
+
+        let mut state = SharedState::new(
+            SectionProofChain::new(elders_info.proof.public_key),
+            elders_info,
+        );
+
+        for p2p_node in state.sections.our().elders.values() {
```
maqi, line 77: as being the first node in the network, then here shall just be self ?

src/node/stage/mod.rs
```diff
@@ -11,337 +11,420 @@ mod bootstrapping;
...
+            State::Joining(state) => state.handle_message(sender, msg),
+            State::Approved(state) => state.handle_message(sender, msg),
+        }
+    }
 
-        Ok(in_dst)
+    async fn handle_timeout(&self, token: u64) -> Result<Vec<Command>> {
+        match &mut *self.state.lock().await {
+            State::Approved(state) => state.handle_timeout(token),
+            _ => Ok(vec![]),
```
maqi, line 356: other stages won't timeout?

src/node/stage/approved.rs
```diff
@@ -51,99 +49,183 @@ pub(crate) struct Approved {
...
         }
 
         match self.decide_message_status(&msg)? {
             MessageStatus::Useful => {
-                trace!("Useful message from {}: {:?}", sender, msg);
-                self.update_section_knowledge(&msg).await?;
-                self.handle_useful_message(Some(sender), msg).await
+                trace!("Useful message from {:?}: {:?}", sender, msg);
+                commands.extend(self.update_section_knowledge(&msg)?);
+                commands.extend(self.handle_useful_message(sender, msg)?);
```
maqi, line 115: any diff to use `extend` here and `push` below?

src/node/executor.rs
madadam: No, `handle_command` (singular) handles only one command and returns a `Vec` of further commands to handle. `handle_commands` (plural) also handles those returned commands (and those returned from them, etc...).

src/node/mod.rs
madadam: Good catch, will remove.

src/node/mod.rs
madadam: No, this was left here by accident. Will remove.

src/node/tests/elder.rs
madadam: They are being converted to the new style. Currently the new tests are all in `node/tests/mod.rs` but we might want to split them up into multiple modules when there is too many of them.

src/node/stage/mod.rs
madadam: That's right. The bootstraping/joining timeout is better handled by the upper layers.

src/node/stage/approved.rs
madadam: Because some function return `Vec<Command>`, some `Option<Command>` and some just `Command`.

src/node/stage/approved.rs
```diff
@@ -51,99 +49,183 @@ pub(crate) struct Approved {
...
+                Ok(vec![])
+            }
+            Vote::ChangeAge(member_info) => {
+                self.handle_change_age_event(member_info, proof);
+                Ok(vec![])
+            }
+            Vote::SendMessage {
+                message,
+                proof_chain,
+            } => Ok(vec![self.handle_send_message_event(
```
maqi, line 186: why not to have the same return type as others ?

PR #2200, opened 2020-10-05
BREAKING CHANGE: remove id.rs file

maqi:


src/crypto.rs
```diff
@@ -30,77 +31,25 @@ pub fn sha3_256(input: &[u8]) -> Digest256 {
...
-    fn hash<H: Hasher>(&self, state: &mut H) {
-        self.0.to_bytes().hash(state);
-    }
-}
-
-impl Ord for Signature {
-    fn cmp(&self, other: &Self) -> Ordering {
-        self.0.to_bytes().cmp(&other.0.to_bytes())
-    }
+pub fn name_from_key(public_key: &PublicKey) -> XorName {
```
dirvine, line 34: Should this not be just `name` (we can see the param is key)

src/error.rs
```diff
@@ -19,6 +19,8 @@ pub enum Error {
     BadLocation,
     #[error(display = "Failed signature check.")]
     FailedSignature,
+    #[error(display = "Cannot covert a public name into public_key.")]
```
dirvine, line 22: Perhaps "Invalid PublicKey bytes," I think this can have a source error from dalek. so s `#[source]` attribute. That will impl a From trait for us

src/messages/mod.rs
```diff
@@ -122,7 +127,7 @@ impl Message {
 
     /// Creates a signed message from single node.
     pub(crate) fn single_src(
```
dirvine, line 129: I still see this confusing, there can be only one node with a keypair and that's us. All others are Peers.

src/messages/src_authority.rs
```diff
@@ -26,7 +26,7 @@ pub enum SrcAuthority {
     /// Authority of a single node.
```
dirvine, line 26: This is a Peer as we are the Node and the only Node.

src/crypto.rs
maqi: just want to make the calling to this function more explicitly.

src/messages/mod.rs
maqi: this passed_in keypair is just used for signing. not being kept and sent_out

src/messages/src_authority.rs
maqi: I think the word `Node` was used only mean to refer to `Single`

How about change this to `Single` ?

I am just concerned that change to `Peer` could cause the same confusion when thinking from the `sender`'s perspective?

src/peer.rs
```diff
@@ -0,0 +1,52 @@
...
+        &self.name
+    }
+
+    /// Returns the `SocketAddr`.
+    pub fn addr(&self) -> &SocketAddr {
+        &self.addr
+    }
+}
+
+impl PartialEq for Peer {
```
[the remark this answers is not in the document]
maqi: `PartialEq` cannot be derived when `Hash` is implemented.
And `Hash` is required as other structs relies on Peer to impl such trait.

src/crypto.rs
```diff
@@ -30,77 +31,25 @@ pub fn sha3_256(input: &[u8]) -> Digest256 {
...
+    XorName(public_key.to_bytes())
 }
 
-impl PartialOrd for Signature {
-    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        Some(self.cmp(other))
-    }
+pub fn pub_key_from_name(name: &XorName) -> Result<PublicKey, SignatureError> {
```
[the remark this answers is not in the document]
maqi: these two functions have already been renamed to `name` and `pub_key` within yesterday's update, as suggested by David previously.
maybe your github page doesn't got refreshed?

btw. do you think `key_from_name` and `name_from_key` are better ?

src/peer.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    /// Returns the `XorName` of the peer.
+    pub fn name(&self) -> &XorName {
+        &self.name
+    }
+
+    /// Returns the `SocketAddr`.
+    pub fn addr(&self) -> &SocketAddr {
+        &self.addr
+    }
```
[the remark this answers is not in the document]
maqi: yes, mainly just to using the previous impl to avoid changing too many places to use the `reference`.
as many places we only need a reference, but by using `pub`, it means we have to put the ref manually.

src/crypto.rs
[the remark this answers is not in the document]
madadam: > btw. do you think key_from_name and name_from_key are better than name and pub_key ?

I think so, but not too worried about it. Your call.

src/peer.rs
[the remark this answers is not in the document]
madadam: Yeah, but why not derive `Hash` too?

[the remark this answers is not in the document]
maqi: Ok then, will derive `Hash` then.

madadam (approved):

PR #2202, opened 2020-10-07
Some structs refactor work

maqi:
This PR contains the work of:
1, rename SectionMembers to SectionPeers, MemberState to PeerState
2, rename quorum to majority
3, make Peer contains the age info

### DIFF
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -68,6 +68,7 @@ impl Message {
             SrcAuthority::Node {
                 public_key,
                 signature,
+                ..
             } => {
                 if public_key.verify(&signed_bytes, signature).is_err() {
                     error!("Failed signature: {:?}", msg);
@@ -123,6 +124,7 @@ impl Message {
     /// Creates a signed message from single node.
     pub(crate) fn single_src(
         keypair: &Keypair,
+        age: u8,
         dst: DstLocation,
         variant: Variant,
         proof_chain: Option<SectionProofChain>,
@@ -136,6 +138,7 @@ impl Message {
         let signature = crypto::sign(&serialized, keypair);
         let src = SrcAuthority::Node {
             public_key: keypair.public,
+            age,
             signature,
         };
 
@@ -179,6 +182,7 @@ impl Message {
             SrcAuthority::Node {
                 public_key,
                 signature,
+                ..
             } => {
                 if public_key.verify(&bytes, signature).is_err() {
                     return Err(Error::FailedSignature);
@@ -357,7 +361,7 @@ pub(crate) struct SignableView<'a> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{consensus, crypto::Keypair, rng, section};
+    use crate::{consensus, crypto::Keypair, rng, section, MIN_AGE};
     use std::iter;
 
     #[test]
@@ -382,6 +386,7 @@ mod tests {
         let variant = Variant::NodeApproval(elders_info);
         let message = Message::single_src(
             &keypair,
+            MIN_AGE,
             DstLocation::Direct,
             variant,
             Some(full_proof_chain.slice(1..)),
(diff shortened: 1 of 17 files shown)


src/messages/mod.rs
```diff
@@ -123,6 +124,7 @@ impl Message {
     /// Creates a signed message from single node.
     pub(crate) fn single_src(
         keypair: &Keypair,
+        age: u8,
```
dirvine, line 127: Should this not take &Node instead of KeyPair + age?

src/messages/mod.rs
```diff
@@ -136,6 +138,7 @@ impl Message {
         let signature = crypto::sign(&serialized, keypair);
         let src = SrcAuthority::Node {
             public_key: keypair.public,
+            age,
```
dirvine, line 141: same here take &Peer instead of key + age ?

maqi: Peer contains `socket addr` as well, which no need to be in `SrcAuthority::Node` ?

src/messages/mod.rs
maqi: keypair is from self, and `age` is fetched from SectionPeers info.
I thought Node won't contain the info of age ?

src/node/stage/bootstrapping.rs
```diff
@@ -162,6 +162,7 @@ impl Bootstrapping {
 
         let message = Message::single_src(
             &self.node_info.keypair,
+            MIN_AGE,
```
madadam, line 165: When we are relocating, this won't be `MIN_AGE`. You can fetch the correct age from `self.relocate_details`.

src/node/stage/joining.rs
```diff
@@ -235,6 +235,7 @@ impl Joining {
         let variant = Variant::JoinRequest(Box::new(join_request));
         let message = Message::single_src(
             &self.node_info.keypair,
+            MIN_AGE,
```
madadam, line 238: Same here

src/peer.rs
```diff
@@ -15,14 +15,15 @@ use xor_name::XorName;
...
+    name: XorName,
+    addr: SocketAddr,
+    pub age: u8,
 }
 
 impl Peer {
-    /// Creates a new `Peer` given a `Name` and a `ConnectionInfo`.
-    pub fn new(name: XorName, addr: SocketAddr) -> Self {
-        Self { name, addr }
+    /// Creates a new `Peer` given `Name`, `ConnectionInfo` and `age`.
```
madadam, line 24: ConnectionInfo -> SocketAddr

src/node/stage/bootstrapping.rs
```diff
@@ -160,9 +160,15 @@ impl Bootstrapping {
             None => self.node_info.name(),
         };
 
+        let age = if let Some(signed_relocation_details) = &self.relocate_details {
+            signed_relocation_details.relocate_details().age
+        } else {
+            MIN_AGE
+        };
```
madadam, line 167: You could have done it together with the previous match:
```rust
let (destination, age) = match &self.relocate_details {
    Some(details) => (*details.destination(), details.relocate_details().age),
    None => (self.node_info.name(), MIN_AGE)
};
```

src/peer.rs
```diff
@@ -36,6 +36,16 @@ impl Peer {
...
 
+    /// Returns the age.
+    pub fn age(&self) -> u8 {
+        self.age
+    }
+
+    // Converts this info into one with the input age.
+    pub fn into_age(self, age: u8) -> Self {
+        Self { age, ..self }
+    }
```
madadam, line 47: The `into_*` convention usually signifies that you are converting one type into another one. Here you are only changing the age which makes it a confusing name.  How about `change_age` or `with_age` ?

dirvine (approved):

madadam (approved):

PR #2203, opened 2020-10-08
More tests

madadam:


Cargo.toml
```diff
@@ -15,7 +15,7 @@ async-recursion = "0.3.1"
 bincode = "1.2.1"
 bls = { package = "threshold_crypto", version = "~0.4.0" }
 bls_dkg = "~0.2.0"
-bls_signature_aggregator = { git = "https://github.com/madadam/bls_signature_aggregator.git", branch = "repetition" }
+bls_signature_aggregator = { git = "https://github.com/maidsafe/bls_signature_aggregator.git" }
```
maqi, line 18: why not use the version number?

src/node/tests/mod.rs
```diff
@@ -614,6 +915,234 @@ async fn handle_bounced_unknown_message() -> Result<()> {
...
+        Ok(Event::EldersChanged { key, elders, .. }) => {
+            assert_eq!(key, pk2);
+            assert_eq!(elders, new_elders);
+        }
+    );
+
+    Ok(())
+}
+
+// TODO: add test that untrusted `Sync` is not applied
```
maqi, line 1091: better do this in the PR if possible?

Cargo.toml
madadam: Wasn't sure it's published. Will check and change.

src/node/tests/mod.rs
madadam: There is still tons of tests we still need to write. We'll do them in the future PRs. This one will be one of them.

maqi (approved):

PR #2207, opened 2020-10-15
Refactor: remove SharedState, introduce Node, Section and Network modules

### LINKED ISSUE #2204: Raise Event::Connected after transitioning to Approved
Currently we raise the event before the transition to `Approved` is complete which might cause surprising behavior for the users or routing where they already receive the `Connected` event yet some function (like `our_prefix`) still return error. 

To fix it, make sure the event is raised only after the transition is complete.

madadam:
Also closes #2204


src/messages/src_authority.rs
```diff
@@ -62,12 +62,22 @@ impl SrcAuthority {
...
+
+    pub(crate) fn to_node_peer(&self, addr: Option<SocketAddr>) -> Result<Peer> {
+        match self {
+            Self::Section { .. } => Err(Error::BadLocation),
             Self::Node {
                 public_key, age, ..
-            } => Ok((name(public_key), *age)),
-            Self::Section { .. } => Err(Error::BadLocation),
+            } => {
+                let addr = addr.ok_or(Error::InvalidSource)?;
```
maqi, line 78: seems the input addr has to be a solid value?
why not just having the passed in parameter as a non-option, so that the error case can be handled immediately at the calling place ?

src/messages/variant.rs
```diff
@@ -41,7 +42,7 @@ pub(crate) enum Variant {
     /// section.
     NodeApproval(Proven<EldersInfo>),
     /// Message sent to all members to update them about the state of our section.
```
maqi, line 44: better a short comment regarding what the `network` part is about ?

src/messages/variant.rs
```diff
@@ -147,10 +148,10 @@ impl Debug for Variant {
                 .finish(),
             Self::UserMessage(payload) => write!(f, "UserMessage({:10})", HexFmt(payload)),
             Self::NodeApproval(payload) => write!(f, "NodeApproval({:?})", payload),
-            Self::Sync(shared_state) => f
+            Self::Sync { section, .. } => f
                 .debug_struct("Sync")
-                .field("elders_info", shared_state.sections.our())
-                .field("section_key", shared_state.our_history.last_key())
+                .field("elders_info", section.elders_info())
+                .field("section_key", section.chain().last_key())
```
maqi, line 154: shall the network (or some info of it) be logged as well?

src/relocation.rs
```diff
@@ -9,53 +9,35 @@
...
-
-// TODO: move this to the xor-name crate as `BitXor` impl.
-fn xor(lhs: &XorName, rhs: &XorName) -> XorName {
-    let mut output = XorName::default();
-    for (o, (l, r)) in output.0.iter_mut().zip(lhs.0.iter().zip(rhs.0.iter())) {
-        *o = l ^ r;
-    }
-
-    output
+/// Find all nodes to relocate after a churn even and create the relocate actions for them.
```
maqi, line 23: churn event

src/routing/mod.rs
```diff
@@ -103,7 +110,9 @@ impl Node {
         } else {
             info!("{} Bootstrapping a new node.", node_name);
             let (comm, bootstrap_addr) = Comm::from_bootstrapping(config.transport_config).await?;
-            let (state, command) = Bootstrapping::new(None, vec![bootstrap_addr], node_info)?;
+            let addr = comm.our_connection_info()?;
+            let node = Node::new(keypair, addr, MIN_AGE, config.network_params, event_tx);
+            let (state, command) = Bootstrapping::new(None, vec![bootstrap_addr], node)?;
```
maqi, line 115: better rename the following `node = Self { }` to `routing = Self { }` ?

src/routing/stage/approved.rs
```diff
@@ -1241,26 +1265,33 @@ impl Approved {
...
                 key,
             );
-            self.try_update_our_section()
+            self.try_update_state()
+        } else if key.value.0 != *self.section.prefix() {
+            let _ = self.network.update_their_key(key);
+            Ok(vec![])
         } else {
-            let _ = self.shared_state.update_their_key(key);
+            // Ignore our key. Should be updated using `OurKey` instead.
```
maqi, line 1280: shall be an error case ?

src/routing/stage/approved.rs
```diff
@@ -1284,14 +1315,14 @@ impl Approved {
     ) -> Result<Vec<Command>> {
         let mut commands = vec![];
 
-        if self.shared_state.our_history.has_key(&public_key) {
+        if self.section.chain().has_key(&public_key) {
             // Our shared state is already up to date, so no need to vote. Just finalize the DKG so
```
maqi, line 1319: `shared state` -> `section state` ?

src/routing/stage/approved.rs
```diff
@@ -1842,70 +1887,69 @@ impl Approved {
...
+                commands.extend(self.vote(Vote::TheirKey {
+                    prefix: *src_prefix,
+                    key: *src_key,
+                })?);
+            }
+        }
+
+        if let Some(dst_key) = msg.dst_key() {
+            let old = self.network.knowledge_by_section(src_prefix);
+            let new = self.section.chain().index_of(dst_key).unwrap_or(0);
```
maqi, line 1910: I am bit confused here.
This `new` is the src knowledge of us, and the `old` is our knowledge of src
but following we are comparing these two separated knowledge ?

src/routing/tests/mod.rs
```diff
@@ -1056,27 +1057,30 @@ async fn handle_sync() -> Result<()> {
     );
     let new_elders: BTreeSet<_> = new_elders_info.elders.keys().copied().collect();
     let proven_new_elders_info = create_proven(sk1_set.secret_key(), new_elders_info)?;
-    let new_shared_state = SharedState::new(chain, proven_new_elders_info);
+    let new_section = Section::new(chain, proven_new_elders_info);
 
     // Create the `Sync` message containing the new shared state.
```
maqi, line 1062: contain the new section state ?

src/section/mod.rs
```diff
@@ -8,27 +8,395 @@
...
+    /// # Panics
+    ///
+    /// Panics if the key used to sign `elders_info` is not present in `section_chain`.
+    pub fn new(chain: SectionProofChain, elders_info: Proven<EldersInfo>) -> Self {
+        assert!(chain.has_key(&elders_info.proof.public_key));
+
+        Self {
+            elders_info,
+            chain,
+            members: SectionPeers::default(),
```
maqi, line 62: would like to confirm that shall the `members` contains the `elders` ?

src/section/mod.rs
```diff
@@ -8,27 +8,395 @@
...
+    /// Update the `EldersInfo` of our section.
+    pub fn update_elders(&mut self, new_elders_info: Proven<EldersInfo>) -> bool {
+        if !new_elders_info.verify(&self.chain) {
+            return false;
+        }
+
+        if new_elders_info != self.elders_info {
+            self.elders_info = new_elders_info;
+            self.members
+                .prune_not_matching(&self.elders_info.value.prefix);
```
maqi, line 146: only `prune` ? no need to update?

src/section/mod.rs
```diff
@@ -8,27 +8,395 @@
...
+            warn!("ignore attempt to reduce the number of elders too much");
+            vec![]
+        } else {
+            let new_info = EldersInfo::new(expected_elders_map, self.elders_info().prefix);
+            vec![new_info]
+        }
+    }
+
+    /// Returns whether the given peer adult or elder.
+    pub fn is_adult_or_elder(&self, name: &XorName) -> bool {
```
maqi, line 231: shall this check be limited to `active_members` only ?

src/section/mod.rs
```diff
@@ -8,27 +8,395 @@
...
+                self.members.is_joined(info.peer.name()) || self.is_elder(info.peer.name())
+            })
+            .map(|info| &info.peer)
+    }
+
+    /// Returns adults from our section.
+    pub fn adults(&self) -> impl Iterator<Item = &Peer> {
+        self.members
+            .adults()
+            .filter(move |peer| !self.is_elder(peer.name()))
```
maqi, line 258: we already called as `adults` , why extra check of `elder` here?

src/section/mod.rs
```diff
@@ -8,27 +8,395 @@
...
+        self.members
+            .get(name)
+            .map(|info| info.peer.age())
+            .unwrap_or(MIN_AGE)
+    }
+
+    fn elders_info_signing_key_index(&self) -> u64 {
+        // NOTE: we assume that the key the current `EldersInfo` is signed with is always
+        // present in our section proof chain. This is currently guaranteed, because we use the
+        // `SectionUpdateBarrier` and so we always update the current `EldersInfo` and the current
```
maqi, line 279: renamed to `UpdateBarrier` ?

src/section/mod.rs
```diff
@@ -8,27 +8,395 @@
...
+            .map(|peer| peer.name().bit(next_bit_index) == next_bit)
+            .fold((0, 0), |(ours, siblings), is_our_prefix| {
+                if is_our_prefix {
+                    (ours + 1, siblings)
+                } else {
+                    (ours, siblings + 1)
+                }
+            });
+
+        // If either of the two new sections will not contain enough entries, return `false`.
```
maqi, line 315: return None

src/section/mod.rs
```diff
@@ -8,27 +8,395 @@
...
+        if our_new_size < network_params.recommended_section_size
+            || sibling_new_size < network_params.recommended_section_size
+        {
+            return None;
+        }
+
+        let our_prefix = self.prefix().pushed(next_bit);
+        let other_prefix = self.prefix().pushed(!next_bit);
+
+        let our_elders = self.members.elder_candidates_matching_prefix(
```
maqi, line 325: looks like here is only place calling `elder_candidates_matching_prefix` twice with different prefix.
why not have one function, say `split_candidates`, which returns pair result directly?

src/section/mod.rs
```diff
@@ -8,27 +8,395 @@
...
+}
+
+fn cmp_section_chain_position<T: Serialize>(
+    lhs: &Proven<T>,
+    rhs: &Proven<T>,
+    section_chain: &SectionProofChain,
+) -> Option<Ordering> {
+    match (lhs.self_verify(), rhs.self_verify()) {
+        (true, true) => (),
+        (true, false) => return Some(Ordering::Greater),
```
maqi, line 388: shall we still check the key of the valid proof is still in the chain ?

src/messages/src_authority.rs
madadam: This code existed before this PR. I think it's there just for convenience but it's a bit confusing. Will change it.

src/messages/variant.rs
madadam: I'd say that probably no. We weren't logging it before this PR and also we are probably soon going to change the way we keep `network` up to date. It's not decided yet, but It's possible it won't be part of `Sync` anymore.

src/network/mod.rs
```diff
@@ -183,8 +110,6 @@ impl SectionMap {
             }
         }
 
-        self.prune_neighbours();
```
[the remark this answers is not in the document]
madadam: > pruning is now being called outside

This.

src/routing/stage/approved.rs
madadam: No, `old` is their knowledge of our section that we currently know of and `new` is their knowledge of our section according to the message received. We simply check whether `new` is newer than `old` and update it if that's the case.

src/section/mod.rs
madadam: No. Non elder nodes have `members` empty and elders populate it separately. We can't populate it given `EldersInfo` alone, because we also need the proofs.

src/section/mod.rs
madadam: It's in the line above: `self.elders_info = new_elders_info`.

src/section/mod.rs
madadam: Because we want nodes that are adults (age > 4) but not elders. I think are usage of the term `adult` is not consistent. Sometimes it just means a node with age > 4, other times it means non-elder node with age > 4. Will fix it.

src/network/mod.rs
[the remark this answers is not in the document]
maqi: I'd prefer call the pruning inside.
However, not a strong preference.
up to you.

src/section/mod.rs
maqi: but `self.elders_info = new_elders_info` only updates self.elders_info, not self.members ?

src/section/mod.rs
madadam: We could do that, but I don't think it would make much difference. The logic would still exist, just in a different place (in `SectionPeers`).

src/section/mod.rs
madadam: We do. We call `index_of` a couple of lines below which returns `None` if the key is not present.

src/network/mod.rs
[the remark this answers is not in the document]
madadam: Can't do that, because we no longer have our section stored in this struct so we can't tell who is neighbour from inside it. Also I think we are going to address https://github.com/maidsafe/sn_routing/issues/2042 soon and this won't matter.

src/section/mod.rs
madadam: Ah, I see. No, members are updated separately (it's the same reason as explained in the previous comment). We only do the prune here, because this is the place where our prefix can change.

maqi (approved):

PR #2211, opened 2020-10-20
Remove state machine

madadam:
This PR removes the `Bootstrapping` and `Joining` states. The bootstrapping process now happens inside `Routing::new`, so when that function returns, the node is already connected to a section.

There is also some general refactoring and code cleanup in this PR.


src/routing/bootstrap.rs
```diff
@@ -0,0 +1,653 @@
...
+
+        debug!("{} Sending BootstrapRequest to {:?}", self.node, recipients);
+
+        let _ = self.send_tx.send((message.to_bytes(), recipients)).await;
+
+        Ok(())
+    }
+
+    async fn receive_bootstrap_response(&mut self) -> Result<(BootstrapResponse, SocketAddr)> {
+        while let Some((message, sender)) = self.recv_rx.recv().await {
```
maqi, line 203: will this be a permanent block if nothing to be received anymore?

src/routing/bootstrap.rs
```diff
@@ -0,0 +1,653 @@
...
+            match message.variant() {
+                Variant::BootstrapResponse(response) => {
+                    if !self.verify_message(&message, None) {
+                        continue;
```
maqi, line 207: could be an issue of block if the sender doesn't send back response anymore?
and same issue for the wild case below

src/routing/bootstrap.rs
```diff
@@ -0,0 +1,653 @@
...
+    // Send `JoinRequest` and wait for the response. If the response is `Rejoin`, repeat with the
+    // new info. If it is `Approval`, returns the initial `Section` value to use by this node,
+    // completing the bootstrap.
+    async fn join(
+        mut self,
+        mut elders_info: EldersInfo,
+        mut section_key: bls::PublicKey,
+        relocate_payload: Option<RelocatePayload>,
+    ) -> Result<(Node, Section)> {
+        loop {
```
maqi, line 263: bit concerned as the only break out of the loop is to be `Approved`, otherwise will keeps sending out join request or waiting for a response forever.

src/routing/bootstrap.rs
```diff
@@ -0,0 +1,653 @@
...
+        relocate_payload: Option<&RelocatePayload>,
+    ) -> Result<(JoinResponse, SocketAddr)> {
+        while let Some((message, sender)) = self.recv_rx.recv().await {
+            match message.variant() {
+                Variant::BootstrapResponse(BootstrapResponse::Join {
+                    elders_info,
+                    section_key,
+                }) => {
+                    if !self.verify_message(&message, None) {
+                        continue;
```
maqi, line 353: concerned whether this will cause a permanent wait, is the sender doesn't send a response anymore.

src/routing/bootstrap.rs
```diff
@@ -0,0 +1,653 @@
...
+
+                Ok(())
+            })
+            .await
+    }
+
+    // TODO: add test for bootstrap as relocated node
+
+    fn gen_keypair() -> Keypair {
+        let mut rng = rng::new();
```
maqi, line 650: in above there is `crate::rng::new()` used, why the diff ?

src/consensus/vote.rs
```diff
@@ -144,19 +143,19 @@ mod tests {
         let mut rng = rng::new();
 
         // Vote::SectionInfo
-        let (elders_info, _) = section::gen_elders_info(&mut rng, Default::default(), 4);
+        let (elders_info, _) = section::test_utils::gen_elders_info(Default::default(), 4);
```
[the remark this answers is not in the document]
madadam: I got the impression we gave up on reproducible tests via rng seeding.

src/routing/bootstrap.rs
```diff
@@ -0,0 +1,653 @@
...
+
+    let (node, section) = future::join(
+        send_messages(comm, send_rx),
+        local_set.run_until(state.run(vec![bootstrap_addr], None)),
+    )
+    .await
+    .1?;
+
+    // Abort the `receive_messages` task and extract the `incoming_connections` out of it by
+    // awaiting its join handle. This cannot fail, unless there is a bug in the code, so it's OK
```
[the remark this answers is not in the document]
madadam: It would make no difference. This task is only to listen to incoming messages and send them to a channel. It doesn't care about how the messages are handled. 

That said, we currently don't reject join requests. What could happen is that the join request gets lost, or fails to accumulate or similar such things. In that case the bootstrapping process would hang. We used to have a timeout for this, but we decided that it would be better if such timeout was handled by the upper layer. They can do something like this, for example:

```rust
tokio::time::timeout(Duration::from_secs(60), Routing::new(...)).await
```

src/routing/bootstrap.rs
```diff
@@ -0,0 +1,653 @@
...
+    // ...).
+
+    let (send_tx, send_rx) = mpsc::channel(1);
+    let (recv_tx, recv_rx) = mpsc::channel(1);
+
+    let state = State::new(node, send_tx, recv_rx)?;
+    let local_set = task::LocalSet::new();
+
+    let (abort_tx, abort_rx) = oneshot::channel();
+    let incoming_connections = comm.listen()?;
```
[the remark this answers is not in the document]
madadam: No, `listen` doesn't block, it only returns a stream of incoming connections.

src/routing/bootstrap.rs
```diff
@@ -0,0 +1,653 @@
...
+    recv_rx: mpsc::Receiver<(Message, SocketAddr)>,
+    bootstrap_addrs: Vec<SocketAddr>,
+    relocate_details: SignedRelocateDetails,
+) -> Result<(Node, Section)> {
+    // NOTE: when we are re-bootstrapping as a relocated node, the main `Executor` is still running.
+    // We don't create a separate one here (like in the `infant` case), because the main `Executor`
+    // would interfere with it. Instead we read the incoming messages from the `message_rx` whose
+    // sending half is stored inside `Approved` which forwards the relevant messages to it.
+
+    let (send_tx, send_rx) = mpsc::channel(1);
```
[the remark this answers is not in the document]
madadam: Should be, because the channel is being read from concurrently in the `send_messages` task.

Cargo.toml
```diff
@@ -11,38 +11,29 @@ version = "0.37.0"
...
 log = "~0.4.8"
 lru_time_cache = "~0.11.0"
-mock-quic-p2p = { git = "https://github.com/maidsafe/qp2p", rev = "602dc620", optional = true }
-qp2p = { git = "https://github.com/maidsafe/qp2p", features = ["upnp"], branch = "master" }
+qp2p = { version = "~0.8.4", features = ["upnp"] }
 rand = "~0.7.3"
-rand_core = "~0.5.1"
-rand_os = "~0.2.2"
 rand_xorshift = "~0.2.0"
-serde = { version = "1.0.111", features = ["derive" ,"rc"] }
```
maqi, line 38: seems there is compile error for testing code?
```
error[E0463]: can't find crate for `serde_derive` which `serde` depends on
  --> src/lib.rs:70:1
   |
70 | extern crate serde;
   | ^^^^^^^^^^^^^^^^^^^ can't find crate
```

madadam: No idea what's up with this error. Seems to happen only on macos.

src/routing/bootstrap.rs
```diff
@@ -0,0 +1,653 @@
...
+                    elders_info,
+                    section_key,
+                } => {
+                    info!(
+                        "{} Joining a section {:?} (given by {:?})",
+                        self.node, elders_info, sender
+                    );
+                    return Ok((elders_info, section_key));
+                }
+                BootstrapResponse::Rebootstrap(new_bootstrap_addrs) => {
```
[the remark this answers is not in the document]
madadam: Good catch. Sending an empty list would be an act of malice from the bootstrap node though. Not sure what's the best thing to do then: re-send the request or terminate?

src/routing/bootstrap.rs
madadam: Yes, as detailed here: https://github.com/maidsafe/sn_routing/pull/2211#discussion_r509056380. I'm going to add a comment about this, to make it clear.

src/routing/bootstrap.rs
madadam: Same issue

src/routing/bootstrap.rs
madadam: Same

src/routing/bootstrap.rs
madadam: Same

Cargo.toml
madadam: Seems to be fixed now.

src/consensus/vote.rs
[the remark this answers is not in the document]
maqi: yeah I know. However, at least previous we can still re-produce with little manual work.
but with this change, it means almost impossible.

I'd prefer keep the `reproduce ability` unless `have to` remove the passed in rng ?

src/routing/bootstrap.rs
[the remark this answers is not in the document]
maqi: maybe worth to put the above as a comment into the routing/mod.rs , above the Routing struct declaration or `fn new` definition ?

src/routing/bootstrap.rs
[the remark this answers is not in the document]
maqi: I'd prefer a termination.
Though, a resend is acceptable, if easier to impl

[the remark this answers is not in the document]
madadam: Both should be equally easy. OK, will implement the termination.

maqi (approved):

src/routing/bootstrap.rs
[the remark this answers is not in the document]
madadam: Already did.

src/consensus/vote.rs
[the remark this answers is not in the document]
madadam: I think most (or all) of those unit test should not be affected by the rng anyway. And if they are, we should change them to proptest/quickcheck which would gives us reproducibility, among other nice things. As for the integration tests - they use real network now so they can't be made reproducible anyway.

PR #2212, opened 2020-10-21
resolve failing example

maqi:
non-first node shall have non-contacts to bootstrap with
spawned handlers shall be collected and joined

On top of PR 2211

### DIFF
--- a/examples/minimal.rs
+++ b/examples/minimal.rs
@@ -33,6 +33,7 @@
 //! command-line option for more details.
 //!
 
+use futures::future::join_all;
 use hex_fmt::HexFmt;
 use log::{info, LevelFilter};
 use sn_routing::{
@@ -42,9 +43,11 @@ use sn_routing::{
 use std::{
     collections::HashSet,
     convert::TryInto,
+    iter,
     net::{IpAddr, Ipv4Addr, SocketAddr},
 };
 use structopt::StructOpt;
+use tokio::task::JoinHandle;
 
 /// Minimal example node.
 #[derive(Debug, StructOpt)]
@@ -52,7 +55,13 @@ struct Options {
     /// Socket address (e.g. 203.0.113.45:6789) of a node(s) to bootstrap against. Multiple
     /// contacts can be specified by passing the option multiple times. If omitted, will try to use
     /// contacts cached from previous run, if any.
-    #[structopt(short, long, name = "bootstrap-contact", value_name = "SOCKET_ADDRESS")]
+    #[structopt(
+        short,
+        long,
+        name = "bootstrap-contact",
+        value_name = "SOCKET_ADDRESS",
+        required_unless = "first"
+    )]
     bootstrap_contacts: Vec<SocketAddr>,
     /// Whether this is the first node ("genesis node") of the network. Only one node can be first.
     #[structopt(short, long, conflicts_with = "bootstrap-contact")]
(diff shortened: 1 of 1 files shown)


examples/minimal.rs
```diff
@@ -202,21 +227,24 @@ async fn start_node(
...
 
-    run_node(index, node, event_stream).await;
-
-    contact_info
+    (contact_info, handler)
 }
 
 // Runs the nodes event loop. Blocks until terminated.
-async fn run_node(index: usize, mut node: Routing, mut event_stream: EventStream) {
+async fn run_node(
```
madadam, line 236: This function doesn't need to be `async`.

examples/minimal.rs
```diff
@@ -98,8 +101,10 @@ async fn main() {
     let opts = Options::from_args();
     init_log(opts.verbosity);
 
-    if opts.count <= 1 {
-        start_single_node(opts.first, opts.bootstrap_contacts, opts.ip, opts.port).await
+    let handlers: Vec<JoinHandle<()>> = if opts.count <= 1 {
```
madadam, line 104: Unnecessary type qualification. Just `let handles: Vec<_> = ...` should be enough.

madadam, line 104: Also, it's `JoinHandle`, not `JoinHandler`, so the variable should probably be called `handles`.

examples/minimal.rs
```diff
@@ -118,8 +124,10 @@ async fn start_single_node(
     contacts: Vec<SocketAddr>,
     ip: Option<IpAddr>,
     port: Option<u16>,
-) {
-    let _contact = start_node(0, first, contacts.into_iter().collect(), ip, port).await;
+) -> JoinHandle<()> {
+    let (_contact, handler): (SocketAddr, JoinHandle<()>) =
```
madadam, line 128: Again, rust can infer the type, no need to specify them. Also `handler` -> `handle`.

examples/minimal.rs
```diff
@@ -132,23 +140,34 @@ async fn start_multiple_nodes(
     mut contacts: Vec<SocketAddr>,
     ip: Option<IpAddr>,
     base_port: Option<u16>,
-) {
+) -> Vec<JoinHandle<()>> {
+    let mut handlers = Vec::new();
     let first_index = if first {
-        let first_contact = spawn_first_node(ip, base_port).await;
+        let (first_contact, first_handler): (SocketAddr, JoinHandle<()>) =
```
madadam, line 146: same

examples/minimal.rs
```diff
@@ -132,23 +140,34 @@ async fn start_multiple_nodes(
...
+            spawn_other_node(index, contacts.clone(), ip, base_port).await;
+        handlers.push(handler);
     }
+    handlers
 }
 
 // Spawns the first (genesis) node in its own thread
-async fn spawn_first_node(ip: Option<IpAddr>, base_port: Option<u16>) -> SocketAddr {
-    start_node(0, true, Vec::default(), ip, base_port).await
+async fn spawn_first_node(
```
madadam, line 164: There doesn't seem to be much that these two functions are doing (`spawn_first_node` and `spawn_other_node`). I'd remove them and call `start_node` directly.

examples/minimal.rs
```diff
@@ -202,21 +227,24 @@ async fn start_node(
...
         .expect("Failed to obtain node's contact info.");
+    let handler: JoinHandle<()> = run_node(index, node, event_stream).await;
 
-    run_node(index, node, event_stream).await;
-
-    contact_info
+    (contact_info, handler)
 }
 
 // Runs the nodes event loop. Blocks until terminated.
```
madadam, line 235: This comment is outdated. Maybe change to something like "spawns a task to run the node until terminated".

examples/minimal.rs
```diff
@@ -170,7 +166,11 @@ async fn start_node(
     contacts: Vec<SocketAddr>,
     ip: Option<IpAddr>,
     base_port: Option<u16>,
-) -> SocketAddr {
+) -> (SocketAddr, JoinHandle<()>) {
+    if !first && contacts.is_empty() {
+        panic!("non-first node shall have non-empty contacts to bootstrap with!");
```
madadam, line 171: Maybe add this also to the structopt thing, so we get nicer output? Something like this:
```rust
#[structopt(..., required_unless = "first")]
bootstrap_contacts: Vec<SocketAddr>,
```
?

madadam: required_unless: https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.required_unless

madadam (approved):

madadam (approved):

PR #2213, opened 2020-10-21
fix: no longer use serde macro derive

maqi:
switching to the formal setup as stated at https://serde.rs/derive.html


S-Coyle (approved):
Seems fine from a QA/CI perspective 👍

dirvine (approved):

PR #2215, opened 2020-10-23
chore(simplify)

dirvine:
Remove rng wrappers and xor_seeded rng.

If needed to seed we can use  rand_pcg or rand_chacha but we will use proptest nd fuzz and set up any seeded rng there.


src/consensus/vote.rs
```diff
@@ -131,16 +131,15 @@ impl Serialize for SignableWrapper {
...
     use super::*;
-    use crate::{
-        rng::{self, MainRng},
-        section,
-    };
-    use rand::Rng;
+    use crate::section;
+    use rand::{Rng, SeedableRng};
+
+    use rand::rngs::SmallRng;
```
maqi, line 137: these can be merged into one bracket?

src/network/mod.rs
```diff
@@ -303,13 +303,9 @@ where
...
-    use crate::{
-        consensus,
-        location::DstLocation,
-        rng::{self, MainRng},
-        section,
-    };
-    use rand::Rng;
+    use crate::{consensus, location::DstLocation, section};
+    use rand::rngs::SmallRng;
+    use rand::{Rng, SeedableRng};
```
maqi, line 308: same here, merge these two into one bracket?

src/routing/tests/mod.rs
```diff
@@ -569,7 +571,7 @@ async fn handle_unknown_message(source: UnknownMessageSource) -> Result<()> {
     let original_message = Message::single_src(
         &sender_keypair,
         MIN_AGE + 1,
-        DstLocation::Section(rng::new().gen()),
+        DstLocation::Section(SmallRng::from_entropy().gen()),
```
maqi, line 574: just wondering, why sometime OsRng, sometime SmallRng ?

dirvine: One is a secure rng and the other is not. The secure rng is much more resource-hungry and best to only use it for csprng

dirvine:
Closing as this is handled in https://github.com/maidsafe/sn_routing/pull/2218/commits/0e811f24d97010978936401cd92ffcc8b1521a91

src/routing/tests/mod.rs
maqi: thx for the clarification

PR #2216, opened 2020-10-24
chore(consts)

dirvine:
Update const values to be clear when using magic numbers and dynamically calculated values. Also set majority to strict majority.

### DIFF
--- a/src/delivery_group.rs
+++ b/src/delivery_group.rs
@@ -8,9 +8,16 @@
 
 //! Utilities for sn_routing messages through the network.
 
-use crate::{ELDER_MAJORITY, error::{Error, Result}, location::DstLocation, network::Network, peer::Peer, section::Section};
-use std::{cmp, iter};
+use crate::{
+    error::{Error, Result},
+    location::DstLocation,
+    network::Network,
+    peer::Peer,
+    section::Section,
+    ELDER_MAJORITY,
+};
 use itertools::Itertools;
+use std::{cmp, iter};
 use std::{cmp, iter};
 use xor_name::XorName;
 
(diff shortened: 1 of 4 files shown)


src/delivery_group.rs
```diff
@@ -107,11 +101,19 @@ fn candidates(
...
     let mut nodes_to_send = Vec::new();
     for (idx, (prefix, len, connected)) in sections.enumerate() {
         nodes_to_send.extend(connected.cloned());
-        dg_size = delivery_group_size(len);
+        // If we don't have enough contacts send to as many as possible
+        // up to majority of Elders
+        dg_size = cmp::min(len, dg_size);
+        if len < ELDER_MAJORITY {
+            warn!(
+                "Delivery group only {:?} when it should be {:?}",
```
maqi, line 113: `when it should be` sounds like this case is a `definitely wrong situation`, and shall terminate and return
However, such case may happen and we do continue.
Maybe change to : `Delivery group only {:?} when expected to be {:?}` ?

maqi (approved):

src/delivery_group.rs
dirvine: It's an area we should talk more about. The consts assume we have that amount of Elders and a fixed majority. So when there are less and the majority is still >50% of remaining nodes then is it still majority? Tricky edge case. I assume we keep going, especially now as we have at least once delivery and lazy messaging. 
I will make this change though @maqi  as what you suggest reads better for sure.

src/lib.rs
```diff
@@ -122,21 +115,25 @@ const RECOMMENDED_SECTION_SIZE: usize = 60;
...
-            "Majority impossible to achieve"
-        );
-        assert!(
-            MAJORITY_NUMERATOR * 2 >= MAJORITY_DENOMINATOR,
-            "Majority does not guarantee agreement"
-        );
+    fn strict_majority() {
+        assert_eq!(majority(4), majority(5));
+        assert_eq!(majority(6), majority(7));
+        assert_eq!(majority(7), 4);
```
madadam, line 136: Maybe worth adding more asserts here:
```rust
assert_eq!(majority(0), 1) // pathological case
assert_eq!(majority(1), 1)
assert_eq!(majority(2), 2)
assert_eq!(majority(3), 2)
assert_eq!(majority(4), 3)
assert_eq!(majority(5), 3)
assert_eq!(majority(6), 4)
```
Possibly even proptest this?

madadam (dismissed):

src/lib.rs
dirvine: I have a pr to add proptest here. I can put that up next (branced off here)

maqi (approved):

PR #2218, opened 2020-10-26
Refactoring + more tests + fix DKG

madadam:


src/routing/comm.rs
```diff
@@ -292,6 +296,87 @@ impl SendState {
     }
 }
 
+// Stream of incoming messages from all existing and incoming connections.
```
maqi, line 299: `from all existing and incoming connections.` ?
it's not for just one connection?

the constructor seems to construct new channel for new incoming_conns ?

src/routing/executor.rs
```diff
@@ -6,101 +6,81 @@
...
+    while let Some(msg) = incoming_msgs.next().await {
+        match msg {
+            qp2p::Message::UniStream { bytes, src, .. } => {
+                trace!(
+                    "New message ({} bytes) received on a uni-stream from: {}",
+                    bytes.len(),
+                    src
+                );
+                // Since it's arriving on a uni-stream we treat it as a Node
+                // message which we need to be processed by us, as well as
```
maqi, line 54: treat it as a Node message which need to be processed by us, and report it to the event stream consumer as well.

src/section/section_peers.rs
```diff
@@ -181,18 +177,14 @@ impl IntoIterator for SectionPeers {
 }
 
 // Returns the nodes that should become the next elders out of the given members.
```
maqi, line 179: better mention the resulted Vec is sorted by name

src/routing/tests/mod.rs
```diff
@@ -1142,6 +1210,30 @@ fn create_section(
     Ok((section, section_key_share))
 }
 
+// Create a `Vote::Online` whose consensus handling triggers relocation of a node with the given age.
+// NOTE: recommended to call this with low `age` (4 or 5), otherwise it might take very long time
```
maqi, line 1214: better to explain the longer time for a higher age is due to make signature having more trailing zeroes?

src/consensus/dkg.rs
```diff
@@ -35,8 +34,9 @@ impl DkgKey {
         // Calculate the hash without involving serialization to avoid having to return `Result`.
         let mut hasher = Sha3::v256();
 
-        for name in elders_info.elders.keys() {
-            hasher.update(&name.0);
+        for peer in elders_info.elders.values() {
+            hasher.update(&peer.name().0);
+            hasher.update(&[peer.age()]);
```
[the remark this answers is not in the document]
dirvine: Again I feel name+age needs to be immutable if that helps.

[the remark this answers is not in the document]
madadam: The reason for this is that during the startup phase, we increase ages on every churn event directly. So currently name+age is **not** immutable. But I think we should change that because it complicates things. We might use relocation instead, but we need to be careful to not trigger relocation loop (a node is relocated and as it re-joins, it triggers more relocations, etc...). I'd prefer to tackle this issue as a separate task.

src/routing/comm.rs
madadam: OK, that wording is clumsy, but what I wanted to say is that it will receive messages on all established connections including the ones that will only be established in the future. In other words, it listen for incoming connections and then multiplex the messages from those connection into a single stream. Will try to reword that.

src/consensus/dkg.rs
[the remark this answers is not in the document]
maqi: But we still have the issue of same participants with different ages will have same dkg_keys ?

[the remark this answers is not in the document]
madadam: Because the `DkgKey` is different if the ages are different, it will start a new dkg session even if we have one for the same participants already running.

maqi (approved):

PR #2224, opened 2020-10-28
Extract most of the DKG logic out of Approved and into DkgVoter + adds some DKG tests

madadam:
This is a part of the effort to simplify the `Approved` struct by extracting parts of it into separate self-contained modules.

### DIFF
--- a/src/consensus/dkg.rs
+++ b/src/consensus/dkg.rs
@@ -10,8 +10,7 @@ use crate::{
     crypto::Digest256,
     error::Result,
     majority,
-    messages::Message,
-    messages::Variant,
+    messages::{Message, Variant},
     node::Node,
     peer::Peer,
     routing::command::{self, Command},
@@ -422,16 +421,12 @@ impl Participant {
     }
 
     fn check(&mut self) -> Option<Command> {
-        let _ = self.elders_info.as_ref()?;
-
         if !self.key_gen.is_finalized() {
             return None;
         }
 
         let (participants, outcome) = self.key_gen.generate_keys()?;
-
-        // This is OK to `unwrap` because we already checked it is `Some`.
-        let elders_info = self.elders_info.take().unwrap();
+        let elders_info = self.elders_info.take()?;
 
         if participants.iter().eq(elders_info.elders.keys()) {
             trace!(
--- a/src/consensus/dkg.rs
+++ b/src/consensus/dkg.rs
@@ -480,7 +480,10 @@ struct Observer {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{section::test_utils::gen_addr, MIN_AGE};
+    use crate::{
+        section::test_utils::{gen_addr, gen_elders_info},
+        ELDER_SIZE, MIN_AGE,
+    };
     use std::iter;
     use xor_name::Prefix;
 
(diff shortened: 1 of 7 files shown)


src/routing/approved.rs
```diff
@@ -397,12 +411,17 @@ impl Approved {
                 }
             }
             Variant::UserMessage(_) => {
-                if !self.should_handle_user_message(msg.dst()) {
+                // If elder, always handle UserMessage, otherwise handle it only if addressed
```
maqi, line 414: Any particular reason to carry out this change (and the same two below) ?
I'd prefer using the previous one to have check and comment in a separate function.

src/routing/approved.rs
```diff
@@ -250,6 +234,72 @@ impl Approved {
...
+        dkg_key: DkgKey,
+        elders_info: EldersInfo,
+        result: Result<DkgOutcome, ()>,
+    ) -> Result<Vec<Command>> {
+        let key = result
+            .as_ref()
+            .map(|outcome| outcome.public_key_set.public_key())
+            .map_err(|_| ());
+
+        if let Ok(outcome) = result {
```
maqi, line 248: the can be merged with the above ?

src/routing/command.rs
```diff
@@ -40,6 +42,17 @@ pub(crate) enum Command {
     HandleVote { vote: Vote, proof_share: ProofShare },
     /// Handle consensus on a vote.
     HandleConsensus { vote: Vote, proof: Proof },
+    /// Handle the result of a DKG session we are participating in
+    HandleDkgParticipationResult {
+        dkg_key: DkgKey,
+        elders_info: EldersInfo,
+        result: Result<DkgOutcome, ()>,
+    },
+    /// Handle the result of a DKG session we observing
```
maqi, line 51: according to the comment, these two entries are very similar and confusing.
maybe expand the comment bit more?

src/consensus/dkg.rs
```diff
@@ -368,6 +357,117 @@ struct Participant {
...
+    }
+
+    fn broadcast(
+        &mut self,
+        node: &Node,
+        dkg_key: DkgKey,
+        dkg_message: DkgMessage,
+    ) -> Result<Vec<Command>> {
+        let recipients: Vec<_> = self
+            .elders_info
```
maqi, line 391: is this `self.elders_info` is the node's current `elders_info` or this dkg_voter's targeted elders_info?

src/consensus/dkg.rs
```diff
@@ -368,6 +357,117 @@ struct Participant {
...
+            recipients.len(),
+            message.to_bytes(),
+        ));
+        commands.extend(self.process_message(node, dkg_key, dkg_message)?);
+
+        Ok(commands)
+    }
+
+    fn check(&mut self) -> Option<Command> {
+        let _ = self.elders_info.as_ref()?;
```
maqi, line 424: better at least have a log for None case ?

src/consensus/dkg.rs
```diff
@@ -368,6 +357,117 @@ struct Participant {
...
+
+        if !self.key_gen.is_finalized() {
+            return None;
+        }
+
+        let (participants, outcome) = self.key_gen.generate_keys()?;
+
+        // This is OK to `unwrap` because we already checked it is `Some`.
```
maqi, line 432: still bit tricky. prefer not to use this.

src/routing/approved.rs
madadam: I thought that as those functions are all one-liners and each used only once, there is not much reason having them. But I can revert if you thing it's cleaner.

src/consensus/dkg.rs
madadam: It's the proposed new elders info.

src/consensus/dkg.rs
madadam: That would happen when the DKG already completed but we are still handling some messages because some of the other participants have not completed yet. Putting a log there might cause log spam and I don't think it's that useful anyway. Or do you think it is?

src/routing/approved.rs
maqi: I'd prefer revert them, as the comments here is making this already lengthy function expanded further.

src/consensus/dkg.rs
maqi: I am mainly agaisnt having an empty `?` , especially it is needed below.
So when this merged with the one below, the log itself is not necessary

src/consensus/dkg.rs
```diff
@@ -422,16 +421,12 @@ impl Participant {
...
-
         if !self.key_gen.is_finalized() {
             return None;
         }
 
         let (participants, outcome) = self.key_gen.generate_keys()?;
-
-        // This is OK to `unwrap` because we already checked it is `Some`.
-        let elders_info = self.elders_info.take().unwrap();
+        let elders_info = self.elders_info.take()?;
```
maqi, line 429: will this cause a termination?

src/consensus/dkg.rs
```diff
@@ -368,6 +357,113 @@ struct Participant {
...
+        Ok(commands)
+    }
+
+    fn check(&mut self) -> Option<Command> {
+        if !self.key_gen.is_finalized() {
+            return None;
+        }
+
+        let (participants, outcome) = self.key_gen.generate_keys()?;
+        let elders_info = self.elders_info.take()?;
```
maqi, line 429: is it safe to do so?

madadam: If `self.elders_info` is `None` that means the DKG has completed. So I think it is. Only issue is we would now still call `generate_keys`. Not sure it's a problem. Maybe just wasted computation? This was solved by the previous approach, but that required `unwrap`.

maqi (approved):

PR #2225, opened 2020-10-29
Remove NetwrokParams and use const value for testnet

maqi:


dirvine (approved):

src/routing/mod.rs
```diff
@@ -50,7 +49,10 @@ pub struct Config {
     /// Configuration for the underlying network transport.
     pub transport_config: TransportConfig,
     /// Global network parameters. Must be identical for all nodes in the network.
-    pub network_params: NetworkParams,
+    /// The number of elders per section
+    pub elder_size: usize,
+    /// Recommended number of nodes in a section.
+    pub recommended_section_size: usize,
```
madadam, line 55: Do we still want this?

maqi: ok. will remove them

madadam (approved):

PR #2228, opened 2020-11-05
Startup phase improvements

madadam:
During startup phase, every infant that joins is instantly relocated back to the same section but with increased age. The new age is pseudo-randomly generated from a given age range to prevent the peers to have the same age which would cause them to be relocated at the same time. Also the relocation calculation is changed so that now we only relocate the oldest matching nodes, not all of them. There are also some new tests for this new behaviour. Finally the `Event` enum is slightly simplified which is an API breaking change.


src/relocation.rs
```diff
@@ -282,4 +336,127 @@ mod tests {
...
+        peers: Vec<Peer>,
+        signature_trailing_zeros: u8,
+        seed: u64,
+    ) -> Result<()> {
+        let mut rng = SmallRng::seed_from_u64(seed);
+
+        let sk: bls::SecretKey = rng.gen();
+        let pk = sk.public_key();
+
+        // Create `Section` with `peers` as its members and set the `ELDER_SIZE` oldest peers as
```
maqi, line 363: fort he peers with same age, shall there be a second order of using XorName ?
as the input peers is a Vec, will this causing issue ?

src/routing/tests/mod.rs
```diff
@@ -1174,7 +1226,7 @@ fn create_relocation_trigger(sk: &bls::SecretKey, age: u8) -> Result<(Vote, Proo
 
         let signature = sk.sign(&bincode::serialize(&vote.as_signable())?);
 
-        if relocation::check(age, &signature) {
+        if relocation::check(age, &signature) && !relocation::check(age + 1, &signature) {
```
maqi, line 1229: why the extra check?

src/relocation.rs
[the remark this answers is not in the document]
madadam: No, it makes no difference here. It only matters when we are deciding who to promote/demote, but this test is not about that.

src/routing/tests/mod.rs
madadam: Because we only want to trigger relocation of nodes with that exact age, not with any higher age.

maqi (approved):

PR #2230, opened 2020-11-10
Improve lost peer detection

madadam:
~~This is draft until https://github.com/maidsafe/qp2p/pull/172 is merged.~~


src/consensus/dkg.rs
```diff
@@ -114,6 +122,20 @@ impl DkgVoter {
             }
         }
 
+        // Special case: only one participant.
+        if elders_info.elders.len() == 1 {
```
maqi, line 126: shall it also be checked that the `elders` is only self ?
or it has already be checked ?

tests/drop.rs
```diff
@@ -9,48 +9,57 @@
...
-            "Event::MemberLeft not received for {}",
-            dropped_name
-        ))
-    };
+        assert!(
+            received,
+            "event stream closed before receiving Event::EldersChanged"
+        );
+    }
+
```
maqi, line 53: I'd suggest to use something like:
```rust
    let mut nodes = create_connected_nodes(3).await?;
    for i in 1..nodes.len() {
        assert_next_event!(nodes[i].1, Event::RelocationStarted { .. });
        assert_next_event!(nodes[i].1, Event::Relocated { .. });
        expect_event!(nodes[i].1, Event::PromotedToElder);
    }
```
As, the above procedure shall be covered by a proper `bootstrap flow` test.

and the expect_event macro is:
```rust
macro_rules! expect_event {
    ($event_stream:expr, $pattern:pat) => {
        loop {
            match tokio::time::timeout($crate::utils::TIMEOUT, $event_stream.next()).await {
                Ok(Some($pattern)) => break,
                Ok(other) => trace!("Received {:?}", other),
                Err(_) => panic!("Timeout when expecting {}", stringify!($pattern)),
            }
        }
    };
}
```

tests/drop.rs
```diff
@@ -9,48 +9,57 @@
...
+    // Drop one node
+    let dropped_name = nodes.remove(1).0.name().await;
+
+    while let Some(event) = nodes[0].1.next().await {
```
maqi, line 62: so there is no more need of sending out a msg to trigger the lost detection anymore, right?

also, maybe better to check all leftover nodes, instead of just node[0], to ensure everything is OK ?

src/routing/comm.rs
```diff
@@ -9,80 +9,76 @@
...
+use std::{
+    fmt::{self, Debug, Formatter},
+    net::SocketAddr,
 };
-
-// Number of Connections to maintain in the cache
-const CONNECTIONS_CACHE_SIZE: usize = 1024;
-
-/// Maximal number of resend attempts to the same target.
-pub const RESEND_MAX_ATTEMPTS: u8 = 3;
```
maqi, line 28: just to confirm, the all these (cache & resend) are moved into QuicP2p ?

src/routing/comm.rs
```diff
@@ -122,295 +127,167 @@ impl Comm {
...
+                Ok(()) => successes += 1,
+                Err(qp2p::Error::Connection(qp2p::ConnectionError::LocallyClosed)) => {
+                    // The connection was closed by us which means we are terminating so let's cut
+                    // this short.
+                    return (Err(SendError), vec![]);
+                }
+                Err(_) => {
+                    failed_recipients.push(*addr);
+
+                    if next < recipients.len() {
```
maqi, line 157: just pop(take) from recipients, so that using of `next` can be avoided?
also, when `next >= recipients.len()` , shall we break out of the loop?

src/consensus/dkg.rs
```diff
@@ -448,6 +481,40 @@ struct Observer {
...
+        }
+
+        self.0.push_back((dkg_key, message))
+    }
+
+    fn take(&mut self, dkg_key: &DkgKey) -> Vec<DkgMessage> {
+        let mut output = Vec::new();
+        let max = self.0.len();
+
+        for _ in 0..max {
```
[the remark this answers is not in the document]
madadam: It would be a couple lines shorter, but this way it's marginally more efficient (avoids allocations). I can change it if you prefer.

src/consensus/dkg.rs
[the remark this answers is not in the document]
madadam: We check that in `Approved::decide_message_status`. Do you think we should check here again, just in case?

src/routing/approved.rs
```diff
@@ -230,6 +250,11 @@ impl Approved {
             let key = outcome.public_key_set.public_key();
             self.section_keys_provider
                 .insert_dkg_outcome(&self.node.name(), &elders_info, outcome);
+
+            if self.section.chain().has_key(&key) {
```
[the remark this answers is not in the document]
madadam: Those are two separate things: one is the BLS public key which we append into our section chain and the other is the new BLS secret key share which we store in `SectionKeysProvider`. We need both to be able to start using the new key share for voting. Problem is the appending of the public key and the completion of the DKG are two independent events that can happen in any order, so we need to handle it both ways.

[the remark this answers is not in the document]
madadam: Note that we were already handling this before this PR, it just wasn't entirely correctly implemented.

src/routing/mod.rs
```diff
@@ -321,6 +320,79 @@ impl Routing {
...
+        self.stage.terminate()
+    }
+}
+
+// Listen for incoming connection events and handle them.
+async fn handle_connection_events(
+    stage: Arc<Stage>,
+    mut incoming_conns: mpsc::Receiver<ConnectionEvent>,
+) {
+    while let Some(event) = incoming_conns.recv().await {
```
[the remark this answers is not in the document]
madadam: No, the loop will break only when all the `Sender`s get dropped which happens when all the connections are closed. We forcefully close all the connection when we drop the `Routing` instance by calling `Comm::terminate` which calls `Endpoint::close` which closes all open connections and also stops accepting new connections. Note that `Endpoint::close` is a new feature added in https://github.com/maidsafe/qp2p/pull/172.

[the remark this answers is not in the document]
madadam: > make it cancellable 

`cancellable` is gone as it's not needed anymore.

src/routing/mod.rs
```diff
@@ -321,6 +320,79 @@ impl Routing {
...
+}
+
+async fn handle_message(stage: Arc<Stage>, msg_bytes: Bytes, sender: SocketAddr) {
+    match Message::from_bytes(&msg_bytes) {
+        Ok(message) => {
+            let command = Command::HandleMessage {
+                message,
+                sender: Some(sender),
+            };
+            let _ = stage.handle_commands(command).await;
```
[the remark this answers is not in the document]
madadam: It's already logged in `Stage::handle_command`.

tests/drop.rs
madadam: Yeah, I was also thinking about adding something like this `expect_event` macro. OK, will do it, but I'll call it `assert_event` for consistency.

tests/drop.rs
madadam: > so there is no more need of sending out a msg to trigger the lost detection

That's right. That's basically the main point of this PR.

> maybe better to check all leftover nodes, instead of just node[0], to ensure everything is OK 

Do you mean to check that the other nodes also raised `Event::MemberLeft` ?

src/routing/comm.rs
madadam: The cache yes. The resend isn't explicitly implemented in qp2p, but according to @dirvine , quinn/quic should do something like that by itself.

src/routing/comm.rs
madadam: > just pop(take) from recipients

`recipients` is slice not `Vec`, so there is no `pop`. But even it there was, that would reverse the order of the recipients (but maybe that doesn't matter?). I could maybe do something with `slice::split_first` or so, but I don't think the code would be simpler.

> when next >= recipients.len() , shall we break out of the loop

We still need to wait for all the ongoing send task to finish.

src/consensus/dkg.rs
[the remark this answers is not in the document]
maqi: up to you then.
efficiency is also preferred.  :)

src/consensus/dkg.rs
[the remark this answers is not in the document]
maqi: Ah, right, we already checked the list contains us over there. 
no need for extra checking then.

src/routing/mod.rs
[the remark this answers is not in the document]
maqi: ah, thx for the explaination. not aware of the new `Endpoint::close` function.

tests/drop.rs
maqi: > Do you mean to check that the other nodes also raised Event::MemberLeft ?

yes

src/routing/comm.rs
maqi: ok then.

src/routing/stage.rs
```diff
@@ -145,18 +154,13 @@ impl Stage {
...
-            Ok(()) => vec![],
-            Err(error) => error
-                .failed_recipients
-                .into_iter()
-                .map(Command::HandlePeerLost)
-                .collect(),
-        }
+            .1
+            .into_iter()
+            .map(Command::HandlePeerLost)
```
maqi, line 162: at this point, I think we shall only calll `HandleConnectionLost` ?
or, call `HandlePeerLost` only when the message is a `Ping` ?

or, you mean if the ConnectionLost is caused during a message lost, we can immediately claim that peer is lost ?

tests/utils/mod.rs
```diff
@@ -95,33 +68,48 @@ pub const TIMEOUT: Duration = Duration::from_secs(60);
...
         match tokio::time::timeout($crate::utils::TIMEOUT, $event_stream.next()).await {
-            Ok(Some($pattern)) => {}
+            Ok(Some($pattern)) $(if $cond)? => {}
             Ok(other) => panic!("Expecting {}, got {:?}", stringify!($pattern), other),
             Err(_) => panic!("Timeout when expecting {}", stringify!($pattern)),
         }
     };
 }
 
+/// Assert that some event in the event stream matches the given pattern.
```
maqi, line 80: Assert that at least one event in ...

src/routing/stage.rs
madadam: >  if an error happend during sending a message, we can immediately claim that peer is lost

Yes, that's the idea.

tests/utils/mod.rs
madadam: Same meaning no? But OK, will change it.

maqi (approved):

PR #2231, opened 2020-11-12
feat: allow rejoin with same name

maqi:
As discussed, rejoin node with same name will be immediately relocated with half of original age.

Close https://github.com/maidsafe/sn_routing/issues/2214

### DIFF
--- a/src/routing/approved.rs
+++ b/src/routing/approved.rs
@@ -29,15 +29,15 @@ use crate::{
         SignedRelocateDetails,
     },
     section::{
-        EldersInfo, MemberInfo, Section, SectionKeyShare, SectionKeysProvider, SectionProofChain,
-        MIN_AGE,
+        EldersInfo, MemberInfo, PeerState, Section, SectionKeyShare, SectionKeysProvider,
+        SectionProofChain, MIN_AGE,
     },
     RECOMMENDED_SECTION_SIZE,
 };
 use bls_dkg::key_gen::outcome::Outcome as DkgOutcome;
 use bytes::Bytes;
 use itertools::Itertools;
-use std::{net::SocketAddr, slice};
+use std::{cmp, net::SocketAddr, slice};
 use tokio::sync::mpsc;
 use xor_name::{Prefix, XorName};
 
@@ -213,7 +213,7 @@ impl Approved {
         }
 
         if let Some(info) = self.section.members().get(&name) {
-            let info = info.clone().leave();
+            let info = info.clone().leave()?;
             self.vote(Vote::Offline(info))
         } else {
             Ok(vec![])
@@ -1086,6 +1086,20 @@ impl Approved {
         self.send_relocate(&peer, details)
     }
 
+    fn relocate_rejoin_peer(&self, peer: Peer, age: u8) -> Result<Vec<Command>> {
+        let details =
+            RelocateDetails::with_age(&self.section, &self.network, &peer, *peer.name(), age);
+
+        trace!(
+            "Relocating {:?} to {} with age {} due to rejoin",
+            peer,
+            details.destination,
+            details.age
+        );
+
+        self.send_relocate(&peer, details)
+    }
+
     // Are we in the startup phase? Startup phase is when the network consists of only one section
     // and it has no more than `recommended_section_size` members.
     fn is_in_startup_phase(&self) -> bool {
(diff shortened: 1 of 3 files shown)


src/routing/approved.rs
```diff
@@ -1158,6 +1176,36 @@ impl Approved {
...
+                    DstLocation::Direct,
+                    variant,
+                    Some(self.section.create_proof_chain_for_our_info(None)),
+                    None,
+                )?;
+
+                commands.push(Command::send_message_to_target(
+                    peer.addr(),
+                    message.to_bytes(),
+                ));
```
madadam, line 1205: The `Relocate` message cannot be sent like this, it needs to be accumulated. Just use `send_relocate` instead of this.

src/routing/approved.rs
```diff
@@ -1158,6 +1176,36 @@ impl Approved {
...
         let signature = proof.signature.clone();
 
+        // A vote of half-aged Left peer means such peer is trying to rejoin with the same name,
+        // and shall be relocated immediately. The flow will return early during the process of
+        // update_member, as `Left -> Relocated` is not allowed.
+        //
+        // TODO: shall we make `Left -> Relocated` to avoid the left node is being repeatedly used
+        //       to generate higher new nodes?
+        if let Some(info) = self.section.members().get(member_info.peer.name()) {
+            if info.peer.age() > age {
```
madadam, line 1186: Should we also check that `info.state == Relocated` ?

madadam (commented):
Looks OK, apart from the `Relocate` message issue, but that should be easy to fix. Would be great to also add some tests for the new logic.

src/routing/approved.rs
maqi: don't need to.
as this is handling a concensused Vote::Offline, and there is logic added that a Relocated node is not allowed to trigger a Vote::Offline.
i.e. A relocated Node is not allowed to rejoin with its origin name

src/routing/approved.rs
maqi: `send_relocate ` requires the receiver to have a vote_accumulator, which a `bootstraping` node doesn't have yet?

src/routing/tests/mod.rs
```diff
@@ -390,6 +390,102 @@ async fn handle_consensus_on_online_of_elder_candidate() -> Result<()> {
...
+        .map(|node| node.with_age(32))
+        .collect();
+    let elders_info = EldersInfo::new(nodes.iter().map(Node::peer), Prefix::default());
+    let (mut section, section_key_share) = create_section(&sk_set, &elders_info)?;
+
+    // Simulate a peer of age 16 is joined.
+    let existing_peer = create_peer().with_age(16);
+    let member_info = MemberInfo::joined(existing_peer);
+    let member_info = proven(sk_set.secret_key(), member_info)?;
+    let _ = section.update_member(member_info);
```
madadam, line 408: You don't need to simulate the first join and offline. You can create the peer in the `Left` state initially and only test the rejoin itself. This is OK, because we already have tests for joining and leaving. It would also make this test shorter and simpler.

In general, these unit tests should always follow this structure, unless there is a good reason not to:
1. Setup the initial state
2. Execute one command
3. Perform assertions on the resulting commands and/or events

src/routing/tests/mod.rs
```diff
@@ -390,6 +340,131 @@ async fn handle_consensus_on_online_of_elder_candidate() -> Result<()> {
     Ok(())
 }
 
+async fn handle_online_command(
+    peer: &Peer,
+    sk_set: &SecretKeySet,
+    stage: &Stage,
+    elders_info: EldersInfo,
+) -> Result<(bool, bool, u8)> {
```
madadam, line 348: Add a comment to document the return type

src/routing/tests/mod.rs
```diff
@@ -390,6 +340,131 @@ async fn handle_consensus_on_online_of_elder_candidate() -> Result<()> {
...
+    let stage = Stage::new(state, create_comm()?);
+
+    // Simulate peer with the same name is rejoin and verify resulted behaviours.
+    let (node_approval_sent, relocate_sent, relocate_age) =
+        handle_online_command(&peer, &sk_set, &stage, elders_info).await?;
+    assert!(event_rx.try_recv().is_err());
+    // After startup, a rejoin node with low age will be rejected.
+    if let NetworkPhase::Regular = phase {
+        if age / 2 <= MIN_AGE {
+            return Ok(());
```
madadam, line 437: Assert that both `node_approval_sent` and `relocate_sent` are false.

src/routing/tests/mod.rs
```diff
@@ -390,6 +340,136 @@ async fn handle_consensus_on_online_of_elder_candidate() -> Result<()> {
...
+    // Simulate peer with the same name is rejoin and verify resulted behaviours.
+    let (node_approval_sent, relocate_sent, relocate_age) =
+        handle_online_command(&peer, &sk_set, &stage, elders_info).await?;
+    assert!(event_rx.try_recv().is_err());
+    // After startup, a rejoin node with low age will be rejected.
+    if let NetworkPhase::Regular = phase {
+        if age / 2 <= MIN_AGE {
+            assert!(!node_approval_sent);
+            assert!(!relocate_sent);
+            assert!(relocate_age == 0);
```
madadam, line 441: Sorry for nitpicking, but this looks a bit weird. How about instead of having separate `relocate_sent` and `relocate_age`, have only `relocate_age: Option<u8>` where `Some` means the `Relocate` message was sent with the given age and `None` means no `Relocate` message was sent?

madadam (approved):

PR #2233, opened 2020-11-18
Indication for section start-up and Age getter

Yoga07:


src/routing/approved.rs
```diff
@@ -1449,6 +1452,10 @@ impl Approved {
             self.send_event(Event::Demoted);
         }
 
+        if self.node.age > MIN_AGE && !self.is_elder() {
```
maqi, line 1455: just use `new_is_elder ` instead of `self.is_elder()` ?

madadam, line 1455: Now this is going to be called every time there is a section change, not just the first time the node goes from `MIN_AGE` to `> MIN_AGE`. That's probably not what we want, is it?

src/event.rs
```diff
@@ -51,6 +51,8 @@ pub enum Event {
         previous_name: Option<XorName>,
         /// Age of the node
         age: u8,
+        /// Indication that is has been relocated during startup.
+        startup_relocation: bool,
```
madadam, line 55: Out of curiosity, but why do we need this?

Yoga07: To differentiate between natural relocation and startup phase induced relocation.

src/routing/approved.rs
Yoga07: Yea, that and `update_state` function triggers only when an event comes by, so yea this might not be the right place to rise this event. And we need to fire it when we get `Demoted` from being an Elder too. Node should become an adult when demoted from an Elder. So sn_node should receive the `PromotedToAdult` event in both cases.

src/routing/approved.rs
```diff
@@ -1201,6 +1203,7 @@ impl Approved {
                 name: *peer.name(),
                 previous_name,
                 age: peer.age(),
+                startup_relocation: false,
```
maqi, line 1206: better use self.is_in_startup_phase() to populate this field ?

maqi (approved):

src/event.rs
madadam: I get that, but why do we need to differentiate it?

madadam (approved):

PR #2234, opened 2020-11-19
feat: joins_allowed to toggle accept new node or not

maqi:
This contains the work to allow node to tell routing whether new node shall be accepted.
This is also being a solution to prevent sybil attack.

### DIFF
--- a/src/consensus/vote.rs
+++ b/src/consensus/vote.rs
@@ -58,6 +58,9 @@ pub(crate) enum Vote {
         message: Box<PlainMessage>,
         proof_chain: SectionProofChain,
     },
+
+    // Voted to concensus whether new node shall be allowed to join
+    JoinsAllowed(bool),
 }
 
 impl Vote {
@@ -95,6 +98,7 @@ impl<'a> Serialize for SignableView<'a> {
             Vote::TheirKey { prefix, key } => (prefix, key).serialize(serializer),
             Vote::TheirKnowledge { prefix, key_index } => (prefix, key_index).serialize(serializer),
             Vote::SendMessage { message, .. } => message.as_signable().serialize(serializer),
+            Vote::JoinsAllowed(joins_allowed) => joins_allowed.serialize(serializer),
         }
     }
 }
--- a/src/routing/approved.rs
+++ b/src/routing/approved.rs
@@ -1731,10 +1731,37 @@ impl Approved {
         Ok(Some(command))
     }
 
+    // Setting the JoinsAllowed has two effectives:
+    // 1. Trigger a round Vote::SetJoinsAllowed to update the flag once concensused. Which will
+    //    allow/disallow new nodes joining once done.
+    // 2, Trigger one infant to be relocated to the same section with age increased by 1. This makes
+    //    it an adult once join back. Using Relocation (not just increase age) to be consistent with
+    //    other process and avoid potential issue.
     pub fn set_joins_allowed(&mut self, joins_allowed: bool) -> Result<Vec<Command>> {
         let mut commands = Vec::new();
         if self.is_elder() && joins_allowed != self.joins_allowed {
             commands.extend(self.vote(Vote::JoinsAllowed(joins_allowed))?);
+
+            // Only trigger infant relocation when update flag from disallow to allow.
+            if joins_allowed {
+                // Members already stored within BTreeMap, so the result is ordered.
+                if let Some(infant) = self
+                    .section
+                    .members()
+                    .joined()
+                    .find(|info| info.peer.age() == MIN_AGE)
+                {
+                    commands
+                        .extend(self.vote(Vote::Offline(infant.relocate(*infant.peer.name())))?);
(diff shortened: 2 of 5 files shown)


src/consensus/vote.rs
```diff
@@ -58,6 +58,9 @@ pub(crate) enum Vote {
         message: Box<PlainMessage>,
         proof_chain: SectionProofChain,
     },
+
+    // Voted to concensus whether new node shall be allowed to join
+    JoinFlag(bool),
```
madadam, line 63: I don't think we need to vote for this. Just setting the join flag individually by each node should be sufficient, no?

src/routing/approved.rs
```diff
@@ -55,6 +55,7 @@ pub(crate) struct Approved {
     relocate_state: Option<RelocateState>,
     msg_filter: MessageFilter,
     pub(super) event_tx: mpsc::UnboundedSender<Event>,
+    join_flag: bool,
```
madadam, line 58: Maybe call it `joins_allowed`?

src/routing/approved.rs
```diff
@@ -895,6 +901,14 @@ impl Approved {
...
     fn handle_join_request(&self, peer: Peer, join_request: JoinRequest) -> Result<Vec<Command>> {
         debug!("Received {:?} from {}", join_request, peer);
+        let pub_id = *peer.name();
+        if !self.join_flag {
+            debug!(
+                "Ignoring JoinRequest from {} - new node not acceptable.",
+                pub_id,
+            );
+            return Ok(vec![]);
+        }
```
madadam, line 911: Now we are ignoring even relocated nodes, which probably isn't what we want.

src/routing/approved.rs
```diff
@@ -1451,6 +1465,9 @@ impl Approved {
 
         if !new_is_elder {
             commands.extend(self.return_relocate_promise());
+        } else if &old_elders_info != self.section.elders_info() {
+            // Whenever there is an elders change, casting a round of join_flag vote to sync.
+            commands.extend(self.vote(Vote::JoinFlag(self.join_flag))?);
```
madadam, line 1470: So it seems to me that either the join flag should not be voted for at all or if it is, it should be part of the `Section` state in which case it would get synced via the `Sync` message already. In either case, we probalby don't need this.

src/routing/mod.rs
```diff
@@ -258,8 +258,14 @@ impl Routing {
         src: SrcLocation,
         dst: DstLocation,
         content: Bytes,
+        join_flag: bool,
```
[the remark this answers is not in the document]
maqi: Currently binding it to send_message just to ensure the vault status will be periodically checked, so that the flag can be updated in time.

If a separate setter is neater, I will make the change.

src/consensus/vote.rs
maqi: This is to reduce the chance of accumulating mis-flag and causing issue.
The vote is also used to notify a newly promoted elder of what flag value shall be.

src/routing/approved.rs
maqi: yeah, sounds better.
will change to it if no other suggestions.

src/routing/approved.rs
maqi: right, will change it so that relocated node will always be accepted.

src/routing/approved.rs
maqi: maybe should be part of the section. I will think about it.

src/routing/mod.rs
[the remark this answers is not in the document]
madadam: I think a separate setter is cleaner. If sn-node wants to set it always when they send a message, they can still do it, but I think they will want to only set it when they detect change in the section storage capacity or things like that. Bundling it together with message sending seems confusing and not that useful.

src/consensus/vote.rs
madadam: We should probably discuss this with the sn-node team to understand how exactly will this be used, but my impression is that the elders will track all the adults in the section and when they detect the average storage capacity (or some other resource) is too low, they will flip the flag to true so the section starts accepting new nodes. All the elders should detect this more or less at the same time, so eventually all the elders will flip the flag equally. But even if there is inconsistency, the majority value will decide what happens. So if the majority has it set to `true`, then the section will accept new nodes, otherwise they won't. Seems good enough already. I'd just like to keep things simple if possible.

dirvine: This is correct. The issue in routing is adding a node will mean Infants (joining queue). As infants don't add to sn_node resources then we need to do more here. 
1 suggestion is when this Flag is set we increment Infants age + promote them to Adults? (promote here is probably OK and no need to change the name as Infants will not be held in Elders or Adult containers (we do need separate containers though, but that's more obvious))

Yoga07 (approved):

madadam (approved):

PR #2235, opened 2020-11-19
feat: notify when key got changed during relocation

maqi:


src/event.rs
```diff
@@ -78,6 +79,8 @@ pub enum Event {
     Relocated {
         /// Old name before the relocation.
         previous_name: XorName,
+        /// Serialised new keypart. Use Keypair::from_bytes to re-construct.
```
madadam, line 82: Why is it serialized?

maqi: cloning will be problematic espically out of the crate ?

madadam: Pass it as `Arc<Keypair>` (it's how it's kept in `Node` anyway). By passing the bytes you are basically bypassing the whole security thing where we are trying to avoid having multiple copies of the key in memory.

madadam (approved):

PR #2238, opened 2020-11-30
Refactor section update

madadam:
This PR refactors the DKG and its outcome handling in order to simplify the codebase:

- ~~DKG failures are no longer reported to the current elders. If a DKG participant detects a failure, it automatically restarts the DKG~~
- DKG failures are still reported, but differently - they are now accumulated by the DKG participants themselves and only when they accumulate a majority are they sent to the current elders. The elders validate it and restart the DKG if necessary. This makes things simpler because we don't need to track potentially multiple DKG sessions or deal with the invalidation of the accumulators. It also offloads the work to the nodes that benefit from it, which seems neater.
- The separate DKG result accumulator is gone. We now use the regular vote accumulator.
- So there is no more special state for "DKG observer".
- Section elders and section key are now naturally updated at the same time, without requiring a separate synchronization
- So the `UpdateBarrier` is now needed only during splits and thus got simplified and renamed to `SplitBarrier`

### DIFF
--- a/src/consensus/dkg.rs
+++ b/src/consensus/dkg.rs
@@ -58,6 +58,9 @@ impl DkgKey {
         Self(output)
     }
 
+    // Create a new `DkgKey` to be used to retry a previously failed DKG session. This key is
+    // deterministically derived from `self`, so different DKG participants that fail at different
+    // times end up using the same retried key.
     pub fn retry(&self) -> Self {
         use tiny_keccak::{Hasher, Sha3};
 
@@ -166,6 +169,8 @@ impl DkgVoter {
         }
     }
 
+    // Try to initialize a DKG session. If it succeeds, returns a list of `DkgCommand`s to apply.
+    // On failure gives back `elders_info` so it can be used again to retry, avoiding a clone.
     fn try_initialize(
         &mut self,
         our_name: XorName,
--- a/src/consensus/dkg.rs
+++ b/src/consensus/dkg.rs
@@ -7,7 +7,7 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::{
-    crypto::Digest256,
+    crypto::{self, Digest256, Keypair, PublicKey, Signature, Verifier},
     error::Result,
     majority,
     messages::{Message, Variant},
@@ -24,10 +24,11 @@ use serde::{Deserialize, Serialize};
 use std::{
     collections::VecDeque,
     fmt::{self, Debug, Formatter},
+    iter, mem,
     net::SocketAddr,
     time::Duration,
 };
-use xor_name::XorName;
+use tiny_keccak::{Hasher, Sha3};
 
 // Interval to progress DKG timed phase
 const DKG_PROGRESS_INTERVAL: Duration = Duration::from_secs(30);
@@ -40,8 +41,6 @@ pub struct DkgKey(Digest256);
 
 impl DkgKey {
     pub fn new(elders_info: &EldersInfo) -> Self {
-        use tiny_keccak::{Hasher, Sha3};
-
         // Calculate the hash without involving serialization to avoid having to return `Result`.
         let mut hasher = Sha3::v256();
         let mut output = Digest256::default();
@@ -53,21 +52,6 @@ impl DkgKey {
 
         hasher.update(&elders_info.prefix.name().0);
         hasher.update(&elders_info.prefix.bit_count().to_le_bytes());
-        hasher.finalize(&mut output);
-
-        Self(output)
-    }
-
-    // Create a new `DkgKey` to be used to retry a previously failed DKG session. This key is
-    // deterministically derived from `self`, so different DKG participants that fail at different
-    // times end up using the same retried key.
-    pub fn retry(&self) -> Self {
-        use tiny_keccak::{Hasher, Sha3};
-
-        let mut hasher = Sha3::v256();
-        let mut output = Digest256::default();
-
-        hasher.update(&self.0);
         hasher.finalize(&mut output);
 
         Self(output)
@@ -91,7 +75,7 @@ impl Debug for DkgKey {
 /// 3. When the `DKGStart` message accumulates, the participants call `start`.
 /// 4. The participants keep exchanging the DKG messages and calling `process_message`.
 /// 5. On DKG completion, the participants send `DKGResult` vote to the current elders (observers)
-/// 6. When the observers accumulate the votesm the can proceed with voting for the section update.
+/// 6. When the observers accumulate the votes, they can proceed with voting for the section update.
 ///
 /// Note: in case of heavy churn, it can happen that more than one DKG session completes
 /// successfully. Some kind of disambiguation strategy needs to be employed in that case, but that
@@ -118,23 +102,24 @@ impl DkgVoter {
     // Starts a new DKG session.
     pub fn start(
         &mut self,
-        our_name: XorName,
+        keypair: &Keypair,
         dkg_key: DkgKey,
         elders_info: EldersInfo,
     ) -> Vec<DkgCommand> {
         if let Some(session) = &self.session {
-            if session.dkg_key == dkg_key && session.elders_info.is_some() {
+            if session.dkg_key == dkg_key && !session.complete {
                 trace!("DKG for {} already in progress", elders_info);
                 return vec![];
             }
         }
 
-        let index = if let Some(index) = elders_info.position(&our_name) {
+        let name = crypto::name(&keypair.public);
+        let index = if let Some(index) = elders_info.position(&name) {
             index
         } else {
             error!(
                 "DKG for {} failed to start: {} is not a participant",
-                elders_info, our_name
+                elders_info, name
             );
             return vec![];
         };
@@ -153,31 +138,6 @@ impl DkgVoter {
             }];
         }
 
-        // Keep trying to initialize until we succeed. This should always terminate in a finite
-        // number of iterations (usually small).
-        let mut next_dkg_key = dkg_key;
-        let mut next_elders_info = elders_info;
-
-        loop {
-            match self.try_initialize(our_name, next_dkg_key, next_elders_info, index) {
-                Ok(commands) => break commands,
-                Err(elders_info) => {
-                    next_dkg_key = next_dkg_key.retry();
-                    next_elders_info = elders_info;
-                }
-            }
-        }
-    }
-
-    // Try to initialize a DKG session. If it succeeds, returns a list of `DkgCommand`s to apply.
-    // On failure gives back `elders_info` so it can be used again to retry, avoiding a clone.
-    fn try_initialize(
-        &mut self,
-        our_name: XorName,
-        dkg_key: DkgKey,
-        elders_info: EldersInfo,
-        index: usize,
-    ) -> Result<Vec<DkgCommand>, EldersInfo> {
         let threshold = majority(elders_info.elders.len()) - 1;
         let participants = elders_info
             .elders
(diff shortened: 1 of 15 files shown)


src/consensus/dkg.rs
```diff
@@ -52,8 +53,18 @@ impl DkgKey {
 
         hasher.update(&elders_info.prefix.name().0);
         hasher.update(&elders_info.prefix.bit_count().to_le_bytes());
+        hasher.finalize(&mut output);
+
+        Self(output)
+    }
 
+    pub fn retry(&self) -> Self {
```
maqi, line 61: I am not quite sure about this part.
The DkgKey of a voting session is supposed to remain same across the nodes.
However, this `retry` stuff actually makes different version of DkgKey, depends on how many times of failure that a paticipant experienced during initialize stage.
I don't get the point why it is needed, and how it works.

src/consensus/dkg.rs
```diff
@@ -73,69 +84,95 @@ impl Debug for DkgKey {
...
+        // number of iterations (usually small).
+        let mut next_dkg_key = dkg_key;
+        let mut next_elders_info = elders_info;
+
+        loop {
+            match self.try_initialize(our_name, next_dkg_key, next_elders_info, index) {
+                Ok(commands) => break commands,
+                Err(elders_info) => {
+                    next_dkg_key = next_dkg_key.retry();
+                    next_elders_info = elders_info;
```
maqi, line 163: this is bit weird. the function is asking for a DkgGen supposed about the input `elders_info`,
but here the DkgGen will be silently started about the returned `error` elders_info.

Also, I think the  `next_dkg_key.retry()` shall not be used as the dkg_key, but shall be a key deduced from the error returned elders_info ?

Also, also, if we do decide the KeyGen shall be about the error returned elders_info, then we shall at least return this back to upper caller to make it be notified ?

src/consensus/dkg.rs
```diff
@@ -228,37 +266,32 @@ impl DkgVoter {
...
 
-                self.participant = None;
-
-                vec![DkgCommand::HandleParticipationResult {
-                    dkg_key,
-                    elders_info,
-                    result: Err(()),
-                }]
+                // Restart on failure.
+                self.start(our_name, dkg_key.retry(), elders_info)
```
maqi, line 281: could cause different node using different version of dkg_key (of the same eldres_info),
depends on the different times of time_out experienced ?

src/routing/approved.rs
```diff
@@ -596,10 +553,22 @@ impl Approved {
...
-    fn should_handle_vote(&self, proof_share: &ProofShare) -> bool {
-        self.section
-            .chain()
-            .has_key(&proof_share.public_key_set.public_key())
+    fn should_handle_vote(&self, vote: &Vote, proof_share: &ProofShare) -> bool {
+        // We always handle a `SectionInfo` vote for our section because it is signed by the new
+        // key produced by the last DKG and it is only about to be inserted into the chain.
+        // Any other vote must be signed by a known key.
+        match vote {
+            Vote::SectionInfo(elders_info)
```
maqi, line 561: will this open a door that any group of adults (or even just unkowns) can send elder a Vote::SectionInfo to claim a split ?

src/routing/approved.rs
```diff
@@ -1261,44 +1219,48 @@ impl Approved {
...
-        } else {
-            // Other section
-            if self.network.update_neighbour_info(elders_info) {
-                self.network.prune_neighbours(self.section.prefix());
+        if prefix_is_equal || prefix_is_extension {
+            // Our section
+            if self
+                .section
+                .promote_and_demote_elders(&self.node.name())
+                .contains(&elders_info.value)
```
maqi, line 1234: As the Vote::SectionInfo could be sent from a non-elder, and the vote_accumulator only carries out aggregate,
may need to check whether the claimed elders_info is the real ones carried out the votes ?

src/routing/approved.rs
```diff
@@ -1653,56 +1575,19 @@ impl Approved {
...
-        trace!("Send DKGStart for {}", new_elders_info);
+    fn send_dkg_start(&mut self, elders_info: EldersInfo) -> Result<Vec<Command>> {
+        trace!("Send DKGStart for {}", elders_info);
 
-        let dkg_key = DkgKey::new(&new_elders_info);
+        let dkg_key = DkgKey::new(&elders_info);
 
         // Send to all participants.
-        let recipients: Vec<_> = new_elders_info.elders.values().copied().collect();
+        let recipients: Vec<_> = elders_info.elders.values().copied().collect();
```
maqi, line 1584: this elders_info could be different to the real KenGen initialized internally, due to the retry attempt.

src/messages/variant.rs
```diff
@@ -89,13 +89,6 @@ pub(crate) enum Variant {
         /// The DKG message.
         message: DkgMessage,
     },
-    /// Message to notify current elders about the DKG result.
-    DKGResult {
```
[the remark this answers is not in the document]
madadam: Yes, this still works the same, the only difference is that now the DKG result is collected as a normal vote (`SectionInfo`) and does not require a separate accumulator.

src/section/mod.rs
```diff
@@ -61,8 +61,6 @@ impl Section {
         let public_key_set = secret_key_set.public_keys();
         let secret_key_share = secret_key_set.secret_key_share(0);
 
-        // Note: `ElderInfo` is normally signed with the previous key, but as we are the first node
-        // of the network there is no previous key. Sign with the current key instead.
```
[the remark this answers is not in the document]
madadam: Because it's not true anymore. The elders info is now always signed with the latest key. This makes things simpler I think.

src/section/mod.rs
```diff
@@ -89,30 +87,24 @@ impl Section {
...
+            &self.chain,
+        ) {
+            Ordering::Less => {
                 self.elders_info = other.elders_info;
             }
-            Some(Ordering::Equal)
-                if self.elders_info.value.elders.len() < other.elders_info.value.elders.len() =>
-            {
-                // Note: our `EldersInfo` is normally signed with the previous key, except the very
-                // first one which is signed with the latest key. This means that the first and
```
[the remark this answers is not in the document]
madadam: First elders info is signed with the first key, second elders info with the second key, and so on.

src/routing/approved.rs
madadam: No, because we check whether the elders_info corresponds to the current elder candidates (the `ELDER_SIZE` oldest members) before actioning the vote (in `handle_section_info_event`).

src/routing/approved.rs
madadam: Do you mean to check whether a node that votes a `SectionInfo` is among the proposed elders in it? Probably a good idea.

src/consensus/dkg.rs
madadam: If a DKG fails, then the DKG participant doesn't necessarily have to notify the elders about it, because all they would do is to vote to restart it. So we can simplify this and the DKG participant can already restart it himself. However, to prevent the messages of the new restarted DKG to mix with the old DKG, we use a unique DKG key. This new key is deterministically derived from the old key, so all the participants will eventually use the same key. 

The participants will keep doing this until either the DKG succeeds or until a new DKG is started by the current elders (when the elder candidates change).

I should probably add a comment to this function explaining this.

src/consensus/dkg.rs
madadam: The elders info returned in the `Err` is the same one that was used to start the DKG. When a DKG fails and is restarted, it uses the same participants, but different DKG key (to avoid mixing the messages). Maybe this too should be commented?

src/consensus/dkg.rs
madadam: Yes, but that's on purpose to avoid the messages from different DKG session (the original one and the restarted one) to mix.

src/routing/approved.rs
madadam: A retry attempt doesn't change the elders info.

src/section/mod.rs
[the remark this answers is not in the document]
maqi: Yeah, I agree it will be simpler.
However, do you mean the new_elders_info is now signed by the new section key ?
if so, the section_keys are no longer forms a chain anymore ?

src/section/mod.rs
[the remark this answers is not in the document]
maqi: same doubt as above. no longer forms a chain?

src/routing/approved.rs
maqi: This is related to the below question, 
the `sender` of the vote is not checked.

src/routing/approved.rs
maqi: yes, that's what I mean. anyone can vote.

src/section/mod.rs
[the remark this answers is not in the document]
madadam: They still form a chain, it's just the elders info is now signed by the last key, not the one-before last as before.

src/section/mod.rs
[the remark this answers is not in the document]
madadam: No, the chain is still the same as it was. It's just the elders info (which isn't part of the chain) is now signed with different key.

src/routing/approved.rs
madadam: Got it. It's addressed now.

src/consensus/dkg.rs
maqi: > Maybe this too should be commented?

yes. better to make it clear.

src/consensus/dkg.rs
maqi: >I should probably add a comment to this function explaining this.

yes. will be good to do so

src/section/mod.rs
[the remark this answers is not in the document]
maqi: then, yes, this links my other question: will the signers to be verified ?
if not, then anyone can carry out this now.

src/consensus/dkg.rs
maqi: I am concerned this may cause a deadlock, say some is always at version `n`, meanwhile some alwasy at version `n+1`, and they keeps growing with the diff of `1` always exists.

src/routing/approved.rs
maqi: ok. so the returned elders_info within the Error is the same one of the input.
got it.

src/consensus/dkg.rs
madadam: Hm, I guess that's not impossible. So say a node A detects a failure and switches to version `n+1` but the other nodes are taking longer to detect the failure themselves, so node A fails again and switches to `n+2` and the other nodes switch to `n+1` only after that, this never catching up.

There is also another issue I realized yesterday, although it's not as serious. Say we start DKG with the participants A, B, C, D, E. Then some time later, but before the first DKG completes, we start another DKG with A, B, C, F, G. Then D and E will still think they are on the first DKG and will keep trying to progress it, fail, restart and so on, basically forever.

OK, I'll try to think about how to resolve this. Will probably have to revert part of this PR.

src/routing/approved.rs
```diff
@@ -450,19 +426,19 @@ impl Approved {
...
-                if !self.should_handle_vote(proof_share) {
-                    // Message will be bounced if we are lagging (not known of the signing key).
-                    return Ok(MessageStatus::Unknown);
+            Variant::Vote {
+                content,
+                proof_share,
+                ..
+            } => {
+                if let Some(status) =
+                    self.decide_vote_status(&msg.src().to_node_name()?, content, proof_share)
```
maqi, line 439: this is bit strange that a returned `Some` is actually a `useless/unknown` message.

src/consensus/dkg.rs
```diff
@@ -473,12 +412,78 @@ impl Participant {
...
+        {
+            self.0.push(proof);
+            true
+        } else {
+            false
+        }
+    }
+
+    fn has_agreement(&self, elders_info: &EldersInfo) -> bool {
+        self.0.len() >= majority(elders_info.elders.len())
```
maqi, line 453: according to the below `verify` function, it seems the failures could be for different elders_info`,
so, shall we here only filter out the failure regarding this elders_info ?

src/routing/approved.rs
```diff
@@ -1090,34 +1104,74 @@ impl Approved {
...
+            );
+            return Ok(vec![]);
+        }
+
+        trace!(
+            "Received DKG failure agreement - restarting: {}",
+            elders_info
+        );
+
+        self.send_dkg_start_to(elders_info, slice::from_ref(sender))
```
maqi, line 1172: shall the start be sent to ALL elders_info ?
otherwise the chance of `out-of-snyc` could got increased?

also, the log above says `restarting elders_info` , however actually we just restart one of it.

src/consensus/vote.rs
```diff
@@ -33,13 +33,10 @@ pub(crate) enum Vote {
...
-    // Voted to update our section key.
-    OurKey {
-        // In case of split, this prefix is used to differentiate the subsections. It's not part of
-        // the proof and thus not serialised when signing.
-        prefix: Prefix,
-        key: bls::PublicKey,
-    },
+    // Voted to update the elders in our section.
+    // NOTE: the `EldersInfo` is already signed with the new key. All that's left to do is to sign
+    // the new key using our current key, which is what this vote is for.
```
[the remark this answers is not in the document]
madadam: No, what's signed is only the key, but the vote contains the other stuff so that when it reaches consensus, we can update the new elders info and the new key at the same time. I'll try to update the comment to make is more clear.

src/crypto.rs
```diff
@@ -59,3 +59,22 @@ pub fn gen_keypair_within_range(range: &RangeInclusive<XorName>) -> Keypair {
...
+    use super::*;
+    use ed25519_dalek::SECRET_KEY_LENGTH;
+    use proptest::prelude::*;
+
+    pub(crate) fn arbitrary_keypair() -> impl Strategy<Value = Keypair> {
+        any::<[u8; SECRET_KEY_LENGTH]>().prop_map(|bytes| {
+            // OK to unwrap because `from_bytes` returns error only if the input slice has incorrect
+            // length. But here we only generate arrays of size `SECRET_KEY_LENGTH` which is the
+            // correct one.
+            let secret = SecretKey::from_bytes(&bytes[..]).unwrap();
```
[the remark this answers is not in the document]
madadam: Because the idea is to generate a `Keypair` out of arbitrary sequence of bytes, so that we can use it for proptest. `SecretKey::from_bytes` works with any input as long as it has the correct length, but `Keypair::from_bytes` does not. See also https://docs.rs/ed25519-dalek/1.0.1/ed25519_dalek/struct.Keypair.html#warning.

src/routing/approved.rs
madadam: The idea is that when this returns `Some` then that's the status to be used, but when it returns `None` that means the status wasn't decided and we need to fall back to the default status processing. What would you suggest to make this more clear?

src/consensus/dkg.rs
madadam: This function assumes the proofs are already valid. Will add a comment clarifying this.

src/routing/approved.rs
madadam: It should be enough to send the DKGStart only to those nodes that observed the failure, because eventually all of the participants should observe it and will send the failure agreement too. Sending to all of them would open us up for spam attack, because we would have a situation where a single message triggers multiple messages.

src/routing/approved.rs
maqi: I'd prefer to return explicit `MessageStatus::...` below, to be consistent with other branches.
though, not a big preference, up to you.

maqi (approved):

PR #2242, opened 2020-12-02
feat: carry out resource proofing during bootstrap

maqi:
This PR contains the work of using resource_proofing to challenge the joining node so that any joined node can now be considered as an adult immediately.

### DIFF
--- a/src/messages/variant.rs
+++ b/src/messages/variant.rs
@@ -19,6 +19,7 @@ use bytes::Bytes;
 use hex_fmt::HexFmt;
 use serde::{Deserialize, Serialize};
 use std::{
+    collections::VecDeque,
     fmt::{self, Debug, Formatter},
     net::SocketAddr,
 };
@@ -215,6 +216,12 @@ pub enum BootstrapResponse {
     /// The new peer should retry bootstrapping with another section. The set of connection infos
     /// of the members of that section is provided.
     Rebootstrap(Vec<SocketAddr>),
+    /// Expecting the new peer to carry out a resouce proofing.
+    ResourceProof {
+        data_size: usize,
+        difficulty: u8,
+        nonce: [u8; 32],
+    },
 }
 
 /// Request to join a section
@@ -224,6 +231,8 @@ pub(crate) struct JoinRequest {
     pub section_key: bls::PublicKey,
     /// If the peer is being relocated, contains `RelocatePayload`. Otherwise contains `None`.
     pub relocate_payload: Option<RelocatePayload>,
+    /// Proof of the resouce proofing
+    pub resource_proof: Option<(u64, VecDeque<u8>)>,
 }
 
 impl Debug for JoinRequest {
@@ -237,6 +246,10 @@ impl Debug for JoinRequest {
                     .relocate_payload
                     .as_ref()
                     .map(|payload| payload.relocate_details()),
+            )
+            .field(
+                "resouce_proof",
+                &self.resource_proof.as_ref().map(|(solution, _)| solution),
             )
             .finish()
     }
--- a/src/messages/variant.rs
+++ b/src/messages/variant.rs
@@ -9,6 +9,7 @@
 use super::{Message, MessageHash, VerifyStatus};
 use crate::{
     consensus::{DkgKey, ProofShare, Proven, Vote},
+    crypto::Signature,
     error::{Error, Result},
     network::Network,
     relocation::{RelocateDetails, RelocatePayload, RelocatePromise},
@@ -101,6 +102,13 @@ pub(crate) enum Variant {
     Vote {
         content: Vote,
         proof_share: ProofShare,
+    },
+    /// Challenge sent from existing elder nodes to the joining peer for resource proofing.
+    Challenge {
+        data_size: usize,
+        difficulty: u8,
+        nonce: [u8; 32],
+        signature: Signature,
     },
 }
 
@@ -200,6 +208,15 @@ impl Debug for Variant {
                 .field("content", content)
                 .field("proof_share", proof_share)
                 .finish(),
+            Self::Challenge {
+                data_size,
+                difficulty,
+                ..
+            } => f
+                .debug_struct("Challenge")
+                .field("data_size", data_size)
+                .field("difficulty", difficulty)
+                .finish(),
         }
     }
 }
@@ -216,12 +233,15 @@ pub enum BootstrapResponse {
     /// The new peer should retry bootstrapping with another section. The set of connection infos
     /// of the members of that section is provided.
     Rebootstrap(Vec<SocketAddr>),
-    /// Expecting the new peer to carry out a resouce proofing.
-    ResourceProof {
-        data_size: usize,
-        difficulty: u8,
-        nonce: [u8; 32],
-    },
+}
+
+/// Joining peer's proof of resolvement of given resource proofing challenge.
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub(crate) struct Proof {
+    pub(crate) solution: u64,
+    pub(crate) data: VecDeque<u8>,
+    pub(crate) nonce: [u8; 32],
+    pub(crate) signature: Signature,
 }
 
 /// Request to join a section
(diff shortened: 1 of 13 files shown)


src/routing/bootstrap.rs
```diff
@@ -21,18 +21,19 @@ use crate::{
...
 const BACKLOG_CAPACITY: usize = 100;
 
-/// Bootstrap into the network as an infant node.
+/// Bootstrap into the network as an adult node.
 ///
 /// NOTE: It's not guaranteed this function ever returns. This can happen due to messages being
 /// lost in transit or other reasons. It's the responsibility of the caller to handle this case,
 /// for example by using a timeout.
-pub(crate) async fn infant(
+pub(crate) async fn adult(
```
madadam, line 36: The name of this function is confusing now, because adult can be both joining for the first time or being relocated. This name should reflect that this is a bootstrap as a new node as opposed to bootstrap as relocated node. Some suggestions: `first`, `new` (or `new_node`), `initial`, ... something like that.

src/messages/variant.rs
```diff
@@ -224,6 +231,8 @@ pub(crate) struct JoinRequest {
     pub section_key: bls::PublicKey,
     /// If the peer is being relocated, contains `RelocatePayload`. Otherwise contains `None`.
     pub relocate_payload: Option<RelocatePayload>,
+    /// Proof of the resouce proofing
+    pub resource_proof: Option<(u64, VecDeque<u8>)>,
```
madadam, line 235: Why is this a `VecDeque` and not a simple `Vec`?

madadam: Does it make sense for a `JoinRequest` to contain both relocate payload and resource proof? I though that relocated nodes don't have to be proven again, do they? But this type now suggests that such state is valid. To avoid confusion and to make it more robust, I suggest replacing the `relocate_payload` and `resource_proof` fields with a single enum field with three possible variants: "nothing", "relocate_payload" and "resource_proof". Let's say:
```rust
enum JoinDetails {
    None,
    RelocatePayload(RelocatePayload),
    ResourceProof(u64, VecDeque<u8>),
}
```
Alternatively, the resource proof can be a separate message variant.

src/routing/approved.rs
```diff
@@ -56,6 +62,8 @@ pub(crate) struct Approved {
     msg_filter: MessageFilter,
     pub(super) event_tx: mpsc::UnboundedSender<Event>,
     joins_allowed: bool,
+    resource_proof: ResourceProof,
+    nonces: LruCache<XorName, [u8; 32]>,
```
madadam, line 66: Can we avoid this LRU cache? How about we make it so the joining node needs to send the nonce back to us together with the solution? To avoid tampering, we could sign the nonce with the ed25519 key.

src/routing/approved.rs
```diff
@@ -1005,14 +1019,33 @@ impl Approved {
...
+                        previous_name,
+                        their_knowledge,
+                    });
+                }
+            }
+        }
+
+        let nonce: [u8; 32] = rand::random();
+        let _ = self.nonces.insert(*peer.name(), nonce);
+        let response = BootstrapResponse::ResourceProof {
```
madadam, line 1040: Two (minor) issues with this:
1. I don't think this should be a `BootstrapResponse` variant, because we use bootstrap request/response just to find the section to join. I'd make it a separate message variant.
2. The name "resource poof" is overloaded in a confusing way. It sometimes means the request/challenge and sometimes the response/solution. We should use distinct names for these two concepts.

src/routing/bootstrap.rs
```diff
@@ -229,16 +236,28 @@ impl<'a> State<'a> {
 
     // Send `JoinRequest` and wait for the response. If the response is `Rejoin`, repeat with the
     // new info. If it is `Approval`, returns the initial `Section` value to use by this node,
-    // completing the bootstrap.
+    // completing the bootstrap. If it is `ResourceProff`, carries out a resource proof calculation.
```
madadam, line 239: typo: `ResourceProff` -> `ResourceProof`

src/routing/approved.rs
```diff
@@ -37,12 +37,18 @@ use crate::{
...
 use bytes::Bytes;
 use itertools::Itertools;
+use lru_time_cache::LruCache;
+use resource_proof::ResourceProof;
 use std::{cmp, net::SocketAddr, slice};
 use tokio::sync::mpsc;
 use xor_name::{Prefix, XorName};
 
+pub(crate) const RESOURCE_PROOF_DATA_SIZE: usize = 256;
+pub(crate) const RESOURCE_PROOF_DIFF: u8 = 8;
```
madadam, line 47: I would not abbreviate "difficulty" to "diff". It's confusing as "diff" typically means "difference".

madadam (commented):
Some better naming suggestions + I don't think we need the LRU cache. Otherwise nice job.

src/messages/variant.rs
maqi: When you got relocated to a new section, that section may need to you to carry out a new round of resource-proofing?
also, the resouce-proofing could be dynamic eventually (i.e. different time/stage may allow nodes to join with different acceptable performance).
So, a re-resour_proofing for relocated node may be necessary ?

src/routing/approved.rs
maqi: > To avoid tampering, we could sign the nonce with the ed25519 key.

Yeah, good point. will change it.

src/routing/approved.rs
maqi: > I'd make it a separate message variant.
> We should use distinct names for these two concepts.

How about `Variant::Challenge` for request/challenge . and `proof` for response/solution ?

src/messages/variant.rs
madadam: OK, if you think resource proofing a relocated node is something we might need, then having two separate fields for it makes sense.

src/routing/approved.rs
madadam: Agree, but I'd make it `Variant::ResourceChallenge` to be more explicit.

src/routing/approved.rs
```diff
@@ -1732,36 +1737,12 @@ impl Approved {
     }
 
     // Setting the JoinsAllowed has two effectives:
```
madadam, line 1739: Seems like there is only one now?

src/routing/bootstrap.rs
```diff
@@ -21,18 +21,19 @@ use crate::{
...
 use futures::future;
+use resource_proof::ResourceProof;
 use std::{collections::VecDeque, mem, net::SocketAddr};
 use tokio::sync::mpsc;
 use xor_name::Prefix;
 
 const BACKLOG_CAPACITY: usize = 100;
 
-/// Bootstrap into the network as an infant node.
+/// Bootstrap into the network as an adult node.
```
madadam, line 31: Maybe "Bootstrap into the network for the first time" or "...as new node" or something like that. Again, it's not important that we are adult, but that we are not being relocated.

src/routing/tests/mod.rs
```diff
@@ -102,6 +110,47 @@ async fn receive_join_request() -> Result<()> {
...
+        commands.next(),
+        Some(Command::SendMessage { message, .. }) => message
+    );
+    let response_message = Message::from_bytes(&response_message)?;
+
+    let (nonce, signature) = assert_matches!(
+        response_message.variant(),
+        Variant::Challenge { nonce, signature, .. } => (*nonce, *signature)
+    );
+
```
madadam, line 123: I'd prefer to stop this test here and to add a separate test for the following logic. So one test for receiving join request without resource proof and another one for with resource proof. Would keep the pattern where we handle only one command per test which makes it simpler to read. 

We should also consider splitting this file into multiple ones, as it has become quite big. Can be done in a different PR though.

src/messages/variant.rs
```diff
@@ -217,13 +235,24 @@ pub enum BootstrapResponse {
     Rebootstrap(Vec<SocketAddr>),
 }
 
+/// Joining peer's proof of resolvement of given resource proofing challenge.
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub(crate) struct Proof {
```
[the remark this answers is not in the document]
maqi: yeah, will change to `ResourceProofResponse`

madadam (approved):

PR #2247, opened 2020-12-07
Do not require resource proof from relocated nodes + bugfixes

madadam:
- Do not require resource proof from relocated nodes
- Add test for handling join request from relocated node
- Fix incorrect test assertion that a newly joined node is never instantly relocated
- Forward `ResourceChallenge` to the bootstrap task (not necessary, but good future proofing)
- Refactor `handle_join_request` by extracting the resource proof logic into separate methods
- Fix a bug with section update being permanently blocked when there is only one elder and he wants to demote himself and promote another single node
- Fix a bug with promotion being blocked when all the current elders are being relocated and there is fewer than `ELDER_SIZE` other joined nodes in the section
- Fix the `test_startup_section_bootstrapping` test by taking relocations into account.

### DIFF
--- a/src/routing/approved.rs
+++ b/src/routing/approved.rs
@@ -1028,39 +1028,64 @@ impl Approved {
                 (MIN_AGE + 1, None, None)
             };
 
-        if let Some(ResourceProofResponse {
-            solution,
-            data,
-            nonce,
-            nonce_signature,
-        }) = join_request.resource_proof_response
-        {
-            let serialized = bincode::serialize(&(*peer.name(), nonce))?;
-            if self
-                .node
-                .keypair
-                .public
-                .verify(&serialized, &nonce_signature)
-                .is_ok()
-                && self.resource_proof.validate_all(&nonce, &data, solution)
-            {
-                return self.vote(Vote::Online {
-                    member_info: MemberInfo::joined(peer.with_age(age)),
-                    previous_name,
-                    their_knowledge,
-                });
+        // Require resource proof only if joining as a new node.
+        if previous_name.is_none() {
+            if let Some(response) = join_request.resource_proof_response {
+                if !self.validate_resource_proof_response(peer.name(), response) {
+                    debug!(
+                        "Ignoring JoinRequest from {} - invalid resource proof response",
+                        pub_id
+                    );
+                    return Ok(vec![]);
+                }
+            } else {
+                return Ok(vec![self.send_resource_proof_challenge(&peer)?]);
             }
         }
 
+        self.vote(Vote::Online {
+            member_info: MemberInfo::joined(peer.with_age(age)),
+            previous_name,
+            their_knowledge,
+        })
+    }
+
+    fn validate_resource_proof_response(
+        &self,
+        peer_name: &XorName,
+        response: ResourceProofResponse,
+    ) -> bool {
+        let serialized = if let Ok(serialized) = bincode::serialize(&(peer_name, &response.nonce)) {
+            serialized
+        } else {
+            return false;
+        };
+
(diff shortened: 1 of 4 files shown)


src/routing/approved.rs
```diff
@@ -1028,39 +1023,64 @@ impl Approved {
...
-                && self.resource_proof.validate_all(&nonce, &data, solution)
-            {
-                return self.vote(Vote::Online {
-                    member_info: MemberInfo::joined(peer.with_age(age)),
-                    previous_name,
-                    their_knowledge,
-                });
+        // Require resource proof only if joining as a new node.
+        if previous_name.is_none() {
+            if let Some(response) = join_request.resource_proof_response {
```
maqi, line 1028: seems we could use something like:
```rust
if let (true, Some(response)) = (previous_name.is_none(), join_request.resource_proof_response) {}
```
to avoid net if

madadam: Do you mean something like this:
```rust
if let (true, Some(response)) = (previous_name.is_none(), join_request.resource_proof_response) {
    self.validate_resource_proof_response(...)
} else {
    self.send_resource_proof_challenge(...)
}
```
?
Because that would send the challenge even to relocated nodes which we don't want.
```

Comment (maqi) [reply src/routing/approved.rs:1028]: Ah, messed up with my previous logic.
you are right, they cannot be bound.

File: src/section/mod.rs [hunk 2]
```diff
@@ -208,6 +208,12 @@ impl Section {
         } else if expected_names.len() < crate::majority(current_names.len()) {
             warn!("ignore attempt to reduce the number of elders too much");
             vec![]
+        } else if expected_names.len() <= 1 {
```
maqi, line 211: don't get this branch, this branch will only be reached when `current_names.len()` is 1 ?
which actually is handling a case not as described by the comment?

src/section/section_peers.rs
```diff
@@ -200,3 +199,32 @@ fn cmp_elder_candidates(
         })
         .then_with(|| lhs.proof.signature.cmp(&rhs.proof.signature))
 }
+
+// Compare candidates for the next elders according to their peer state. The one comparing `Less`
+// wins. `Joined` is preferred over `Relocated` which is preferred over `Left`.
+// NOTE: we only consider `Relocated` peers as elder candidates if we don't have enough `Joined`
+// members to reach `ELDER_SIZE`.
```
maqi, line 206: not sure about this.
I thought `only` Joined shall be considered as candidates.

src/section/mod.rs
madadam: This is to handle a pretty specific edge case: there is only one elder (typically this will be the genesis node) and it's being relocated and there is only one other node that is joined. So this elder demotes itself and promotes the other node. So we go from one elder to also just one elder, but a different one. There is nothing necessarily wrong about this, except it clashes with the way `UpdateBarrier` handles single elder sections. It seemed easier to simply disallow this transition than to try to fix `UpdateBarrier` itself, especially when this whole special handling of single elder section is going to be removed in #2238. 

However, I just realized that with the other change in this PR where we allow relocated current elders as elder candidates, this is no longer necessary because we would never go from A to B, but rather from A to A, B. Going to remove it.

src/section/section_peers.rs
madadam: Yes, that's the change. If there is not enough `Joined` members to create `ELDER_SIZE` elders, we fill them up with the current elders that are `Relocated`. That's OK, because such elders are still participating in the section until they are demoted and this will simply delay their relocation further.

maqi (approved):

PR #2248, opened 2020-12-08
Test node rejoin

maqi:
This PR also contains the work of :
rejoin node with low age will be relocated with new name and default age


maqi:
After further discussion, we decide that it will be better to leave the `rejoin node with low age` be silently rejected.
As a punishment of `less work for us, more work for them`.

Closing this PR as the existing unit tests shall be enough covering the current handling logic.

PR #2259, opened 2020-12-15
Use age assigned by section

madadam:


src/routing/approved.rs
```diff
@@ -1564,26 +1566,39 @@ impl Approved {
...
             self.section.prefix(),
-            peer
+            member_info.value.peer
         );
 
-        let their_knowledge = their_knowledge.and_then(|key| self.section.chain().index_of(&key));
-        let proof_chain = self
+        let addr = *member_info.value.peer.addr();
+
+        // Attach proof chain that includes the key the approved node knows (if any), the key its
```
maqi, line 1583: what does `the key its` mean ?

maqi (approved):

src/routing/approved.rs
madadam: it continues on the next line: "...the key its MembersInfo is signed with..."

PR #2266, opened 2020-12-29
api: returns section info matches target name

maqi:


src/network/mod.rs
```diff
@@ -170,6 +170,17 @@ impl Network {
         self.keys.get_matching(name).map(|entry| &entry.value.1)
     }
 
+    /// Returns the elders_info and the latest known key for the prefix that matches `name`,
+    /// excluding self section.
```
dirvine, line 174: If possible we should include self if it is the case?

maqi: This is inside network (which only holds the knowledge of neighbours), which is an `internal` field hold by routing.
the API function that routing provides to outside caller does has the self section included.

PR #2269, opened 2020-12-31
chore(tokio): upgrade tokio to v1.0.1

bochaco:


examples/stress.rs
```diff
@@ -126,7 +127,8 @@ async fn main() -> Result<()> {
     let mut churn_events = schedule.events();
 
     let probe_interval = Duration::from_secs_f64(1.0 / opts.probe_frequency);
-    let mut probes = time::interval(probe_interval);
+    let probes = time::interval(probe_interval);
+    let mut probe_stream = IntervalStream::new(probes);
```
madadam, line 131: We don't need this. Just use [`tick`](https://docs.rs/tokio/1.2.0/tokio/time/struct.Interval.html#method.tick) instead of `next`. We probably don't need `tokio_stream` at all.

src/routing/bootstrap.rs
```diff
@@ -643,7 +649,10 @@ mod tests {
             task::yield_now().await;
```
madadam, line 649: This `yield_now` (and all the others in this file) might not be necessary after the switch from `try_recv` to `recv`. A disadvantage of this switch is that this test can now hang if there is a bug where previously it would fail normally. Probably something we have to live with as `try_recv` has been removed from tokio, if I understand it correctly.

src/routing/tests/mod.rs
```diff
@@ -1350,7 +1350,9 @@ async fn handle_untrusted_sync() -> Result<()> {
     }
 
     assert!(bounce_sent);
-    assert_matches!(event_rx.try_recv(), Err(_));
+    assert!(timeout(Duration::from_secs(5), event_rx.recv())
```
madadam, line 1353: I understand there is no good way to do this now, but at least make the timeout much shorter so this test doesn't take 5 seconds more than it used to for no good reason. I think 100ms should be more than enough.

src/routing/bootstrap.rs
lionel1704: I think we still need some of the occourences since we use `try_send()` in some places

Cargo.toml
```diff
@@ -19,7 +19,8 @@ futures = "~0.3.12"
 hex_fmt = "~0.3.0"
 itertools = "~0.9.0"
 lru_time_cache = "~0.11.0"
-qp2p = "~0.9.22"
+qp2p = { git = "https://github.com/bochaco/qp2p.git", branch = "upgrade-tokio" }
```
maqi, line 22: mark to avoid this got merged accidently

lionel1704 (approved):

PR #2270, opened 2021-01-02
feat(check) Add cargo deny. \n Allows us to check licenses plus more …

dirvine:
…to meet our requirements.
Please see https://embarkstudios.github.io/cargo-deny/index.html to add to GHA. We will use this as a template for all crates to ensure we have deps in line with our requirements. i.e. we can ban openssl and such like that prevent musl.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -166,7 +166,8 @@ deny = [
     #{ name = "ansi_term", version = "=0.11.0" },
     #
     { name = "openssl" },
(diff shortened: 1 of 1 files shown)


lionel1704 (approved):
This looks good. I like how `openssl` and `rust-openssl`  banned :smile: :100:

dirvine:
@S-Coyle  has a better way to do this.

PR #2274, opened 2021-01-06
Run cargo intraconv to change markdown link to intra-doc link

S-Coyle:
Run cargo intraconv to change markdown link to intra-doc link.
Originally this PR also added `cargo deadlinks` scan to CI but there were several broken doc links from external deps so would have been too much noise - removed that commit.


S-Coyle:
Set as draft PR because the `deadlinks` check hits `deny(bad_style) incompatible with previous forbid` errors, also being seen with the unused dep check scan in this repo.

dirvine (approved):

PR #2276, opened 2021-01-12
Fix Sync messages being untrusted by non-elders

madadam:


src/routing/tests/mod.rs
```diff
@@ -1437,6 +1443,119 @@ async fn message_to_self(dst: MessageDst) -> Result<()> {
...
+
+    // Create `HandleConsensus` command for an `OurElders` vote. This will demote one of the
+    // current elders and promote the oldest peer.
+    let elders_info1 = EldersInfo::new(
+        iter::once(node.peer())
+            .chain(other_elder_peers.clone())
+            .chain(iter::once(promoted_peer)),
+        Prefix::default(),
+    );
+    let elder_names1: BTreeSet<_> = elders_info1.elders.keys().copied().collect();
```
maqi, line 1483: that's public keys collected?

src/routing/tests/mod.rs
```diff
@@ -1437,6 +1443,119 @@ async fn message_to_self(dst: MessageDst) -> Result<()> {
...
+        assert_eq!(section.chain().last_key(), &pk1);
+
+        // The message is trusted even by peers who don't yet know the new section key.
+        assert_matches!(
+            message.verify(iter::once((&Prefix::default(), &pk0))),
+            Ok(VerifyStatus::Full)
+        );
+
+        // Merging the section contained in the message with the original section succeeds.
+        assert_matches!(section0.clone().merge(section.clone()), Ok(()));
```
maqi, line 1533: do we also need to check the `merged section` now contains the latest key (pk1) ?

src/section/mod.rs
```diff
@@ -150,8 +153,14 @@ impl Section {
         self.members.update(member_info)
     }
 
-    pub fn to_minimal(&self) -> Self {
-        let first_key_index = self.elders_info_signing_key_index();
+    // Returns a trimmed version of this `Section` which contains only the elders info and the
+    // section chain truncated to the given length (the chain is truncated from the end, so it
+    // always contains the latest key). If `chain_len` is zero, it is silently replaced with one.
```
maqi, line 158: if `chain_len > existing_len`, the whole chain will be returned, right?

src/section/mod.rs
```diff
@@ -164,6 +173,15 @@ impl Section {
         &self.chain
     }
 
+    // Extend the section chain so it starts at `new_first_key` while keeping the last key intact.
```
maqi, line 176: it's actually: update the chain to be : from the new_first_key to the current_last_key within the full_chain, right?
I think the name of `extend` is bit improper, `slice` is more accurate ?
anyway, I am ok with stay with the current name, just thinking the comment can be more clear & accurate?

src/routing/approved.rs
```diff
@@ -1621,33 +1621,41 @@ impl Approved {
...
+        let mut commands = vec![];
+
+        let (elders, non_elders): (Vec<_>, _) = section
+            .active_members()
+            .filter(|peer| peer.name() != &self.node.name())
+            .copied()
+            .partition(|peer| section.is_elder(peer.name()));
+
+        // Send the trimmed state to non-elders. The trimmed state contains only the latest
+        // section key and one key before that which is the key the recipients should know so they
```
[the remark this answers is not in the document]
madadam: Yes, when we append a new section key to the chain, the members need to receive a chain that contains the key they already know and the new key so that they can verify the new key. Under normal circumstances, the key they already know is the one before the latest key. In case they are lagging and their latest known key is further behind, they will bounce the message and it will be resent back to them with extended chain. The bouncing of the `Sync` messages were not handled correctly which is what this PR fixes.

src/routing/tests/mod.rs
madadam: No, `XorName`s. `EldersInfo::elders` is a `BTreeMap<XorName, Peer>` so `elders.keys()` is an iterator of `XorName`s.

src/routing/tests/mod.rs
madadam: Maybe, but we already checked that the `Section` in the message contains the latest key and I think it's OK to assume here the `merge` function works correctly because this is an unit test, not an integration test. The `merge` function should have its own unit test (although it currently doesn't :grimacing: )

src/section/mod.rs
madadam: Yes.

src/section/mod.rs
madadam: I don't think `slice` is appropriate, but I think I can see the confusion. There are two chains here: first the chain we want to extend (`self.chain`) and second the full section chain we use to lookup the missing keys (`full_chain`). We are "extending" the first chain but "slicing" the second chain. However because it's the first chain that's the one being modified, the function name is `extend` which I think is appropriate.

src/routing/tests/mod.rs
maqi: Yeah, I agree the merge function shall be covered within its own unit test.
we can add it in a separate PR later on.

src/section/mod.rs
maqi: ok then, make sense.

maqi (approved):

PR #2285, opened 2021-01-20
feat: do not create connection when failed to send to client

maqi:
Close https://github.com/maidsafe/sn_routing/issues/2283


src/routing/stage.rs
```diff
@@ -173,14 +174,28 @@ impl Stage {
...
-            .map(Command::HandlePeerLost)
-            .collect()
+        if to_client {
+            if self
+                .comm
+                .send_message_to_client(&recipients[0], message)
+                .await
+                .is_err()
+            {
+                vec![Command::HandlePeerLost(recipients[0])]
```
madadam, line 186: Do we need to handle peer lost if it is a client?

src/routing/command.rs
```diff
@@ -59,6 +59,7 @@ pub(crate) enum Command {
         recipients: Vec<SocketAddr>,
         delivery_group_size: usize,
         message: Bytes,
+        to_client: bool,
```
madadam, line 62: While this works, I think it would be a bit cleaner to have a separate command for that. Because when sending to a client, we don't use `delivery_group_size` and `recipients` has always exactly one element so this makes it ever so slightly prone to bugs. So say:

```rust
enum Command {
    // ...
    SendMessageToNodes {
        recipients: Vec<SocketAddr>,
        delivery_group_size: usize,
        message: Bytes,
    },
    SendMessageToClient {
        recipient: SocketAddr,
        message: Bytes,
    }
}
```
Up to you though.

src/routing/stage.rs
maqi: As I understand it will be handled by above layer?
here we are returning the error.

src/routing/command.rs
maqi: right, I will make them two separate commands

madadam (approved):

PR #2286, opened 2021-01-20
Better error handling

madadam:
Improve error handling in routing:

- Remove redundant and ambiguous variants from `Error`
- Add more specific variants and make them contain more information
- Use multiple small and specialized error types instead of one big error type everywhere. There is still only one public error type though.

Additionally and unrelated to error handling:
- fix a bug in the stress test
- log more info in some cases
- minor refactoring

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -16,6 +16,7 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 // TODO: consider removing those variants that cannot occur when calling the public API.
 #[derive(Debug, Error)]
 #[allow(missing_docs)]
+#[non_exhaustive]
 pub enum Error {
     #[error("network layer error: {}", .0)]
     Network(#[from] qp2p::Error),
--- a/src/error.rs
+++ b/src/error.rs
@@ -6,37 +6,43 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use crate::{messages, section};
 use thiserror::Error;
 
 /// The type returned by the sn_routing message handling methods.
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
-/// Internal error.
+/// Routing error.
+// TODO: consider removing those variants that cannot occur when calling the public API.
 #[derive(Debug, Error)]
 #[allow(missing_docs)]
 pub enum Error {
-    #[error("Failed signature check.")]
-    FailedSignature,
-    #[error("Cannot route.")]
-    CannotRoute,
-    #[error("Network layer error: {}", .0)]
+    #[error("network layer error: {}", .0)]
     Network(#[from] qp2p::Error),
-    #[error("The node is not in a state to handle the action.")]
-    InvalidState,
-    #[error("Bincode error: {}", .0)]
-    Bincode(#[from] bincode::Error),
-    #[error("Invalid source location.")]
+    #[error("node bootstrap was interrupted")]
+    BootstrapInterrupted,
+    #[error("failed to create genesis section: {}", .0)]
+    CreateGenesisSection(#[source] section::CreateError),
+    #[error("failed to merge sections: {}", .0)]
+    MergeSection(#[from] section::MergeError),
+    #[error("failed to create message: {}", .0)]
+    CreateMessage(#[from] messages::CreateError),
+    #[error("received invalid message: {}", .0)]
+    InvalidMessage(#[from] messages::IntegrityError),
+    #[error("failed to send a message")]
+    SendMessage,
+    #[error("failed to create vote proof share: {}", .0)]
+    CreateVoteProofShare(#[source] bincode::Error),
+    #[error("received invalid vote proof share")]
+    InvalidVoteProofShare,
+    #[error("failed to create resource challenge: {}", .0)]
+    CreateResourceChallenge(#[source] bincode::Error),
+    #[error("invalid message source location")]
     InvalidSrcLocation,
-    #[error("Invalid destination location.")]
+    #[error("invalid message destination location")]
     InvalidDstLocation,
-    #[error("Content of a received message is inconsistent.")]
-    InvalidMessage,
-    #[error("A signature share is invalid.")]
-    InvalidSignatureShare,
-    #[error("The secret key share is missing.")]
+    #[error("the secret key share is missing")]
     MissingSecretKeyShare,
-    #[error("Failed to send a message.")]
-    FailedSend,
-    #[error("Invalid vote.")]
-    InvalidVote,
+    #[error("cannot route")]
+    CannotRoute,
 }
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -316,34 +316,19 @@ pub enum MessageStatus {
     Unknown,
 }
 
-#[derive(Debug, Error)]
-pub enum ParseError {
-    #[error("failed to deserialize message: {}", .0)]
-    Deserialize(#[from] bincode::Error),
-    #[error("signature check failed")]
-    FailedSignature,
-}
-
 /// Error when creating a message.
 #[derive(Debug, Error)]
+#[non_exhaustive]
+#[allow(missing_docs)]
 pub enum CreateError {
     #[error("failed to serialize message: {}", .0)]
     Serialize(#[from] bincode::Error),
 }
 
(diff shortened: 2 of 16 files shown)


src/error.rs
```diff
@@ -6,37 +6,43 @@
...
+    #[error("received invalid message: {}", .0)]
+    InvalidMessage(#[from] messages::IntegrityError),
+    #[error("failed to send a message")]
+    SendMessage,
+    #[error("failed to create vote proof share: {}", .0)]
+    CreateVoteProofShare(#[source] bincode::Error),
+    #[error("received invalid vote proof share")]
+    InvalidVoteProofShare,
+    #[error("failed to create resource challenge: {}", .0)]
+    CreateResourceChallenge(#[source] bincode::Error),
```
maqi, line 39: Maybe for `CreateResourceChallenge` and `SendMessage`, it shall includ a detailed reason of detailed error? like what `CreateMessage` does ?

src/routing/bootstrap.rs
```diff
@@ -209,35 +207,35 @@ impl<'a> State<'a> {
             }
         }
 
-        error!("{} Message sender unexpectedly closed", self.node);
```
maqi, line 212: still keeping this log ?

src/routing/bootstrap.rs
```diff
@@ -436,9 +445,7 @@ impl<'a> State<'a> {
...
```
maqi, line 439: same here, maybe keep the log?

src/section/mod.rs
```diff
@@ -44,10 +40,13 @@ pub(crate) struct Section {
 impl Section {
     /// Creates a minimal `Section` initially containing only info about our elders
     /// (`elders_info`).
-    pub fn new(chain: SectionProofChain, elders_info: Proven<EldersInfo>) -> Result<Self, Error> {
+    pub fn new(
+        chain: SectionProofChain,
+        elders_info: Proven<EldersInfo>,
+    ) -> Result<Self, CreateError> {
         if !chain.has_key(&elders_info.proof.public_key) {
             // TODO: consider more specific error here.
```
maqi, line 48: this TODO can now be removed?

src/section/mod.rs
```diff
@@ -352,12 +346,37 @@ impl Section {
...
+    #[error("failed to serialize data for signing: {}", .0)]
+    Serialize(#[from] bincode::Error),
+    #[error("failed to combine signature shares: {}", .0)]
+    // TODO: add the `#[from]` attribute here once threshold_crypto publishes the version where
+    // their `Error` implements `std::error::Error`.
+    CombineSignatures(bls::error::Error),
+}
+
+/// Error when merging sections.
+#[derive(Debug, Error)]
```
maqi, line 364: MergeError doesn't require `#[allow(missing_docs)]` like others ?

src/messages/mod.rs
```diff
@@ -170,15 +170,16 @@ impl Message {
...
     where
         I: IntoIterator<Item = (&'a Prefix, &'a bls::PublicKey)>,
     {
         let bytes = bincode::serialize(&SignableView {
             dst: &self.dst,
             dst_key: self.dst_key.as_ref(),
             variant: &self.variant,
-        })?;
+        })
+        .map_err(|_| IntegrityError::FailedSignature)?;
```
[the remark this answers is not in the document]
madadam: Yes but we serialize it only so we can calculate the signature from it. So failure to serialize here results in failure to verify. We already do something like this in other places, for example `Proven::self_verify` returns `false` even on serialize failure.

That said, I don't have a strong opinion on this. So if you think we should rather return a serialization error then I don't mind that much changing it.

src/messages/mod.rs
```diff
@@ -336,20 +317,19 @@ pub enum MessageStatus {
...
 
-impl From<CreateError> for Error {
-    fn from(src: CreateError) -> Self {
-        match src {
-            CreateError::Bincode(inner) => Self::Bincode(inner),
-            CreateError::FailedSignature => Self::FailedSignature,
-        }
-    }
+/// Error when creating a message.
+#[allow(missing_docs)]
```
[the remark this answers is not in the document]
madadam: Thing is the comments would say the same thing the error descriptions already do, so they would be redundant.

src/error.rs
madadam: Agree about `SendMessage`, but the resource challenge creation can fail only due to serialization.

src/routing/bootstrap.rs
madadam: Don't think it's needed. There is now separate error variant for this and we log those errors already.

src/section/mod.rs
madadam: No, because it's not exposed outside of the crate. Hm, but the other types aren't exposed either, so maybe I can remove the attribute there too...

src/error.rs
maqi: Ok.
Then just change for SendMessage

src/messages/mod.rs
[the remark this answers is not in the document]
maqi: Well, as it's aimed to make the returned error more accurate, I think it will help if return with a serialisation error.
Which may help the caller to carry out restore work ?

src/section/mod.rs
maqi: also maybe consider make them `pub(crate)` ?

madadam: That won't work because those types are part of `Error` which is itself exposed. This is a bit confusing but it seems that in rust there is a difference between something being public and it being exposed. You can have things that are `pub` but still not exposed. That means the users of the crate can still use those types, but for example calling methods on instances of them, but they can't construct or deconstruct (`match`) them. 

So in this case we have the `Error` type which is exposed, but some of its variants contain types that are not. So the users of routing can do something like:

```rust
match error {
    Error::InvalidMessage(inner_error) => println!("invalid message: {}", inner_error); // you can use `inner_error` even though you don't know its type.
    // ...
}
```

but they can't do this:

```rust
match error {
    Error::InvalidMessage(inner_error) => {
        match inner_error {
            IntegrityError::Untrusted => ...,
            // ..
        }
    }
    // ...
}
```

because the `IntegrityError` type is not exposed. 

I'm not sure this is the best thing to do. Maybe we should expose those inner errors as well? The reason I decided to to it this way is that those inner errors are mostly interesting for routing itself and not that much for its users. But not sure.

dirvine: I reckon if any part of an Error is in the API then the whole thing needs to be pub visible. I am much more a fan of pub(crate) for readability TBH, otherwise, we need all devs to know the backflips ;-) 

If any of these sub errors do expose then we need to make certain they are in a module that is very well contained. Reason being we can have boundaries between modules as you would expect but we don't want Error types to create another set of boundaries, they need to be the same.

madadam: Right, so then we have these options:

1. Do not use the inner errors at all. So instead of `Error::InvalidMessage(IntegrityError)` how just `Error::InvalidMessage`.
2. Use type-erased inner errors. So `Error::InvalidMessage(Box<dyn std::error::Error>)` or even `Error::InvalidMessage(String)`
3. Use concrete inner errors, but don't expose them (this is what's in the PR now)
4. Use concrete inner errors and expose them

I don't really like option 1, because that's basically what we have now and I don't think the information in the error is sufficient. As for option 4, this would be my preference, but it would be also this "mini errors" thing which you said you don't want? That leaves us with options 2 or 3. I don't have a strong preference among them as they are effectively very similar. 

Btw, we can't use `pub(crate)` on the inner errors as that doesn't compile.

dirvine: Can we not have InvalidMessage with a `From` clause that is IntegrityError where IntegrityError has more info if required? 

I am cool with mini Errors iff they are module specific and we can test that module independently etc. It's a good look, my issue is mini errors that are not also code modules.

madadam: > Can we not have InvalidMessage with a From clause that is IntegrityError where IntegrityError has more info if required?

Do you mean `impl From<IntegrityError> for Error` which returns a `Error::InvalidMessage` ? Because we already have that. The question is whether to also expose `IntegrityError` as part of the public API.

madadam: OK, I exposed the inner errors and updated the PR. I also made them all `#[non_exhaustive]`.

madadam:
This has fallen too far behind. Will close, but we should come back to it when the more important issues are resolved.

PR #2292, opened 2021-01-26
feat: log send to client error

maqi:


src/routing/mod.rs
```diff
@@ -411,6 +411,7 @@ async fn handle_connection_events(
                 stage.send_event(event).await;
             }
             ConnectionEvent::Disconnected(addr) => {
+                info!("Lost connection to {:?}", addr);
```
madadam, line 414: This is probably not important enough to use `info`, is it? I'd use `debug` or even `trace` to avoid spamming the logs.

madadam (approved):

src/routing/mod.rs
maqi: right, I will change it to trace

madadam (approved):

src/routing/mod.rs
dirvine: Nice catch @maqi  that kind of .is_ok() is possible and hides errors, so nice one.

src/routing/comm.rs
```diff
@@ -105,7 +105,12 @@ impl Comm {
         msg: Bytes,
     ) -> Result<(), SendError> {
         if let Some(conn) = self.endpoint.get_connection(client) {
-            if conn.send_uni(msg.clone()).await.is_ok() {
+            if let Err(err) = conn.send_uni(msg.clone()).await {
```
dirvine, line 108: We are still swallowing an error here, should we? I imagine as this returns a Result we need to return the error as a result here and not swallow it (prob why we have the issue in the first place)

src/routing/comm.rs
```diff
@@ -105,7 +105,12 @@ impl Comm {
...
+                error!(
+                    "Sending message to client {:?} failed with error {:?}",
+                    client, err
+                );
```
dirvine, line 112: possibly return Err(err) here

src/routing/comm.rs
maqi: This is to be addressed within the PR https://github.com/maidsafe/sn_routing/pull/2286

src/routing/comm.rs
maqi: This is to be addressed within the PR https://github.com/maidsafe/sn_routing/pull/2286

PR #2295, opened 2021-01-27
refactor(event): move promoted/demoted to elderschanged

oetyng:
- Removes `PromotedToAdult` and `Demoted` variants of `Event` enum.
- Adds `NodeElderChange` enum.
- Adds the flag `self_status_change` to `EldersChanged` variant, indicating whether the node got promoted, demoted or did not change.

### DIFF
--- a/src/routing/approved.rs
+++ b/src/routing/approved.rs
@@ -14,7 +14,7 @@ use crate::{
     },
     crypto, delivery_group,
     error::{Error, Result},
-    event::Event,
+    event::{Event, NodeElderChange},
     location::{DstLocation, SrcLocation},
     message_filter::MessageFilter,
     messages::{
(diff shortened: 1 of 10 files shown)


tests/bootstrap.rs
```diff
@@ -29,7 +29,7 @@ async fn test_genesis_node() -> Result<()> {
 
     assert_eq!(pub_key, node.public_key().await);
 
-    assert_next_event!(event_stream, Event::PromotedToElder);
+    assert_next_event!(event_stream, Event::EldersChanged { .. });
```
madadam, line 32: Maybe better to do
```rust
assert_next_event!(event_stream, Event::EldersChanged { change: NodeElderChange::Promoted, .. })
```
to keep the original semantics? Same in the other places.

oetyng: Absolutely, good point.

src/routing/approved.rs
```diff
@@ -1534,26 +1534,27 @@ impl Approved {
...
+                NodeElderChange::Demoted
+            } else {
+                NodeElderChange::None
+            };
+
             self.send_event(Event::EldersChanged {
                 prefix: *self.section.prefix(),
                 key: *self.section.chain().last_key(),
                 elders: self.section.elders_info().elders.keys().copied().collect(),
+                change: node_elder_change,
```
maqi, line 1554: the flag is only about self status change, may be name as `self_status_change` to make it more explicit ?
as the Event `EldersChanged` is actually about the whole section's elders.

oetyng: Yes, I agree the name was not good. I was wondering what the convention is now when referring to ourselves as node, is it `self` or `node`?

maqi: I think both works. though I may prefer `self` more.

maqi (approved):

PR #2298, opened 2021-01-28
Infrastructure queries (take 2)

madadam:


src/routing/approved.rs
```diff
@@ -993,15 +1015,12 @@ impl Approved {
         }
 
         if join_request.section_key != *self.section.chain().last_key() {
-            let response = BootstrapResponse::Join {
+            let variant = Variant::Rejoin {
```
maqi, line 1018: I am still feeling this is bit confusing with the `rejoined node` concept.
I understand here it means `ask the peer to try re-join with new key`.
However, actually the `re-join process` won't help to update the key.
The peer has to go through `re-bootstrap`.

Just wondering, why not use the word of `ReBootstrap`?

src/routing/approved.rs
```diff
@@ -1918,23 +1937,20 @@ impl Approved {
...
-            recipient,
-            message.to_bytes(),
-        ))
+        Ok(Command::send_message_to_node(recipient, message.to_bytes()))
     }
 
     // TODO: consider changing this so it sends only to a subset of the elders
     // (say 1/3 of the ones closest to our name or so)
-    fn send_message_to_our_elders(&self, msg_bytes: Bytes) -> Command {
+    fn send_message_to_our_elders(&self, msg: Bytes) -> Command {
```
maqi, line 1945: I see some place change from `msg` to `msg_bytes`, and some place change from `msg_bytes` to `msg`
any guidline to follow?

src/routing/command.rs
```diff
@@ -25,6 +27,7 @@ use std::{
 use tokio::sync::mpsc;
 
 /// Command for node.
+#[allow(clippy::large_enum_variant)]
```
maqi, line 30: got one removed, here adding a new one.  :)

src/routing/approved.rs
madadam: I see. `Rebootstrap` might work, but it's still not entirely accurate. Let me think about it.

src/routing/command.rs
madadam: Right. We could either keep the `allow` or `Box` the `message` field in the `SendMessage` variant. I don't have a strong preference either way. What do you think?

src/routing/approved.rs
madadam: I guess `msg` is when it is a concrete type, like say `Message` and `msg_bytes` when it's just `Bytes`. But I'm not sure we are being consistent. Not sure it matters too much though, unless we deal with both `Message` and `Bytes` in the same scope.

src/routing/approved.rs
madadam: How about `JoinRetry` ?

bochaco (commented):
This looks good to me @madadam, I'm sending a PR (https://github.com/madadam/sn_routing/pull/4) now which just adds to the client messages test to verify it can receive both on uni and bi streams.

src/routing/command.rs
madadam: Actually, prefer to keep the `allow`, as boxing the large variant introduces lot of noise all over the codebase.

src/routing/approved.rs
maqi: ok. just being curious as seeing both changes within the same PR  :)

src/routing/command.rs
maqi: Yeah, I'd prefer leave the `allow` as well.

maqi (approved):

PR #2299, opened 2021-01-30
feat(keycache): adds a key cache and removes exposure of secret key

oetyng:


src/section/section_keys.rs
```diff
@@ -50,12 +65,92 @@ impl SectionKeysProvider {
...
-                self.current = self.pending.take();
+    ///
+    pub fn has_key_share(&self) -> bool {
+        !self.list.is_empty()
+    }
+
+    ///
+    pub fn get_most_recent(&self) -> Result<&SectionKeyShare> {
+        if !self.list.is_empty() {
+            if let Some(public_key) = self.list.get(self.list.len() - 1) {
```
Yoga07, line 107: Can use `self.list.back()` here.

src/section/section_keys.rs
```diff
@@ -50,12 +65,92 @@ impl SectionKeysProvider {
...
+    }
+}
+
+/// Implementation of simple cache.
+#[derive(Debug)]
+pub struct KeyCache {
+    map: BTreeMap<Vec<u8>, SectionKeyShare>,
+    list: VecDeque<bls::PublicKey>,
+    capacity: usize,
+}
```
madadam, line 87: I'd consider using `LruCache` from the `lru_time_cache` crate instead of rolling your own. We already use it in other places and it works fine.

src/section/section_keys.rs
```diff
@@ -50,12 +65,92 @@ impl SectionKeysProvider {
...
+    ///
+    pub fn has_key_share(&self) -> bool {
+        !self.list.is_empty()
+    }
+
+    ///
+    pub fn get_most_recent(&self) -> Result<&SectionKeyShare> {
+        if !self.list.is_empty() {
+            if let Some(public_key) = self.list.get(self.list.len() - 1) {
+                if let Some(share) = self.map.get(&bincode::serialize(public_key)?) {
```
madadam, line 108: `bls::PublicKey` already implements `Ord` so there should be no need to serialize it.

src/section/section_keys.rs
```diff
@@ -50,12 +65,92 @@ impl SectionKeysProvider {
...
+        }
+    }
+}
+
+/// Implementation of simple cache.
+#[derive(Debug)]
+pub struct KeyCache {
+    map: BTreeMap<Vec<u8>, SectionKeyShare>,
+    list: VecDeque<bls::PublicKey>,
+    capacity: usize,
```
madadam, line 86: Why do we need a separate `capacity` here? `VecDeque` already has a capacity.

src/section/section_keys.rs
```diff
@@ -50,12 +65,92 @@ impl SectionKeysProvider {
...
+            }
+            trace!("finalised DKG: {:?}", public_key);
+        }
+    }
+}
+
+/// Implementation of simple cache.
+#[derive(Debug)]
+pub struct KeyCache {
+    map: BTreeMap<Vec<u8>, SectionKeyShare>,
```
madadam, line 84: The cache is going to have only a few entries in practice (less that 100 I imagine?). So maybe we don't even need this map at all as simple linear lookup through the `list` should be efficient enough (and due to cache locality might be even more efficient than the map lookup, if the number of entries is small).

src/section/section_keys.rs
oetyng: It's just a strange thing to include time here. The impl is more or less the same as that of LruCache, only simpler.

src/section/section_keys.rs
oetyng: Wasn't accepted as key though, needed Eq.

src/section/section_keys.rs
oetyng: Yep, agree.

src/section/section_keys.rs
oetyng: Think it was just left by mistake from before the vecdequeue was used. :+1: Removing it.

src/section/section_keys.rs
madadam: The time is optional. `LruCache` can be time-based, capacity-based or both. So in this case you would use capacity only: https://docs.rs/lru_time_cache/0.11.5/lru_time_cache/struct.LruCache.html#method.with_capacity

madadam (approved):

Yoga07 (approved):

PR #2302, opened 2021-02-04
Update routing to use the new qp2p API

lionel1704:


src/routing/comm.rs
```diff
@@ -30,23 +32,30 @@ pub(crate) struct Comm {
...
+        ));
+
+        let _ = task::spawn(handle_disconnection_events(
+            disconnections,
             event_tx.clone(),
         ));
 
         Ok(Self {
             _quic_p2p: quic_p2p,
+            _incoming_connections: incoming_connections,
```
madadam, line 58: Wouldn't this leak memory if we are not reading from this stream?

src/routing/comm.rs
```diff
@@ -104,9 +112,20 @@ impl Comm {
...
+            .await
+            .is_err()
+        {
+            // If the sending of a message failed the connection would no longer
+            // exist in the pool. Calling send_message agail will use a new connection.
+            if let Err(err) = self.endpoint.send_message(msg, recipient).await {
+                error!(
+                    "Sending message to client {:?} failed with error {:?}",
+                    recipient, err
+                );
```
madadam, line 128: This is not quite right. The idea was that we would send the message on a connection that already exists or fail otherwise. But with this change we are establishing new connection on failure which is not going to work for clients in the general case.

src/routing/comm.rs
```diff
@@ -424,15 +432,10 @@ mod tests {
...
-            if let Some(mut incoming_msgs) =
-                time::timeout(TIMEOUT, recv_incoming_connections.next()).await?
-            {
-                if let Some(msg) = time::timeout(TIMEOUT, incoming_msgs.next()).await? {
-                    assert_eq!(msg.get_message_data(), msg0);
-                    msg0_received = true;
-                }
+            if let Some((_src, msg)) = time::timeout(TIMEOUT, incoming_msgs.next()).await? {
+                assert_eq!(msg, msg0);
+                msg0_received = true;
```
madadam, line 437: To keep the original purpose of this test I think we need to now explicitly disconnect from the peer here: 

```rust
recv_enpoint.disconnect_from(src)
```
Also change the comment above to something like: "Receive one message and disconnect from the peer"

src/routing/comm.rs
```diff
@@ -406,12 +413,13 @@ mod tests {
...
     async fn send_after_reconnect() -> Result<()> {
         let (tx, _rx) = mpsc::channel(1);
-        let send_comm = Comm::new(transport_config(), tx)?;
+        let send_comm = Comm::new(transport_config(), tx).await?;
 
         let recv_transport = QuicP2p::with_config(Some(transport_config()), &[], false)?;
-        let recv_endpoint = recv_transport.new_endpoint()?;
-        let recv_addr = recv_endpoint.socket_addr().await?;
-        let mut recv_incoming_connections = recv_endpoint.listen();
+        #[allow(unused)]
```
madadam, line 419: Better prefix the unused variables with underscore, so it's obvious which are actually unused.

src/routing/comm.rs
madadam: So, could we simply drop this stream if we are not going to use it, or would that cause problems?

src/routing/comm.rs
madadam: Although... the reason this didn't work before was that clients weren't listening for incoming connections. But after upgrading clients to use the new qp2p as well, that's not longer the case, is it? So maybe this works in the end.

madadam: But then this makes this function almost identical to the `send_to` below, except here we also log the error. Could we at least remove the duplication somehow?

src/routing/comm.rs
lionel1704: Yes we can do that. We'll need to ignore the result of `tx.send()` in qp2p but I guess that's okay?

src/routing/comm.rs
lionel1704: Ah I see, this is for clients. I can change this :+1:

src/routing/comm.rs
lionel1704: This one sneaked in, my bad :)

src/routing/comm.rs
madadam: Yep, sounds good.

joshuef (dismissed):

src/routing/comm.rs
```diff
@@ -30,18 +31,23 @@ pub(crate) struct Comm {
...
+    pub async fn new(
         transport_config: qp2p::Config,
         event_tx: mpsc::Sender<ConnectionEvent>,
     ) -> Result<Self> {
         let quic_p2p = QuicP2p::with_config(Some(transport_config), Default::default(), true)?;
 
         // Don't bootstrap, just create an endpoint where to listen to
         // the incoming messages from other nodes.
-        let endpoint = quic_p2p.new_endpoint()?;
+        let (endpoint, _, incoming_messages, disconnections) = quic_p2p.new_endpoint().await?;
```
maqi, line 42: better refactor the above comment, also explain what's the ignored returned item and what disconnections for?

src/routing/comm.rs
```diff
@@ -104,14 +108,10 @@ impl Comm {
...
-                error!("Sending message to {} failed: {}", recipient, err);
-            } else {
-                return Ok(());
-            }
-        }
-        Err(SendError)
+        self.endpoint
+            .send_message(msg, recipient)
+            .await
+            .map_err(|_| SendError)
```
maqi, line 114: better log this error during the map

src/routing/comm.rs
```diff
@@ -197,28 +197,20 @@ impl Comm {
 
     // Low-level send
     async fn send_to(&self, recipient: &SocketAddr, msg: Bytes) -> Result<(), qp2p::Error> {
-        let conn = self.connect_to(recipient).await?;
-
-        if conn.send_uni(msg.clone()).await.is_ok() {
+        // This will attempt to used a cached connection
```
maqi, line 200: `attempt to use`

maqi (approved):

PR #2303, opened 2021-02-04
chore: update bls_signature_aggregator

madadam:


src/consensus/vote.rs
```diff
@@ -116,13 +117,21 @@ impl VoteAccumulator {
         &mut self,
         vote: Vote,
         proof_share: ProofShare,
-    ) -> Result<(Vote, Proof), AccumulationError> {
-        self.0
-            .add(SignableWrapper(vote), proof_share)
-            .map(|(vote, proof)| (vote.0, proof))
+    ) -> Result<(Vote, Proof), VoteAccumulationError> {
+        let bytes = bincode::serialize(&SignableView(&vote))?;
+        let proof = self.0.add(&bytes, proof_share)?;
```
maqi, line 122: just wondering why changed the pervious `.map` approach ?

maqi (approved):

src/consensus/vote.rs
madadam: I now have to wrap the error in `VoteAccumulationError` and this way the code is a bit nicer I think.

PR #2317, opened 2021-02-15
Error back when detect client using incorrect section_key or there is DKG in progress

maqi:


src/routing/approved.rs
```diff
@@ -1834,6 +1847,34 @@ impl Approved {
         Ok(Some(command))
     }
 
+    pub fn check_key_status(&self, bls_pk: &bls::PublicKey) -> Result<(), InfrastructureError> {
+        if self.dkg_voter.has_ongoing_dkg() {
```
madadam, line 1851: This only returns `DkgInProgress` if *this node* currently participates in an ongoing DKG. But what if there is an ongoing DKG, but some of the current elders are not participating in it. Then they would not return `DkgInProgress`, would they? Is that what we want?

joshuef (dismissed):

src/routing/approved.rs
maqi: >but some of the current elders are not participating in it

madadam: OK, but just to be clear we are talking about the same thing: let's say the current elders are `{A, B, C, D}` and then churn happens and we are going to pick new elders `{A, E, F, G}`. Then of the current elders, only `A` is also among the new elders. So now when a client sends a message to the current elders, only `A` responds with `DkgInProgress`, the others (`B`, `C, `D`) happily process the request. And because they are still in supermajority, their response "wins". So, is this really what we want (not saying it isn't, just trying to understand)?

maqi: first, I think your above scenario only happens during split (with majority elders replaced), and B,C,D shall also participating in the other DKG process?
second, even in your case a churn causing majority elders replaced, at the time B,C,D got queried, they are still valid elders and compose a majority and using a `valid sectoin key` to response. It might be not 100% perfect as the key may get updated soon, but the response shall be considered as `valid` ?

madadam: > I think your above scenario only happens during split

I think it can happen when `E`, `F` and `G` joins the section in rapid succession and all are older than `B`, `C`, and `D`. Not very likely scenario, but not impossible either.

> they are still valid elders and compose a majority and using a valid sectoin key to response. It might be not 100% perfect as the key may get updated soon, but the response shall be considered as valid

I agree with you, but I think @dirvine specifically wanted to cover this case (from https://maidsafe.slack.com/archives/GHRBKK6R0/p1613380862205200?thread_ts=1613313804.196200&cid=GHRBKK6R0):

> Prob is nodes/clients trying to get consensus are now using the wrong set of Elders and consensus will be harder (less room for error). i.e. we imagine it's a group of 5 when it's only 3 etc.

madadam (approved):

PR #2321, opened 2021-02-19
All acc in routing

oetyng:


src/routing/approved.rs
```diff
@@ -39,22 +38,83 @@ use bytes::Bytes;
...
 use tokio::sync::mpsc;
 use xor_name::{Prefix, XorName};
 
 pub(crate) const RESOURCE_PROOF_DATA_SIZE: usize = 64;
 pub(crate) const RESOURCE_PROOF_DIFFICULTY: u8 = 2;
 const KEY_CACHE_SIZE: u8 = 5;
 
+type SocketId = XorName;
+
+struct EndUserRegistry {
```
maqi, line 64: better put this `SocketId` and `EndUserRegistry` into a separate file?
this approved.rs is already too huge.

src/routing/approved.rs
```diff
@@ -102,9 +163,18 @@ impl Approved {
...
             resource_proof: ResourceProof::new(RESOURCE_PROOF_DATA_SIZE, RESOURCE_PROOF_DIFFICULTY),
+            end_users: EndUserRegistry::new(),
         }
     }
 
+    pub fn get_enduser_by_addr(&self, sender: &SocketAddr) -> Option<EndUser> {
+        self.end_users.get_enduser_by_addr(sender)
+    }
+
+    pub fn get_socket_addr(&self, id: &SocketId) -> Option<&SocketAddr> {
```
maqi, line 174: better name the function to show it is only for enduser?

src/routing/approved.rs
```diff
@@ -200,18 +270,18 @@ impl Approved {
...
+    pub async fn handle_networkinfo_msg(
         &mut self,
         sender: SocketAddr,
-        message: InfrastructureMessage,
+        message: NetworkInfoMsg,
     ) -> Vec<Command> {
         match message {
-            InfrastructureMessage::GetSectionRequest(name) => {
-                debug!("Received GetSectionRequest({}) from {}", name, sender);
+            NetworkInfoMsg::GetSectionQuery(name) => {
```
maqi, line 279: just to double confirm that this renaming is agreed by others, right?

src/routing/approved.rs
```diff
@@ -678,7 +784,13 @@ impl Approved {
     // If elder, always handle UserMessage, otherwise handle it only if addressed directly to us
     // as a node.
     fn should_handle_user_message(&self, dst: &DstLocation) -> bool {
-        self.is_elder() || dst.as_node().ok() == Some(&self.node.name())
+        let is_elder = self.is_elder();
+        let are_we_dst = if let DstLocation::Node(name) = dst {
```
maqi, line 788: better having a `as_node` function as previous?

or can we do:  `DstLocation::Node(self.node.name()) == dst` ?

src/routing/approved.rs
```diff
@@ -746,6 +858,7 @@ impl Approved {
         let src_name = match src {
             SrcLocation::Node(name) => name,
             SrcLocation::Section(prefix) => prefix.name(),
+            _ => unimplemented!(),
```
maqi, line 861: same as previous. try not to use wild card match and `unimplemented`

src/routing/mod.rs
```diff
@@ -312,17 +311,24 @@ impl Routing {
         dst: DstLocation,
         content: Bytes,
     ) -> Result<()> {
+        if let DstLocation::EndUser(EndUser::Client { socket_id, .. }) = dst {
```
maqi, line 314: bit confused here.
with this block of code, do we still need the following `send_message_to_client` ?

src/routing/mod.rs
```diff
@@ -312,17 +311,24 @@ impl Routing {
...
+            if let Some(socket_addr) = self.stage.state.lock().await.get_socket_addr(&socket_id) {
```
maqi, line 315: if a socket_addr cannot be fetched for the socket_id, we will still try to send this UserMessage?
does it mean a `relay` ? or shall we return with an error?

examples/stress.rs
```diff
@@ -354,6 +355,7 @@ impl Network {
                         DstLocation::Direct => {
                             return Err(format_err!("unexpected probe message dst: {:?}", dst))
                         }
+                        _ => unimplemented!(),
```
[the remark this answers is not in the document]
oetyng: Yup, aware of this 👍 . Is why, among things, the PR is still in draft. The unimplemented is used when it is simply not implemented yet ;), which the aim is that it shall be, before making the PR ready for merge.

src/routing/approved.rs
oetyng: Yep, also the aim (`file organisation`) before making this ready for PR. I should update the draft PR message with that into. (It was written in one of the commit msgs, f78a508)

src/routing/approved.rs
oetyng: Not sure how you mean.

src/routing/approved.rs
oetyng: Not more agreed than the first choice `InfrastructureMessage` was (not at all I think). Simply my view of what it should be,

src/routing/approved.rs
oetyng: > DstLocation::Node(self.node.name()) == dst

This seems better IMO.

src/routing/mod.rs
oetyng: The assumption here is that if there is no `socket_addr` here, that means this message is for a client in _another section_, and thus it should be relayed. I was thinking that sending it with `UserMessage` would make it be relayed, is that so?
If we check that our prefix matches this client `public_key`, then a missing `socket_addr` would be some sort of error.

src/routing/mod.rs
oetyng: Yeah IDK, this was the _super simple_ `first-get-it-working` code, to expose one single `send_message` API.
It's because I didn't know exactly how the internals should be refactored.

src/routing/approved.rs
oetyng: Yep :+1:, still wip code, so this is the unimplemented part of the work in progress.

Cargo.toml
```diff
@@ -60,8 +66,11 @@ structopt = "~0.3.17"
...
 yansi = "~0.5.0"
-sn_data_types = "0.14.1"
 
   [dev-dependencies.rand]
   version = "~0.7.3"
   features = [ "small_rng" ]
+
+  [dev-dependencies.tokio]
+  version = "~0.2.24"
+  features = [ "stream", "udp" ]
```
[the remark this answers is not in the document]
oetyng: Yeah, that's a duplication.
This was a fix @madadam told me of, to the compile errors that came with IGD 0.12.0 (in qp2p 0.9.18).

Cargo.toml
```diff
@@ -19,13 +19,19 @@ futures = "~0.3.6"
 hex_fmt = "~0.3.0"
 itertools = "~0.9.0"
 lru_time_cache = "~0.11.0"
-qp2p = "~0.9.16"
+qp2p = "~0.9.18"
+# qp2p = { git = "https://github.com/maidsafe/qp2p", rev = "2fd0bd1" } # 2fd0bd12590ea1a0b9f91947d1764a56ad142d19
```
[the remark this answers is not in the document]
oetyng: Of course. All that is always done at end when making the PR ready, still draft.

src/routing/approved.rs
oetyng: > you can have a map of `BTreeMap<SocketAddr, (EndUserPk, HashSet<SocketAddr>)>`
> and then impl the `get` function for EndUserRegistry` to avoid having two maps?

I think the two maps are needed since we use both socket_addr and socket_id to key into the maps (at different places).

oetyng:
> you can have a map of `BTreeMap<SocketAddr, (EndUserPk, HashSet<SocketAddr>)>`
> and then impl the `get` function for EndUserRegistry` to avoid having two maps?

I think the two maps are needed since we use both socket_addr and socket_id to key into the maps (at different places).

PR #2322, opened 2021-02-22
fix: use unbounded channel to avoid long duration locks

madadam:


madadam:
The problem of long duration locks turned out to be elsewhere so this change is not needed. Closing.

PR #2323, opened 2021-02-22
Handle client registration

oetyng:
A single API is used for sending messages to the network, taking SrcLocation and DstLocation, removing the api intended for client messages.
Clients register at their section by signing the socket address they use. An obfuscated key us used in messaging across the network, to map against this socket address.


joshuef (approved):

src/routing/mod.rs
```diff
@@ -312,17 +311,42 @@ impl Routing {
...
+            public_key,
+        }) = dst
+        {
+            let mut socket = None;
+            {
+                if let Some(socket_addr) = self.stage.state.lock().await.get_socket_addr(&socket_id)
+                {
+                    socket = Some(socket_addr.clone());
+                }
+            }
```
lionel1704, line 325:
```suggestion
            let socket_addr = self.stage.state.lock().await.get_socket_addr(&socket_id).clone();
```

Cargo.toml
```diff
@@ -60,8 +61,11 @@ structopt = "~0.3.17"
...
 yansi = "~0.5.0"
-sn_data_types = "0.14.1"
 
   [dev-dependencies.rand]
   version = "~0.7.3"
   features = [ "small_rng" ]
+
+  [dev-dependencies.tokio]
+  version = "~0.2.24"
+  features = [ "stream", "udp" ]
```
maqi, line 71: missing a line break at the end ?

src/routing/approved.rs
```diff
@@ -39,22 +38,83 @@ use bytes::Bytes;
...
 use tokio::sync::mpsc;
 use xor_name::{Prefix, XorName};
 
 pub(crate) const RESOURCE_PROOF_DATA_SIZE: usize = 64;
 pub(crate) const RESOURCE_PROOF_DIFFICULTY: u8 = 2;
 const KEY_CACHE_SIZE: u8 = 5;
 
+type SocketId = XorName;
+
+struct EndUserRegistry {
```
madadam: I'd put this whole struct and its impl into its own file, to prevent this one from growing too big (it already is, but no need to make it even worse).

src/routing/approved.rs
```diff
@@ -39,22 +38,83 @@ use bytes::Bytes;
...
+
+impl EndUserRegistry {
+    pub fn new() -> Self {
+        Self {
+            clients: BTreeMap::default(),
+            socket_id_mapping: BTreeMap::default(),
+        }
+    }
+
+    pub fn get_enduser_by_addr(&self, socketaddr: &SocketAddr) -> Option<EndUser> {
```
madadam: Minor, but why are we returning by value here but by reference in the next function?

src/routing/approved.rs
maqi, line 64: better move into a separate file ?

src/routing/approved.rs
```diff
@@ -39,22 +38,83 @@ use bytes::Bytes;
...
+            return Err(Error::InvalidState);
+        };
+        let end_user = EndUser::Client {
+            public_key: end_user_pk,
+            socket_id,
+        };
+        match self.socket_id_mapping.entry(socket_id) {
+            Entry::Vacant(_) => {
+                let _ = self.clients.insert(sender, end_user);
+                let _ = self.socket_id_mapping.insert(socket_id, sender);
```
madadam: Use the `entry` API all the way:

```rust
        match self.socket_id_mapping.entry(socket_id) {
            Entry::Vacant(entry) => {
                let _ = self.clients.insert(sender, end_user);
                let _ = entry.insert(sender);
            }
             // ...
```

src/routing/mod.rs
joshuef:
```let socket_addr = self.stage.state.lock().await.get_socket_addr(&socket_id).clone();
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                    - temporary value is freed at the end of this statement
    |                               |
    |                               creates a temporary which is freed while still in use
320 | 
321 |             if let Some(socket_addr) = socket_addr {
    |                                        ----------- borrow later used here```

File: src/routing/approved.rs [hunk 6]
```diff
@@ -678,7 +781,13 @@ impl Approved {
     // If elder, always handle UserMessage, otherwise handle it only if addressed directly to us
     // as a node.
     fn should_handle_user_message(&self, dst: &DstLocation) -> bool {
-        self.is_elder() || dst.as_node().ok() == Some(&self.node.name())
+        let is_elder = self.is_elder();
+        let are_we_dst = if let DstLocation::Node(name) = dst {
```
maqi, line 785: `DstLocation::Node(self.node.name()) == dst` ?

src/routing/approved.rs
```diff
@@ -200,18 +270,18 @@ impl Approved {
         Ok(commands)
     }
 
-    pub async fn handle_infrastructure_message(
+    pub async fn handle_sectioninfo_msg(
```
madadam, line 214: It should be two words: `section_info` not `sectioninfo`.

src/routing/mod.rs
lionel1704: That was the case with the if condition. Once this is assigned to socket_addr (after .clone() or copy `*`)
The mutex lock will be released.

```
if let Some(...) = self.stage.state.lock().do_something().await {
     self.do_something_else().await()
}
```
Above, the lock is held in the if condition so within the block ( `self.do_something_else()` in this case) if we try to get a hold of the mutex, then it will hang since there is a deadlock.

src/routing/mod.rs
```diff
@@ -312,17 +311,42 @@ impl Routing {
         dst: DstLocation,
         content: Bytes,
     ) -> Result<()> {
+        if let DstLocation::EndUser(EndUser::Client {
+            socket_id,
+            public_key,
+        }) = dst
+        {
+            let mut socket = None;
```
madadam, line 320: This can be written more succinctly:

```rust
let socket = self.stage.state.lock().await.get_socket_addr(&socket_id).copied();
if let Some(socket) = socket {
    return self.send_message_to_client(...);
} else {
   ...
}
```

src/routing/mod.rs
lionel1704: I think we can dereference it and it will no longer be borrowed then? :thinking:

src/routing/mod.rs
```diff
@@ -312,17 +311,42 @@ impl Routing {
...
+            if let Some(socket_addr) = socket {
+                return self
+                    .send_message_to_client(socket_addr, ClientMessage::from(content)?)
+                    .await;
+            } else {
+                debug!(
+                    "Could not find socketaddr corresponding to socket_id {:?} and public_key {:?}",
+                    socket_id, public_key
+                );
+                debug!("Sending user message instead.. (Command::SendUserMessage)");
```
madadam, line 337: Just to confirm, but does this make sense? Shouldn't we return an error instead?

src/routing/mod.rs
lionel1704: Just tried this locally, and this can be done as:
`let socket_addr = self.stage.state.lock().await.get_socket_addr(&socket_id).copied();`

src/routing/enduser_registry.rs
```diff
@@ -0,0 +1,71 @@
...
+        };
+        let end_user = EndUser::Client {
+            public_key: end_user_pk,
+            socket_id,
+        };
+        match self.socket_id_mapping.entry(socket_id) {
+            Entry::Vacant(_) => {
+                let _ = self.clients.insert(sender, end_user);
+                let _ = self.socket_id_mapping.insert(socket_id, sender);
+            }
```
madadam, line 66: Use the `entry` API all the way:
```rust
            Entry::Vacant(entry) => {
                let _ = self.clients.insert(sender, end_user);
                let _ = entry.insert(sender);
            }
```

src/routing/enduser_registry.rs
```diff
@@ -0,0 +1,71 @@
...
+        socketaddr_sig: EndUserSig,
+    ) -> Result<()> {
+        if let Ok(data) = &bincode::serialize(&sender) {
+            end_user_pk
+                .verify(&socketaddr_sig, data)
+                .map_err(|_e| Error::InvalidState)?;
+        } else {
+            return Err(Error::InvalidState);
+        }
+        let socket_id = if let Ok(socket_id_src) = &bincode::serialize(&socketaddr_sig) {
```
madadam, line 53: You can avoid this serialization (and the subsequent error handling) by using `Signature::to_bytes` instead.

src/routing/enduser_registry.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub fn try_add(
+        &mut self,
+        sender: SocketAddr,
+        end_user_pk: EndUserPK,
+        socketaddr_sig: EndUserSig,
+    ) -> Result<()> {
+        if let Ok(data) = &bincode::serialize(&sender) {
+            end_user_pk
+                .verify(&socketaddr_sig, data)
+                .map_err(|_e| Error::InvalidState)?;
```
madadam, line 49: `Error::FailedSignature` seems like a better choice here.

src/routing/enduser_registry.rs
```diff
@@ -0,0 +1,71 @@
...
+        } else {
+            return Err(Error::InvalidState);
```
madadam, line 51: `Error::Bincode` or `Error::FailedSignature` both seem like better choices too.

src/routing/enduser_registry.rs
```diff
@@ -0,0 +1,71 @@
...
+};
+use xor_name::XorName;
+
+pub type SocketId = XorName;
+pub(crate) struct EndUserRegistry {
+    clients: BTreeMap<SocketAddr, EndUser>,
+    socket_id_mapping: BTreeMap<SocketId, SocketAddr>,
+}
+
+impl EndUserRegistry {
```
madadam, line 24: Do we eveer remove stuff from this registry?

src/routing/mod.rs
madadam: `get_socket_addr` returns `Option<&SocketAddr>`. Calling `clone` on it just clones the whole `Option`, not the thing in it. So it still returns `Option<&SocketAddr>` and therefore is still borrowing from a temporary. @lionel1704 's suggestion is the correct one.

src/routing/enduser_registry.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub fn new() -> Self {
+        Self {
+            clients: BTreeMap::default(),
+            socket_id_mapping: BTreeMap::default(),
+        }
+    }
+
+    pub fn get_enduser_by_addr(&self, socketaddr: &SocketAddr) -> Option<EndUser> {
```
madadam, line 32: Why are we returning by value here but by reference in the following function?

src/routing/mod.rs
oetyng: It makes sense, because if we are at section A, and want to send to a client which may be anywhere, we first check locally, and if we don't have the socket addr we relay it to the network.

src/routing/enduser_registry.rs
oetyng: That's only for bls::Signature. This is a sn_data_types::Signature.

src/routing/enduser_registry.rs
oetyng: Great, missed that one (think the naming is a bit unclear, FailedSignatureCheck is more to the point, failed signature sounds a bit like the signing failed...)

src/routing/enduser_registry.rs
oetyng: It would be removed on ClientLost. But I suggest that flow be added in a new PR, how's that sound?

src/routing/enduser_registry.rs
oetyng: Unclear. But fixed now anyway.

src/routing/enduser_registry.rs
```diff
@@ -0,0 +1,65 @@
...
+        self.socket_id_mapping.get(socket_id)
+    }
+
+    pub fn try_add(
+        &mut self,
+        sender: SocketAddr,
+        end_user_pk: EndUserPK,
+        socketaddr_sig: EndUserSig,
+    ) -> Result<()> {
+        let data = &bincode::serialize(&sender).map_err(Error::Bincode)?;
```
madadam, line 46: No need for the `map_err(Error::Bincode)` as `Error` already has `impl From<bincode::Error>`.

src/routing/mod.rs
madadam: OK, but is that relaying implemented though? That is, when a node receives a `UserMessage` whose `dst` is `EndUser`, does it try to look up the socket addr and send the message to the client? Unless I'm missing something it doesn't seem to be implemented in this PR. Is that on purpose?

oetyng: The assumption in the above implementation is that the message is relayed to closest section by the next lines:

```
let command = Command::SendUserMessage { src, dst, content };
        self.stage.clone().handle_commands(command).await
``` 

Is it not?

oetyng: `approved.handle_user_message`, is where a relayed msg would turn up, if it iss deemed useful... And there it is raised to upper level by `Event::MessageReceived`.
So, there is no evaluation of if this is really a msg with enduser dst, and call to end_user_registry to send it using the socketaddr.

Where do you suggest that eval goes in? In `approved.handle_user_message`?

maqi (dismissed):

S-Coyle (approved):

maqi (approved):

PR #2328, opened 2021-02-24
Add support for accumlation at dest node

lionel1704:
BREAKING CHANGE: this uses a new version of sn_messaging with a breaking
change


src/messages/mod.rs
```diff
@@ -76,6 +77,16 @@ impl Message {
                     return Err(CreateError::FailedSignature);
                 }
             }
+            SrcAuthority::BlsShare { proof_share, .. } => {
+                if !proof_share
```
maqi, line 81: you can call `proof_share.verify(signed_bytes)` directly?

src/messages/mod.rs
```diff
@@ -122,6 +133,44 @@ impl Message {
         Ok(msg)
     }
 
+    /// Creates a message signed using a BLS KeyShare for
+    /// destination accumulation
```
maqi, line 137: can put in one line?

src/routing/approved.rs
```diff
@@ -665,7 +668,38 @@ impl Approved {
...
+                if msg.dst() == &DstLocation::AccumulatingNode(self.node().name()) {
+                    if let SrcAuthority::BlsShare { proof_share, .. } = msg.src() {
+                        let signed_bytes = bincode::serialize(&msg.signable_view())?;
+                        match self
+                            .message_accumulator
+                            .add(&signed_bytes, proof_share.clone())
+                        {
+                            Ok(proof) => {
+                                trace!("Successfully accumulated message: {:?}", msg);
+                                let key = msg.proof_chain_last_key()?;
```
oetyng, line 680: Should we only check the last key?

src/messages/mod.rs
```diff
@@ -192,6 +241,32 @@ impl Message {
...
+                ..
+            } => {
+                // Proof chain is required for accumulation at destination.
+                let proof_chain = if let Some(proof_chain) = self.proof_chain.as_ref() {
+                    proof_chain
+                } else {
+                    return Err(Error::InvalidMessage);
+                };
+
+                if !proof_share
```
maqi, line 256: same as above, call `proof_share.verify` function directly?

src/messages/src_authority.rs
```diff
@@ -32,6 +34,16 @@ pub enum SrcAuthority {
         /// ed-25519 signature of the message corresponding to the public key of the source peer.
         signature: SimpleSignature,
     },
+    /// Authority of a single peer that used
```
maqi, line 37: `uses` ? 
also, can put into one line?

src/messages/src_authority.rs
```diff
@@ -45,6 +57,7 @@ impl SrcAuthority {
     pub(crate) fn src_location(&self) -> SrcLocation {
         match self {
             Self::Node { public_key, .. } => SrcLocation::Node(name(public_key)),
+            Self::BlsShare { public_key, .. } => SrcLocation::Node(name(public_key)),
```
maqi, line 60: is this `mapping` really wanted ?

src/messages/src_authority.rs
```diff
@@ -74,7 +91,7 @@ impl SrcAuthority {
     pub(crate) fn as_section_prefix(&self) -> Result<&Prefix> {
         match self {
             Self::Section { prefix, .. } => Ok(prefix),
-            Self::Node { .. } => Err(Error::InvalidSrcLocation),
+            _ => Err(Error::InvalidSrcLocation),
```
maqi, line 94: prefer not to use wildcard match.
using `Self::Node { .. } | Self::BlsShare { .. }` ?

src/routing/message_accumulator.rs
```diff
@@ -0,0 +1,10 @@
+use bls_signature_aggregator::{Error, Proof, ProofShare, SignatureAggregator};
```
maqi, line 1: don't think we need to intruduce this struct? 
also, the copy right text is missed?

src/routing/approved.rs
madadam: It's enough. The proof chain itself is already validated by this point (meaning each key is validly signed by the previous key) so all that's left to do is to verify that the last key signs the message.

src/routing/approved.rs
```diff
@@ -665,7 +668,38 @@ impl Approved {
                 self.handle_join_request(msg.src().to_node_peer(sender)?, *join_request.clone())
             }
             Variant::UserMessage(content) => {
-                self.handle_user_message(msg.src().src_location(), *msg.dst(), content.clone())
+                if msg.dst() == &DstLocation::AccumulatingNode(self.node().name()) {
```
madadam, line 671: I'd put this new logic into `handle_user_message`, to keep this function small (-ish).

src/routing/approved.rs
```diff
@@ -665,7 +668,38 @@ impl Approved {
...
+                                        msg.src().src_location(),
+                                        *msg.dst(),
+                                        content.clone(),
+                                    )
+                                } else {
+                                    trace!("Aggregated signature is invalid. Handling message {:?} skipped", msg);
+                                    Ok(vec![])
+                                }
+                            }
+                            Err(err) => {
```
madadam, line 692: The signature aggregator also returns `Error::NotEnoughShares` which is not really an error so should probably not be logged to avoid spamming the log? See also `Approved::handle_vote`.

src/routing/message_accumulator.rs
```diff
@@ -0,0 +1,10 @@
...
+
+#[derive(Default)]
+pub(crate) struct MessageAccumulator(SignatureAggregator);
```
madadam, line 4: Do we need this wrapper?

src/messages/src_authority.rs
madadam: I think this is fine. It's still a single node sending the message, just using a different signature scheme. So `SrcLocation::Node` seems a good choice to me. Don't think we need a one-to-one mapping between `SrcAuthority` and `SrcLocation`.

src/messages/mod.rs
```diff
@@ -122,6 +129,43 @@ impl Message {
         Ok(msg)
     }
 
+    /// Creates a message signed using a BLS KeyShare for destination accumulation
+    pub(crate) fn for_dst_accumulation(
+        node: &Node,
+        key_share: &SectionKeyShare,
+        dst: DstLocation,
+        variant: Variant,
```
madadam, line 137: If I understand it correctly, we are currently using this dst accumulation only for `UserMessage`, are we? Although in theory we could use it for other variants as well, that's not how it's implemented in this PR, correct? So that's probably fine, but could be a potential source of confusion as it's now easy for someone to accidentally create invalid message. So why not make this restriction explicit? I'd suggest to change this function so that instead of taking `Variant`, it would take only `Bytes` and construct the `Variant::UserMessage` internally. Similarly, it seems the dst must be only `DstLocation::AccumulatingNode`. So again, construct it internally and only take the `XorName` here. That way the message should be correct by construction. Does it make sense?

src/routing/approved.rs
```diff
@@ -99,6 +101,7 @@ impl Approved {
             network: Network::new(),
             section_keys_provider,
             vote_accumulator: Default::default(),
+            message_accumulator: Default::default(),
```
madadam, line 104: Suggestion for a possible future refactoring (not necessary to be addressed in this PR): instead of adding another accumulator for this, we could have reused `vote_accumulator` instead, and just created a new `Vote` variant for the dst-accumulated messages. That way we would avoid duplicating the accumulation logic.

maqi: the `vote_accumulator` is just a wrapper of `SignatureAggregator` , which I think shall be OK to be used directly ?
maybe we shall re-use it in this PR ?

maqi (approved):

src/routing/approved.rs
madadam: > it may little bit consusing with the existing SendMessage entry

True, but I think it's just a matter of naming. So we would have two vote variants both dealing with message accumulation. One is for accumulating section-src messages and one for node-src messages. So if we reflect this distinction in the name of the respective vote variants, then there shouldn't be much confusion.

madadam (approved):

PR #2329, opened 2021-02-24
Add sibling key to EldersChanged event + API to retrieve any section key

madadam:


joshuef (approved):
LGTM

src/routing/approved.rs
```diff
@@ -160,6 +160,24 @@ impl Approved {
...
 
+    /// Returns the latest known public key of the section with `prefix`.
+    pub fn section_key(&self, prefix: &Prefix) -> Option<&bls::PublicKey> {
+        if prefix == self.section.prefix() || prefix.is_extension_of(self.section.prefix()) {
+            Some(self.section.chain().last_key())
+        } else {
+            self.network.key_by_prefix(prefix).or_else(|| {
+                if self.is_elder() {
+                    // We are elder - the first key is the genesis key
+                    Some(self.section.chain().first_key())
```
maqi, line 171: what if we dont know that prefix section?
maybe better remain as return with None , instead of replace with the genesis key, in case the prefix is not our current prefix ?

src/routing/approved.rs
```diff
@@ -1678,6 +1696,12 @@ impl Approved {
                 commands.extend(self.send_sync(self.section.clone(), self.network.clone())?);
             }
 
+            let sibling_key = if new_prefix != old_prefix {
+                self.section_key(&new_prefix.sibling()).copied()
```
maqi, line 1700: at this momenet, the section_key of the sibling shall be presented, right?

src/network/mod.rs
```diff
@@ -170,6 +170,14 @@ impl Network {
         self.keys.get_matching(name).map(|entry| &entry.value.1)
     }
 
+    /// Returns the latest known key for a section with `prefix`.
+    /// If this returns `None` that means the latest known key is the genesis key.
+    pub fn key_by_prefix(&self, prefix: &Prefix) -> Option<&bls::PublicKey> {
```
maqi, line 175: we have `section_by_name` already, why not just use `section_by_name(prefix.name())` ?

src/routing/approved.rs
madadam: Because this doesn't return the current key of that section, but the last one we know. If we don't know the section at all then the last known key is the genesis key.

src/routing/approved.rs
madadam: Yes, it should be `Some` if that's what you are asking. It's tricky to enforce that statically though.

src/network/mod.rs
madadam: Hm, I'm not sure that's a good idea. So say we call it with prefix `01` but we already progressed and have `010` and `011` in the key map. Then `name()` of `01` is `01000000....` which would be matched by `010` but not `011` even though both are valid successors to `01`. So why would we prefer one over the other?

src/routing/approved.rs
maqi: bit confused here.
the function name is `section_key(prefix)` , indicating the key is for the section of prefix.
meanwhile here we use `our section's first_key(gensis)` to return

madadam: It could have been called `last_known_section_key` or something like that. Decided to keep the name short as there are no other key-retrieval functions this could be confused with. Plus the doc comment should explain it.

> meanwhile here we use our section's first_key(gensis) to return

Thing is that "our section's first key" is also every other section's first key.

PR #2335, opened 2021-02-25
Fork resistant section chain

madadam:
This PR changes the section chain implementation to be resistant to forks. This is implemented by accepting the forks internally, but always presenting an unique and deterministic order of the chain blocks to the outside.


src/routing/tests/mod.rs
```diff
@@ -1279,59 +1279,6 @@ async fn handle_sync() -> Result<()> {
 
 // TODO: add test that untrusted `Sync` is not applied
 
-#[tokio::test]
```
maqi, line 1282: better to have this test refactored to show the forked chain can now be handled?

src/section/mod.rs
```diff
@@ -91,34 +87,20 @@ impl Section {
     /// Try to merge this `Section` with `other`. Returns `InvalidMessage` if `other` is invalid or
     /// its chain is not compatible with the chain of `self`.
     pub fn merge(&mut self, other: Self) -> Result<()> {
-        if !other.chain.self_verify() || !other.elders_info.verify(&other.chain) {
+        if !other.elders_info.self_verify() {
+            error!("can't merge sections: other elders_info failed verification");
```
maqi, line 91: `failed self verification` ?

src/routing/approved.rs
```diff
@@ -494,7 +499,7 @@ impl Approved {
             content: vote.clone(),
             proof_share: proof_share.clone(),
         };
-        let proof_chain = self.section.create_proof_chain_for_our_info(None);
+        let proof_chain = self.section.chain().truncate(1);
```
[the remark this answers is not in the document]
madadam: The thing is that our info is always signed by the last key in the chain, so `proof_chain_for_our_info` would always return a chain containing just the last key. Doesn't seem useful to have such function. 

Can't hurt to add a comment though.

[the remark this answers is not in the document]
madadam: Actually, we don't seem to be using the proof chain at all for the `Vote` variant, so I'm just going to remove it.

src/routing/approved.rs
```diff
@@ -1662,13 +1672,14 @@ impl Approved {
                 sibling.section.elders_info()
             );
 
+            // FIXME: this is duplicated in `update_state`. Do we need it here as well?
```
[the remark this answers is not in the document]
madadam: It's an optimization but quite useful one. If we didn't do this, then after a split, we wouldn't have any entry in the `knowledge` for our sibling section, which means we would have to include the full section chain with any message sent to them until they sent us a proof of their knowledge. So I'd prefer to keep it in some way. 

The question is why is this logic duplicated though. I realized this while working on this PR which is why I put the comment there to remind us in the future. Don't think it needs to be solved in this PR though.

src/section/section_chain.rs
```diff
@@ -0,0 +1,1114 @@
...
+            .enumerate()
+            .skip(new_block.parent_index)
+            .find(|(_, block)| {
+                block.parent_index != new_block.parent_index || block.key >= new_block.key
+            })
+            .map(|(index, _)| index)
+            .unwrap_or(self.tree.len());
+
+        // If the key already exists in the chain, do nothing but still return success to make the
+        // `insert` operation idempotent.
```
maqi, line 343: but the returned value will be `insert_at + 1`, which is not the position of the existing one ?
though curretnly the returned value is ignored by the caller, does it need to be corrected?

src/section/section_chain.rs
```diff
@@ -0,0 +1,1114 @@
...
+        if self.tree.get(insert_at).map(|block| &block.key) != Some(&new_block.key) {
+            self.tree.insert(insert_at, new_block);
+
+            // Adjust the parent indices of the keys whose parents are above the inserted key.
+            for block in &mut self.tree {
```
maqi, line 348: this update only need to be carried out for the blocks AFTER the inserted one AND whose parent is AFTER the inserted one as well ?

i.e. you can just iterate start from self.tree[insert_at + 2] ?

src/section/section_chain.rs
```diff
@@ -0,0 +1,1114 @@
...
+            (Some(lhs), Some(rhs)) => lhs.cmp(&rhs),
+            (Some(_), None) => Ordering::Greater,
+            (None, Some(_)) => Ordering::Less,
+            (None, None) => Ordering::Equal,
+        }
+    }
+
+    /// Returns the number of blocks in the chain. This is always >= 1.
+    pub fn len(&self) -> usize {
+        1 + self.tree.len()
```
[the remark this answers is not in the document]
madadam: Yes, the root block is stored separately from the other blocks because it doesn't have the parent signature so it's a different type and so cannot be stored in the same container.

src/section/section_chain.rs
```diff
@@ -0,0 +1,1114 @@
...
+    }
+
+    fn insert_block(&mut self, new_block: Block) -> usize {
+        // Find the index into `self.tree` to insert the new key at. All the keys above will be
+        // pushed upwards.
```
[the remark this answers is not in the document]
madadam: Maybe the "All the keys above will be pushed upwards." bit can be removed as that's the default behaviour of `Vec::insert`?

src/section/section_chain.rs
```diff
@@ -0,0 +1,1114 @@
...
+        if index == 0 {
+            None
+        } else {
+            self.tree.get(index - 1).map(|block| block.parent_index)
+        }
+    }
+
+    // Is the key at `lhs` an ancestor of the key at `rhs`?
+    fn is_ancestor(&self, lhs: usize, rhs: usize) -> bool {
+        let mut index = rhs;
```
maqi, line 375: what if lhs is 0, i.e. does it mean the root ?

src/section/section_chain.rs
```diff
@@ -0,0 +1,1114 @@
...
+    fn insert_block(&mut self, new_block: Block) -> usize {
+        // Find the index into `self.tree` to insert the new key at. All the keys above will be
+        // pushed upwards.
+        let insert_at = self
+            .tree
+            .iter()
+            .enumerate()
+            .skip(new_block.parent_index)
+            .find(|(_, block)| {
+                block.parent_index != new_block.parent_index || block.key >= new_block.key
```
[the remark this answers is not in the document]
madadam: Blocks that have the same parent are ordered according to the `Ord` impl of `PublicKey`. This is how the chain achieves the total order of the blocks even in the presence of forks. So here we are looking for the first block that either has a different parent or the same parent but a greater key. Then we insert the new block before this block. This is so the new block is inserted at the correct position so the whole chain is still correctly ordered.

src/section/section_chain.rs
```diff
@@ -0,0 +1,1114 @@
...
+
+            expected_keys.push(pk);
+            last_sk = sk;
+        }
+
+        assert_eq!(chain.keys().copied().collect::<Vec<_>>(), expected_keys);
+    }
+
+    #[test]
+    fn insert_fork() {
```
maqi, line 555: what if the fork is 
first `0->1->2` then insert `0->3` ?
will the chain becomes `0 , 3 , 1 , 2` (when 1>3) ,   `0, 1, 2 , 3` (when 1<=3) or  `0, 1 , 3 , 2` (when 1<=3 AND 2 > 3) ?

src/section/section_chain.rs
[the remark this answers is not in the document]
maqi: yeah the current sentence can be removed, as it's vectore::insert default behaviour.
However, it will be better to explain the `rule` used to find the insertion index ? 
i.e. something like the one I got above ?

src/section/section_chain.rs
```diff
@@ -0,0 +1,1114 @@
...
+    pub fn truncate(&self, count: usize) -> Self {
+        let count = count.max(1);
+
+        // Indices of the blocks to include in the result, in reverse order.
+        let mut indices = Vec::with_capacity(count - 1);
+
+        // Index of the currently processed block.
+        let mut index = self.len() - 1;
+
+        // Walk the chain starting from the last key and following the parent links.
```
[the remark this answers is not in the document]
madadam: Yes, this is a concern I'm still thinking about how best to solve. The issue with this is that it can make a node that has only part of the chain get stuck if the section switches to another branch of the chain. I have a couple of ideas how to solve this. I think the simplest would be to never truncate the chain to just the last key, but always keep at least a couple of keys (say 10). That would work most of the time, but it's not 100% fix. A simple 100% fix would be to never truncate the chain at all and always use the full chain. There are other options, but they are more complex. I'll probably add one of those fixes in the next PR (the one with the other bugfixes). I might even put up a post on slack or discourse explaining this in more details.

src/section/section_chain.rs
madadam: No, it actually is used in `merge` so this can be an issue. Good catch!

src/section/section_chain.rs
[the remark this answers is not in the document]
maqi: Ok, got it. thx for the clarification.

src/section/section_chain.rs
madadam: Sorry, I was wrong. It's correct the way it is. It's because `index_at` is an index into `self.tree` but the function returns the "logical" index which is the index when we consider the `root` block as well.

src/section/section_chain.rs
madadam: Good point, but if the new block is the last block then `insert_at + 2` would be an invalid index. I'm going to use `insert_at + 1` instead which should always work.

src/section/section_chain.rs
madadam: Yes, that should work as expected.

src/section/section_chain.rs
madadam: If `1 > 3` then the chain will be `0, 3, 1, 2`. If `1 < 3` then it will be `0, 1, 3, 2`. It will never be `0, 1, 2, 3`.

maqi: will it be worth to have a test, or just expand the current one, to confirm this behaviour ?

maqi: also, I am wondering the case that if there is `3->4` to be inserted later on.
if the chain was `0 , 3 , 1 , 2`, it becomes  `0 , 3 , 4 , 1 , 2` and truncate(2) gives `1, 2`
when the chain was  `0 , 1 , 3 , 2`, it becomes  `0 , 1 , 3 , 4 , 2` and truncate(2) shall still give `1, 2`
right?

src/section/section_chain.rs
```diff
@@ -0,0 +1,1132 @@
...
+            self.tree.get(index - 1).map(|block| block.parent_index)
+        }
+    }
+
+    // Is the key at `lhs` an ancestor of the key at `rhs`?
+    fn is_ancestor(&self, lhs: usize, rhs: usize) -> bool {
+        let mut index = rhs;
+        loop {
+            if index == lhs {
+                return true;
```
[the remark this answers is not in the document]
madadam: Yeah, I guess we do. There is no deep reason for this, it was kind of arbitrary decision. We can change it anytime, as this is not part of the public API. Don't think there is a reason to change it currently though.

src/section/section_chain.rs
madadam: That's right, because `truncate` only follows the parents and ignores the siblings.

src/section/section_chain.rs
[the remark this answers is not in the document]
maqi: ok.

maqi (approved):

PR #2337, opened 2021-02-25
refactor(messaging): add expected aggregation scheme, and use an itin…

oetyng:
…erary

### DIFF
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -40,7 +40,7 @@ pub(crate) struct Message {
     src: SrcAuthority,
     /// Destination location.
     dst: DstLocation,
-    ///
+    /// The aggregation scheme to be used.
     aggregation: Aggregation,
     /// The body of the message.
     variant: Variant,
@@ -311,15 +311,13 @@ impl Message {
     }
 
     /// Elders will aggregate a group sig before
-    /// they all all send one copy of it each to dst.
-    pub fn aggregate_at_src(&self) -> bool {
-        matches!(self.src, SrcAuthority::Section { .. })
-    }
-
-    /// Elders will send their signed message, which
-    /// recipients aggregate.
-    pub fn aggregate_at_dst(&self) -> bool {
-        matches!(self.src, SrcAuthority::BlsShare { .. })
+    /// they all send one copy of it each to dst.
+    pub fn aggregation(&self) -> Aggregation {
+        match self.src {
+            SrcAuthority::Section { .. } => Aggregation::AtSource,
+            SrcAuthority::BlsShare { .. } => Aggregation::AtDestination,
+            SrcAuthority::Node { .. } => Aggregation::None,
+        }
     }
 
     /// Returns the attached proof chain, if any.
--- a/src/messages/mod.rs
+++ b/src/messages/mod.rs
@@ -25,7 +25,7 @@ use crate::{
 use bls_signature_aggregator::ProofShare;
 use bytes::Bytes;
 use serde::{Deserialize, Serialize};
-use sn_messaging::DstLocation;
+use sn_messaging::{Aggregation, DstLocation};
 use std::fmt::{self, Debug, Formatter};
 use thiserror::Error;
 use xor_name::{Prefix, XorName};
@@ -40,6 +40,8 @@ pub(crate) struct Message {
     src: SrcAuthority,
     /// Destination location.
     dst: DstLocation,
+    ///
+    aggregation: Aggregation,
     /// The body of the message.
     variant: Variant,
     /// Proof chain to verify the message trust. Does not need to be signed.
@@ -116,6 +118,7 @@ impl Message {
         let mut msg = Message {
             dst,
             src,
+            aggregation: Aggregation::None,
             proof_chain,
             variant,
             dst_key,
@@ -133,12 +136,12 @@ impl Message {
     pub(crate) fn for_dst_accumulation(
         node: &Node,
         key_share: &SectionKeyShare,
-        dst_node_name: XorName,
+        dst: DstLocation,
         user_msg: Bytes,
         proof_chain: SectionProofChain,
         dst_key: Option<bls::PublicKey>,
+        src_section: XorName,
     ) -> Result<Self, CreateError> {
-        let dst = DstLocation::AccumulatingNode(dst_node_name);
         let variant = Variant::UserMessage(user_msg);
         let serialized = bincode::serialize(&SignableView {
             dst: &dst,
@@ -152,6 +155,7 @@ impl Message {
             signature_share,
         };
         let src = SrcAuthority::BlsShare {
+            src_section,
             proof_share,
             public_key: node.keypair.public,
             age: node.age,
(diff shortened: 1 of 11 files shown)


Yoga07 (dismissed):

src/messages/src_authority.rs
```diff
@@ -36,6 +36,8 @@ pub enum SrcAuthority {
     },
     /// Authority of a single peer that uses it's BLS Keyshare to sign the message.
     BlsShare {
+        /// Section name at the time
+        src_section: XorName,
```
lionel1704, line 40: Should this be prefix instead?

oetyng: Talked with Adam last week about it, really no point using Prefix. From what I understood the idea was to ensure the message goes to the section it came from. But only if there's split would it go anywhere else, and that is the same with prefix as xorname - old section is not there anymore. So can just as well go with xorname.

src/messages/mod.rs
```diff
@@ -40,6 +40,8 @@ pub(crate) struct Message {
     src: SrcAuthority,
     /// Destination location.
     dst: DstLocation,
+    ///
```
maqi, line 43: put some words in?

src/messages/mod.rs
```diff
@@ -306,6 +310,18 @@ impl Message {
         &self.hash
     }
 
+    /// Elders will aggregate a group sig before
+    /// they all all send one copy of it each to dst.
```
maqi, line 314: addition `all`

src/routing/approved.rs
```diff
@@ -2031,73 +2036,62 @@ impl Approved {
...
-    pub fn send_user_message(
-        &mut self,
-        src: SrcLocation,
-        dst: DstLocation,
-        content: Bytes,
-    ) -> Result<Vec<Command>> {
-        if !src.contains(&self.node.name()) {
+    pub fn send_user_message(&mut self, itry: Itinerary, content: Bytes) -> Result<Vec<Command>> {
+        let are_we_src =
+            matches!(itry.src, SrcLocation::Node(_)) && itry.src.name() == self.node.name();
```
maqi, line 2041: just to confirm so this means user_message shall only be sent as `SrcLocation::Node(own_name)`, right?

src/routing/approved.rs
```diff
@@ -2031,73 +2036,62 @@ impl Approved {
...
-                "Not sending user message {:?} -> {:?}: not part of the source location",
-                src, dst
+                "Not sending user message {:?} -> {:?}: we are not the source location",
+                itry.src, itry.dst
             );
             return Err(Error::InvalidSrcLocation);
         }
-
-        if matches!(dst, DstLocation::Direct) {
+        if matches!(itry.src, SrcLocation::EndUser(_)) {
```
maqi, line 2049: this check won't be hit as the above check already filter this out?

src/routing/approved.rs
```diff
@@ -2031,73 +2036,62 @@ impl Approved {
...
-                        content,
-                        self.section().create_proof_chain_for_our_info(None),
-                        None,
-                    )?
-                } else {
-                    let variant = Variant::UserMessage(content);
-                    Message::single_src(&self.node, dst, variant, None, None)?
-                };
-                let mut commands = vec![];
+        // If the source is a single node, we don't even need to vote, so let's cut this short.
```
maqi, line 2060: the comment needs to be updated as now the source can only be a single node?

src/messages/mod.rs
```diff
@@ -152,6 +155,7 @@ impl Message {
             signature_share,
         };
         let src = SrcAuthority::BlsShare {
+            src_section,
```
madadam, line 158: Are we ever going to use anything else than `name(&public_key)` (that is, the name of the sending node) here?

src/messages/mod.rs
```diff
@@ -306,6 +310,18 @@ impl Message {
...
+    /// they all all send one copy of it each to dst.
+    pub fn aggregate_at_src(&self) -> bool {
+        matches!(self.src, SrcAuthority::Section { .. })
+    }
+
+    /// Elders will send their signed message, which
+    /// recipients aggregate.
+    pub fn aggregate_at_dst(&self) -> bool {
+        matches!(self.src, SrcAuthority::BlsShare { .. })
+    }
```
madadam, line 323: Minor suggestion: instead of these two methods, have one which returns `Aggregation`.

src/routing/approved.rs
```diff
@@ -1005,8 +1002,16 @@ impl Approved {
                 message: MessageType::ClientMessage(ClientMessage::from(content)?),
             }]);
         }
-        if let DstLocation::AccumulatingNode(_name) = &dst {
-            if let SrcAuthority::BlsShare { proof_share, .. } = &src {
+        if msg.aggregate_at_dst() {
+            if !matches!(dst, DstLocation::Node(_)) {
+                return Err(Error::InvalidDstLocation);
+            }
```
madadam, line 1008: Why is this an error? Can't we have a dst-accumulated message sent to a section?

src/routing/command.rs
```diff
@@ -71,11 +70,7 @@ pub(crate) enum Command {
         message: MessageType,
     },
     /// Send `UserMessage` with the given source and destination.
-    SendUserMessage {
-        src: SrcLocation,
-        dst: DstLocation,
-        content: Bytes,
-    },
+    SendUserMessage { itry: Itinerary, content: Bytes },
```
madadam, line 73: Minor suggestion which applies to the whole PR, not just this line: I'd spell out `itinerary` in full instead of abbreviating it. It's not very commonly used word and definitely not a commonly used abbreviation, so it just makes the codebase ever so slightly more difficult to read.

src/routing/approved.rs
oetyng: Mm, mistake, should include Section as well.

src/routing/approved.rs
oetyng: No, this is an error, fixing.

src/routing/approved.rs
oetyng: Yep. No change required with the fix resulting from the above comment though 👍

src/messages/mod.rs
oetyng: We now only send in `self.section().prefix().name()` to the method where this is constructed, and I suppose that is the only thing we want there, as passing in a node name would(?) give different behaviour if a response were to be sent after the src section splits. Actually, if sending to a parent name, would both children be equally close? Is the msg sent to both the children sections then?

src/routing/command.rs
oetyng: Yeah, thought of that as well.

src/messages/mod.rs
madadam: If you set it to `prefix.name()` and the section splits, then the response will be received only by one of the subsections (the one last prefix bit is 0). Maybe that's OK?

> Actually, if sending to a parent name, would both children be equally close? Is the msg sent to both the children sections then?

No, would still be sent only to one subsection. The sections are disjoint which means there is always only one section that matches any given name. So there is no way to send to multiple sections while giving the destination as a name only.

madadam: Maybe best to handle this on a case-by-case basis. So for example, if a section is relaying a message from a client, then this should perhaps be set to the client's name. So the response goes to whichever section is still responsible for the client?

madadam (approved):

maqi (approved):

PR #2349, opened 2021-03-03
Bugfixes, part 1

madadam:
First batch of assorted bugfixes: fixes some issues the bootstrap and the stress-test.

### DIFF
--- a/src/routing/bootstrap.rs
+++ b/src/routing/bootstrap.rs
@@ -145,7 +145,7 @@ impl<'a> State<'a> {
             self.send_get_section_request(mem::take(&mut bootstrap_addrs), relocate_details)
                 .await?;
 
-            let (response, sender) = self.receive_get_section_response().await?;
+            let (response, sender) = self.receive_get_section_response(relocate_details).await?;
 
             match response {
                 GetSectionResponse::Success(SectionInfo {
@@ -188,10 +188,9 @@ impl<'a> State<'a> {
     ) -> Result<()> {
         debug!("{} Sending GetSectionQuery to {:?}", self.node, recipients);
 
-        let destination = match relocate_details {
-            Some(details) => *details.destination(),
-            None => self.node.name(),
-        };
+        let destination = relocate_details
+            .map(|details| *details.destination())
+            .unwrap_or_else(|| self.node.name());
 
         let message = SectionInfoMsg::GetSectionQuery(destination);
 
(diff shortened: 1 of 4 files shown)


src/routing/bootstrap.rs
```diff
@@ -156,12 +160,19 @@ impl<'a> State<'a> {
...
-                        "Bootstrapping redirected to another set of peers: {:?}",
-                        new_bootstrap_addrs,
-                    );
-                    bootstrap_addrs = new_bootstrap_addrs.to_vec();
+                GetSectionResponse::Redirect(mut new_bootstrap_addrs) => {
+                    // Ignore already used addresses
+                    new_bootstrap_addrs.retain(|addr| !used_addrs.contains(addr));
+
+                    if new_bootstrap_addrs.is_empty() {
+                        debug!("Bootstrapping redirected to the same set of peers we already contacted - ignoring");
```
maqi, line 168: is that expecting `mem::take` to throw an error on an empty value of bootstrap_addrs to break the loop ?

src/routing/bootstrap.rs
```diff
@@ -213,7 +230,7 @@ impl<'a> State<'a> {
                             return Ok((response, sender))
                         }
                     }
-                }
+                },
```
maqi, line 233: is this added by the cargo-fmt automatically?

src/routing/bootstrap.rs
```diff
@@ -137,11 +137,15 @@ impl<'a> State<'a> {
...
+        // Avoid sending more than one request to the same peer.
+        let mut used_addrs = HashSet::new();
+
         loop {
+            used_addrs.extend(bootstrap_addrs.iter().copied());
             self.send_get_section_request(mem::take(&mut bootstrap_addrs), relocate_details)
                 .await?;
 
-            let (response, sender) = self.receive_get_section_response().await?;
+            let (response, sender) = self.receive_get_section_response(relocate_details).await?;
```
maqi, line 148: just to confirm. so this is to resolve a bug previous that:
keeps contact old section to bootstrap even got relocated to new section.
right?

src/routing/bootstrap.rs
madadam: No. If `bootstrap_addrs` is empty, we won't send any more requests but we'll still wait for more responses.

src/routing/bootstrap.rs
madadam: No. Don't know how it got there. I'll remove it, but unless there are other changes in this PR I'll do it in the next one.

src/routing/bootstrap.rs
madadam: The bug was that as a relocated node, we were checking the response against our current name, but we should have been checking it against the destination name instead. So we were incorrectly rejecting valid responses.

src/routing/bootstrap.rs
maqi: ok. then there will be high chance that after a `time_out` error, the loop got break, i.e. no more GetSectionResponse come in.
right?

maqi (dismissed):

src/routing/bootstrap.rs
madadam: We don't implement any timeout in routing for this, so unless the upper layer adds one, this loops might run forever if we never receive a valid `GetSectionResponse`.

PR #2358, opened 2021-03-04
feat: use supermajority agreement + increase elder size to 7

### LINKED ISSUE #2313: Change quorum to supermajority (> 2/3) and increase elder size to 7
The supermajority is necessary for some aspects of the network like AT2, BRB or fork detection. 7 elders is so we can still lose 2 and be OK.

madadam:
Closes #2313

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,23 +107,38 @@ pub const RECOMMENDED_SECTION_SIZE: usize = 2 * ELDER_SIZE;
 /// Number of elders per section.
 pub const ELDER_SIZE: usize = 7;
 
-/// Number of votes required to agree
-/// with a strict majority (i.e. > 50%)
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -105,25 +105,40 @@ mod section;
 pub const RECOMMENDED_SECTION_SIZE: usize = 10;
```
bochaco, line 105: Perhaps we need to increase this as well now?

madadam: Good point. What we be a good value? Maybe `2 * ELDER_SIZE` ?

bochaco: yeh perhaps it's better to make it a factor of ELDER_SIZE so we don't leave one of them unchanged when one of them changes.

Yoga07 (dismissed):

joshuef (approved):
Works w/ node amster and client master 👍

PR #2366, opened 2021-03-18
Dst aggregation for all variants

madadam:
Closes https://github.com/maidsafe/sn_routing/issues/2315

### DIFF
--- a/src/routing/approved.rs
+++ b/src/routing/approved.rs
@@ -382,10 +382,10 @@ impl Approved {
                 self.handle_their_knowledge_event(prefix, key, proof);
                 Ok(vec![])
             }
-            Vote::SendMessage {
+            Vote::AccumulateAtSrc {
                 message,
                 proof_chain,
-            } => Ok(vec![self.handle_send_message_event(
+            } => Ok(vec![self.handle_accumulate_at_src_event(
                 *message,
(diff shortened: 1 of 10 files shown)


src/routing/approved.rs
```diff
@@ -2085,27 +2110,26 @@ impl Approved {
             return Err(Error::InvalidDstLocation);
         }
 
+        let variant = Variant::UserMessage(content);
+
         // If the msg is to be aggregated at dst, we don't vote among our peers, wemsimply send the msg as our vote to the dst.
```
maqi, line 2115: `wemsimply` -> we simply

src/routing/approved.rs
```diff
@@ -2085,27 +2110,26 @@ impl Approved {
...
         let msg = if itinerary.aggregate_at_dst() {
             Message::for_dst_accumulation(
-                &self.node,
                 self.section_keys_provider.key_share()?,
+                self.section().prefix().name(),
```
maqi, line 2119: why not use `itinerary.src.name()` here?

madadam: It was that before, but I think your suggestion makes more sense. I'll change it.

maqi (approved):

PR #2370, opened 2021-03-22
Remove neighbours restriction

madadam:
Keep all sections, not just neighbours.

### DIFF
--- a/src/delivery_group.rs
+++ b/src/delivery_group.rs
@@ -84,11 +84,8 @@ fn section_candidates(
         .min_by(|lhs, rhs| lhs.prefix.cmp_distance(&rhs.prefix, target_name))
         .unwrap_or_else(|| section.elders_info());
 
-    if info.prefix == *section.prefix() || info.prefix.is_neighbour(section.prefix()) {
+    if info.prefix == *section.prefix() {
         // Exclude our name since we don't need to send to ourself
-
-        // FIXME: only doing this for now to match RT.
-        // should confirm if needed esp after msg_relay changes.
         let section: Vec<_> = info
             .peers()
             .filter(|node| node.name() != our_name)
--- a/src/delivery_group.rs
+++ b/src/delivery_group.rs
@@ -176,3 +176,278 @@ where
     list.truncate(cmp::min(list.len(), ELDER_SIZE));
     list
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        consensus::test_utils::proven,
+        section::{
+            test_utils::{gen_addr, gen_elders_info},
+            EldersInfo, MemberInfo, SectionChain, MIN_AGE,
+        },
+    };
+    use anyhow::{Context, Result};
+    use rand::seq::IteratorRandom;
+    use xor_name::Prefix;
+
+    #[test]
+    fn delivery_targets_elder_to_our_elder() -> Result<()> {
+        let (our_name, section, network, _) = setup_elder()?;
+
+        let dst_name = *section
+            .elders_info()
+            .elders
+            .keys()
+            .filter(|&&name| name != our_name)
+            .choose(&mut rand::thread_rng())
+            .context("too few elders")?;
+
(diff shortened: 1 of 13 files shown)


src/delivery_group.rs
```diff
@@ -179,3 +176,278 @@ where
...
+    #[test]
+    fn delivery_targets_adult_to_our_elder() -> Result<()> {
+        let (our_name, section, network) = setup_adult()?;
+
+        let dst_name = choose_elder_name(section.elders_info())?;
+        let dst = DstLocation::Node(dst_name);
+        let (recipients, dg_size) = delivery_targets(&dst, &our_name, &section, &network)?;
+
+        // Send to all elders
+        assert_eq!(dg_size, section.elders_info().elders.len());
```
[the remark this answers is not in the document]
madadam: That's how worked before this PR. This PR doesn't change the routing logic, it only adds tests for it. That said, perhaps we should change this? Seems like it should be enough to send just to that one elders in that case.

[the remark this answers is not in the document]
maqi: yes. probably previous this behaviour is `hidden`, so that we don't notice it.
I'd prefer to change it unless it got any test broken.

src/routing/lazy_messaging.rs
```diff
@@ -0,0 +1,353 @@
...
+    dst: XorName,
+    nonce: MessageHash,
+    dst_key: Option<bls::PublicKey>,
+) -> Result<Message> {
+    let dst_knowledge = network
+        .knowledge_by_name(&dst)
+        .unwrap_or_else(|| section.chain().root_key());
+    let proof_chain = section
+        .chain()
+        .minimize(vec![section.chain().last_key(), dst_knowledge])?;
```
maqi, line 99: shall the order of these two swapped ?
our own chain shall always being not shorter than other sections' knowledge of us?

src/routing/lazy_messaging.rs
```diff
@@ -0,0 +1,353 @@
...
+        }
+    }
+
+    if let Some(new) = msg.dst_key() {
+        if let Some(src_prefix) = src_prefix {
+            let old = network
+                .knowledge_by_name(&src_name)
+                .unwrap_or_else(|| section.chain().root_key());
+
+            if section.chain().cmp_by_position(new, old) == Ordering::Greater {
```
maqi, line 56: Better put a comment:
When detect other section trusts a newer key from us, casting a TheirKnowledge vote to update own section elders.

src/routing/lazy_messaging.rs
```diff
@@ -0,0 +1,353 @@
...
+                actions.vote.push(Vote::TheirKnowledge {
+                    prefix: *src_prefix,
+                    key: *new,
+                });
+            }
+        }
+
+        if new != section.chain().last_key() {
```
maqi, line 65: better have a comment:
Notify the sender section our latest key when their knowledge of us is outdated.

src/routing/lazy_messaging.rs
```diff
@@ -0,0 +1,353 @@
...
+        variant,
+        Some(proof_chain),
+        dst_key,
+    )?)
+}
+
+#[derive(Default)]
+pub(crate) struct Actions {
+    // Messages to send.
+    pub send: Vec<Message>,
```
maqi: sends, votes ?

src/routing/lazy_messaging.rs
madadam: The order is not significant. `minimize` gives you the shortest sub-chain that contains all the given keys, regardless of in what order you pass them.

src/routing/lazy_messaging.rs
```diff
@@ -0,0 +1,353 @@
...
+            assert_matches!(message.proof_chain(), Ok(chain) => {
+                assert_eq!(chain, env.section.chain());
+            })
+        });
+
+        Ok(())
+    }
+
+    #[test]
+    #[ignore]
```
maqi: not passing yet?

maqi (approved):

PR #2376, opened 2021-03-25
feat!: support adding additional proof chain keys to user messages

madadam:
BREAKING CHANGE: Added `additional_proof_chain_key` parameter to  `Routing::send_message`, added `proof_chain` field to `Event::MessageReceived`.

### DIFF
--- a/src/routing/approved.rs
+++ b/src/routing/approved.rs
@@ -470,7 +470,7 @@ impl Approved {
     }
 
     pub fn handle_dkg_failure(&mut self, proofs: DkgFailureProofSet) -> Result<Command> {
-        let variant = Variant::DKGFailureAgreement(proofs);
+        let variant = Variant::DkgFailureAgreement(proofs);
         let message = Message::single_src(&self.node, DstLocation::Direct, variant, None, None)?;
         Ok(self.send_message_to_our_elders(message.to_bytes()))
     }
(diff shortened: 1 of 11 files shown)


src/routing/approved.rs
```diff
@@ -1046,45 +1044,12 @@ impl Approved {
...
-                        );
-                    }
-                }
-                Err(AggregatorError::NotEnoughShares) => {}
-                Err(err) => {
-                    trace!("Error accumulating message at destination: {:?}", err);
-                }
-            }
-            return Ok(vec![]);
-        }
```
madadam, line 1082: Note for reviewers: Don't mind this removal. It should have been done in a previous PR but somehow slipped through. It's unrelated to the purpose of this PR.

oetyng (approved):

oetyng (approved):

src/routing/mod.rs
```diff
@@ -304,7 +304,16 @@ impl Routing {
     /// Send a message.
     /// Messages sent here, either section to section or node to node are signed
     /// and validated upon receipt by routing itself.
-    pub async fn send_message(&self, itinerary: Itinerary, content: Bytes) -> Result<()> {
+    ///
+    /// `additional_proof_chain_key` is a key to be included in the proof chain attached to the
+    /// message. This is useful when the message contains some data that is signed with a different
+    /// key than the whole message is so that the recipient can verify such key.
```
maqi, line 310: what if the `different key` is a key not from our section ? 
include it into our section proof_chain may incur problem?

madadam: Then the message creation fails and an error is returned.

PR #2377, opened 2021-03-25
feat: keep the genesis key and use it for fallback proofs

### LINKED ISSUE #2372: Fork resolution failure when using truncated chain
Currently only the elders have the full chain (going all the way to the genesis key). Other nodes have only the last couple of keys (often only the very last key). This can cause problems on fork resolution where some nodes might end up permanently out of sync with the rest of the section. Consider this example:

1. The current chain is `A->B->C`
2. A new node joins and gets only the latest key `C`
3. A fork happens and is resolved, so the new chain becomes:
    ```
    A->B->C
       |
       +->D
    ```
    Where D is now the last key.
4. The elders send a `Sync` message to the new node which contains a truncated chain `B->D` (so the last key and its parent, which is usually enough to update the node)
5. The new node knows only `C` so it bounced the message back as untrusted
6. The elders have no way to extend the proof which would make it trusted by the new node so they discard the bounce
7. Thus the new node becomes permanently out if sync because it never progressed from the key `C`

NOTE: the reason why `B->D` cannot be extended to include `C` is that if we extended it to:
```
B->D
|
+->C
```
then it would be impossible to verify it if the node known only `C` as anyone can create a key that signs `C` and pretend it's the real chain:
```
X->Y
|
+->C
```

## Idea for a fix: don't truncate chain

Probably the simplest way to fix this is to never truncate the chain and also always provide the full chain to new nodes via the `NodeApproval` message. Doing this it would always be possible to find a common ancestor of both the latest key and the key the new node known.

## Alternative idea 1: fallback to the full chain

If we assume the genesis key is implicitly trusted by everyone, we can always fall back to sending the full chain in case the proof cannot be extended otherwise. We would just have to make sure everyone known the genesis key. So either hardcode it into the source code, pass it via the config or obtain it by some other means.

## Alternative idea 2: bidirectional chain

Modify the chain so that every block contains also a signature of the parent key created with the child key. That way the chain is verifiable in both directions and so it's always possible to extend the proof.

madadam:
Every node now always has the genesis key. This means we can now always provide valid proof chain for anyone because if all else fails we can always fall back to sending the full section chain.

Closes #2372.

### DIFF
--- a/src/routing/approved.rs
+++ b/src/routing/approved.rs
@@ -824,7 +824,8 @@ impl Approved {
             .section
             .chain()
             .keys()
-            .chain(self.network.keys().map(|(_, key)| key));
+            .chain(self.network.keys().map(|(_, key)| key))
+            .chain(iter::once(self.section.genesis_key()));
 
         match msg.verify(known_keys) {
             Ok(VerifyStatus::Full) => Ok(true),
@@ -1836,6 +1837,7 @@ impl Approved {
         )?;
 
         let variant = Variant::NodeApproval {
+            genesis_key: *self.section.genesis_key(),
             elders_info: self.section.proven_elders_info().clone(),
             member_info,
         };
--- a/src/routing/bootstrap.rs
+++ b/src/routing/bootstrap.rs
@@ -55,7 +55,7 @@ pub(crate) async fn initial(
     let state = State::new(node, send_tx, recv_rx);
 
     future::join(
-        state.run(vec![bootstrap_addr], None),
+        state.run(vec![bootstrap_addr], None, None),
         send_messages(send_rx, comm),
     )
     .instrument(span)
@@ -73,6 +73,7 @@ pub(crate) async fn relocate(
     comm: &Comm,
     recv_rx: mpsc::Receiver<(MessageType, SocketAddr)>,
     bootstrap_addrs: Vec<SocketAddr>,
+    genesis_key: bls::PublicKey,
     relocate_details: SignedRelocateDetails,
 ) -> Result<(Node, Section, Vec<(Message, SocketAddr)>)> {
     let (send_tx, send_rx) = mpsc::channel(1);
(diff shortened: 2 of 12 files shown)


src/routing/approved.rs
```diff
@@ -824,7 +824,8 @@ impl Approved {
             .section
             .chain()
             .keys()
-            .chain(self.network.keys().map(|(_, key)| key));
+            .chain(self.network.keys().map(|(_, key)| key))
```
maqi, line 827: `self.network.keys()` may already contains the `genesis_key` ?

src/routing/bootstrap.rs
```diff
@@ -449,6 +456,13 @@ impl<'a> State<'a> {
                         continue;
                     }
 
+                    if let Some(expected_genesis_key) = expected_genesis_key {
+                        if expected_genesis_key != genesis_key {
+                            error!("Unexpected Genesis key");
```
maqi, line 461: does this have any `real` usage?

src/routing/approved.rs
madadam: Yes, also `section.chain()` as well but there is no harm in that.

src/routing/bootstrap.rs
madadam: it's just to make sure we are joining the right network. Currently this check only applies to relocated nodes but it's not hard to extend it to newly joining nodes too. We could for example bundle the expected genesis key in the release binaries, or something like that.

src/routing/approved.rs
maqi: the type is `vec` ? or it will be `filtered out` later on when map to `BTreeSet` ?

maqi (approved):

src/routing/approved.rs
madadam: It's converted to `HashSet` in `SectionChain::check_trust`.

PR #2378, opened 2021-03-25
feat: last byte of node's name represents its age

maqi:


src/crypto.rs
```diff
@@ -43,18 +43,37 @@ pub fn name(public_key: &PublicKey) -> XorName {
     XorName(public_key.to_bytes())
 }
 
-/// Construct a random `Keypair`
-pub fn gen_keypair() -> Keypair {
-    Keypair::generate(&mut rand::thread_rng())
+/// Construct a random `Keypair` whose name's last byte represents the targeted age.
```
dirvine, line 46: Should this not be part of the gen keypair in a range?

maqi: `gen keypair in a range` is about the `address`

src/section/elders_info.rs
```diff
@@ -103,22 +103,26 @@ pub(crate) mod test_utils {
     }
 
     // Create `count` Nodes sorted by their names.
-    pub(crate) fn gen_sorted_nodes(prefix: &Prefix, count: usize) -> Vec<Node> {
+    pub(crate) fn gen_sorted_nodes(prefix: &Prefix, count: usize, age_diff: bool) -> Vec<Node> {
```
madadam, line 106: I'd add a short comment explaining the `age_diff` parameter.

src/peer.rs
```diff
@@ -78,14 +64,14 @@ pub(crate) mod test_utils {
     // Generate Vec<Peer> where no two peers have the same name.
     pub(crate) fn arbitrary_unique_peers(
         count: impl Into<SizeRange>,
-        age: impl Strategy<Value = u8>,
     ) -> impl Strategy<Value = Vec<Peer>> {
-        proptest::collection::btree_map(arbitrary_xor_name(), (any::<SocketAddr>(), age), count)
-            .prop_map(|peers| {
+        proptest::collection::btree_map(arbitrary_xor_name(), any::<SocketAddr>(), count).prop_map(
```
madadam, line 68: Maybe `arbitrary_xor_name` should take the age strategy as a parameter? Because right now it generates names with completely random ages, including invalid ages like 0, 1, 2 and 3. That might break some tests.

src/messages/src_authority.rs
```diff
@@ -74,9 +74,7 @@ impl SrcAuthority {
     // If this location is `Node`, returns the corresponding `Peer` with `addr`. Otherwise error.
     pub(crate) fn peer(&self, addr: SocketAddr) -> Result<Peer> {
         match self {
-            Self::Node {
-                public_key, age, ..
-            } => Ok(Peer::new(name(public_key), addr, *age)),
+            Self::Node { public_key, .. } => Ok(Peer::new(name(public_key), addr)),
```
madadam, line 77: Seems we don't need the `age` field in the `Node` variant anymore.

madadam (approved):

PR #2383, opened 2021-03-29
Anti Entropy updates

Yoga07:
Supersedes https://github.com/maidsafe/sn_routing/pull/2379

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ qp2p = "~0.11.3"
 rand = "~0.7.3"
 rand_chacha = "~0.2.2"
 resource_proof = "0.8.0"
-sn_messaging = "~13.1.0"
+sn_messaging = { git = "https://github.com/maidsafe/sn_messaging", branch = "anti-entropy" }
 sn_data_types = "~0.18.0"
 thiserror = "1.0.23"
 tokio = "1.3.0"
--- a/src/event.rs
+++ b/src/event.rs
@@ -52,8 +52,6 @@ pub enum Event {
         /// The proof chain for the message, if any.
         proof_chain: Option<SectionChain>,
     },
-    /// The node has been promoted to adult
-    PromotedToAdult,
     /// A new peer joined our section.
     MemberJoined {
         /// Name of the node
@@ -123,7 +121,6 @@ impl Debug for Event {
                 src,
                 dst
             ),
-            Self::PromotedToAdult => write!(formatter, "PromotedToAdult"),
             Self::MemberJoined {
                 name,
                 previous_name,
--- a/src/event.rs
+++ b/src/event.rs
@@ -52,6 +52,8 @@ pub enum Event {
         /// The proof chain for the message, if any.
         proof_chain: Option<SectionChain>,
     },
+    /// The node has been promoted to adult
+    PromotedToAdult,
     /// A new peer joined our section.
     MemberJoined {
         /// Name of the node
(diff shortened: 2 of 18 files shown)


src/routing/approved.rs
```diff
@@ -1996,8 +2103,19 @@ impl Approved {
...
+        let targets: Vec<_> = targets
+            .into_iter()
+            .map(|node| (*node.addr(), *node.name()))
+            .collect();
+        let command = Command::send_message_to_nodes(
+            targets,
+            dg_size,
+            msg.to_bytes(),
+            HeaderInfo {
+                dest: XorName::random(),
```
maqi, line 2115: if it always involve using a random input then later on replacing, 
shall we consider to make the field option?
filling with `random` does looks quite suspicious

src/routing/tests/mod.rs
```diff
@@ -88,26 +93,43 @@ async fn receive_matching_get_section_request_as_elder() -> Result<()> {
...
+    let mut random_pk = keypair.public_key();
+    let mut new_node_name = XorName::from(random_pk);
+
+    while new_node_name.bit(0) {
+        keypair = Keypair::new_ed25519(&mut rng);
+        random_pk = keypair.public_key();
+        new_node_name = XorName::from(random_pk);
+    }
+
+    // new_node_name = bad_prefix.substituted_in(rand::random());
```
maqi, line 120: remove this line?

src/routing/tests/mod.rs
```diff
@@ -882,14 +914,20 @@ async fn handle_unknown_message(source: UnknownMessageSource) -> Result<()> {
         })
         .await?;
 
+    println!("COMMANDSSSSSSSSSSSSSSSS:\n {:?}", commands);
```
maqi, line 917: remove

Cargo.toml
```diff
@@ -22,12 +22,12 @@ lru_time_cache = "~0.11.0"
...
 rand = "~0.7.3"
 rand_chacha = "~0.2.2"
-resource_proof = "0.8.0"
-sn_messaging = "~9.0.0"
-sn_data_types = "~0.17.0"
 thiserror = "1.0.23"
 tokio = "1.3.0"
 xor_name = "1.1.0"
+resource_proof = "0.8.0"
+sn_messaging = "^10.0.0"
```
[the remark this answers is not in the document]
Yoga07: That was because there were breaking changes in sn_messaging at `v10.0.0`. So we need to enforce versions that are only above `10.0.0`.

src/event.rs
```diff
@@ -82,6 +82,8 @@ pub enum Event {
         /// The proof chain for the message.
         proof_chain: SectionChain,
     },
+    /// The node has been promoted to adult
```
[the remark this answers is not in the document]
Yoga07: Must've been from an older rebase, agree we don't need that node anymore.

src/routing/mod.rs
```diff
@@ -357,12 +357,45 @@ impl Routing {
...
+                    msg: message,
+                    hdr_info: HeaderInfo {
+                        dest: XorName::from(end_user_pk),
+                        dest_section_pk: section_pk,
+                    },
+                },
+            };
+            self.stage.clone().handle_commands(command).await
+        } else {
+            warn!(
```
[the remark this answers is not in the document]
Yoga07: >shall we forward to the closest section we know about?

Yea, we do that above at src/routing/mod.rsL#378, but just incase we don't have a section at all we log a message for that. 
Though makes sense to raise an Error too.

src/routing/approved.rs
Yoga07: Yea, makes sense. Makes the code cleaner as well.

PR #2387, opened 2021-03-31
Fix msgs error when sending msgs as Adult

oetyng:
All msg sending expected having a section key share, which made it impossible for Adults to send messages using the api from `sn_node`.


oetyng:
Superseeded by wip

PR #2396, opened 2021-04-01
feat(event): add genesis event

oetyng:
BREAKING CHANGE: Adds new event.


dirvine:
This would not be a breaking change if that enum was declared as #[non_exhustive]

dirvine:
Can you check @oetyng  if this is still required?

oetyng:
> Can you check @oetyng if this is still required?

It's not a very important PR, but would be a bit cleaner with this.

oetyng:
> This would not be a breaking change if that enum was declared as #[non_exhaustive]

Guys mostly in routing didn't want wildcard enum matchings before, so probably don't want to introduce it for this.

oetyng:
Fork branch was overwritten.

PR #2400, opened 2021-04-05
feat: nodes using different ages

maqi:
genesis node use age of 255
nodes of first section use age randomized between 6 and 100
later on node has to be age of 5

### DIFF
--- a/src/routing/bootstrap.rs
+++ b/src/routing/bootstrap.rs
@@ -610,7 +610,7 @@ mod tests {
     use super::*;
     use crate::{
         agreement::test_utils::*, routing::tests::SecretKeySet, section::test_utils::*,
-        section::MemberInfo, ELDER_SIZE, MIN_AGE,
+        section::MemberInfo, ELDER_SIZE, MIN_ADULT_AGE, MIN_AGE,
     };
     use anyhow::{anyhow, Error, Result};
     use assert_matches::assert_matches;
@@ -636,7 +636,7 @@ mod tests {
         let sk = sk_set.secret_key();
         let pk = sk.public_key();
 
-        // Node in first section has to have an age higher than MIN_AGE + 1
+        // Node in first section has to have an age higher than MIN_ADULT_AGE
         // Otherwise during the bootstrap process, node will change its id and age.
         let node_age = MIN_AGE + 2;
         let node = Node::new(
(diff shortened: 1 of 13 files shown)


src/routing/bootstrap.rs
```diff
@@ -123,6 +124,19 @@ impl<'a> State<'a> {
             .bootstrap(bootstrap_addrs, relocate_details.as_ref())
             .await?;
 
+        // For the first section, using age random among 6 - 100 to avoid relocating too many nodes
+        // at the same time.
+        if prefix == Prefix::default() {
```
Yoga07, line 129: Can this be prefix.is_empty()?

src/routing/bootstrap.rs
```diff
@@ -123,6 +124,19 @@ impl<'a> State<'a> {
...
+            let age: u8 = (6..100)
+                .choose(&mut rand::thread_rng())
+                .unwrap_or_else(|| 100);
+            let new_keypair = crypto::gen_keypair(&Prefix::default().range_inclusive(), age);
+            let new_name = crypto::name(&new_keypair.public);
+
+            info!("Changing name to {}", new_name);
```
dirvine, line 136: Should be setting name, not changing name IMO

dirvine (dismissed):

src/routing/core.rs
```diff
@@ -1260,6 +1260,29 @@ impl Core {
...
+                debug!(
+                    "Ignoring JoinRequest from {} - first-section node having wrong age {:?}",
+                    peer,
+                    peer.age(),
+                );
+                return Ok(vec![]);
+            } else {
+                age = peer.age();
+            }
+        } else if peer.age() != MIN_AGE + 1 {
```
Yoga07, line 1277: Can this be a case? New nodes with `age=MIN_AGE+1` can join after startup no?

maqi: After section split, new node `can only` join with age of MIN_AGE+1.
But this reminds me the relocated node shall be the exception.
I will push a fix commit to fix this.

dirvine: Best to identify Infant here as age=4 or joining infant as age = 5 and use that const. That way we keep code simple to read.

maqi: changed to use the const of MIN_ADULT_AGE

Yoga07 (approved):

PR #2421, opened 2021-04-14
Use separate endpoint to test reachability

lionel1704:
Using the same endpoint to check reachability will result in a false positive even when a node is not externally reachable. This fixes this by keeping a separate endpoint to verify external reachability of new peers.
﻿

### DIFF
--- a/src/routing/comm.rs
+++ b/src/routing/comm.rs
@@ -23,8 +23,6 @@ use tokio::{sync::mpsc, task};
 pub(crate) struct Comm {
     _quic_p2p: QuicP2p,
     endpoint: Endpoint,
-    // Additional endpoint used to check if peers are externally reachable
-    connectivity_endpoint: Endpoint,
     // Sender for connection events. Kept here so we can clone it and pass it to the incoming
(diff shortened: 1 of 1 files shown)


Yoga07 (dismissed):

Yoga07 (approved):

src/routing/comm.rs
```diff
@@ -123,7 +141,7 @@ impl Comm {
 
     /// Tests whether the peer is reachable.
     pub async fn is_reachable(&self, peer: &SocketAddr) -> Result<(), SendError> {
-        self.endpoint
+        self.connectivity_endpoint
```
maqi, line 144: just to confirm that this means we have to carry this `test only endpoint` along the whole life.
though qp2p/quinn may take care of the idle issue, still a waste for some underneath keep-alive resources.
shall we instead here create a temp endpoint to test ? i.e. discard after the test immediately ?

lionel1704: Yeah that is the other option. Let's discuss this on Slack with the team.

maqi (dismissed):

Yoga07 (approved):

PR #2432, opened 2021-04-21
feat(api): add new API for an Elder node to propose that a node has gone offline

lionel1704:


src/routing/core.rs
```diff
@@ -491,7 +491,7 @@ impl Core {
         self.propose_offline(name)
     }
 
-    fn propose_offline(&self, name: XorName) -> Result<Vec<Command>> {
+    pub fn propose_offline(&self, name: XorName) -> Result<Vec<Command>> {
```
maqi, line 494: this will actually cause an error when a non-elder calls it.
better have a safe-guard that only `elder` makes such call.
also, the commit/PR title better to be : new API for elder sn_node to propose a node has gone offline

maqi (approved):

PR #2458, opened 2021-04-27
fix: only send to client directly when it belongs to self section

maqi:


src/routing/core.rs
```diff
@@ -1099,8 +1100,10 @@ impl Core {
                 }
             };
             if recipients.is_empty() {
+                trace!("cannot routing user message {:?}", msg);
```
Yoga07, line 1103:
```suggestion
                trace!("Cannot route user message, recipient list empty: {:?}", msg);
```

src/routing/mod.rs
```diff
@@ -357,24 +357,38 @@ impl Routing {
...
-                .lock()
-                .await
-                .get_socket_addr(socket_id)
-                .copied();
-
-            if let Some(socket_addr) = socket_addr {
-                return self
-                    .send_message_to_client(socket_addr, ClientMessage::from(content)?)
-                    .await;
+            let name = if let Some(pub_key) = public_key.ed25519() {
```
Yoga07, line 360:
```suggestion
            let name =XorName::from(public_key)
```
As the client can have any type of Keypair(BLS/Ed). And we do have a `From` implemented for `sn_data_types::PublicKey` [here](https://github.com/maidsafe/sn_data_types/blob/095e22eb36b381a01397bb7708878d0d56582974/src/keys/public_key.rs#L177).

Yoga07 (approved):

PR #2466, opened 2021-04-28
chore(err): remove qp2p error as a public type

Dylan-DPC:
removes qp2p errors as being exported and converts them into custom errors

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -19,13 +19,18 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 /// Internal error.
 #[derive(Debug, Error)]
 #[allow(missing_docs)]
+#[allow(clippy::large_enum_variant)]
 pub enum Error {
     #[error("Failed signature check.")]
     FailedSignature,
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -19,13 +19,18 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
...
     #[error("Failed signature check.")]
     FailedSignature,
     #[error("Cannot route.")]
     CannotRoute,
-    #[error("Network layer error: {0}")]
-    Network(#[from] qp2p::Error),
+    #[error("The config is invalid")]
+    InvalidConfig,
+    #[error("Cannot connect to the endpoint")]
+    CannotConnectEndpoint,
```
lionel1704, line 31: Perhaps we could add some context to these errors to help identify the cause using:
`.map_err(|err| ErrorVariant(err.to_string()))`

Dylan-DPC: yeh will do that as part of the next step as I need to do it to the other variants as well

PR #2468, opened 2021-04-29
Anti entropy

Yoga07:


src/messages/src_authority.rs
```diff
@@ -41,7 +41,7 @@ pub enum SrcAuthority {
     },
     /// Authority of a whole section.
     Section {
-        /// Name in the source section.
+        /// Name of the source section.
```
maqi, line 44: I think `in` is more accurate. `name of a section` will be just prefix tailed with all zero.

src/routing/comm.rs
```diff
@@ -179,11 +173,29 @@ impl Comm {
         // Run all the sends concurrently (using `FuturesUnordered`). If any of them fails, pick
         // the next recipient and try to send to them. Proceed until the needed number of sends
         // succeeds or if there are no more recipients to pick.
-        let send = |recipient, msg| async move { (self.send_to(recipient, msg).await, recipient) };
-
+        let send = |recipient: (SocketAddr, XorName), mut msg: MessageType| async move {
+            msg.update_dest_info(None, Some(recipient.1));
```
maqi, line 177: I do think make that `DestInfo` an optional field , and only fill it at the constructing place when necessary will be better than here that have to `update` it separately later on.

src/routing/comm.rs
```diff
@@ -302,9 +316,11 @@ mod tests {
     use assert_matches::assert_matches;
     use futures::future;
     use qp2p::Config;
+    use sn_messaging::{DestInfo, WireMsg};
     use std::{net::Ipv4Addr, slice, time::Duration};
     use tokio::{net::UdpSocket, sync::mpsc, time};
 
+    #[allow(unused)]
```
maqi, line 323: remove if no longer used?

src/routing/comm.rs
```diff
@@ -442,24 +495,29 @@ mod tests {
...
+                assert_eq!(WireMsg::deserialize(msg)?, msg0);
                 msg0_received = true;
                 recv_endpoint.disconnect_from(&src)?;
             }
             assert!(msg0_received);
         }
 
         // Send the second message.
-        let msg1 = Bytes::from_static(b"one");
+        // Send the first message.
```
maqi, line 506: comment got mixed up with above line?

src/routing/comm.rs
```diff
@@ -530,10 +600,12 @@ mod tests {
                 rx,
                 _incoming_connections: incoming_connections,
                 _disconnections: disconnections,
+                _name: XorName::random(),
             })
         }
     }
 
+    #[allow(unused)]
```
maqi, line 608: remove if no longer used.

src/routing/mod.rs
```diff
@@ -400,14 +400,47 @@ impl Routing {
...
-            delivery_group_size: 1,
-            message: MessageType::ClientMessage(message),
+        let end_user = self
+            .dispatcher
+            .core
+            .lock()
+            .await
+            .get_enduser_by_addr(&recipient)
+            .copied();
+        let end_user_pk = match end_user {
```
maqi, line 412: This function is being called by `send_message` , which shall already have the `public_key` from the `itinerary.dst` at the calling place.
why not just pass that in, instead carry out another round of fetch?

src/agreement/dkg.rs
```diff
@@ -635,9 +638,13 @@ impl DkgCommand {
                 let message = Message::single_src(node, DstLocation::Direct, variant, None, None)?;
 
                 Ok(Command::send_message_to_nodes(
-                    &recipients,
+                    recipients.clone(),
                     recipients.len(),
                     message.to_bytes(),
+                    DestInfo {
```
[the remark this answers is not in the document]
Yoga07: >using a random dest

The random dest is just a placeholder as when we send our message to multiple nodes, it gets updated to each node (here)[https://github.com/maidsafe/sn_routing/blob/41b090874a7e93c5e9ac8577d9ec7553dbaa8253/src/routing/comm.rs#L177].

>a passed in dest_section_pk seems bit strange

I believe the passed in key is the PK for that session DKG, hence that would be the dest_section_key for all. Do correct me if I am wrong here :)

src/delivery_group.rs
```diff
@@ -39,16 +39,16 @@ pub(crate) fn delivery_targets(
     our_name: &XorName,
     section: &Section,
     network: &Network,
-) -> Result<(Vec<Peer>, usize)> {
+) -> Result<(Vec<Peer>, usize, bls::PublicKey)> {
```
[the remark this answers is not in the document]
Yoga07: The `delivery_targets` funtion now returns the relaying section's dst_section_pk also, as to add that in the DestInfo. Will update the comment 👍

src/delivery_group.rs
```diff
@@ -39,16 +39,16 @@ pub(crate) fn delivery_targets(
...
     if !section.is_elder(our_name) {
         // We are not Elder - return all the elders of our section, so the message can be properly
         // relayed through them.
         let targets: Vec<_> = section.elders_info().peers().copied().collect();
         let dg_size = targets.len();
-        return Ok((targets, dg_size));
+        return Ok((targets, dg_size, *section.chain().last_key()));
```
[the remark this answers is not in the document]
Yoga07: >why not just use it at the calling place, then no need to manipulate this function API?

The issue here is that we need to find the correct destination section's PK for all kinds of scenarios i.e. 
1) If we are not Elders we route the msg to our elders(requires our section PK).
2) If we are Elders then:
    a) check if the msg is for other section and fetch their dst section PK. 
    b) check if the msg is for a node and forward it to that node's section, which also requires their section PK.
    c) check if the msg is for a client, which again should contain our section PK.
    
    
So this API covers a lot of scenarios for relaying the messages, so it's better to return the dst_section_pk along with the other details as well. Morever we only use this API during relaying messages only, so it isn't affecting much of the code :)

src/routing/comm.rs
Yoga07: Agreed 👍 But I would like to do this in a separate PR as the changes would span all across the repos.

src/routing/mod.rs
Yoga07: We could do even better here, since we are the client's own section(proved by us having the client's socket_addr in calling function), we can just simply add our section PK instead of matching and finding. Will update 👍

src/routing/bootstrap.rs
```diff
@@ -218,16 +224,38 @@ impl<'a> State<'a> {
...
+                PublicKey::from(details.relocate_details()?.destination_key),
+                *details.destination()?,
+            ),
+            None => (PublicKey::from(self.node.keypair.public), self.node.name()),
         };
 
-        let message = SectionInfoMsg::GetSectionQuery(destination);
+        let message = SectionInfoMsg::GetSectionQuery(PublicKey::from(self.node.keypair.public));
+
+        // Group up with our XorName as we do not know their name yet.
```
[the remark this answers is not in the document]
Yoga07: Yep, it is our name, but with AE we are starting to send our `sn_data_types::PublicKey`, which the network then converts to Xor gives us the appropriate section.

src/routing/bootstrap.rs
```diff
@@ -218,16 +224,38 @@ impl<'a> State<'a> {
...
+        let recipients = recipients
+            .iter()
+            .map(|addr| (*addr, dest_xorname))
+            .collect();
 
+        let dest_info = DestInfo {
+            dest: dest_xorname,
+            dest_section_pk: PublicKey::bls(&dest_pk).unwrap_or_else(|| {
+                // Create a random PK as we'll be getting the right one with the response
+                bls::SecretKey::random().public_key()
```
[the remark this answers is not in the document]
Yoga07: True, though I'd like to do this optimization along with the one mentioned (here)[https://github.com/maidsafe/sn_routing/pull/2468#discussion_r623195833] as they cover the same scope, and would be easier to get them done in one separate PR.

src/routing/bootstrap.rs
```diff
@@ -417,15 +472,17 @@ impl<'a> State<'a> {
...
-                }
-                MessageType::Ping | MessageType::ClientMessage(_) | MessageType::SectionInfo(_) => {
-                    continue
-                }
+            let (message, dest_info) = match message {
+                MessageType::Routing {
+                    msg: RoutingMsg(msg_bytes),
+                    dest_info,
+                } => (Message::from_bytes(Bytes::from(msg_bytes))?, dest_info),
+                MessageType::Node { .. }
```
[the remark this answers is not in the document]
Yoga07: Yup, although we send sn_node messages through `SendUserMessage` in routing now, this flow is yet to come into action.

src/routing/core.rs
```diff
@@ -432,30 +461,43 @@ impl Core {
...
 
-        // Try to send a "ping" message to probe the peer connection. If it succeeds, the
-        // connection loss was just temporary. Otherwise the peer is assumed lost and we will
-        // propose it offline.
-        Some(Command::SendMessage {
-            recipients: vec![addr],
-            delivery_group_size: 1,
-            message: MessageType::Ping,
-        })
+        self.propose_offline(name)
```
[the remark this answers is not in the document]
Yoga07: An error during the rebase there, added the ping message back in now.

src/routing/lazy_messaging.rs
```diff
@@ -275,13 +280,12 @@ mod tests {
...
-            *env.section.chain().last_key(),
-        )?;
-
-        let actions = process(&env.node, &env.section, &env.network, &msg)?;
+        let msg = env.create_message(&env.their_prefix, proof_chain)?;
+        let dest_info = DestInfo {
+            dest: XorName::random(),
+            dest_section_pk: *env.section.chain().last_key(),
+        };
+        let actions = process(&env.node, &env.section, &env.network, &msg, dest_info)?;
```
maqi, line 288: better have the tests to check the `dest` field has been updated after process?

src/routing/mod.rs
```diff
@@ -472,28 +497,41 @@ async fn handle_message(dispatcher: Arc<Dispatcher>, bytes: Bytes, sender: Socke
...
+            Err(error) => {
+                error!("Failed to deserialize node message: {}", error);
+            }
+        },
+        MessageType::Node {
+            msg: _,
+            dest_info: _,
+            src_section_pk: _,
+        } => {
+            unimplemented!()
```
maqi, line 532: better remove it if not implemented yet?

tests/messages.rs
```diff
@@ -100,14 +106,15 @@ async fn test_messages_client_node() -> Result<()> {
...
-        let expected_bytes = query.serialize()?;
+        let expected_bytes = query.serialize(XorName::from(pk), section_key)?;
         assert_eq!(resp, expected_bytes);
     }
 
     Ok(())
 }
 
 #[tokio::test]
+#[ignore]
```
maqi, line 117: why the test ignored?

src/routing/core.rs
```diff
@@ -2041,25 +2169,32 @@ impl Core {
...
         };
 
         let mut commands = vec![];
 
-        let adults: Vec<_> = self.section.live_adults().copied().collect();
+        let adults: Vec<_> = self
+            .section
+            .adults()
+            .map(|peer| (*peer.addr(), *peer.name()))
+            .collect();
```
maqi, line 2191: shall use `live_adults()`

Yoga07:
Superseded by #2491

PR #2469, opened 2021-04-29
section authority provider

maqi:
replaces https://github.com/maidsafe/sn_routing/pull/2393

TODO: move section_key into SectionAuthorityProvider
as discussed, this will be left over to the upcoming PR


src/event.rs
```diff
@@ -33,6 +33,17 @@ pub enum NodeElderChange {
     None,
 }
 
+/// Bound name of elders and section_key, section_prefix info together.
+#[derive(Debug, Clone, PartialEq)]
```
dirvine, line 37: This is a `SectionAuthorityProvider` here. So this struct can go along with SectionInfo AFAIK

PR #2485, opened 2021-05-06
WIP: refactor(AE): remove voting on AE messages

Yoga07:


src/routing/lazy_messaging.rs
```diff
@@ -97,9 +103,9 @@ fn create_other_section_message(
     node: &Node,
     section: &Section,
     network: &Network,
+    variant: Variant,
     dst: DstLocation,
-    nonce: MessageHash,
-    // dst_key: Option<bls::PublicKey>,
+    // nonce: MessageHash,
 ) -> Result<Message> {
     let dst_knowledge = dst
```
maqi, line 110: better have `dst_knowledge` as an input parameter.
as now knowledge is updated individually. so our record only reflects the most advanced one.

src/routing/lazy_messaging.rs
```diff
@@ -97,9 +103,9 @@ fn create_other_section_message(
     node: &Node,
     section: &Section,
     network: &Network,
+    variant: Variant,
```
maqi, line 106: as mentioned above, I think we can retain the OtherSection, without introducing those new variants.

src/routing/lazy_messaging.rs
```diff
@@ -36,58 +35,65 @@ pub(crate) fn process(
...
-            //
-            // NOTE: the reason why we update the key only after receiving their `OtherSection` and
-            // not right here is to keep their key and their elders info in sync.
-            trace!("lazy messaging: src key unknown");
-            send_other_section = true;
+    if let Ok(src_chain) = msg.proof_chain() {
+        if let Some(key) = network.key_by_name(&src_name) {
+            match src_chain.cmp_by_position(src_chain.last_key(), key) {
+                Ordering::Greater => {
+                    trace!("Anti-Entropy: We do not know source's key, need to update ourselves");
```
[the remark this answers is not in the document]
dirvine: >send an OtherSection to the sender to notify it that we now get the knowledge updated to the newer key

Are you sure? We don't need to tell anyone we updated? We may resend a message that required us to be updated but we don't tell folk when we update.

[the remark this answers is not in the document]
Yoga07: >notify it that we now get the knowledge updated to the newer key

This should be with the `knowledge` map we are using. Since we'll be removing that in this PR, we don't need to notify other sections now and keep it just as individual messages.

src/routing/core.rs
```diff
@@ -2204,7 +2238,7 @@ impl Core {
         // We need to construct a proof that would be trusted by the destination section.
         let known_key = self
```
maqi, line 2239: this shall be `our` key that trusted by destination section.
not `destination section's key` .

src/routing/core.rs
```diff
@@ -2557,7 +2594,7 @@ impl Core {
         // the actual last key of the resulting proof chain because it's the key that will be used
         // to sign the message.
         let additional_key = additional_key
-            .or_else(|| self.network.knowledge_by_name(&dst.name()?))
+            .or_else(|| self.network.key_by_name(&dst.name()?))
```
maqi, line 2597: this shall be our key that trusted by destination section.
not destination section's key .

dirvine: We don't now want this @maqi  is that right? Other sections key is not our concern, AE will update if it has to but we cannot know what they know? (anybody can update them)

Yoga07:
Superseded by #2491

PR #2488, opened 2021-05-07
[WIP] refactor(core): remove Section::genesis_key field since it's redundant with the Section::chain::root value

bochaco:


src/messages/variant.rs
```diff
@@ -44,9 +44,9 @@ pub(crate) enum Variant {
     /// Message sent to newly joined node containing the necessary info to become a member of our
     /// section.
     NodeApproval {
-        genesis_key: bls::PublicKey,
         section_auth: Proven<SectionAuthorityProvider>,
         member_info: Proven<MemberInfo>,
+        section_chain: SectionChain,
```
maqi, line 49: the message itself already contains the section_chain, which is already used: https://github.com/maidsafe/sn_routing/blob/master/src/routing/bootstrap.rs#L501
no need to have a separate field for it.

PR #2489, opened 2021-05-07
feat: move section_key into SectionAuthorityProvider

maqi:


src/section/section_authority_provider.rs
```diff
@@ -61,6 +65,18 @@ impl SectionAuthorityProvider {
     pub(crate) fn position(&self, name: &XorName) -> Option<usize> {
         self.elders.keys().position(|other_name| other_name == name)
     }
+
+    pub(crate) fn set_pk_set(&mut self, pk_set: ReplicaPublicKeySet) {
+        self.pk_set = Some(pk_set);
+    }
+
+    pub(crate) fn elders_only(&self) -> Self {
```
bochaco, line 73: I think the name is misleading, I was about to ask "but the SAP is supposed to contain only elders!?", so maybe call it `clone_without_pk_set` ?

bochaco (commented):
I've gone through it all, I don't have any other comments with my current knowledge, apart from what we discussed @maqi about decoupling the node's state from what is exchanged on messages, that we'd need to remove the `Option<ReplicaPublicKeySet>`

src/section/section_authority_provider.rs
```diff
@@ -25,11 +26,13 @@ pub struct SectionAuthorityProvider {
     pub elders: BTreeMap<XorName, SocketAddr>,
     /// The section prefix. It matches all the members' names.
     pub prefix: Prefix,
+    /// Public key set of the section.
+    pub pk_set: Option<ReplicaPublicKeySet>,
```
bochaco, line 30: I'm not sure we should use `sn_data_types::ReplicaPublicKeySet` for the pk-set, I see they are the same but sn_data_types::ReplicaPublicKeySet is meant as a type for the transfers types and messages, so perhaps we should use directly the `threshold_crypto::PublicKeySet` ?

dirvine: This should not be a PublicKey set at all but the SectionKey only. So BLS::PublicKey

src/section/section_authority_provider.rs
```diff
@@ -61,6 +65,18 @@ impl SectionAuthorityProvider {
     pub(crate) fn position(&self, name: &XorName) -> Option<usize> {
         self.elders.keys().position(|other_name| other_name == name)
     }
+
+    pub(crate) fn set_pk_set(&mut self, pk_set: ReplicaPublicKeySet) {
+        self.pk_set = Some(pk_set);
+    }
+
+    pub(crate) fn clone_without_pk_set(&self) -> Self {
```
dirvine, line 73: Without the Section key this is useless as an Elder with no Section Key to identify him is unable to tell us anything. We should never have this and if we use it that place has bugs ;-)

maqi: The place we use this is at https://github.com/maidsafe/sn_routing/blob/master/src/routing/core/messaging/handling/agreement.rs#L192
when handling SectionInfo, to ensure the `agreement` is not about an out-dated candidates.
Stuff from `promote_and_demote_elders` cannot to have key-related info.
Hence this mapping util function is required.

dirvine: However Adults,clients and Elders will now all hold the SAP for the current key. So can we now use that? So promote / demote etc. have this and get a new one and can see if they are Edlers or not from the difference?

PR #2494, opened 2021-05-09
refactor(event): add SectionSplit, increase granularity

oetyng:
- Removes the `sibling_elders: Option<Elders>` field from `EldersChanged`.
- Increases event data granularity by detailing the actual changes.
- Same granularity applied to `AdultsChanged`.
BREAKING CHANGE: `Event` enum variants changed and new added.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -40,8 +40,12 @@ pub struct Elders {
     pub prefix: Prefix,
     /// The BLS public key of a section.
     pub key: bls::PublicKey,
-    /// The set of elders of a section.
-    pub elders: BTreeSet<XorName>,
+    /// Remaining Elders in our section.
+    pub remaining: BTreeSet<XorName>,
+    /// New Elders in our section.
+    pub added: BTreeSet<XorName>,
+    /// Removed Elders in our section.
+    pub removed: BTreeSet<XorName>,
 }
 
 /// An Event raised by a `Node` or `Client` via its event sender.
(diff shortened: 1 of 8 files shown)


src/event.rs
```diff
@@ -118,7 +128,14 @@ pub enum Event {
     /// Failed in sending a message to client, or connection to client is lost
     ClientLost(SocketAddr),
     /// Notify the current list of adult nodes, in case of churning.
-    AdultsChanged(BTreeSet<XorName>),
+    AdultsChanged {
+        /// Existing Adults in our section.
+        existing: BTreeSet<XorName>,
```
maqi, line 133: according to the calculating
```
let existing = old_adults.intersection(&current_adults).copied().collect();
```
naming of `remaining` shall be more accurate. `existing` is little bit mis-leading.

src/routing/core/messaging/handling/mod.rs
```diff
@@ -452,13 +453,22 @@ impl Core {
...
+                .map(|p| p.name())
+                .copied()
                 .collect();
-            if old_adults != new_adults {
-                self.send_event(Event::AdultsChanged(new_adults));
+            let added = current_adults.difference(&old_adults).copied().collect();
+            let removed = old_adults.difference(&current_adults).copied().collect();
+            let existing = old_adults.intersection(&current_adults).copied().collect();
+
+            if old_adults != current_adults {
```
maqi, line 466: `if !added.is_empty() && !removed.is_empty()` for less computation?

src/routing/tests/mod.rs
```diff
@@ -1493,7 +1495,9 @@ async fn handle_elders_update() -> Result<()> {
         event_rx.recv().await,
         Some(Event::EldersChanged { elders, .. }) => {
             assert_eq!(elders.key, pk1);
-            assert_eq!(elders.elders, elder_names1);
+            assert!(elders.added.iter().all(|a| elder_names1.contains(a)));
+            assert!(elders.existing.iter().all(|a| elder_names1.contains(a)));
```
maqi, line 1499: `assert_eq!(elder_names1, elders.added.union(&elders.existing))` shall be enough and more accurate for the above two `all` check?

PR #2497, opened 2021-05-10
feat(smoke): make smoke test interactive

Dylan-DPC:


dirvine:
Currently, this requires the schedule to be added on cmd line but ignores it in interactive mode. Also in interactive mode pressing return exits prog. It would be nice to be able to add/remove several nodes at once.

PR #2511, opened 2021-05-15
refactor(messaging): remove the RegisterEndUser messaging handling and flows

bochaco:
BREAKING CHANGE: this version uses a non backward-compatbile version of sn_messaging


dirvine:
LGTM

PR #2520, opened 2021-05-24
Digging

joshuef:


S-Coyle:
Closing as repo is being archived - to be raised against https://github.com/maidsafe/safe_network if still required.

PR #2523, opened 2021-05-25
Remove proofchain from all messages and send only when necessary

Yoga07:


src/routing/core/anti_entropy.rs
```diff
@@ -43,41 +41,6 @@ pub(crate) fn process(
 
     let dst = msg.src().src_location().to_dst();
 
-    if let Ok(src_chain) = msg.proof_chain() {
-        if let Some(key) = network.key_by_name(&src_name) {
```
maqi, line 47: do we need to check whether the incoming src_key is known to us?
and if unknown, shall we ask for an update?

src/routing/core/messaging/handling/bad_msgs.rs
```diff
@@ -82,11 +82,11 @@ impl Core {
                         section,
                         network: network.clone(),
                     },
-                    None,
+                    self.section.authority_provider().section_key,
                 )?
             }
             _ => bounced_msg
-                .extend_proof_chain(&dst_key, self.section.chain())
+                .updated_with_latest_key(self.section.authority_provider().section_key)
```
maqi, line 89: for `AggregationAtSrc` message, updating to a new section_key means they have to carry out a new round of vote using the new section_key_share .

src/routing/core/messaging/sending.rs
```diff
@@ -362,7 +356,12 @@ impl Core {
         variant: Variant,
         dst_pk: bls::PublicKey,
     ) -> Result<Command> {
-        let message = Message::single_src(&self.node, DstLocation::Direct, variant, None)?;
+        let message = Message::single_src(
+            &self.node,
+            DstLocation::Direct,
+            variant,
+            self.section.authority_provider().section_key,
```
maqi, line 363: What's the message to update destion section's knowledge of us? 
i.e. what will happen when we shifted from key_A to key_B ? which means we will then send messages with key_B, which is unknown to the destination section. Then how it knows this new key_B is a trust-worth key ?

src/routing/core/messaging/handling/bad_msgs.rs
Yoga07: We only send it back with just the latest keys that we know since the original message could have been sent with a later key and is now being bounced back,

src/messages/variant.rs
```diff
@@ -0,0 +1,300 @@
...
+    /// Inform other sections about our section or vice-versa.
+    SectionKnowledge {
+        /// `SectionAuthorityProvider` and `SectionChain` of the sender's section, with the proof chain.
+        src_info: (Proven<SectionAuthorityProvider>, SectionChain),
+        /// Message
+        msg: Option<Box<Message>>,
+        // Nonce that is derived from the incoming message that triggered sending this
+        // message. It's purpose is to make sure that `OtherSection`s that are identical
+        // but triggered by different messages are not filtered out.
+        // nonce: MessageHash,
```
joshuef, line 44: lets remove this

src/messages/mod.rs
```diff
@@ -92,27 +92,23 @@ pub trait RoutingMsgUtils {
...
-    // Extend the current message proof chain so it starts at `new_first_key` while keeping the
-    // last key (and therefore the signature) intact.
-    // NOTE: This operation doesn't invalidate the signatures because the proof chain is not part of
-    // the signed data.
-    fn extend_proof_chain(
-        self,
-        new_first_key: &bls::PublicKey,
-        full_chain: &SecuredLinkedList,
-    ) -> Result<RoutingMsg, Error>;
+    // // Extend the current message proof chain so it starts at `new_first_key` while keeping the
```
joshuef, line 95: lets clean up these comments

src/messages/mod.rs
```diff
@@ -405,69 +398,54 @@ impl RoutingMsgUtils for RoutingMsg {
...
-
-        RoutingMsg::new_signed(self.src, self.dst, self.variant, self.proof_chain)
-    }
-
-    fn verify_variant<'a, I>(
-        &self,
-        proof_chain: Option<&SecuredLinkedList>,
-        trusted_keys: I,
-    ) -> Result<VerifyStatus>
+    // // Extend the current message proof chain so it starts at `new_first_key` while keeping the
```
joshuef, line 401: and here

src/routing/core/messaging/handling/mod.rs
```diff
@@ -53,7 +53,11 @@ impl Core {
 
         // Check if the message is for us.
         let in_dst_location = msg.dst().contains(&self.node.name(), self.section.prefix());
-        if !in_dst_location || msg.dst().is_section() {
```
maqi, line 56: removing `msg.dst().is_section()` means when the sender knows only part of the receivers, the msg won't reach all.
For node->section msgs, it may cause issue.
Meanwhile, I agree this will reduce certain amount of traffic. 
Worth to at least put a comment here, mark as `TODO` or `FIXME`

src/routing/core/mod.rs
```diff
@@ -201,6 +201,7 @@ impl Core {
                         dest: XorName::random(),
                         dest_section_pk: sap.section_key,
                     };
+                    info!("Sending updated SectionInfo to all known sections");
```
maqi, line 204: better put this trace out of the for clause to avoid un-necessary log.

Cargo.toml
```diff
@@ -23,7 +23,8 @@ qp2p = "~0.11.9"
 rand = "~0.7.3"
 rand_chacha = "~0.2.2"
 resource_proof = "0.8.0"
-sn_messaging = "27.0.0"
+# sn_messaging = "27.0.0"
+sn_messaging = { git = "https://github.com/yoga07/sn_messaging", branch = "remove-proofchain" }
```
maqi, line 27: just a mark, change before merge

src/messages/mod.rs
```diff
@@ -424,31 +426,26 @@ impl RoutingMsgUtils for RoutingMsg {
...
-        &self,
-        proof_chain: Option<&SecuredLinkedList>,
-        trusted_keys: I,
-    ) -> Result<VerifyStatus>
+    fn verify_variant<'a, I>(&self, trusted_keys: I) -> Result<VerifyStatus>
     where
         I: IntoIterator<Item = &'a bls::PublicKey>,
     {
         let proof_chain = match &self.variant {
             Variant::NodeApproval {
```
maqi, line 396: SectionKnowledge now contains a chain as well.
does it need to be verified?

src/routing/core/messaging/handling/mod.rs
```diff
@@ -53,13 +53,9 @@ impl Core {
 
         // Check if the message is for us.
         let in_dst_location = msg.dst().contains(&self.node.name(), self.section.prefix());
+        // TODO: Broadcast message to our section of src is a Node.
```
maqi, line 56: better expand the comment little bit to explain why need to do so.
also
`to our section of src is a Node` => `to our section when src is a Node`

maqi (approved):

PR #2527, opened 2021-05-26
feat: use message_id instead of hash for message_filter

maqi:


src/crypto.rs
```diff
@@ -19,17 +19,6 @@ use xor_name::{XorName, XOR_NAME_LEN};
 /// SHA3-256 hash digest.
 pub type Digest256 = [u8; 32];
```
dirvine, line 21: Can this be removed now?

src/message_filter.rs
```diff
@@ -6,38 +6,16 @@
...
-            Self::NewMessage => true,
-            Self::KnownMessage => false,
-        }
-    }
-}
-
 // Structure to filter (throttle) incoming and outgoing messages.
 pub(crate) struct MessageFilter {
-    incoming: LruCache<MessageHash, ()>,
-    outgoing: LruCache<(MessageHash, XorName), ()>,
```
dirvine, line 40: What happened to filtering outgoing?

src/routing/mod.rs
```diff
@@ -466,10 +466,29 @@ async fn handle_message(dispatcher: Arc<Dispatcher>, bytes: Bytes, sender: Socke
     };
     let _span_guard = span.enter();
 
-    let message_type = match WireMsg::deserialize(bytes) {
+    let wire_msg = match WireMsg::from(bytes) {
```
dirvine, line 469: Are we deserialising the whole message here as the point is not to and just read dst an dif it's us then process it (deserialise) or forward it. 
Also msg_id to be read from msg without deserializing the whole message and added to filter / checked we have handled it.

src/routing/dispatcher.rs
```diff
@@ -42,6 +42,10 @@ impl Dispatcher {
         }
     }
 
+    pub async fn contains_incoming(&self, msg_id: &MessageId) -> bool {
+        self.core.lock().await.contains_incoming(msg_id)
+    }
+
```
dirvine, line 48: Possibly better to have `add_to_filter() ->bool` if added then cool, if not added then drop message (it's filtered)

src/routing/mod.rs
maqi: `WireMsg::from` only restore the header part, which contains the msg_id
`WireMsg::deserialize` or `WireMsg::to_message` will deserialize the whole message

src/crypto.rs
maqi: we use it in DKGKey

src/message_filter.rs
maqi: as each WireMsg will be given a random msg_id, there is no point to filtering outgoing (based on random msg_id).

dirvine: Why did we filter outgoing previously and if there was a valid reason are we now creating an issue? Or was it always useless?

src/routing/mod.rs
```diff
@@ -466,10 +466,29 @@ async fn handle_message(dispatcher: Arc<Dispatcher>, bytes: Bytes, sender: Socke
...
+        Ok(wire_msg) => wire_msg,
+        Err(error) => {
+            error!("Failed to deserialize wired_message: {}", error);
```
bochaco, line 472: I would here log something like "Failed to deserialize message header: {}"

src/routing/mod.rs
```diff
@@ -466,10 +466,29 @@ async fn handle_message(dispatcher: Arc<Dispatcher>, bytes: Bytes, sender: Socke
...
+        );
+        return;
+    }
+
+    let message_type = match wire_msg.to_message() {
         Ok(message_type) => message_type,
         Err(error) => {
-            error!("Failed to deserialize message: {}", error);
+            error!(
+                "Failed to deserialize message({:?}): {}",
```
bochaco, line 488: ...and may here log something like "Failed to deserialize message payload ({:?}): {}"

src/routing/core/messaging/handling/mod.rs
```diff
@@ -62,12 +62,6 @@ impl Core {
             return Ok(commands);
         }
 
-        // Filter messages which were already handled
-        if self.msg_filter.contains_incoming(&msg) {
-            trace!("not handling message - already handled: {:?}", msg);
```
dirvine, line 67: Should this not still be the case?

dirvine (approved):

PR #2533, opened 2021-05-28
Feat message update

joshuef:
BREAKING CHANGE: updates messaging


src/lib.rs
```diff
@@ -30,38 +30,18 @@
...
-    unused_allocation,
-    unused_attributes,
-    unused_comparisons,
-    unused_features,
-    unused_parens,
-    while_true,
-    clippy::unicode_not_nfc,
-    clippy::wrong_pub_self_convention,
-    deprecated,
-    warnings
```
lionel1704, line 55: Are these `warn` by default?

Cargo.toml
```diff
@@ -23,7 +23,8 @@ qp2p = "~0.11.9"
 rand = "~0.7.3"
 rand_chacha = "~0.2.2"
 resource_proof = "0.8.0"
-sn_messaging = "25.1.0"
+sn_messaging = {git="https://github.com/maidsafe/sn_messaging"}
+# sn_messaging = "26.0.0"
```
Yoga07, line 27: sn_messaging `26.0.0` has been published, we can uncomment this now :)

src/lib.rs
lionel1704: Ah in that case do we want to move these to the `warn` section below?

lionel1704 (approved):

PR #2548, opened 2021-06-02
refactor(locking): replace mutex witch rwlock

oetyng:
- Unnecessary locking on read operations degrade performance. Using an `RwLock`
we allow multiple readers, where a `Mutex` only ever allowed one.
- Also replaces the `lru_time_cache` with a wip light
cache (to be extended with size limit).


src/cache/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+        };
+    }
+
+    #[tokio::test]
+    async fn set_do_not_get_expired_value() {
+        let cache = Cache::new(Some(Duration::from_secs(0)));
+        let _ = cache.set(KEY, VALUE, None).await;
+        let value = cache.get(&KEY).await;
+        if value.is_some() {
+            panic!("found expired value in cache")
```
dirvine, line 167: Cold we move panics to asserts please? `assert!(value.is_some())` or similar

oetyng: Mm, yes, missed this one. I copy pasted code :+1:

oetyng: There, fixed.

dirvine (approved):

src/cache/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+
+use self::item::Item;
+use std::collections::BTreeMap;
+use std::hash::Hash;
+use std::time::Duration;
+use tokio::sync::RwLock;
+
+///
+#[derive(Debug)]
+pub struct Cache<T, V>
```
maqi, line 19: ideally, this shall be moved into lru_time_cache , maybe adding it as an additional facility.
btw. just wondering is it possible to make LruCache to be RwLock ? 
so that this Cache becomes a RwLock warrper of LruCache ?

src/message_filter.rs
```diff
@@ -6,15 +6,15 @@
...
 use crate::messages::RoutingMsgUtils;
-use lru_time_cache::LruCache;
 use sn_messaging::{node::RoutingMsg, DstLocation, MessageId};
 use std::time::Duration;
 use xor_name::XorName;
 
 const INCOMING_EXPIRY_DURATION: Duration = Duration::from_secs(20 * 60);
 const OUTGOING_EXPIRY_DURATION: Duration = Duration::from_secs(10 * 60);
-const MAX_ENTRIES: usize = 15_000;
+//const MAX_ENTRIES: usize = 15_000;
```
maqi, line 17: removed? or at least put a TODO comment

src/message_filter.rs
```diff
@@ -36,34 +36,39 @@ impl FilteringResult {
...
 }
 
 impl MessageFilter {
     pub fn new() -> Self {
         Self {
-            incoming: LruCache::with_expiry_duration_and_capacity(
-                INCOMING_EXPIRY_DURATION,
-                MAX_ENTRIES,
+            incoming: Cache::new(
+                Some(INCOMING_EXPIRY_DURATION),
```
maqi, line 47: better as to the LruCache, avoid passing Option and make the constructor name more explicit

src/routing/connectivity_complaints.rs
```diff
@@ -6,63 +6,69 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::{supermajority, ELDER_SIZE};
-use lru_time_cache::{Entry, LruCache};
-use std::{collections::BTreeSet, iter, time::Duration};
+//use crate::{supermajority, ELDER_SIZE};
```
maqi, line 9: remove

src/routing/core/public_api.rs
```diff
@@ -145,10 +145,18 @@ impl Core {
         let target_name = msg.dst().name().ok_or(Error::CannotRoute)?;
         let dest_pk = *self.section_key_by_name(&target_name);
 
-        let targets: Vec<_> = targets
-            .into_iter()
-            .filter(|peer| self.msg_filter.filter_outgoing(msg, peer.name()).is_new())
-            .collect();
+        let mut targets = vec![];
+
+        for peer in presumed_targets {
```
maqi, line 150: using filter_map will be better than this current block of code

oetyng: (using async, so cannot)

src/cache/mod.rs
oetyng: Will be addressing all cache things (still wip) in follow up PR.

src/message_filter.rs
oetyng: Will be addressing all cache things (still wip) in follow up PR.

src/message_filter.rs
oetyng: Will be addressing all cache things (still wip) in follow up PR.

src/routing/connectivity_complaints.rs
oetyng: Will be addressing all cache things (still wip) in follow up PR.

PR #2552, opened 2021-06-03
chore(cache): implement capacity

oetyng:
- Reintroduces the api used with lru_time_cache.

### DIFF
--- a/src/cache/item.rs
+++ b/src/cache/item.rs
@@ -17,6 +17,13 @@ impl<T> Item<T> {
             .map(|expiry| expiry < Instant::now())
             .unwrap_or(false)
     }
+
+    pub fn elapsed(&self) -> u128 {
+        self.expiry
+            .map(|expiry| Instant::now() - expiry)
+            .unwrap_or_default()
+            .as_millis()
+    }
 }
 
 #[cfg(test)]
--- a/src/cache/item.rs
+++ b/src/cache/item.rs
@@ -3,24 +3,36 @@ use std::time::{Duration, Instant};
 #[derive(Clone, Debug)]
 pub struct Item<T> {
     pub object: T,
-    expiry: Option<Instant>,
+    time: Option<Time>,
+}
+
+#[derive(Clone, Copy, Debug)]
+struct Time {
+    pub(crate) start: Instant,
+    pub(crate) expiry: Instant,
 }
 
 impl<T> Item<T> {
     pub fn new(object: T, item_duration: Option<Duration>) -> Self {
-        let expiry = item_duration.map(|duration| Instant::now() + duration);
-        Item { object, expiry }
+        let time = item_duration.map(|duration| {
+            let start = Instant::now();
+            Time {
+                start,
+                expiry: start + duration,
+            }
+        });
+        Item { object, time }
     }
 
     pub fn expired(&self) -> bool {
-        self.expiry
-            .map(|expiry| expiry < Instant::now())
+        self.time
+            .map(|time| time.expiry < Instant::now())
             .unwrap_or(false)
     }
 
     pub fn elapsed(&self) -> u128 {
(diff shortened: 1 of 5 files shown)


src/cache/item.rs
```diff
@@ -17,6 +17,13 @@ impl<T> Item<T> {
             .map(|expiry| expiry < Instant::now())
             .unwrap_or(false)
     }
+
+    pub fn elapsed(&self) -> u128 {
+        self.expiry
+            .map(|expiry| Instant::now() - expiry)
```
maqi, line 23: `expiry` is `now + duration` , will it cause a `overflow` here?
say, at 13:00:00 set a expiry after 10 mins, i.e. expiry is at 13:10:00
and check at 13:05:00, then that will be `13:05:00 - 13:10:00` ?

src/cache/mod.rs
```diff
@@ -72,34 +93,48 @@ where
...
+    }
+
+    async fn drop_excess(&self) {
+        let len = self.len().await;
+        if len > self.capacity {
+            let excess = len - self.capacity;
+            let read_items = self.items.read().await;
+            let mut items = read_items.iter().collect_vec();
+
+            items.sort_by(|(_, item_a), (_, item_b)| item_a.elapsed().cmp(&item_b.elapsed()));
```
maqi, line 133: elapsed with less time will be at beginning of the vector, which will be removed?

src/cache/mod.rs
```diff
@@ -22,18 +23,38 @@ where
...
     T: Ord + Hash,
 {
-    ///
-    pub fn new(item_duration: Option<Duration>) -> Self {
-        Cache {
+    /// Creating capacity based `Cache`.
+    pub fn with_capacity(capacity: usize) -> Self {
+        Self {
+            items: RwLock::new(BTreeMap::new()),
+            item_duration: None,
```
maqi, line 38: to confirm that will this mean to set expiry as to Now, which means the item is `expired` immediately?

oetyng: No.

Condition to consider an item expired as follows:

```rust
pub fn expired(&self) -> bool {
        self.expiry
            .map(|expiry| expiry < Instant::now())
            .unwrap_or(false)
    }
```

src/cache/item.rs
oetyng: Yeah, that could overflow. WIll fix.

src/cache/mod.rs
oetyng: Yep, will fix.

src/cache/mod.rs
```diff
@@ -72,34 +93,51 @@ where
...
+            .map(|(key, _)| key)
             .collect();
 
         for key in expired_keys {
-            let _ = self.items.write().await.remove(&key);
+            let _ = self.items.write().await.remove(key);
+        }
+    }
+
+    /// removs keys beyond capacity
```
maqi, line 126: removes

src/cache/mod.rs
```diff
@@ -72,34 +93,51 @@ where
...
+    async fn drop_excess(&self) {
+        let len = self.len().await;
+        if len > self.capacity {
+            let excess = len - self.capacity;
+            let read_items = self.items.read().await;
+            let mut items = read_items.iter().collect_vec();
+
+            // sort
```
maqi, line 134: no need to comment.
unless you want to use `reverse sort`  directly, to avoid the `.rev()` call below
i.e. `items.sort_by(|(_, item_a), (_, item_b)| item_b.elapsed().cmp(&item_a.elapsed()))`

oetyng: yeah, thought of that one, seemed less readable to me
but I'll actually go for it now

maqi (dismissed):

PR #2557, opened 2021-06-04
Ensure that Adults are externally reachable too

lionel1704:


src/routing/comm.rs
```diff
@@ -136,30 +141,7 @@ impl Comm {
...
-            .is_reachable(peer)
-            .await
-            .map_err(|err| {
-                info!("Peer {} is NOT externally reachable: {}", peer, err);
-                Error::AddressNotReachable { err }
-            })
-            .map(|()| {
-                info!("Peer {} is externally reachable.", peer);
-            })
+        is_reachable(self.endpoint.local_addr().ip(), peer).await
```
bochaco, line 144: Why do we move this logic to a helper function? the body is not that large and we are not breaking up anyway. Also, I'm getting a weird error due to this as it seems something becomes not `Send`

src/routing/core/messaging/handling/mod.rs
```diff
@@ -94,44 +98,53 @@ impl Core {
...
-                debug!("Received GetSectionQuery({}) from {}", name, sender);
-
-                let response = if let (true, Ok(pk_set)) =
-                    (self.section.prefix().matches(&name), self.public_key_set())
+            SectionInfoMsg::GetSectionQuery {
+                public_key,
+                is_node,
+            } => {
+                let name = XorName::from(public_key);
+                let response = if is_node && comm::is_reachable(our_local_ip, &sender).await.is_ok()
```
bochaco, line 114: I'm getting a weird error I think due to this call to `is_reachable` as it seems something becomes not `Send`

lionel1704: I was able to fix the compilation issue by changing `max_recursion` in sn_node.

joshuef (approved):

PR #2565, opened 2021-06-07
CI: add stress test to PR workflow

joshuef:


.github/workflows/pr.yml
```diff
@@ -160,6 +160,9 @@ jobs:
       - shell: bash
         run: ./scripts/tests
 
+      - name: Stress tests
+        run: cargo run --release --example stress -- -l "-" 10 5 0 20 15 0 30 20 0
```
dirvine, line 164: As this has no return type and needs a human to read the output, I wonder if we should just leave it out of CI? Esp is we mono lib the whole routing/node/client?

PR #2567, opened 2021-06-07
refactor(bootstrap): changes to new messaging flow for peers joining the network

bochaco:


src/routing/bootstrap.rs
```diff
@@ -336,70 +173,125 @@ impl<'a> State<'a> {
...
+                    if section_auth.section_key() == section_key {
                         continue;
                     }
 
-                    if section_auth.prefix().matches(&self.node.name()) {
+                    // Ignore already used recipients
+                    let new_recipients: Vec<(XorName, SocketAddr)> = section_auth
+                        .elders
+                        .iter()
+                        .filter(|(name, addr)| !used_recipient.contains(&(**name, **addr)))
```
maqi, line 242: 1, for the case of `retry` you may need to contact the same guy again ?
2, only need to check name or address? why need to check the pair?

src/routing/bootstrap.rs
```diff
@@ -336,70 +173,125 @@ impl<'a> State<'a> {
...
+                            "Joining redirected to another set of peers: {:?}",
+                            new_recipients,
+                        );
+                    }
+
+                    let prefix = section_auth.prefix;
+
+                    // For the first section, using age random among 6 to 100 to avoid
+                    // relocating too many nodes at the same time.
+                    if prefix.is_empty() && self.node.age() < FIRST_SECTION_MIN_AGE {
```
maqi, line 259: will relocate happen during the first section?
if so, then we cannot carry out this step when under relocation?

src/routing/bootstrap.rs
```diff
@@ -336,70 +173,125 @@ impl<'a> State<'a> {
...
+                            crypto::gen_keypair(&Prefix::default().range_inclusive(), age);
+                        let new_name = crypto::name(&new_keypair.public);
+
+                        info!("Setting Node name to {}", new_name);
+                        self.node = Node::new(new_keypair, self.node.addr);
+                    }
+
+                    // if we are relocating, and we didn't generate
+                    // the relocation payload yet, we do it now
+                    if relocate_payload.is_none() {
```
maqi, line 274: This `relocate_payload` is just be set to None at the beginning of the loop .
or does the flow changed to sending back `relocate_payload` via the JoinResponse ?

src/routing/bootstrap.rs
```diff
@@ -336,70 +173,125 @@ impl<'a> State<'a> {
...
+                    // Ignore already used recipients
+                    let new_recipients: Vec<(XorName, SocketAddr)> = section_auth
+                        .elders
+                        .iter()
+                        .filter(|(name, addr)| !used_recipient.contains(&(**name, **addr)))
+                        .map(|(name, addr)| (*name, *addr))
+                        .collect();
+
+                    if new_recipients.is_empty() {
+                        debug!("Joining redirected to the same set of peers we already contacted - ignoring response");
```
maqi, line 247: then no need to carry on the following steps, just call `continue` to loop ?

src/routing/bootstrap.rs
```diff
@@ -476,72 +368,77 @@ impl<'a> State<'a> {
...
-                MessageType::Routing { msg, dest_info } => (msg, dest_info),
+            // we are interested only in `JoinResponse` type of messages
+            let (routing_msg, dest_info, join_response) = match message {
                 MessageType::Node { .. }
                 | MessageType::Client { .. }
                 | MessageType::SectionInfo { .. } => continue,
+                MessageType::Routing { msg, dest_info } => {
+                    if let Variant::JoinResponse(resp) = &msg.variant {
+                        let join_response = resp.clone();
+                        (msg, dest_info, *join_response)
```
maqi, line 385: no need to clone a reference. use `(msg, dest_info, *resp)` shall work

src/routing/bootstrap.rs
```diff
@@ -476,72 +368,77 @@ impl<'a> State<'a> {
...
-            match message.variant() {
-                Variant::JoinRetry {
-                    section_auth,
-                    section_key,
-                } => {
-                    if !self.verify_message(&message, None) {
+            match join_response {
+                JoinResponse::Rejected(JoinRejectionReason::NodeNotReachable(_))
+                | JoinResponse::Rejected(JoinRejectionReason::JoinsDisallowed) => {
+                    return Ok((join_response, sender, dest_info));
```
maqi, line 396: the above `used_recipients` filter shall only be carried out for the case of `NodeNotReachable`

src/routing/bootstrap.rs
```diff
@@ -476,72 +368,77 @@ impl<'a> State<'a> {
...
-                    return Ok((
-                        JoinResponse::Retry {
-                            section_auth: section_auth.clone(),
-                            section_key: *section_key,
-                        },
-                        sender,
-                        dest_info,
-                    ));
+                    if section_auth.elders.is_empty() {
+                        error!("Invalid JoinResponse missing peers: {:?}", join_response);
```
maqi, line 406: Invalid JoinResponse with empty elders

src/routing/bootstrap.rs
bochaco: good catch @maqi , thanks, I'll fix 1).
For 2), I guess I can check the address only.

src/routing/bootstrap.rs
bochaco: Yes, but it's outside of the loop. 
So the idea is that the first request will won't send the relocate payload since we don't know the section prefix (similar to what it used to happen when we would be first getting info from the section with a GetSectionInfo request, but now we send a JoinRequest to grab that same info), once we get the first JoinResponse::Retry (which contains the prefix) we will populate it with the relocate payload.

src/routing/bootstrap.rs
bochaco: No that doesn't work because I'm moving the `msg`.

src/routing/bootstrap.rs
maqi: maybe the boostrap flow got changed, however the relocate flow as I remember is:

bochaco: yes, but previously we still send a GetSectionInfo to grab the prefix from the target section, so here we do the same

src/routing/bootstrap.rs
bochaco: we currently do this step when relocating, we grab the prefix from the GetSectionInfo request

src/routing/bootstrap.rs
maqi: yes, but previously `relocate_details` is a passed_in populated value, before the GetSectionInfo call.
now it's set to None before the call.

src/routing/tests/mod.rs
```diff
@@ -164,7 +159,7 @@ async fn receive_mismatching_get_section_request_as_adult() -> Result<()> {
     assert_matches!(
         message,
         SectionInfoMsg::GetSectionResponse(GetSectionResponse::Redirect(addrs)) => {
-            assert_eq!(addrs, elders_addrs)
+            assert_eq!(addrs, section_auth)
```
Yoga07, line 162: Can rename addrs to received_auth to reduce confusion.

bochaco: ah yeah, sure I'll do that!

Yoga07 (approved):

maqi (approved):

PR #2570, opened 2021-06-09
refactor(reachability): verify reachablility before handling JoinRequest

lionel1704:


src/routing/dispatcher.rs
```diff
@@ -100,6 +104,28 @@ impl Dispatcher {
                 message,
                 dest_info,
             } => {
+                if let (Variant::JoinRequest(join_request), Some(sender)) =
```
maqi, line 107: you can have this block of code inside the `handle_join_request` function, 
so that the handling logic can be clustered ?

src/routing/core/api.rs
```diff
@@ -304,23 +301,10 @@ impl Core {
...
-
-            trace!("Sending {:?} to {}", variant, peer);
-            Ok(vec![self.send_direct_message(
-                (*peer.name(), *peer.addr()),
-                variant,
-                *self.section.chain().last_key(),
-            )?])
-        }
+        self.propose(Proposal::Online {
+            member_info: MemberInfo::joined(peer),
```
maqi, line 305: the reachable flag of this input peer was set separately at https://github.com/maidsafe/sn_routing/blob/master/src/routing/dispatcher.rs#L185

Though the chance is tiny, still better to check the flag is true here?

Meanwhile, with your current change, I think we can now have `Command::ProposeOnline` contains the peer have reachable flag set to true directly.
So that we can avoid having extra check here.

i.e. at the place at  https://github.com/maidsafe/sn_routing/blob/master/src/routing/core/messaging/handling/mod.rs#L676
the reachable flag shall be set to true already.

src/routing/dispatcher.rs
lionel1704: The issue I faced there was we don't have access to the `comm` object to call `is_reachable()`

maqi: Ah, right, it's in the separate object.
ok then

src/routing/dispatcher.rs
```diff
@@ -182,7 +208,7 @@ impl Dispatcher {
                 previous_name,
                 destination_key,
             } => {
-                peer.set_reachable(self.comm.is_reachable(peer.addr()).await.is_ok());
+                peer.set_reachable(true);
```
maqi, line 211: better put a comment to explain when can setting to `true` directly.

maqi (approved):

PR #2571, opened 2021-06-09
Notify other Elders to start a connectivity check when a peer goes offline

lionel1704:
- feat(connectivity): refactor handling of lost connections


src/routing/core/connectivity.rs
```diff
@@ -134,7 +134,9 @@ impl Core {
             return Ok(vec![]);
         }
 
-        self.propose_offline(name)
+        let mut commands = self.propose_offline(name)?;
```
bochaco, line 137: Shouldn't we `propose_offline` only after we did `StartConnectivityTest` and considered it offline?

lionel1704: This call is within `handle_peer_lost` so it's _after_ we know that the peer is offline. The signal is for other nodes to start the connectivity test

src/routing/dispatcher.rs
```diff
@@ -217,6 +220,48 @@ impl Dispatcher {
...
+                Ok(self.core.read().await.send_or_handle(msg, &peers))
+            }
+            Command::TestConnectivity(name) => {
+                let mut commands = vec![];
+                if let Some(peer) = self
+                    .core
+                    .read()
+                    .await
+                    .section()
+                    .members
```
maqi, line 254: better use `.members().get()` instead of two lines of `.members`

src/routing/mod.rs
```diff
@@ -192,6 +192,15 @@ impl Routing {
         self.dispatcher.clone().handle_commands(command).await
     }
 
+    /// Signals the Elders of our section to test connectivity to a node.
+    pub async fn start_connectivity_test(&self, name: XorName) -> Result<()> {
+        if !self.is_elder().await {
```
maqi, line 197: thought it's said that Adult shall notify Elder to start a connectivity test?

tests/drop.rs
```diff
@@ -50,7 +50,10 @@ async fn test_node_drop() -> Result<()> {
 
     tracing::info!("Dropped {} at {}", dropped_name, dropped_addr);
 
-    for (node, _) in &mut nodes {
+    //  A failed send_message from any node should
+    // trigger voting by all nodes
+    {
+        let (node, _) = nodes.get(0).ok_or_else(|| anyhow!("Missing Node"))?;
```
maqi, line 56: better pick the `last` node, not the first one.
as the first one is normally a genesis, which may have some privilege.

src/routing/mod.rs
lionel1704: Ah yes that's right. I'll remove this condition :+1:

maqi (approved):

PR #2577, opened 2021-06-12
Dkg

dirvine:
Rename Agreement -> Dkg as it is Dkg we do the agreements via messages and here we should only be updating section chain and SAP with different members (handling churn)

BREAKING CHANGE


.gitignore
```diff
@@ -14,3 +14,5 @@ target/
 *.sublime-*
 .cargo/
 .DS_Store
+.core
+core
```
maqi, line 18: missing end of file?
also, why now need to add these two in?

src/dkg/commands.rs
```diff
@@ -0,0 +1,123 @@
+// Copyright 2020 MaidSafe.net limited.
```
maqi, line 1: 2021

src/dkg/voter.rs
```diff
@@ -0,0 +1,181 @@
+// Copyright 2020 MaidSafe.net limited.
```
maqi, line 1: 2021

src/dkg/session.rs
```diff
@@ -6,25 +6,21 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use super::{DkgFailureSignedSetUtils, DkgFailureSignedUtils};
 use crate::{
+    dkg::dkg_msgs_utils::DkgFailureSignedUtils,
+    dkg::{commands::DkgCommand, dkg_msgs_utils::DkgFailureSignedSetUtils},
```
maqi, line 11: Line 10 can be merged into Line 11 ?

PR #2582, opened 2021-06-14
Add documentation to Cache and fix a possible bug

connec:
- e8ed35b8 **docs(cache): Populate empty doc comments**

- a67164c7 **refactor(cache): Avoid suppressing lints**

- b5dab42c **fix(cache): Prevent exposing expired values from Cache::set**

  `Cache::set` was not testing the removed item's expiry, meaning expired
  values could be returned.

- 93de8c87 **ci: Install cargo-udeps without --locked**

  This works around a compilation issue caused by
  https://github.com/rust-lang/rust/issues/85574. A more permanent fix may
  come in future.

- 37fc9daf **chore: Remove unused dashmap dependency**

  This has been unused since #2556.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -115,7 +115,7 @@ jobs:
       # Install and run cargo udeps to find unused cargo dependencies
       - name: cargo-udeps unused dependency check
         run: |
-          cargo install cargo-udeps --locked
+          cargo install cargo-udeps
           cargo +nightly udeps --all-targets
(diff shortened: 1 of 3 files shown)


dirvine (dismissed):
Nice one @connec you're more than welcome in these parts :-)

connec:
Had to fix a couple of actions issues @dirvine, but it's green across the board now!

dirvine (approved):

PR #2584, opened 2021-06-14
refactor(relocate): simplifying relocation flow and logic

bochaco:


bochaco:
This will be sent against the new git@github.com:maidsafe/sn.git repo.

PR #2597, opened 2021-06-18
fix(filter): ignore JoinRequests from filtering

Yoga07:


src/routing/core/mod.rs
```diff
@@ -92,8 +92,11 @@ impl Core {
     // Miscellaneous
     ////////////////////////////////////////////////////////////////////////////
 
-    pub async fn add_to_filter(&mut self, msg_id: &MessageId) -> bool {
-        self.msg_filter.add_to_filter(msg_id).await
+    pub async fn add_to_filter(&mut self, wire_msg: &WireMsg) -> bool {
+        match wire_msg.is_join_request() {
+            Ok(false) => self.msg_filter.add_to_filter(&wire_msg.msg_id()).await,
+            _ => false,
```
lionel1704, line 98:
```suggestion
        wire_msg.is_join_request()? && self.msg_filter.add_to_filter(&wire_msg.msg_id()).await
```

Yoga07: `wire_msg.is_join_request()?` This can't be done as the caller method return only a `bool` :)

S-Coyle (approved):

